PR #12, opened 2016-07-18
SPLICE-696 Fixing Small Memory Leak in Spark Stream Processing

jleach4:
There is a path where close() may not have been called which will leak Streams.


genedavis:
+1 Good find

ammolitor:
:sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles:

PR #14, opened 2016-07-19
SPLICE-698 Decreasing the clean internal in case there is signifant r…

jleach4:
…ate of Spark Queries


scottfines:
Does this have any negative performance implications? Just for the sake of the future when we are wondering why this number is set this way

jleach4:
It is just a week reference background cleaner (Limited Effect).  Spark 2.0 has removed it entirely (i.e. just clean up when you are done).  This configuration is saying that we will keep results (accumulators, results) for an hour worth of queries.

scottfines:
excellent. Thank you for the explanation  :)

PR #15, opened 2016-07-19
DB-5011: pop statement context properly when a trigger fails

jyuanca:
The code change is to fix two problems:
1) when a trigger throws an exception, we pop up all statement context for trigger on stack and clean them up. There could be more than one statement contexts because a trigger may fire another trigger for another table. Pop the stack until the parent statement is not in a trigger, which is the parent statement context(GenericStatementContext.java).
2) Always look at the context at the top of the stack(GenericLanguageConnectionContext.java).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
@@ -2286,7 +2286,7 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
 
                 parentStatementContext=statementContexts[0];
             }else{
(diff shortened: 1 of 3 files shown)


scottfines:
Could we add a bit more description about what went wrong, what the fix is, etc.? Just a sentence or two is fine, doesn't have to be great detail, but that way we can get a quick idea of what's being fixed just by looking at the error message. 

Previously, it didn't matter too much, since I could always just call and ask, but since these are public, we want to make sure we explain what we are doing and why we are doing it to anyone who may be watching.

jyuanca:
More description added.

scottfines:
+1 looks good to me.

jeffcunningham:
+1

jleach4:
+1

PR #17, opened 2016-07-19
SPLICE-700: Import requires write access to data file on HDFS

jeffcunningham:
We were catching the wrong exception from fs#access(path, mode).
We were catching AccessDeniedException but was throwing org.apache.hadoop.security.AccessControlException.
I changed to catch IOException which is the ancestor of all exceptions thrown by the method. The rational is that, if we get any IOException, the access mode we were checking is not available.


jyuanca:
+1

jleach4:
+1, but we may have to revisit using local filesystem.  I think my original strategy there is starting to bite us.

jeffcunningham:
Monitor? Monitor? Where art thou Monitor?

PR #18, opened 2016-07-19
SPLICE-700: Import requires write access to data file on HDFS

jeffcunningham:
We were catching the wrong exception from fs#access(path, mode).
We were catching AccessDeniedException but was throwing org.apache.hadoop.security.AccessControlException.
I changed to catch IOException which is the ancestor of all exceptions thrown by the method. The rational is that, if we get any IOException, the access mode we were checking is not available.
Modified ImportUtils.validateReadable(path, isDirectory) to include an existence check since the logic in HNIOFileSystem#getInfo(path) changed.
Also found an incorrect error msg being thrown in validateReadable(). Changed its handle name and added a true "file not found" msg.
Fixed the ImportErrorIT#testCannotFindFile(), which expects a file not found exception.


jyuanca:
+1

scottfines:
+1

jeffcunningham:
Clean IT run on cdh5.4.10

scottfines:
Considering the clean IT run report, and the fact that the previous PR garnered the necessary votes, I'm okay with performing this merge.

scottfines:
Build looks good, deleting branch

PR #20, opened 2016-07-20
SPLICE-704 Project Tungsten pre-cursor code and build configuration

jleach4:
Initial Drop of Project Tungsten SerDe to and from DataValueDescriptors along with using HBase's OrderedBytes library (Nick D's Lib).  I have written some basic tests and added the functionality.  It does not change the on disk representation yet.


ammolitor:
Does this code require Java 8 to build?

jleach4:
It is the Yahoo sketching algorithms that require java 8...  I am waiting on those until I get an all clear for installation on our platform partners.

db-engine/pom.xml
```diff
@@ -33,6 +33,16 @@
     <artifactId>db-engine</artifactId>
     <dependencies>
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
```
ammolitor: is there a non-vendoer specific version of these we can depend on at this level?

genedavis:
SQLDateTest.java and other new *Test.java files need copyright statement at the top?

jleach4:
Total rookie move...  Will get the copyright on top...

scottfines:
Please rebase into a single commit. It's a good habit for us to get into, since we will be asking new contributors to do the same.

scottfines:
Also, since this requires Java 8 to function, I think we should be really careful about when this drops, since our next point release would then require an upgrade script (contrary to our stated desires about not requiring upgrades for patch releases).

jleach4:
It does not need Java 8...

PR #22, opened 2016-07-21
Splice 707 708

jleach4:
Applies Derby Fixes to Order By Logic.  These changes will handle with exceptions calculations in the order by with values clauses.  Also, it allows a simple values order by to work.


jleach4:
Closing...  Need to merge with master first...

PR #23, opened 2016-07-21
Splice 707 splice 708

jleach4:
Applying two derby patches to support order by error checking and simple execution of values clauses...

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
@@ -17,6 +17,7 @@ package com.splicemachine.derby.impl.sql.execute.operations;
 
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.SQLSyntaxErrorException;
 import java.sql.Timestamp;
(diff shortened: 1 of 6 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
```diff
@@ -263,6 +264,27 @@ public void testSortOperationByKey2Descending() throws Exception {
...
 	}
 
+	@Test(expected = SQLSyntaxErrorException.class)
+	public void testDerby6027() throws Exception {
+		methodWatcher.executeQuery("values 1,2 order by int(1)");
+	}
+
+	@Test(expected = SQLSyntaxErrorException.class)
+	public void orderByWithComputation() throws Exception {
+		methodWatcher.executeQuery("values 1 order by 1+0");
```
jeffcunningham: You could catch the exception here and assert its sql code is the new one you added.

jeffcunningham:
Other than the commented test improvement, +1

scottfines:
Please create a separate pull request for each distinct ticket. That way we can keep our patch-sets clean, and other people wanting to merge patches into their fork will know what they are getting.

jyuanca:
+1

PR #25, opened 2016-07-21
DB-4198:throw an exception when aggregate function is applied to rowid Column

jyuanca:
Throw an exception and return an error message to a client when an aggregate funciton is applied to a rowid column, instead of failing an assertion and dropping client connection.


jeffcunningham:
+1

scottfines:
+1

PR #27, opened 2016-07-22
SPLICE-114: clear table cache when dropping a table failed

jyuanca:
All objects defined on a table(such as indexes, constraints) are dropped before dropping a table. For any reasons, if an exception is thrown when dropping a table, the txn will be rolled back, and table definition is not changed on disk. However, the table object in cache may have been modified. This will cause erroneous behavior when the table is accessed again.


jeffcunningham:
+1, nice.

scottfines:
+1 very cool

scottfines:
Have you verified the ITs locally? Just wanting to know if I should before merging.

jyuanca:
I've run ITs locally. All ITs passed.

PR #28, opened 2016-07-22
DB-5007: Adding correct error message and test

scottfines:
Here's a revised version which I tested BEFORE pushing :)

It's basically the same change as before, but I reverted a few bad changes in messages.xml that I apparently let IDEA do for me.


jyuanca:
+1

jeffcunningham:
+1

PR #29, opened 2016-07-22
SPLICE-713 NPE in left join whose inner table has a generated column

jleach4:
Forwarding a derby patch for a NPE when using left joins with generated columns.  Funny we have not seen this one since we have a decent number of generated column tests/POCS.


jeffcunningham:
+1

scottfines:
+1

jyuanca:
+1

PR #30, opened 2016-07-22
SPLICE-714 NullPointerException when using more than one parameter in…

jleach4:
NullPointerException when using more than one parameter in COALESCE.  This is another mindless vacation special.  

Corresponds to this derby bug fix.

https://issues.apache.org/jira/browse/DERBY-6273


scottfines:
I'm sorry, but I'm having trouble seeing the actual fix here--it looks like an extra IT and the changing of some indentation?

jleach4:
The bug is the "break" line that is removed.

jyuanca:
+1

jeffcunningham:
+1

scottfines:
Ah, missed that, thanks. +1

PR #31, opened 2016-07-22
Fix explain for spark

jyuanca:
Fixes two issues:
1) Pick up useSpark property in connection context
2) Serialize/deserialize explain plan when executing explain on spark, because it's stored in a ThreadLocal variable.


scottfines:
Please split into a Pull Request for each distinct issue; doing so will make it easier to track patch sets.

PR #32, opened 2016-07-22
SPLICE-28: bad file creation broken on hdfs

jeffcunningham:
Bad record accumulators create ".bad" files in <bad_record_directory>/.tmp directory but do not merge.

In the case where there's only one bad record accumulator, accumulator#addInPlace(BadRecordsRecorder r1, BadRecordsRecorder r2), which forwards to merge(), never gets called by spark.
Removed the merge. Now BadRecordRecorder writes to only its uniquely named file.
Tested on cluster with small (10 record) file and with TPCC 1000 where import date format was changed to force import errors.

See more details in https://splice.atlassian.net/browse/SPLICE-28


jeffcunningham:
You guys are going to have to finish this one without me.  I need to leave in 1/2 hour to catch a flight.
Thanks!

scottfines:
+1 on the patch: enjoy vacation!

jyuanca:
+1

PR #36, opened 2016-07-24
SPLICE-704 Project Tungsten pre-cursor code and build configuration

jleach4:
Squashed Project Tungsten to a single commit.  This pull request lays down code for Project Tungsten...

### DIFF
--- a/db-engine/pom.xml
+++ b/db-engine/pom.xml
@@ -33,6 +33,16 @@
     <artifactId>db-engine</artifactId>
     <dependencies>
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <version>2.6.0-cdh5.4.10</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-common</artifactId>
+            <version>1.0.0-cdh5.4.10</version>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>db-shared</artifactId>
             <version>${project.version}</version>
@@ -47,13 +57,6 @@
             <artifactId>${spark-assembly-id}</artifactId>
             <version>${spark.version}</version>
         </dependency>
-        <!--
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>11.0.2</version>
-        </dependency>
--->
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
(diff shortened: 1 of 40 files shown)


jyuanca:
+1

db-engine/pom.xml
```diff
@@ -33,6 +33,16 @@
...
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <version>2.6.0-cdh5.4.10</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-common</artifactId>
+            <version>1.0.0-cdh5.4.10</version>
+        </dependency>
+        <dependency>
```
scottfines: Don't we need to make this profile-dependent, so that we build against the right versions of hadoop/hbase?

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -1118,10 +1129,131 @@ private static int getWholeDigits(BigDecimal decimalValue) {
...
+	 *
+	 * TODO JL: This is not accurate...
+	 *
+	 * @return
+	 * @throws StandardException
+     */
+	@Override
+	public int encodedKeyLength() throws StandardException {
+		throw new RuntimeException("not implemented correctly");
+//		return isNull()?1:9;
```
scottfines: This appears to be the only data type that you removed; why remove this one and only this one?

db-testing/src/test/java/com/splicemachine/db/iapi/types/SQLRealTest.java
```diff
@@ -0,0 +1,111 @@
...
+            SQLReal float1 = new SQLReal(200.0f);
+            SQLReal float2 = new SQLReal(100.0f);
+            Assert.assertEquals("Integer subtract Fails",100.0f,float1.minus(float1, float2, null).getFloat(),0.0f);
+        }
+
+        @Test(expected = StandardException.class)
+        public void testPositiveOverFlow() throws StandardException {
+                SQLReal float1 = new SQLReal(Float.MAX_VALUE);
+                SQLReal float2 = new SQLReal(1.0f);
+                float1.plus(float1,float2,null);
```
scottfines: You should probably check for the correct type of StandardException as well.

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceConglomerate.java
```diff
@@ -215,4 +219,29 @@ public boolean equals(Object o){
     public int hashCode(){
         return (int)(containerId^(containerId>>>32));
     }
+
+    @Override
+    public void write(UnsafeRowWriter unsafeRowWriter, int ordinal) throws StandardException {
+        throw new RuntimeException("Not Implemented");
```
scottfines: Don't use RuntimeException to denote not being implemented: use UnsupportedOperationException

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
jleach4: Not clear currently how to compute that length.

jyuanca:
If no major issues, I will merge it by end of day, agree?

PR #37, opened 2016-07-25
SPLICE-712: Adding additional source paths for ColumnReference

scottfines:
We weren't catching ColumnReference versions, assuming everything was a VirtualColumnNode instead. This corrects that and adds an IT to ensure the flattening works as desired.


jyuanca:
+1

PR #38, opened 2016-07-25
SPLICE-128: Fix SYSCS_UTIL.SYSCS_PEEK_AT_SEQUENCE

jyuanca:
DataDictionaryImpl.peekAtSequence() tries to peek next sequence number from SequenceUpdater, which is obsolete. The code changes make it work by peeking SpliceSequence.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/sequence/AbstractSequence.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/sequence/AbstractSequence.java
@@ -52,6 +52,10 @@ public abstract class AbstractSequence implements Sequence, Externalizable{
         return currPosition.getAndAdd(incrementSteps);
     }
 
+    public long peekAtCurrentValue() throws StandardException {
+        return currPosition.get();
+    }
+
     protected abstract long getCurrentValue() throws IOException;
 
     protected abstract boolean atomicIncrement(long nextValue) throws IOException;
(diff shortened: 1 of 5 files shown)


jfilali:
+1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/sequence/AbstractSequence.java
```diff
@@ -52,6 +52,10 @@ public long getNext() throws StandardException{
         return currPosition.getAndAdd(incrementSteps);
     }
 
+    public long peekAtCurrentValue() throws StandardException {
+        return currPosition.get();
```
dgomezferro: If the current block is consumed this will return a wrong result. We should perform the same check as in getNext() and allocate a new block if necessary.

jyuanca: The current block can only be consumed by getNext(). PeekAtCurrentValue does not consume anything, only returns the current value.

dgomezferro: Quoting from the docs:

> The SYSCS_UTIL.SYSCS_PEEK_AT_SEQUENCE function shows you the very next value that will be returned by a NEXT VALUE FOR clause. Users should never directly query the SYSSEQUENCES table, because that will cause sequence generator concurrency to slow drastically.

If the sequence has a state such that the next call to `getNext()` will allocate a new block, with the current change `peekAtCurrentValue()` will return `currPosition`, but that won't necessarily be the value that `getNext()` will return if called just afterwards, because the `currPosition` will be (possibly) updated by the new block allocation.

PR #39, opened 2016-07-25
Fixed SPLICE 709 - Use Kryo pools in result streamer / stream listener

jfilali:
I didn't change ResultStreamer because this code is executed in Spark for each request by deserializing the class. It doesn't make sense then to spin up a pool each time.

 Made SpliceSparkKryoRegistrator to implement KryoPool.KryoRegistry so we can reuse the Kryo Pool.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSparkKryoRegistrator.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSparkKryoRegistrator.java
@@ -24,6 +24,8 @@ import com.esotericsoftware.kryo.serializers.DefaultSerializers;
 import com.esotericsoftware.kryo.serializers.FieldSerializer;
 import com.esotericsoftware.kryo.serializers.MapSerializer;
 import com.google.common.collect.ArrayListMultimap;
+import com.splicemachine.EngineDriver;
+import com.splicemachine.SpliceKryoRegistry;
 import com.splicemachine.derby.ddl.DDLChangeType;
 import com.splicemachine.derby.ddl.TentativeAddColumnDesc;
 import com.splicemachine.derby.ddl.TentativeAddConstraintDesc;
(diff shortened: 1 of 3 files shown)


scottfines:
+1 for this particular change. I'm not sure that we should cross off the KryoPool idea entirely though--we could use a global KryoPool that is shared across the JVM and save on construction during each ResultStreamer execution.

jfilali:
Thanks Scott, please could you explain to me more in detail you thinking. 
The Spark Executor are running in Yarn, so if i understand correctly you would want to have a Kryo Pool attached to each Spark Executor and have it accessible during the execution of the ResultStreamer rdd ?

dgomezferro:
I think that's what he meant, yes. I also agree, we should have a KryoPool reusable on the executors. The executors are usually long lived, so we don't end up creating too many KryoPools.

I'd add that part and then it'd be ready for merging in my opinion.

jyuanca:
Please open a new branch and pull request for each jira issue.

jfilali:
Sorry i don't know that end up there,

jfilali:
Daniel, the difficulty here is that we don't have access to external object in a rdd. Unless we modify the spark code i can't come around the simple serialization system and create a Kryo Pool that will be unique to the JVM and accessible for all the executors in each rdd ... does that make sense or i miss something ,

dgomezferro:
I'm not sure why you think it's not possible, we can meet today if you'd like to discuss. The basic idea would be to replicate what you did on `StreamListenerServer` in `ResultStreamer`, that is, getting an `KryoPool` instance (which is shared, since it's stored in a static field in `SpliceSparkKryoRegistrator`) and then using that `KryoPool` to generate `Kryo` objects.

jfilali:
Ok Daniel, let meet, we will discuss also others stuff i wanted to ask you.

jfilali:
creating a new pull request , because my repo got corrupted

PR #40, opened 2016-07-26
DB-5561: revert spark.cleaner.ttl setting

jyuanca:
revert changes to spark.cleaner.ttl setting, because it did not fix DB-5252, but may cause long running job to fail.


dgomezferro:
+1

ammolitor:
:+1:

PR #43, opened 2016-07-27
SPLICE-726 Set start/stop keys properly in Spark

dgomezferro:
This change makes sure we set start/stop keys properly on Spark scans when they are bounded, for instance when we have an inequality on an index access. Previously we would scan the whole region.

Furthermore, when computing cutpoints we didn't take into account the start/stop scan keys, so we could create more partitions than necessary that did no useful work.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/mrio/api/core/HBaseSubregionSplitter.java
+++ b/hbase_sql/src/main/java/com/splicemachine/mrio/api/core/HBaseSubregionSplitter.java
@@ -16,6 +16,7 @@
 package com.splicemachine.mrio.api.core;
 
 import com.google.protobuf.ByteString;
+import com.google.protobuf.ZeroCopyLiteralByteString;
 import com.splicemachine.access.HConfiguration;
 import com.splicemachine.access.hbase.HBaseTableInfoFactory;
 import com.splicemachine.coprocessor.SpliceMessage;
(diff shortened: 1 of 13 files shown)


hbase_sql/src/main/java/com/splicemachine/mrio/api/core/HBaseSubregionSplitter.java
```diff
@@ -67,11 +67,15 @@
...
                                 if (LOG.isDebugEnabled()) {
-                                    LOG.debug(String.format("Original split [%s,%s]", CellUtils.toHex(startKey), CellUtils.toHex(stopKey)));
+                                    LOG.debug(String.format("Original split [%s,%s] with scan [%s,%s]",
+                                            CellUtils.toHex(startKey), CellUtils.toHex(stopKey),
+                                            CellUtils.toHex(scanStartRow), CellUtils.toHex(scanStopRow)));
                                 }
 
                                 SpliceMessage.SpliceSplitServiceRequest message = SpliceMessage.SpliceSplitServiceRequest.newBuilder()
-                                        .setBeginKey(ByteString.copyFrom(startKey)).setEndKey(ByteString.copyFrom(stopKey)).build();
+                                        .setBeginKey(ByteString.copyFrom(scanStartRow))
```
scottfines: Isn't there a ZeroCopyLiteralByteString that we can use to wrap the byte array without performing the copy?

scottfines:
+1

jleach4:
+1 on the commit...

How could we have found this issue in a debug mode, etc.?  We have had a few issues where we scan more than we should (Merge Join, etc.) but no clear way to catch the issue...

PR #45, opened 2016-07-27
DB-5252 Fix ContextManager leak

dgomezferro:
Close SpliceTransactionResourceImpl always after marshalling a
transaction and make sure to remove the ContextManager from the list of
active CMs


jyuanca:
+1

genedavis:
+1

PR #46, opened 2016-07-27
DB-5586: add warning message for restore

jyuanca:
Add warning message for restore, and print out any warnings when restore completes.


dgomezferro:
+1

PR #47, opened 2016-07-27
Fixed SPLICE-724 : ORDER BY with UNION ignored

jfilali:
Should not return the operation if distinct is set to true but continue to traverse the execution tree.


genedavis:
+1

dgomezferro:
+1

PR #48, opened 2016-07-28
SPLICE-729 : Added unit test that make sure distinct is to true in So…

jfilali:
SPLICE-729 : Added unit test that make sure distinct is to true in SortOperation#getDataSet
this is a unit test for SPLICE-724  : ORDER BY with UNION ignored

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
@@ -40,12 +40,16 @@ public class SortOperationIT extends SpliceUnitTest {
 	public static final String TABLE_NAME_2 = "PERSON";
     public static final String TABLE_NAME_3 = "BOOL_TABLE";
 	public static final String TABLE_NAME_4 = "A";
+	public static final String TABLE_NAME_5 = "B";
+	public static final String TABLE_NAME_6 = "C";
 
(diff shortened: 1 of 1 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
```diff
@@ -140,6 +146,27 @@ protected void starting(Description description) {
...
                     ps.executeBatch();
+
+					ps = spliceClassWatcher.prepareStatement("insert into "+ spliceTableWatcher5+" values (?)");
+					ps.setInt(1,1); ps.execute();
+					ps.setInt(1,1); ps.execute();
+					ps.setInt(1,1); ps.execute();
+					ps.setInt(1,5); ps.execute();
+					ps.setInt(1,2); ps.execute();
+					ps.setInt(1,7); ps.execute();
+					ps.setInt(1,4); ps.execute();
```
dgomezferro: Minor suggestion, feel free to ignore: I'd add a value bigger than some on the other table just to make sure we aren't returning  something like ( ( B order by id) union ( C order by id) )

dgomezferro:
+1

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SortOperationIT.java
jfilali: Right on, let me add that.

jfilali:
done, rebase and pushed the changes

PR #50, opened 2016-07-29
SPLICE-731 Fix maven exclusions

dgomezferro:
Some exclusions were wrong (transposed artifactId/groupId) and some were missing. This caused the standalone version to sometimes fail to start, as reported on the user forum: http://community.splicemachine.com/community/forum/topic/getting-error-on-startup/


ryanwoodsmall:
+1

PR #51, opened 2016-07-29
Fixed SPLICE 709 - Use Kryo pools in result streamer / stream listener

jfilali:
- Added Kryo Pool in Result Streamer and Streamer listener.
  I have added a OpenHandler and will do more cleaning  with the other classes as it get confusing to have class handling multiple handlers via anonymous classes


hbase_sql/src/main/java/com/splicemachine/stream/handlers/OpenHandler.java
```diff
@@ -0,0 +1,51 @@
+package com.splicemachine.stream.handlers;
```
dgomezferro: Missing license/copyright notice

dgomezferro:
+1 pending license on new file. Good job!

jyuanca:
+1 after adding licese info

PR #52, opened 2016-08-01
SPLICE-732 Fixing Race Condition in Finding Partition during splits, …

jleach4:
Fixing the handling of a race condition for getting partition information.  I kept the 100 attempt check in there in case something is very wrong.


jyuanca:
+1

jfilali:
+1

PR #53, opened 2016-08-02
DB-5595: Parallelize backup/restore

jyuanca:
The current backup/restore is not efficient because it backup/restore one table a time. The code changes parallelize backup/restore by creating callables and submitting to ExecutorCompletionService.  This speeds up backup/restore by 3-4 times on bare splice db.

I include some other minor changes in the commit:
1) rename cancel_backup to syscs_cancel_backup for naming consistency.
2) Add an error message for backup cancelation.
2) code clean up.

**Note: please review code changes on both spliceengine and spliceengine_ee.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
+++ b/splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
@@ -84,6 +84,7 @@ public class BackupSystemProcedures {
         } catch (Throwable t) {
             resultSets[0] = generateResult("Error", t.getLocalizedMessage());
             SpliceLogUtils.error(LOG, "Database backup error", t);
+            t.printStackTrace();
         }
     }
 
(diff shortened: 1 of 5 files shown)


splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
```diff
@@ -84,6 +84,7 @@ public static void SYSCS_BACKUP_DATABASE(String directory, String type, ResultSe
         } catch (Throwable t) {
             resultSets[0] = generateResult("Error", t.getLocalizedMessage());
             SpliceLogUtils.error(LOG, "Database backup error", t);
+            t.printStackTrace();
```
scottfines: Let's lose the printStackTrace(): those are annoying

splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
```diff
@@ -269,7 +270,7 @@ public static void SYSCS_DELETE_OLD_BACKUPS(int backupWindow, ResultSet[] result
         }
     }
 
-    public static void CANCEL_BACKUP(ResultSet[] resultSets) throws StandardException, SQLException {
+    public static void SYSCS_CANCEL_BACKUP(ResultSet[] resultSets) throws StandardException, SQLException {
```
scottfines: Making this change will require additional steps when performing upgrades to this version: not severe (just need to reload stored procedures and change application code which uses it), but painful enough I have to ask if it's worth it to make the change.

scottfines:
OK, I'm probably blind, but where is the concurrency code? the part that creates the callables and submits them, I mean.

jyuanca:
Look at spliceengine-ee for code that actually creates callables

PR #54, opened 2016-08-02
Fixed SPLICE-735

jfilali:
Trying to drop a column with a trigger depending on that column will throw a error but behind the scene the column is effectively deleted and the table descriptor will become stall.
This fix makes sure we handling triggers and constraints before we effectively deleting the column.


jyuanca:
The column was not actually dropped if an exception was thrown when dropping triggers or constraints, because the transaction was rolled back. It appeared that the column was dropped because the table descriptor in the cache was modified.  All you should do is to invalidate table descriptor in the cache if an exception was thrown. See DropTableConstantOperation.executeConstantAction() for an example.

jyuanca:
+1

PR #55, opened 2016-08-03
SPLICE-682: Adding IT check to make sure message is well formatted

scottfines:
Modifying an IT to check for table and constraint names in error messages to avoid regressing in SPLICE-682.


jyuanca:
+1

jfilali:
+1

PR #56, opened 2016-08-04
SPLICE-739: Correcting WholeTextInputFormat.

scottfines:
WholeTextInputFormat was only processing a single path out of a combined
list, which meant that multiple files were not being processed. This is
easily detected in a UT(WholeTextInputFormatTest), and corrected here.

It also includes an IT because why not.

I also included a few (minor) code cleanups, but it passes ITs on my box.


jleach4:
+1

genedavis:
+1, except need copyright text on top of new Test file.

jyuanca:
+1

PR #58, opened 2016-08-08
SPLICE-757 Removing Ignore from Timeout Test

jleach4:
Removing Ignore from test, making sure we can load the jar correctly,  adding to serial transactions since we are loading a jar.


scottfines:
+1

jyuanca:
+1

jeffcunningham:
+1

PR #59, opened 2016-08-08
DB-5592: Making HDP2.3.6 compile

scottfines:
This makes HDP-2.3.6 compile. I ran the ITs against it and a CDH branch to make sure there was no contamination, but it should be pretty clean


ammolitor:
does this cause any issue with hdp2.4.2 compilation?

scottfines:
I don't believe so, but I will confirm and adjust if necessary.

Thanks for the catch.

Scott

scottfines:
HDP 2.4.2 compiles successfully. This isn't too much of a surprise--2.3.x and 2.4.x appear to run on different minor release versions of HBase (1.2.x versus 1.1.x), but it is good to double check

ammolitor:
um, check your assumptions @scottfines, last I looked they are they are both using <envHbase>1.1.2</envHbase> and hence have the same code underneath.

scottfines:
Well, I do get into trouble working from memory :)

ammolitor:
well, your branch has this as I expected and so if they both compile here, then I am :+1:

scottfines:
They don't. I don't know why, but it compiled and ran through ITs the first time through, but after your assumptions comment I re-checked, and it *is* showing problems. Will close this and re-open a new PR once I've resolved the issue.

PR #62, opened 2016-08-09
SPLICE-728 Increasing limit for the max number of columns

scottfines:
SPLICE-728: Making select column limit equal the table column limit. There isn't a test here because, well, it's a number change.

### DIFF
--- a/db-shared/src/main/java/com/splicemachine/db/iapi/reference/Limits.java
+++ b/db-shared/src/main/java/com/splicemachine/db/iapi/reference/Limits.java
@@ -31,12 +31,12 @@ public interface Limits
         * Various fixed Limits. DB2 related limits are prefixed with "DB2_".
         */
 
-	public static final int DB2_MAX_TRIGGER_RECURSION = 16; /* Maximum nesting level for triggers */
+	int DB2_MAX_TRIGGER_RECURSION = 16; /* Maximum nesting level for triggers */
 
 	/** Maximum number of indexes on a table */
-	public static final int DB2_MAX_INDEXES_ON_TABLE = 32767;
+	int DB2_MAX_INDEXES_ON_TABLE = 32767;
 	/* Maximum number of columns in a table */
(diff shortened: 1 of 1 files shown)


jfilali:
+1

jyuanca:
+1

db-shared/src/main/java/com/splicemachine/db/iapi/reference/Limits.java
```diff
@@ -31,12 +31,12 @@
...
 
-	public static final int DB2_MAX_TRIGGER_RECURSION = 16; /* Maximum nesting level for triggers */
+	int DB2_MAX_TRIGGER_RECURSION = 16; /* Maximum nesting level for triggers */
 
 	/** Maximum number of indexes on a table */
-	public static final int DB2_MAX_INDEXES_ON_TABLE = 32767;
+	int DB2_MAX_INDEXES_ON_TABLE = 32767;
 	/* Maximum number of columns in a table */
-	public static final int DB2_MAX_COLUMNS_IN_TABLE = 1012;
+	int DB2_MAX_COLUMNS_IN_TABLE =1<<17 ; //set to 131K
```
jeffcunningham: Was there some reason to do the bit shift here rather than the actual int?

ammolitor: why is it not `public static final` anymore?

scottfines:
LImits is a public interface. Any variables in a public interface are automatically public static final, so there is no need to include the extra text.

ammolitor:
@scott, thanks for the lesson :-) 

have we tried to create a table that big yet?

scottfines:
Yeah. I ran a test up to 75K columns (it got brutal really quickly, but it did work), and Erin ran a benchmark against 15K tables.

Scott

ammolitor:
so then we haven't tested 131072 columns yet?

genedavis:
Let's make sure we get the feedback resolved and get this in for the Wednesday cutoff.  Alliant is waiting...

PR #63, opened 2016-08-09
 SPLICE 741

jfilali:
Use Kryo pools in result streamer   stream listener
 Added Kryo Pool master
 This version added  ThreadLocal to avoid race condition with multiple handlers used with Netty.
 It should now use  different kryo objects by thread basis and then the kryo object will be return when the thread close and perform a clean of its ThreadLocal


jfilali:
wait for it, need to make change , thanks to Scott.

jfilali:
@scottfines please look at it now

jeffcunningham:
+1, nice!

genedavis:
+1

scottfines:
+1 from me. Let's merge this into 2.0 as well; can you create a second PR against branch-2.0 and assign it to me?

PR #66, opened 2016-08-10
SPLICE-812 Removing Dead SQL Package Wrappers

jleach4:
Removing Dead SQL Wrappers.  These cause warnings, etc. and we are not using it anymore.


jeffcunningham:
+1

jyuanca:
+1

PR #68, opened 2016-08-11
SPLICE-817 Using the old mechanism for MapR

jleach4:
Adding a check for mapr since it does not handle starting zoo and hbase directly.


genedavis:
+1

jfilali:
+1

PR #69, opened 2016-08-11
SPLICE-816: Correcting error in SparseBitIndex

scottfines:
It turns out that, in the (incredibly rare) situation where the bits in
a SparseBitIndex exactly line up with a byte boundary, SparseBitIndex
may end up over-estimating the number of bytes required by 1, which
results in a trailing 0. This trailing 0 can throw off deserialization
of that row, resulting in all kind of odd errors.

This ensures that the trailing zero is not present by dropping it if the
over-estimation happens, and adds a unit test to verify.


genedavis:
Are there any upgrade issues or is this just a silent fix for all existing customers/data?

jeffcunningham:
+1

scottfines:
it doesn't have impact unless you encounter the situation (the customer did), in which case you'll need to reload that data. No true "upgrade" is necessary, though.

PR #70, opened 2016-08-11
SPLICE-818 Code Cleanup: Hbase Patch now in all of our versions

jleach4:
Trying to get rid of the copied CoprocessorChannel code.  Scott take a look at this and see if I am missing something.


jeffcunningham:
+1

hbase_storage/hbase1.1.2/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
```diff
@@ -65,7 +66,11 @@ private HBaseConnectionFactory(SConfiguration configuration){
         this.namespace=configuration.getNamespace();
         this.namespaceBytes=Bytes.toBytes(namespace);
         try{
-            this.connection=ConnectionFactory.createConnection((Configuration) configuration.getConfigSource().unwrapDelegate());
+            Configuration config = (Configuration) configuration.getConfigSource().unwrapDelegate();
+            this.connection=ConnectionFactory.createConnection(config);
+            Configuration clonedConfig = new Configuration(config);
+            clonedConfig.setInt("hbase.client.retries.number",1);
```
scottfines: Should this be 1, or 0? i.e. does it reference only retries, or all tries?

jleach4: Good question.  0 blows chunks immediately and 1 seems to work as advertised.

Regards,
John

scottfines:
+1. Like the reduction in surface area

scottfines:
I'm happy to merge this over; was it also intended for 2.0? If so, please create a second PR so that we can merge it there as well.

PR #71, opened 2016-08-12
SPLICE-816: Correcting error in SparseBitIndex

scottfines:
2.0 version of the 3.0 pull request.

It turns out that, in the (incredibly rare) situation where the bits in
a SparseBitIndex exactly line up with a byte boundary, SparseBitIndex
may end up over-estimating the number of bytes required by 1, which
results in a trailing 0. This trailing 0 can throw off deserialization
of that row, resulting in all kind of odd errors.

This ensures that the trailing zero is not present by dropping it if the
over-estimation happens, and adds a unit test to verify.


jleach4:
+1

PR #75, opened 2016-08-12
DB-4457: incremental backup

jyuanca:
First checkin for incremental backup
1) As soon as an HFile is flushed to disk, record its name in backup medata directory by creating an empty file with the same name. The file indicates that an HFile with the same name in either archived directory or data directory should be included in next incremental backup
2) During incremental backup, for each region, look for HFile names in the backup medata directory, and copy HFiles from hbase data directory or archived directory.
3) Prevent an archived HFile from being purged if there is a file with the same name in backup metadata directory.

Also included are code cleanup, refactoring, backup system table cleanup. Please review code changes in both spliceengin and spliceengine-ee repositories.

### DIFF
--- a/hbase_sql/hbase1.0.0.x/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
+++ b/hbase_sql/hbase1.0.0.x/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
@@ -17,8 +17,16 @@ package com.splicemachine.hbase;
 
 import com.google.protobuf.Service;
 import com.splicemachine.access.HConfiguration;
+import com.splicemachine.backup.BackupRestoreConstants;
+import com.splicemachine.backup.BackupUtils;
 import com.splicemachine.coprocessor.SpliceMessage;
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.pipeline.Exceptions;
+import com.splicemachine.si.constants.SIConstants;
 import com.splicemachine.utils.SpliceLogUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.Coprocessor;
 import org.apache.hadoop.hbase.CoprocessorEnvironment;
 import org.apache.hadoop.hbase.coprocessor.CoprocessorService;
@@ -26,6 +34,7 @@ import org.apache.hadoop.hbase.coprocessor.ObserverContext;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
 import org.apache.hadoop.hbase.regionserver.*;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.FSUtils;
 import org.apache.hadoop.hbase.zookeeper.RecoverableZooKeeper;
 import org.apache.log4j.Logger;
 import org.apache.zookeeper.CreateMode;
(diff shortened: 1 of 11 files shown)


hbase_sql/hbase1.0.0.x/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
```diff
@@ -219,9 +247,74 @@ else if (Bytes.compareTo(status, HConfiguration.BACKUP_DONE) == 0) {
...
+            }
+            else if (BackupUtils.existsDatabaseBackup(fs, rootDir)) {
+                shouldRegister = true;
+            }
+            if (shouldRegister) {
+                registerHFile(fileName);
+            }
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
jeffcunningham: probably want to get rid of this printStackTrace()

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceHFileCleaner.java
```diff
@@ -0,0 +1,51 @@
+package com.splicemachine.hbase;
+
```
jeffcunningham: copyright notice...

splice_machine/src/main/java/com/splicemachine/backup/BackupRestoreConstants.java
```diff
@@ -0,0 +1,16 @@
+package com.splicemachine.backup;
+
```
jeffcunningham: copyright notice.

splice_machine/src/main/java/com/splicemachine/backup/BackupUtils.java
```diff
@@ -0,0 +1,57 @@
+package com.splicemachine.backup;
+
```
jeffcunningham: copyright notice.

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceHFileCleaner.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.master.cleaner.BaseHFileCleanerDelegate;
+import org.apache.hadoop.hbase.util.FSUtils;
+import org.apache.log4j.Logger;
+
+/**
+ * Created by jyuan on 1/29/15.
+ */
+public class SpliceHFileCleaner extends BaseHFileCleanerDelegate {
```
jeffcunningham: Seems like this could have gone into spliceengine-ee

jeffcunningham:
It seem like some of this code is backup specific enough to have gone into spliceengine-ee, no?

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceHFileCleaner.java
jyuanca: SpliceTestPlatformConfig needs it.

jeffcunningham: Why?

jyuanca:
Some backup code are here because BackupEndpointObserver requires them. Unless we can move  BackupEndpointObserver to spliceengine-ee, they will cause compiling errors.

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceHFileCleaner.java
jyuanca: It is needed to comfigure hbase.master.hfilecleaner.plugins parameter

jeffcunningham:
I see, and we can only move the endpoint observer if we relax the requirement that we don't require restart to enable enterprise backup.

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceHFileCleaner.java
jeffcunningham: This one I think can be moved and config changed in spliceengine-ee.  We have to be able to modify/override config there anyway.

jyuanca: Changing configuration requires a restart?

jeffcunningham: Not if the component being config'd is started after the config is modified.

PR #77, opened 2016-08-16
SPLICE-823 - update kafka versions

ammolitor:
Updating poms for correct Kafka versions for each platform, fixing a method call in the TestKafkaCluster related to the update.


jfilali:
+1

scottfines:
I'm not the moderator this week, but generally we should avoid having more than 1 commit per JIRA ticket (and 1 JIRA ticket per Pull Request). This will make it easier to revert changes if they cause problems, and will also simplify the acceptance process.

The change itself is fine, I would just advocate rebasing before merging.

ammolitor:
@scottfines you are right, I should know better. Will squash and re-submit.

PR #81, opened 2016-08-17
SPLICE-838: Making olap server tick time configurable

scottfines:
This makes the olap server tick time configurable, so that we can wait longer if we need to.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/olap/OlapServer.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/OlapServer.java
@@ -56,7 +56,9 @@ public class OlapServer {
         ServerBootstrap bootstrap = new ServerBootstrap(factory);
 
         // Instantiate handler once and share it
-        OlapJobRegistry registry = new MappedJobRegistry(1L,TimeUnit.SECONDS); //TODO -sf- make this tick time configurable
+        OlapJobRegistry registry = new MappedJobRegistry(config.getOlapClientTickTime(),
+                config.getOlapServerTickLimit(),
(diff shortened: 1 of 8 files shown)


hbase_sql/src/main/java/com/splicemachine/olap/OlapServer.java
```diff
@@ -56,7 +56,9 @@ public void startServer(SConfiguration config) {
         ServerBootstrap bootstrap = new ServerBootstrap(factory);
 
         // Instantiate handler once and share it
-        OlapJobRegistry registry = new MappedJobRegistry(1L,TimeUnit.SECONDS); //TODO -sf- make this tick time configurable
+        OlapJobRegistry registry = new MappedJobRegistry(config.getOlapClientTickTime(),
+                config.getOlapServerTickLimit(),
+                TimeUnit.MILLISECONDS);
```
jeffcunningham: Changed from seconds to milliseconds but default time stayed the same.  Is that what you wanted?

scottfines:
YIKES! good catch, I'll fix and re-push

scottfines:
oh wait, my bad. No, it actually IS intended--the configured setting is 1000 (1 second in milliseconds), but for some reason we weren't paying attention to the setting, so changing to Millis is actually the correct thing to do. Sorry, misread the comment to meaning that I *didn't* make the change.

jeffcunningham:
Ok, good.  +1

jfilali:
+1

jeffcunningham:
I'm gonna merge this branch unless someone screams.

PR #82, opened 2016-08-18
SPLICE-839 Set default handler for uncaught exceptions

dgomezferro:
Improve logging in case an uncaught exception is thrown


dgomezferro:
I'm not sure that's the best place to add this static piece of code, it should be something used in any Splice process (RegionServer, OlapServer, SparkExecutor...)

Would it be better to have an explicit call for each type of server on their startup sequence?

scottfines:
+1 I'm okay with it. It's not ideal, but I can't think of anyplace else that would be better.

jeffcunningham:
+1

jleach4:
+1

PR #84, opened 2016-08-18
SPLICE-836: Removing recursive cogroups.

scottfines:
The implementation of cogrouping was (accidentally?) recursive: it
created a wrapper Iterable around each pair of iterables, resulting in a
wrapper Iterable around each row on the left-hand side. If you get
enough such rows, you'll blow off with a StackOverflowError
(SPLICE-836).

This removes the recursive implementation and uses an iterative approach
instead by flattening out the excess Iterables and using just a single
while() loop. This should have the added benefit of reducing extraneous
object creation, reducing memory overhead and improving performance (in
a pretty small way, to be sure, but still).

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/function/AntiJoinRestrictionFlatMapFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/function/AntiJoinRestrictionFlatMapFunction.java
@@ -54,7 +54,7 @@ public class AntiJoinRestrictionFlatMapFunction<Op extends SpliceOperation> exte
             op.setCurrentRow(mergedRow);
             if (op.getRestriction().apply(mergedRow)) { // Has Row, abandon
                 operationContext.recordFilter();
-                return Collections.EMPTY_LIST;
+                return Collections.emptyList();
             }
         }
         // No Rows Matched...
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/function/CogroupAntiJoinRestrictionFlatMapFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/function/CogroupAntiJoinRestrictionFlatMapFunction.java
@@ -18,17 +18,20 @@ package com.splicemachine.derby.stream.function;
 import com.splicemachine.derby.iapi.sql.execute.SpliceOperation;
 import com.splicemachine.derby.impl.sql.execute.operations.LocatedRow;
 import com.splicemachine.derby.stream.iapi.OperationContext;
+import com.splicemachine.derby.stream.utils.ConcatenatedIterable;
 import org.sparkproject.guava.collect.Iterables;
 import org.sparkproject.guava.collect.Sets;
 import scala.Tuple2;
 import java.util.ArrayList;
+import java.util.LinkedList;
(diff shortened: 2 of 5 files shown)


genedavis:
Can we add an IT (or is there one already)?

scottfines:
There are a number of ITs which verify regression behavior, and preventing
a stackoverflow doesn't require an IT, simply analyzing the algorithm is
enough.

jeffcunningham:
Wut?!  You're shitting me?

scottfines:
Hey, don't forget you are in public, Jeff :)

Is there a difficulty, or objection?

jeffcunningham:
No, I think the code is good.  I just don't have the mental capacity to reliably reason about algorithms that deeply.  That's why I write ITs.
Besides that, +1

jfilali:
+1 ,  IT's are always useful but i can understand for that particular case.

scottfines:
In this case the analysis is far superior to an IT. To reliably force a
potential SO would require millions of rows, and as the JVM evolves the
number of rows required would change as well(most likely increasing), so
the test itself is inherently fragile.

On the other hand, analyzing this code block, we can clearly see that the
number of stack additions is finite(in fact just 2 additions: one for the
wrapping iterator.next() and one for the underlying iterator.next() call).
Since its finite, there is no risk in this causing a SO to occur.

Of course, that reasoning doesn't protect us from someone regressing
this change, but as I indicated above, an IT wouldn't really do that
either. The only thing to worry about in my mind is that I accidentally
screwed up something unrelated; the hope is that our existing ITs are
comprehensive enough to catch those.

jeffcunningham:
A plain unit test would have been easy to write for this.

genedavis:
I only found a couple of simple existing tests that specifically exercised this join strategy.  Didn't think an IT could/should simulate the stack overflow but wondered if the code changes might inspire to add more tests.

genedavis:
+1

jleach4:
Checkout Iterables.consumingIterable(), we can then get rid of new class (I believe)...

jeffcunningham:
Scott, do you think you can address John's comment by today's cutoff (25min?)

scottfines:
No. Testing alone will take longer than that time window.

Scott

genedavis:
Can you determine if John's suggestion is applicable in the next 20 min?  If it is not applicable, you are done and Jeff can merge.  But if it is applicable, the fix will have to go into the next Sprint

jeffcunningham:
John's comment does not regard testing, only replacing the custom class with a guava class.

jleach4:
John's comment is wrong since it does not work either.  It wraps in an object (i.e. stack)...

```
  public static <T> Iterable<T> consumingIterable(final Iterable<T> iterable) {
        if(iterable instanceof Queue) {
            return new FluentIterable() {
                public Iterator<T> iterator() {
                    return new Iterables.ConsumingQueueIterator((Queue)iterable, null);
                }
            };
        } else {
            Preconditions.checkNotNull(iterable);
            return new FluentIterable() {
                public Iterator<T> iterator() {
                    return Iterators.consumingIterator(iterable.iterator());
                }
            };
        }
    }
```

jleach4:
+1

PR #85, opened 2016-08-18
SPLICE-836: Removing recursive cogroups.

scottfines:
This is the 3.0-branch version of the same commit.

The implementation of cogrouping was (accidentally?) recursive: it
created a wrapper Iterable around each pair of iterables, resulting in a
wrapper Iterable around each row on the left-hand side. If you get
enough such rows, you'll blow off with a StackOverflowError
(SPLICE-836).

This removes the recursive implementation and uses an iterative approach
instead by flattening out the excess Iterables and using just a single
while() loop. This should have the added benefit of reducing extraneous
object creation, reducing memory overhead and improving performance (in
a pretty small way, to be sure, but still).


jeffcunningham:
+1

genedavis:
+1

PR #87, opened 2016-08-18
SPLICE-846 Remove Metrics Computation from Standard Scans

jleach4:
Scans should not compute size unless it is a statistics run.


jeffcunningham:
+1

PR #88, opened 2016-08-18
SPLICE-848 Fixing tests that used text as column name now that it is …

jleach4:
Fixing failing tests.  _Text_ is now a reserved keyword.  As a reminder, this is so Spark Integration works with create table foo (col1 text);


genedavis:
What is the "upgrade" for customers who have TEXT as a column name?  Alter table rename?  Should we have an IT?

genedavis:
Confirmed that the following works to rename the column if present - the key is to put double-quotes around "TEXT":

splice> RENAME COLUMN A."TEXT" to FOO;

PR #91, opened 2016-08-24
2.0 version of SPLICE-856 fix

scottfines:
This is the 2.0 version of SPLICE-856. 

Updates push multiple versions into a cell. When scanning, HBase
defaults to only scanning a single version of the cell. Thus, if you do
an update, you have to make sure that any subsequent scans are using
non-default hbase behavior (i.e. reading all versions). There is a
method on DataScan to provide this, but statistics collection was not
calling it. This resulted in statistics collections only reading the
updates, and not the underlying insertions (or past updates), instead of
constructing the entire row.

This corrects that issue, and adds an IT for verification. Note that the
IT would only fail in the HBase architecture, as the memory architecture
ignores version limits.


genedavis:
+1

jeffcunningham:
+1

jfilali:
+1

PR #92, opened 2016-08-24
SPLICE-857 Handling Versioned Lock Implementations

jleach4:
Tried to write an IT but I could not get Mockito to really run it through the Region Flow.  I think the challenge is that there is no no-arg constructor.  

This handles the hbase change from "waitFor boolean" in the lock logic to "readLock boolean."


genedavis:
+1

genedavis:
merge into both branches, right?

jfilali:
+1

PR #93, opened 2016-08-25
SPLICE-859 Having System Tables Compact Outside of Spark Since we Loo…

jleach4:
The UI lookup for the region location causes compaction to fail for the meta table since it cannot perform a meta lookup on itself.  This is a catastrophic failure during splits of meta.


mzweben:
+1 MZ and GD - Looks like all system tables will now use Hbase compaction

PR #94, opened 2016-08-25
SPLICE-859 Not Allowing Spark to Perform System table compaction: met…

jleach4:
Forgot all the versions of HBase and that the compactor code is copied.  Clearly needs to get cleaned up.


genedavis:
+1

mzweben:
+1

PR #101, opened 2016-08-29
SPLICE-875 Moving Master Branch to Spark 2.0

jleach4:
Mongo commit moving Splice Machine from 1.6.2 to Spark 2.0 without shading.


splice_machine/src/main/java/com/splicemachine/derby/stream/function/StreamFileFunction.java
```diff
@@ -57,9 +56,9 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
 
     @Override
-    public Iterable<LocatedRow> call(final InputStream s) throws Exception {
+    public Iterator<LocatedRow> call(final InputStream s) throws Exception {
         if (operationContext.isFailed())
-            return Collections.emptyList();
+            return Collections.<LocatedRow>emptyList().iterator();
         checkPreference();
 
         return new Iterable<LocatedRow>() {
```
dgomezferro: We can remove this Iterable, just return the iterator we are creating in line 69 directly.

db-engine/src/main/java/com/splicemachine/db/impl/ast/CorrelatedPushDown.java
```diff
@@ -26,15 +26,15 @@
...
-import com.google.common.base.Function;
+import org.spark_project.guava.base.Function;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.compile.Optimizable;
 import com.splicemachine.db.iapi.sql.compile.Visitable;
 import com.splicemachine.db.impl.sql.compile.*;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.log4j.Logger;
-import org.sparkproject.guava.collect.Collections2;
-import org.sparkproject.guava.collect.Iterables;
```
dgomezferro: I think it would be a better idea to use our own Guava (possibly shading it ourselves under com.splicemachine.guava) rather than depending on Spark's shaded guava:
- They could change their guava version or we might want to upgrade.
- Their use of guava is entirely internal, they don't expose Guava on their public API, and they could change it at any time.

What do you think?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/StreamFileFunction.java
jleach4: Nice catch.

db-engine/src/main/java/com/splicemachine/db/impl/ast/CorrelatedPushDown.java
jleach4: Good point.  Would you object to starting out with Spark's and then moving to our own a bit later?  I think it would just be a find and replace.

jeffcunningham:
+1 for pushing this upgrade and creating a bug for Guava versioning stuff.

dgomezferro:
+1

mzweben:
The new entry for Spark is SparkSession. Perhaps it would have better to make SpliceSpark have a reference to a SparkSession so that Stored Procedures can use the new 2.0 API such as createDataFrame. It is true that SparkContext API is backward compatible but not the right thing to use anymore. Looking to see if DataSet can be used without SparkSession.

SparkSession is now the new entry point of Spark that replaces the old SQLContext and HiveContext. Note that the old SQLContext and HiveContext are kept for backward compatibility. A new catalog interface is accessible from SparkSession - existing API on databases and tables access such as listTables, createExternalTable, dropTempView, cacheTable are moved here.

Dataset API and DataFrame API are unified. In Scala, DataFrame becomes a type alias for Dataset[Row], while Java API users must replace DataFrame with Dataset<Row>. Both the typed transformations (e.g. map, filter, and groupByKey) and untyped transformations (e.g. select and groupBy) are available on the Dataset class. Since compile-time type-safety in Python and R is not a language feature, the concept of Dataset does not apply to these languages’ APIs. Instead, DataFrame remains the primary programing abstraction, which is analogous to the single-node data frame notion in these languages.

PR #102, opened 2016-08-30
VTI Support for Orc File 

jramineni:
Closes SPLICE-879


splice_machine/pom.xml
```diff
@@ -23,6 +23,20 @@
         <version>3.0.0.3-SNAPSHOT</version>
     </parent>
     <dependencies>
+
```
scottfines: the splice_machine module is specifically designed to NOT have this dependency. Any code which required hadoop or hbase should be moved to the hbase_sql module.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+				throw new RuntimeException(cE);
+			}
+			throw new RuntimeException(e);
+		}
+
+	}
+
+	@Override
+	public Iterator<LocatedRow> iterator() {
+		return this;
```
scottfines: This violates to contract of the iterator() method in Iterable. Specifically, doing this:

`
for(LocatedRow lr: orcFileIterator){
    doSomething(lr);
}
for(LocatedRow lr: orcFileIterator){
    doSomethingElse(lr);
}
`

must work. As currently written, this is unlikely to work correctly.

The solution is to separate the Iterator and Iterable interface implementation, and making the iterator() method return a new Iterator with each invocation.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	/*
+	 * This formats the record in the ORC file and return as LocatedRow
+	 */
+	@Override
+	public LocatedRow next() {
+		try {
+			// Return Row Format
+			ExecRow returnRow = execRow.getClone();
```
scottfines: Why are you cloning the row here? This seems like a waste of object creation.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+			// Get the row
+			row = records.next(row);
+
+			// Get the row details
+			List value_lst = inspector.getStructFieldsDataAsList(row);
```
scottfines: Is there no Generic type information available here?

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+	/*
+	 * Returns Records in the next file
+	 */
+	public void getNextFileRecords() throws IOException {
+		// Close the current records iterator to release the resources for
+		// current file
+		records.close();
+
+		// Get the Record Reader for the next file
+		Reader reader = getReader(this.nextFile());
```
scottfines: Resource leak: You don't close the Reader after you finish reading it.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+		}
+		reader = OrcFile.createReader(filesystem, toProcessPath);
+
+		return reader;
+
+	}
+
+	@Override
+	public void remove() {
+		throw new RuntimeException("Not Implemented");
```
scottfines: Please throw new UnsupportedOperationException() instead, since we won't be supporting removal.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
```diff
@@ -0,0 +1,246 @@
...
+	private static final Logger LOG = Logger.getLogger(ORCRecordIterator.class);
+
+	// The format of the Record to be returned
+	private ExecRow execRow;
+
+	// This is contains the details of the ORC File Record format
+	private StructObjectInspector inspector;
+
+	// Records in Orc File to be processed
+	RecordReader records;
```
scottfines: Is there a reason this is package-scoped? We are letting a lot of state leak here.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/VTIOperationIT.java
```diff
@@ -182,6 +183,18 @@ public void testVTIConversion() throws Exception {
         }
         Assert.assertEquals(5, count);
     }
+    //
+    @Test
+    public void testORCFileVTI() throws Exception {
+        String location = getResourceDirectory()+"orcVtiFile.in";
+        String sql = String.format("select * from new com.splicemachine.derby.vti.SpliceORCFileVTI('%s') as b (id bigint,name varchar(128),active char(1), start_dt date, update_dt timestamp )", location);
+        ResultSet rs = spliceClassWatcher.executeQuery(sql);
```
scottfines: Resource leak: You aren't closing the ResultSet. In tests it's not overly critical (we have a few backups in case stuff doesn't close cleanly), but we still should follow good practices and close what we open.

scottfines:
-1 Overall. 

Specifically:

1. We very much do NOT want splice_machine to have ANY references to Hadoop or hbase; it is a core module allowing multiple architectures. This code should be located in hbase_sql (along with its IT); as it stands, I find it pretty unlikely that the mem profile will even compile with this.

2. The Iterable contract is not held correctly.

3. It's extremely minor, but the indentation and spacing seems to be messed up. SInce people tend to get annoyed at me when I do it, you probably should fix that and avoid their annoyance as well. :)

4. This needs to be rebased to a single commit before merging; without that, it's very difficult to revert the change if it breaks something.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
jramineni: This is done to support Multiple files (i.e. when directory is passed). In case of single file, the close was not in this file.

dgomezferro: This is an artifact of Spark's API that imposes returning Iterables where only Iterators where needed. We do this (implementing Iterable and Iterator on the same object) all over the place. It will go away when we move to Spark 2.0 (hopefully today).

dgomezferro:
> This needs to be rebased to a single commit before merging; without that, it's very difficult to revert the change if it breaks something.

I'm actually not a fan of rebasing always into one commit (although here it probably makes sense). It shouldn't be that hard to revert a merge: https://git-scm.com/blog/2010/03/02/undoing-merges.html

If that's our policy we should stick with it though, just wanted to raise my concerns.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
scottfines: Fair enough--although I still hold that it is not correct to do so, even if Spark gets away with it. In this particular case, calling iterator() twice seems unlikely to work.

scottfines:
@dgomezferro The reason I push for rebasing into a single commit is that it makes it easy to revert _features_, rather than individual commits. It also ensures that we don't have weird side-effect situations where we can't (for example) bisect into the middle of a series of commits because it's a broken state. With a single commit, each feature and bug is self-contained. One commit and one pull request per JIRA ticket means that we have one _patch_ per JIRA ticket.

I don't know that it was ever formally adopted, but I am a strong advocate for it personally, and it's been the de facto policy since we went OS, so at this point it's kind of solidified :)

jyuanca:
Jebril, wondering what is the status of this pull request. Have you reworked, pushed to another branch and merged to master? If so, then we should close this one.

splice_machine/src/main/java/com/splicemachine/derby/vti/ORCRecordIterator.java
jramineni: Without Clone, insert with spliceproperties (i.e. to import), resulted in inserting all records with values from last row. So Clone is required.

jleach4:
What is the status on this?

genedavis:
Have all change suggestions been incorporated?  Do we have enough +1's?

hbase_sql/src/main/java/com/splicemachine/derby/vti/ORCRecordReader.java
```diff
@@ -0,0 +1,250 @@
...
+		 */
+		@Override
+		public LocatedRow next() {
+			try {
+				// Return Row Format
+				// Changed for performance
+				//Reverting it back, since on import i.e. using insert with splice properties, inserted rows
+				// with all rows identical, the values of the last row.
+				 ExecRow returnRow = execRow.getClone();
+				//ExecRow returnRow = execRow;
```
dgomezferro: Cleanup this part (we only need to know why .getClone() is needed)

hbase_sql/src/main/java/com/splicemachine/derby/vti/SpliceORCFileVTI.java
```diff
@@ -0,0 +1,161 @@
...
+			filesystem = FileSystem.get(HConfiguration.unwrapDelegate());
+			// filesystem = FileSystem.get(new Configuration());
+
+			// Create the iterator to read the records in ORC files.
+			ORCRecordReader it;
+			it = new ORCRecordReader(filesystem, new Path(fileName), execRow);
+
+			// Changed to split Iterable and iterator, and close is implemented
+			// by Iterator
+			// op.registerCloseable(it);
```
dgomezferro: Why is this commented out?

hbase_sql/src/test/java/com/splicemachine/derby/vti/SpliceORCFileVTIT.java
```diff
@@ -0,0 +1,85 @@
...
+		/*
+		 * new TableCreator(conn) .withCreate(
+		 * "create table user_orc (id BIGINT, name VARCHAR(100), activity char(1), is_fte BOOLEAN,"
+		 * +
+		 * " test_tinyint  SMALLINT, test_smallint  SMALLINT, test_int  INT, test_float  FLOAT,"
+		 * +
+		 * " test_double  DOUBLE, test_decimal DECIMAL(10,0), role      VARCHAR(64), "
+		 * + " salary    DECIMAL(8,2), START_DT  DATE, UPDATE_DT TIMESTAMP)")
+		 * .create();
+		 */
```
dgomezferro: Remove

hbase_sql/src/main/java/com/splicemachine/derby/vti/ORCRecordReader.java
```diff
@@ -0,0 +1,250 @@
...
+import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector;
+import org.apache.log4j.Logger;
+
+import com.splicemachine.db.iapi.sql.execute.ExecRow;
+import com.splicemachine.derby.impl.sql.execute.operations.LocatedRow;
+
+/**
+ * This is iterator for the records in the ORC files. It supports either a
+ * single file or Files in a folder. In case of file the next() will return the
+ * rows from one file, and then move to the next file
```
dgomezferro: It'd be better to create a split per file and read them in parallel, it can be done after merging this is though.

jleach4:
please re-submit.

PR #103, opened 2016-08-30
SPLICE-868: Correcting Race condition around splits.

scottfines:
It turns out that a Region is not to be considered "available" until it
is added to the HBase RegionServerServices, instead of when a
coprocessor start() is called. For most operations this doesn't matter,
because RegionServerServices is where the HBase network layer determines
if the Region is present or not; if you sent an operation to that region
and it wasn't in RegionServerServices, you'd get a
NotServingRegionException instead of allowing through an operation.

However, in the Write Pipeline we send writes to multiple regions
simultaneously in a single network call (if there's a network call at
all--in some cases we send the data directly to the same JVM where we
are writing data). Therefore, we allow the possibility of bypassing the
network stack and its protections.

When operating outside the stack, there is a race condition around
splits where a daughter region starts all of its coprocessors up before
it is finished fully initializing (in fact, start() is called on all
coprocessors BEFORE preOpen() is called on any RegionObserver). This
means that we were allowing some writes to proceed to a daughter region
before it was fully opening, leading to a race condition and (in turn)
causing the abort sequence to occur.

This corrects that issue by requiring that the region be present in the
RegionServerServices before allowing a BulkWrite to proceed.

The awkward raciness of this prevents a decent IT; thankfully, Jenkins
will fail periodically with an ABORT message during builds if this
regresses, which will serve as a decent IT (since it breaks the build).

### DIFF
--- a/hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexEndpoint.java
+++ b/hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexEndpoint.java
@@ -42,8 +42,10 @@ import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.hadoop.hbase.Coprocessor;
 import org.apache.hadoop.hbase.CoprocessorEnvironment;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.coprocessor.CoprocessorService;
 import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.coprocessor.RegionCoprocessorEnvironment;
+import org.apache.hadoop.hbase.regionserver.RegionServerServices;
 import org.apache.hadoop.util.StringUtils;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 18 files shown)


jfilali:
Impressive +1

dgomezferro:
+1,

hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexEndpoint.java
```diff
@@ -168,12 +169,12 @@ public void stop(CoprocessorEnvironment env) throws IOException{
...
+//    @Override
     public BulkWritesResult bulkWrite(BulkWrites bulkWrites) throws IOException{
         return pipelineWriter.bulkWrite(bulkWrites);
     }
 
-    @Override
+//    @Override
     public byte[] bulkWrites(byte[] bulkWriteBytes) throws IOException{
         assert bulkWriteBytes!=null;
         BulkWrites bulkWrites=compressor.decompress(bulkWriteBytes,BulkWrites.class);
```
jyuanca: remove Override instead of commeting out

jyuanca:
+1

PR #104, opened 2016-08-31
SPLICE-883 Don't block preClose() in case of abort

dgomezferro:
When closing regions due to an abort we shouldn't block waiting for scanners to finish, since they might never be able to complete.


jeffcunningham:
+1

jyuanca:
+1

jleach4:
+1

jfilali:
+1

PR #105, opened 2016-08-31
SPLICE-771: ddl dump utility. 

jeffcunningham:
There's a lot of code here, only some of which I wrote.  Best place to start would be with the ITs.
Thanks.

Spliceified by adding splice platform classes. They extend Derby platform classes.
Schema object added to model.
SQL being generated, including create schema with owner (authorizationId).
Fixed view DDL SQL generation.
Round trip IT testing; create schema(s), interrogate mode, build schema and compare with created, drop schema.
Round trip (dump ddl, export model, drop model in DB, create ddl, import model, verify) successful on small model.
Created Foreign Key round trip test but seeing error in setup; cannot create FK.
Added User, Role, Permission objects to database model.  Create, drop for database users, roles, permissions with the following caveats:
  There's no way to get a user's password from a JDBC client.  May have to open up a hole (and encrypt passwords so that they can be queried).
  Role/Permission grantor is not being preserved.  To do so we would have to connect "as" the grantor and issue the grant statement, but there's no way to get the grantor's password.
  Not yet handling column permissions.
  Not yet handling function or procedure permissions.
Integrated ddlutils into splice build.

Much more testing/fixing needs to be done.
Not getting all required entities.
Test coverage lacking.
Need to pull in DDLUtils unit tests but it's a large job after introducing Schema to model.


jeffcunningham:
Me amigos, have a look.  Let's get this merged.

dgomezferro:
Could we move the review to Crucible? Github isn't behaving well for me with such a big change.

jeffcunningham:
Created a [Crucible Review] (https://codereview.splicedev.com:8443/cru/CR-665) for this as Daniel suggested.

jyuanca:
Guys please review and comment at here(https://codereview.splicedev.com:8443/cru/CR-665)

PR #107, opened 2016-09-02
SPLICE-891: Correcting early cancellation race condition.

scottfines:
This corrects the PlaceholderTasks' early cancellation race condition. See SPLICE-891 for more details about why things are done the way they are.

This occurs in the ITs for CDH5.7.2, so no additional tests are necessary (even if it were possible as written).


jleach4:
+1

jfilali:
+1

dgomezferro:
+1

PR #113, opened 2016-09-06
SPLICE-892: Correcting NPE during boot compactions.

scottfines:
If the EngineDriver has not yet booted, then we can't use Spark for
compactions (because we don't know where the OlapServer is). This
ensures that we default back to HBase's default compaction strategy when
a compaction is requested during this time, rather than failing.


jyuanca:
+1

jleach4:
+1

dgomezferro:
+1

PR #114, opened 2016-09-06
SPLICE-886 Don't wait indefinetely for subpartitions

dgomezferro:
Make sure we don't retry indefinitely looking for subpartitions and raise an exception if we ever construct a Scan that doesn't make sense in SMRecordReaderImpl.


hbase_sql/src/main/java/com/splicemachine/mrio/api/core/SMRecordReaderImpl.java
```diff
@@ -202,7 +207,9 @@ public void restart(byte[] firstRow) throws IOException {
 			SplitRegionScanner srs = new SplitRegionScanner(scan,
 					htable,
 					clock,
-                    clientPartition);
+                    clientPartition,
+					driver.getConfiguration()
```
jfilali: not important, just a weird tab space there,

jfilali:
+1

genedavis:
+1

jyuanca:
+1

PR #115, opened 2016-09-07
SPLICE-886 Don't wait indefinetely for subpartitions

dgomezferro:
Make sure we don't retry indefinitely looking for subpartitions and raise an exception if we ever construct a Scan that doesn't make sense in SMRecordReaderImpl.

Fixed other platforms.


jleach4:
+1

jfilali:
+1

PR #117, opened 2016-09-07
SPLICE-838 Increase OlapClient timeout

dgomezferro:
Add logging when cancelling an Olap job

I've increased the Phi failure detector's threshold since 0.1 seems to be too conservative. 
As taken from http://doc.akka.io/docs/akka/snapshot/scala/remoting.html and http://fubica.lsd.ufcg.edu.br/hp/cursos/cfsc/papers/hayashibara04theaccrual.pdf a more reasonable threshold seems to be 10. We can revisit this based on the logged information when cancelling running jobs.


jleach4:
+1

jfilali:
+1

jyuanca:
+1

PR #120, opened 2016-09-07
SPLICE-909 Remove wrong preconditions

dgomezferro:
This fixes some test failures on the recent builds. The preconditions were wrong, sometimes we build scans with a startRow greater than a stopRow, for instance if we do

`create table a ( i int primary key);
select * from a where i > 3 and i < 1;`

We can't raise an exception in that case.


genedavis:
+1

jfilali:
+1

jyuanca:
+1

PR #127, opened 2016-09-09
SPLICE-915 Don't expect read() to return more than 1 byte

dgomezferro:
SQLClob.readExternal still uses read(), complicated code path

Should I try to fix SQLClob.readExternal too? I'm not so sure about the invariants there, or whether that method is used at all. If we assume the header will always have a 10.5 format we can just read 5 bytes and be done with it.


jleach4:
+1...  I think you will have to fix ReadExternal as well.

ryanwoodsmall:
+1

PR #131, opened 2016-09-09
SPLICE-919 Remove Object Creation from Join Operation

jleach4:
Little nit, removing object creation on a per row basis.


jfilali:
+1

ryanwoodsmall:
+1

PR #134, opened 2016-09-11
SPLICE-924 Fixing Ternary Operator Node so it chooses it's variant ty…

jleach4:
Using Variants correctly for TernaryOperatorNodes so qualifiers do not have to be recomputed when they are in fact constants.  

This impacts query 1 in TPCH significantly.

l_shipdate <= date({fn TIMESTAMPADD(SQL_TSI_DAY, -90, cast('1998-12-01 00:00:00'


ammolitor:
:+1:

jyuanca:
+1

PR #137, opened 2016-09-12
SPLICE-897 Limit cutpoints when scans are unbounded

dgomezferro:
The important change is line 95, previously if we had an unbounded scan we wouldn't bound the cutpoint search, meaning we could add cutpoints outside the region in case of a recent split, for instance.


genedavis:
+1

jyuanca:
+1

PR #138, opened 2016-09-12
SPLICE-915 Fix SQLCLob.readExternal assumption of .read()

dgomezferro:
As requested per John, fix also SQLClob.readExternal assumptions about .read() returning all needed bytes.


jyuanca:
+1

jleach4:
+1

PR #140, opened 2016-09-12
DB-5633: fix timestamp column for system tables

jyuanca:
Timestamp column for system tables is encoded by V1 serializer, but is decoded by v2 serializer. In general, you get incorrect values for a timestamp column if you do a select on the column. Fix this problem by using a V1 serializer to decode system tables.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/BaseDataDictionary.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/BaseDataDictionary.java
@@ -52,6 +52,9 @@ public abstract class BaseDataDictionary implements DataDictionary, ModuleContro
 	protected static final int		SYSCOLUMNS_CORE_NUM = 2;
 	protected	static final int		SYSSCHEMAS_CORE_NUM = 3;
 	protected static final int        NUM_CORE = 4;
+	// NOTE: JC - When adding a new sys table, this
+	// number will need to be increased in BOTH places.
+	public static final long FIRST_USER_TABLE_NUMBER = 1458;
 	/**
 	* SYSFUN functions. Table of functions that automatically appear
 	* in the SYSFUN schema. These functions are resolved to directly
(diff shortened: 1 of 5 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/BaseDataDictionary.java
```diff
@@ -52,6 +52,9 @@
 	protected static final int		SYSCOLUMNS_CORE_NUM = 2;
 	protected	static final int		SYSSCHEMAS_CORE_NUM = 3;
 	protected static final int        NUM_CORE = 4;
+	// NOTE: JC - When adding a new sys table, this
+	// number will need to be increased in BOTH places.
+	public static final long FIRST_USER_TABLE_NUMBER = 1458;
```
dgomezferro: This constant is also defined in DataDictionary, I think we should use just that one and remove this.

db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
```diff
@@ -1942,5 +1942,6 @@ The message identifier (ie. the value that matches a field in this class)
 	String BACKUP_CANCELED                                         = "BR011";
 	String BACKUP_DOESNOT_EXIST                                    = "BR012";
 	String PARENT_BACKUP_MISSING                                   = "BR013";
+	String EXISTS_CONCURRENT_BACKUP                                = "BR014";
```
dgomezferro: Isn't this change unrelated?

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -8841,6 +8841,11 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
                <arg>backupId</arg>
                <arg>parentBackupId</arg>
            </msg>
+           <msg>
+               <name>BR014</name>
+               <text>A concurrent {0} backup is running.</text>
+               <arg>backupType</arg>
+           </msg>
```
dgomezferro: Unrelated.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/BaseDataDictionary.java
jyuanca: Yes, I did not find that. Thanks!

db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
jyuanca: A code clean up.

PR #143, opened 2016-09-13
DB-5633: fix timestamp column for system tables

jyuanca:
Timestamp column for system tables is encoded by V1 serializer, but is decoded by v2 serializer. In general, you get incorrect values for a timestamp column if you do a select on the column. Fix this problem by using a V1 serializer to decode system tables.


dgomezferro:
+1

ryanwoodsmall:
+1

PR #144, opened 2016-09-13
SPLICE-869 Make sure operation context is properly setup on index job

dgomezferro:
Spark was failing to access Accumulators because the SparkOperationContext wasn't being properly set up.


jleach4:
+1

jyuanca:
+1

PR #146, opened 2016-09-13
 SPLICE-932 Make WholeTextInputFormat never split a file

dgomezferro:
Files processed by WholeTextInputFormat should be handled as a whole, we can't split them into blocks.


scottfines:
+1

jyuanca:
+1

PR #148, opened 2016-09-14
SPLICE-934 Remove exceptions from PoolSlotBooker

dgomezferro:
Since we are cancelling the job anyway it goes directly to the Failed jobs list, no need to raise exceptions and pollute the logs.


jyuanca:
+1

jleach4:
+1

PR #149, opened 2016-09-14
SPLICE-858: batch nested loop join

jyuanca:
This is a performance improvement for nested loop join. For a batch (10 by default) of left rows, submit tasks to return table scans for inner table.

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
@@ -200,6 +200,9 @@ public class SQLConfiguration implements ConfigurationDefault {
     public static final String PARTITIONSERVER_JMX_PORT = "hbase.regionserver.jmx.port";
     private static final int DEFAULT_PARTITIONSERVER_JMX_PORT = 10102;
 
+    public static final String NESTEDLOOPJOIN_BATCH_SIZE = "splice.nestedLoopJoin.batchSize";
+    private static final int DEFAULT_NESTEDLOOPJOIN_BATCH_SIZE = 10;
+
     @Override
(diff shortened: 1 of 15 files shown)


splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
```diff
@@ -200,6 +200,9 @@
     public static final String PARTITIONSERVER_JMX_PORT = "hbase.regionserver.jmx.port";
     private static final int DEFAULT_PARTITIONSERVER_JMX_PORT = 10102;
 
+    public static final String NESTEDLOOPJOIN_BATCH_SIZE = "splice.nestedLoopJoin.batchSize";
+    private static final int DEFAULT_NESTEDLOOPJOIN_BATCH_SIZE = 10;
```
scottfines: we didn't want to make this a URL hint, so that we could modify it on a per-query basis?

genedavis:
+1

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJInnerJoinFunction.java
```diff
@@ -122,6 +172,105 @@ public OperationContext getOperationContext() {
...
+            try {
+                ExecRow row = from.getCurrentRow();
+                RowLocation location = from.getCurrentRowLocation();
+                to.setCurrentRow(row);
+                to.setCurrentRowLocation(location);
+            }
+            catch (Exception e) {
+                return;
+            }
+            List<SpliceOperation> fromSubOperations = from.getSubOperations();
```
dgomezferro: Jun mentioned the recursive copy is needed for RowID, a comment here would be good.

PR #153, opened 2016-09-14
SPLICE-937 Tungsten for Sort and Distinct

jleach4:
Initial Implementation of Data Frames for Distinct, BroadcastJoin, and MergeSortJoin.

### DIFF
--- a/db-engine/pom.xml
+++ b/db-engine/pom.xml
@@ -159,7 +159,7 @@
                     ===================================================================================================== -->
                     <execution>
                         <id>generate-class-size-catalog-source</id>
-                        <phase>prepare-package</phase>
+                        <phase>process-classes</phase>
                         <goals>
                             <goal>java</goal>
                         </goals>
@@ -225,7 +225,7 @@
                 <executions>
                     <execution>
                         <id>compile-class-size-catalog</id>
-                        <phase>prepare-package</phase>
+                        <phase>process-classes</phase>
                         <goals>
                             <goal>compile</goal>
                         </goals>
@@ -237,6 +237,19 @@
                     </execution>
                 </executions>
             </plugin>
+             <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <version>2.19.1</version>
+                        <configuration>
+                            <additionalClasspathElements>
+                                <additionalClasspathElement>path/to/additional/resources</additionalClasspathElement>
+                                <additionalClasspathElement>path/to/additional/jar</additionalClasspathElement>
+                                <additionalClasspathElement>path/to/csv/jar1, path/to/csv/jar2</additionalClasspathElement>
+                            </additionalClasspathElements>
+                        </configuration>
+                    </plugin>
+
         </plugins>
     </build>
 </project>
--- a/db-engine/pom.xml
+++ b/db-engine/pom.xml
@@ -237,19 +237,6 @@
                     </execution>
                 </executions>
             </plugin>
-             <plugin>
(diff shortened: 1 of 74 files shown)


genedavis:
ValueRow.java, line 380 - no problem if size of row < i (since that is possible with the size computation from line  378)?

genedavis:
ValueRow.java - did you want all of those System.out.printlns?

db-engine/pom.xml
```diff
@@ -237,6 +237,19 @@
                     </execution>
                 </executions>
             </plugin>
+             <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <version>2.19.1</version>
+                        <configuration>
+                            <additionalClasspathElements>
+                                <additionalClasspathElement>path/to/additional/resources</additionalClasspathElement>
```
dgomezferro: Is this needed?

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDate.java
```diff
@@ -79,6 +82,13 @@
 	 +     * The JodaTime has problems with all the years before 1884
 	 +     */
 
+	public static final ThreadLocal<GregorianCalendar> GREGORIAN_CALENDAR =
+			new ThreadLocal<GregorianCalendar>() {
+				@Override
+				protected GregorianCalendar initialValue() {
+					return new GregorianCalendar();
+				}
```
dgomezferro: I think we should .clear() the instance in ThreadLocal.get(), otherwise some fields could be polluted from previous usages.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -113,10 +120,16 @@ public SparkDataSet(JavaRDD<V> rdd, String rddname) {
     public DataSet<V> distinct(String name, boolean isLast, OperationContext context, boolean pushScope, String scopeDetail) {
         pushScopeIfNeeded(context, pushScope, scopeDetail);
         try {
-            JavaRDD rdd1 = rdd.distinct();
-            rdd1.setName(name /* MapPartitionsRDD */);
-            RDDUtils.setAncestorRDDNames(rdd1, 2, new String[]{"Shuffle Data" /* ShuffledRDD */, "Prepare To Find Distinct" /* MapPartitionsRDD */}, null);
+            final ValueRow rowDefinition = (ValueRow) context.getOperation().getExecRowDefinition();
+            JavaRDD rdd1 = SpliceSpark.getSession().createDataFrame(rdd.map(new LocatedRowToRowFunction()),
+                    context.getOperation().getExecRowDefinition().schema())
```
dgomezferro: Is it possible to retain the UI names?

db-engine/pom.xml
jleach4: Yes unfortunately, I want my DataValueDescriptor tests to run in db-engine but since they are dependent on ClassSize codegen it is a bit more involved.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
jleach4: Yes but it is currently a bit of a mess on the UI side.  I need to re-evaluate.

db-engine/pom.xml
dgomezferro: What I'm saying is additionalClasspathElement should receive an path, and "path/to/additional/resources" or "path/to/additional/jar" don't look like real paths.

PR #154, opened 2016-09-15
Fixed SPLICE-742

jfilali:
In EntryPredicateFilter.java; line 79  we remove 1 to the limit.
When we build the Table Hash Key in a join , we end up with a hash key desalign between the left operand and the right operand.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
@@ -1955,16 +1955,16 @@ public class SQLChar
         {
             if (!(this instanceof SQLVarchar))
             {
-                StringBuffer    strbuf;
+                StringBuilder    strBuilder;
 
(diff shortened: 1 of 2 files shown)


jyuanca:
+1

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
```diff
@@ -1959,7 +1959,7 @@ public void setWidth(int desiredWidth,
 
                 strbuf = new StringBuffer(getString());
     
-                for ( ; sourceWidth < desiredWidth; sourceWidth++)
```
jleach4: Can we remove StringBuffer and replace with StringBuilder?

genedavis:
+1 - John can help fix afterwards

jfilali:
@jyuanca  done, please merge,

dgomezferro:
+1

PR #158, opened 2016-09-15
SPLICE-942 SmallScanners bug in HBase where they return extra row

jleach4:
This is causing SelfInsertIT to add an extra row.  This could hit us on Batch NLJ...


jyuanca:
+1

genedavis:
+1

ryanwoodsmall:
+1

PR #162, opened 2016-09-19
SPLICE-428 Support for With Clause Array above a select

jleach4:
This covers with clause support for analytical queries.  It supports an array of with clauses prior to the query and under the covers it behaves like a view.  I went with the view decision (Inline) since that is the default of most competitors (Oracle, Vertica).


jyuanca:
+1

genedavis:
+1 after header added to WithStatementIT.java

dgomezferro:
+1
merged manually due to conflict

PR #167, opened 2016-09-19
DB-5636

jyuanca:
This fix is for 1.5.x only.

When calculating grouped distinct and non-distinct aggregation in one query, a row is emitted from non-distinct buffer in first phase. The columns that are involved in distinct aggregation in this row should be null out. Otherwise, they may be accidentally impact distinct aggregation.


jleach4:
+1

genedavis:
+1

dgomezferro:
+1

PR #169, opened 2016-09-19
SPLICE-950 Make activation classes eligible for GC

dgomezferro:
Invalidate statements when pushed out of the statement cache and store
only weak references on the DependencyManager

I've tested this running a script similar to what Gene uploaded concurrently with an IT run. A test for it will be added to the QA runs.

Tested it against JDK 7 & 8.


jleach4:
+1

jyuanca:
+1

genedavis:
+1

genedavis:
need version for branch-2.0 and make sure it works on Java 7

PR #170, opened 2016-09-19
SPLICE-950 Make activation classes eligible for GC

dgomezferro:
Invalidate statements when pushed out of the statement cache and store
only weak references on the DependencyManager

Tested against JDK 7.


jyuanca:
+1

jleach4:
+1

genedavis:
+1

PR #175, opened 2016-09-21
SPLICE-954: fix an issue with BatchOnceOperation

jyuanca:
BatchOnceOperation pulls a batch of rows to a buffer until the buffer is full (by default 50000). There is a bug that caused the operation to lose one row for each batch, because it pulls a row first and then check whether the buffer is full. If the buffer is full, the row that it just pulled will be ignored. 

The fix is very straightforward: check whether the buffer is full first. If not full, pull a row and buffer it.


jleach4:
+1 Nice one June

genedavis:
+1 - this is not an issue in later releases?

jyuanca:
Not an issue for 2.0 or 3.0, because an iterator is created for a row source. The availability of next row can be checked without consuming it.

PR #176, opened 2016-09-21
SPLICE-960 close parent scanner

dgomezferro:
We were leaking resources in StoreScanner, which MemStoreFlushAwareScanner inherits from. We must close it properly.


jleach4:
+1...  Thanks for cleaning this up.

jyuanca:
+1

PR #185, opened 2016-09-27
SPLICE-977 Cleanup Segment ID Leftover from Page Implementation

jleach4:
Removing Segment ID and Container Key concepts from Conglomerates.  This was an old cleanup today in a branch that I finally finished.


jyuanca:
+1

PR #190, opened 2016-09-28
SPLICE-984 : Implement Except and Intersect with DataSet.

jfilali:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -243,14 +242,91 @@ public long count() {
...
+            Dataset<Row> other = SpliceSpark.getSession()
+                    .createDataFrame(
+                            ((SparkDataSet)dataSet).rdd
+                                   .map(new LocatedRowToRowFunction()),
+                            context.getOperation()
+                                   .getExecRowDefinition()
+                                   .schema());
+
+            //Do the intesect
+            Dataset<Row> result = that.intersect(other);
```
snowspeeder: can we keep dataset names as left and right instead of that and other

start-splice-cluster
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+#!/bin/bash
```
snowspeeder: line repeated again?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -243,14 +242,91 @@ public long count() {
         }
     }
 
+    @SuppressWarnings({ "unchecked", "rawtypes" })
     @Override
     public DataSet< V> intersect(DataSet< V> dataSet) {
-        return new SparkDataSet<>(rdd.intersection(((SparkDataSet<V>) dataSet).rdd));
+        return subtract(dataSet,"Intersect Operator",null,false,null);
+    }
```
jyuanca: Use subtract to implement intersect?

jyuanca:
Add ITs to execute on Spark and control mode.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
jfilali: Good catch thanks, i will as ITs

jfilali:
@snowspeeder  @mzweben @jyuanca , thanks,  done, i have updated the change you have requested, i will merge it now

PR #192, opened 2016-09-30
SPLICE-993 Make shared INSTANCE fields volatile

dgomezferro:
I got an NPE while accessing these fields a couple of times during the boot process. We might be accessing the field before initializing it, but if we are sharing it across threads it has to be volatile or have a synchronized access, I opted for the former.


jyuanca:
+1

jfilali:
+1

PR #193, opened 2016-09-30

jyuanca:
We may pass a timestamp format that's not compatible with SQLTimestamp to Timestamp.ValueOf() to parse. If that happens, we should catch the exception and parse using Joda DateTimeFormatter.


genedavis:
Per my email - we lose microseconds if we do this

PR #194, opened 2016-09-30

jyuanca:
For a timestamp that contains nanosecond, but is not compatible with SQLTimestamp format, we can parse timestamp using Joda DateTimeFormatter for millisecond precision. Then obtain nanosecond from timestamp string and set nanosecond.


genedavis:
+1

dgomezferro:
Please add a test.

PR #195, opened 2016-09-30

jyuanca:
For a timestamp that contains nanosecond, but is not compatible with SQLTimestamp format, we can parse timestamp using Joda DateTimeFormatter for millisecond precision. Then obtain nanosecond from timestamp string and set nanosecond. Backport the fix to 2.0


jfilali:
+1 i guess there is no other way except replacing all the dot by semi columns

snowspeeder:
btw, would be nice to add a unit test for this later

genedavis:
+1

PR #197, opened 2016-10-04
SPLICE-983: Carrying defaults through imports.

scottfines:
It turns out, Imports are really just inserts over subselects; to
support default value during the import, we can just re-write the import
SQL clauses to include a case statement which replaces null entries with
default values.

However, in order to support this case statement effectively, we also
need to make sure that we properly treat NULL strings within the parsing
stage. Per customer request, we treat null, NULL, nUll, etc. as actually
being null; we treat "null" as the string 'null'.

An IT was added to confirm overall behavior.


genedavis:
+1

jyuanca:
+1

jfilali:
+1

PR #198, opened 2016-10-05
SPLICE-983: Carrying defaults through imports.

scottfines:
2.0 version of SPLICE-983 fix.

It turns out, Imports are really just inserts over subselects; to
support default value during the import, we can just re-write the import
SQL clauses to include a case statement which replaces null entries with
default values.

However, in order to support this case statement effectively, we also
need to make sure that we properly treat NULL strings within the parsing
stage. Per customer request, we treat null, NULL, nUll, etc. as actually
being null; we treat "null" as the string 'null'.

An IT was added to confirm overall behavior.

Conflicts:
	splice_machine/src/main/java/com/splicemachine/derby/stream/function/FileFunction.java
	splice_machine/src/main/java/com/splicemachine/derby/stream/function/StreamFileFunction.java

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
@@ -25,9 +25,18 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
+import com.google.common.base.Predicate;
+import com.google.common.collect.Collections2;
+import com.google.common.collect.Lists;
+import com.splicemachine.db.catalog.DefaultInfo;
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
+import com.splicemachine.db.iapi.sql.dictionary.*;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.log4j.Logger;
 
@@ -52,6 +61,8 @@ import com.splicemachine.derby.utils.EngineUtils;
 import com.splicemachine.derby.utils.SpliceAdmin;
 import com.splicemachine.pipeline.ErrorState;
 import com.splicemachine.utils.SpliceLogUtils;
+
+import javax.annotation.Nullable;
 
 /**
  * Imports a delimiter-separated file located in HDFS in a parallel way.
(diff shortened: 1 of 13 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
```diff
@@ -407,6 +420,65 @@ private static void doImport(String schemaName,
...
+            SchemaDescriptor sd = dd.getSchemaDescriptor(schemaName,lcc.getTransactionExecute(),true);
+            assert sd!=null: "Programmer error: schema is not found!";
+            TableDescriptor td = dd.getTableDescriptor(tableName,sd,lcc.getTransactionExecute());
+            assert td!=null: "Programmer error: table is not found!";
+
+            ColumnDescriptorList columnDescriptorList=td.getColumnDescriptorList();
+            if(insertColumnList!=null){
+                boolean isFirst = true;
+                for(String col:insertColumnList){
+                    if(isFirst) isFirst = false;
```
snowspeeder: where is it using isFirst flag, bit confusing

splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
```diff
@@ -180,6 +188,14 @@ void checkPreference() {
         }
     }
 
+    /* ****************************************************************************************************************/
+    /*private helper methods*/
+    @SuppressWarnings("SimplifiableIfStatement") //the logic is clearer this way, without a performance penalty
+    private boolean shouldBeNull(String value,boolean wasQuoted){
```
snowspeeder: add static?

genedavis:
+1

jyuanca:
+1

jfilali:
+1

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -0,0 +1,282 @@
...
+
+        // read a line (ignoring empty lines/comments if necessary)
+        String line;
+        do{
+            line=readLine();
+            if(line==null){
+                return false; // EOF
+            }
+        }
+        while(ignoreEmptyLines && line.length()==0 || (commentMatcher!=null && commentMatcher.isComment(line)));
```
snowspeeder: is it possible that line may contain other white spaces, or it needs trim before length check?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -0,0 +1,282 @@
...
+                        appendSpaces(currentColumn,potentialSpaces);
+                    }
+                    columns.add(currentColumn.length()>0?currentColumn.toString():null); // "" -> null
+                    potentialSpaces=0;
+                    currentColumn.setLength(0);
+                    if(recordQuotes){
+                        quotedColumns.append(wasQuoted);
+                    }
+                    wasQuoted = false;
+                }else if(c==SPACE){
```
snowspeeder: how about other whitespace chars, \t,\r etc

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -0,0 +1,282 @@
...
+                    if(!surroundingSpacesNeedQuotes || currentColumn.length()>0){
+                        appendSpaces(currentColumn,potentialSpaces);
+                    }
+
+                    potentialSpaces=0;
+                    currentColumn.append(c);
+                }
+
+            }else{
+				/*
```
snowspeeder: nice to fix mis-alignments in comments

PR #199, opened 2016-10-06
SPLICE-1034 Removing expensive checks when possible, incremental move

jleach4:
Small optimization that removes a few seconds from TPCH1, query 1.  

aggregator.isInitialized(?) is a bit slow especially when the query has a decent number of aggregates.


jyuanca:
+1

ryanwoodsmall:
+1

PR #201, opened 2016-10-07
SPLICE-1035 Re-using Kryo via a Pool mechanism.  This effects both ex…

jleach4:
Making sure we re-use Kryo objects.  We were currently creating the Kryo objects for each execution on the OLAP server and on the executors.


jyuanca:
+1

ammolitor:
+1

ryanwoodsmall:
+1

PR #206, opened 2016-10-07
SPLICE-466 Make with data the default for create table as (Oracle style)

jleach4:
Adhering to Oracle Syntax as well for create table.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -11583,7 +11583,7 @@ tableDefinition() throws StandardException :
             [ orderCols = orderByClause(null) ]
                hasJDBClimitClause = offsetFetchFirstClause( offsetClauses )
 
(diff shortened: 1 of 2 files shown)


jyuanca:
+1

scottfines:
I'm nervous about this: We've repeatedly had problems where people have tried complex queries in the with data clause only to encounter bizarre errors. I feel like a single IT isn't enough to justify moving to making this default behavior. Specifically, I'd like to see tests on 

1. subqueries which are flattened
2. Subqueries which are not flattened
3. All join strategies
4. Use both Spark and Control-side execution/insertion
5. Multi-table joins

as well as any other "interesting" queries that we can determine. I'd hate to accidentally break things because we moved to a faulty implementation on accident.

jleach4:
I agree we need more testing on "Create table as".  I have filed SPLICE-1048 with the tests you mentioned.  I would however ask you to reconsider my feature addition.  It is only validating that the default behaviour of omitting "with no data" and "with data" is "with data".

dgomezferro:
+1
I'm ok with adding those tests later on.

PR #207, opened 2016-10-07
SPLICE-466 Make with data the default for create table as (Oracle style)

jleach4:
Adhering to Oracle Syntax 2.0 commit.


jyuanca:
+1

dgomezferro:
Should we change this behaviour mid-release in 2.0?

ammolitor:
@dgomezferro, the docs already indicate this behavior. 

http://doc.splicemachine.com/SQLReference/Statements/CreateTable.html

dgomezferro:
@ammolitor Thanks, this makes sense then!
+1

PR #208, opened 2016-10-11
SPLICE-927: Use Batch Nested Loop Join Operation for all NestedLoop Joins

jyuanca:
Use batch nested loop join to execute left outer NLJ, anti NLJ, and one row NLJ. This is a performance improvement, so no need to add ITs. All existing ITs passed.


jfilali:
+1

snowspeeder:
+1

PR #209, opened 2016-10-11
SPLICE-927: Use Batch Nested Loop Join Operation for all NestedLoop Joins

jyuanca:
backport to 2.0


jfilali:
the only thing that would be great is more comments in the code

jfilali:
+1

PR #210, opened 2016-10-11
SPLICE-1049: Dealing with mixed-case columns in import correctly.

scottfines:
This corrects the regression introduced by SPLICE-983, and adds an IT to confirm


jyuanca:
+1

PR #214, opened 2016-10-13
Attempt to fix DB-5653 :

jfilali:
Pulling up hbase.regionserver.kerberos.principal and hbase.regionserver.keytab.file from HBase configuration
and set spark.yarn.keytab and spark.yarn.principal for Spark


snowspeeder:
Jebril, did we test this change.. I see few more properties http://hbase.apache.org/0.94/book/security.html, not sure if they are applicable or not with spark, but worth checking.

jfilali:
not i didn't, The kerberized cluster is currently down

@ryanwoodsmall  Ryan let me know when you get the cluster up and we can test it, either way we push on master or we wait after the test, let me know

jfilali:
I only made sure the property was passed to Spark properly if the hbase property was set.

ryanwoodsmall:
@snowspeeder - good catch. just updated the JIRA issue with some settings from the HBase Master. @jfilali apologies for missing those.

PR #216, opened 2016-10-17
WIndow Functions on top of Dataset.

jfilali:
Added a window method support in com.splicemachine.derby.stream.iapi.DataSet
and pushed down the execution implementation in both engines.

Ignored tests with a decimal precision difference between current implementation and Spark.
This is going to be fixed afterward.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
@@ -30,6 +30,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+
+import com.splicemachine.db.iapi.sql.compile.*;
 import org.apache.log4j.Logger;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.reference.ClassName;
@@ -40,14 +42,6 @@ import com.splicemachine.db.iapi.services.io.FormatableArrayHolder;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.sql.LanguageFactory;
 import com.splicemachine.db.iapi.sql.ResultColumnDescriptor;
-import com.splicemachine.db.iapi.sql.compile.C_NodeTypes;
-import com.splicemachine.db.iapi.sql.compile.CostEstimate;
-import com.splicemachine.db.iapi.sql.compile.NodeFactory;
-import com.splicemachine.db.iapi.sql.compile.Optimizable;
-import com.splicemachine.db.iapi.sql.compile.OptimizablePredicate;
-import com.splicemachine.db.iapi.sql.compile.OptimizablePredicateList;
-import com.splicemachine.db.iapi.sql.compile.Optimizer;
-import com.splicemachine.db.iapi.sql.compile.RowOrdering;
 import com.splicemachine.db.iapi.sql.dictionary.ColumnDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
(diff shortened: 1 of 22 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -248,6 +250,53 @@ public long count() {
...
+                DataType resultDataType = dataset.schema().fields()[aggregator.getResultColumnId()-1].dataType();
+                // We define the window specification and we get a back a spark.
+                // Simply provide all the information and spark window will build it for you
+                Column col = SparkWindow.partitionBy(aggregator.getPartitions())
+                        .function(aggregator.getType())
+                        .inputs(aggregator.getInputColumnIds())
+                        .orderBy(aggregator.getOrderings())
+                        .frameBoundary(aggregator.getFrameDefinition())
+                        .specificArgs(aggregator.getFunctionSpecificArgs())
+                        .specificArgs(aggregator.getFunctionSpecificArgs())
```
jyuanca: redundant invocation of specificArgs

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -180,7 +180,7 @@ public void setSchedulerPool(String pool) {
                     String.class,
                     InputStream.class,
                     HConfiguration.unwrapDelegate());
-            // RDDUtils.setAncestorRDDNames(rdd, 1, new String[] {fileInfo.toSummary()}, null);
+            // SparkUtils.setAncestorRDDNames(rdd, 1, new String[] {fileInfo.toSummary()}, null);
```
jyuanca: Remove commented out code

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkPairDataSet.java
```diff
@@ -112,7 +112,7 @@ public SparkPairDataSet(JavaPairRDD<K, V> rdd,String rddName){
     public PairDataSet< K, V> sortByKey(Comparator<K> comparator, String name) {
         JavaPairRDD rdd2 = rdd.sortByKey(comparator);
         rdd2.setName(name);
-        // RDDUtils.setAncestorRDDNames(rdd2, 2, new String[]{"tbd", "tbd"}, new String[] {"MapPartitionsRDD", "MapPartitionsRDD"});
+        // SparkUtils.setAncestorRDDNames(rdd2, 2, new String[]{"tbd", "tbd"}, new String[] {"MapPartitionsRDD", "MapPartitionsRDD"});
```
jyuanca: ditto

snowspeeder:
+1
Nice edit but I do not see ITs for all the window functions, maybe you're planning to add it later.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
```diff
@@ -962,7 +957,10 @@ public ResultSetNode optimize(DataDictionary dataDictionary,
      */
     public void generate(ActivationClassBuilder acb, MethodBuilder mb) throws StandardException {
 
-		/* Get the next ResultSet#, so we can number this ResultSetNode, its
+
+        dataSetProcessorType= CompilerContext.DataSetProcessorType.FORCED_SPARK;
+        acb.setDataSetProcessorType(dataSetProcessorType);
```
dgomezferro: Are we forcing Spark side execution here? Why is that?

jfilali: this change went out of my scope when i change the profile of my intelliJ, thank for catching the issue

jfilali:
@snowspeeder : I didn't added IT , there is some others in  LeadLagBufferTest but you are right i should probably add more of it . I'll do that after merge, witch ones doesn't have ITs ?

PR #224, opened 2016-10-24
SPLICE-1081: calculate statistics correctly for scan qualifier with a cast

jyuanca:
The optimizer did not expect a scan qualifier may have a CastNode, so it did not look down to CastNode.

Fixed and added an IT.


jfilali:
+1

PR #231, opened 2016-10-27
Added Spark type to ResultColumnDescriptor and now resultSetToDF uses…

mzweben:
… new Tungsten Row interface on ExecRow

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumn.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumn.java
@@ -25,6 +25,7 @@
 
 package com.splicemachine.db.impl.sql.compile;
 
+import java.sql.Types;
 import java.util.Collections;
 import java.util.List;
 
@@ -47,6 +48,8 @@ import com.splicemachine.db.iapi.types.DataValueFactory;
 import com.splicemachine.db.iapi.types.StringDataValue;
 import com.splicemachine.db.iapi.types.TypeId;
 import com.splicemachine.db.iapi.util.StringUtil;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructField;
 
 /**
  * A ResultColumn represents a result column in a SELECT, INSERT, or UPDATE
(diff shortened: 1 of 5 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkUtils.java
```diff
@@ -210,81 +214,20 @@ public ExecRow call(LocatedRow row) throws Exception {
...
         EmbedResultSet40 ers = (EmbedResultSet40) rs;
-        com.splicemachine.db.iapi.sql.ResultSet serverSideRs = ers.getUnderlyingResultSet();
-        JavaRDD<LocatedRow> rdd = ((SparkDataSet) ((SpliceBaseOperation) serverSideRs).getDataSet(EngineDriver.driver().processorFactory().distributedProcessor())).rdd;
-        // The schema is encoded in a string
-        ResultDescription rd = serverSideRs.getResultDescription();
-        final ResultColumnDescriptor[] columns = rd.getColumnInfo();
-
-        // Generate the schema based on the string of schema
+        com.splicemachine.db.iapi.sql.ResultSet serverSideResultSet = ers.getUnderlyingResultSet();
+        SpliceBaseOperation operation = (SpliceBaseOperation) serverSideResultSet;
```
jfilali: any reason we have this package naming in the code ?

jfilali:
What is the use case for converting a resulSet to a dataframe. Usually we want to use Spark to scan the data directly in HBase instead of using the resultSet so we avoid round trip network collection and use collocated node managed by Spark ?

mzweben:
Yes. Thanks. I'll either create an abstract method or encapsulate in
SparkUtils ...

---
Monte Zweben, CEO • Splice Machine
612 Howard Street, SF, CA 94105
415-377-5695 • www.SpliceMachine.com
LinkedIn <http://www.linkedin.com/in/mzweben/> @mzweben
<https://twitter.com/mzweben>

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumn.java
```diff
@@ -1988,5 +1991,49 @@ public ConglomerateDescriptor getBaseConglomerateDescriptor() {
...
+			case Types.TIME:
+				return DataTypes.createStructField(getName(), DataTypes.TimestampType, true); /* TODO: (MZ) Not sure if this is the right conversion */
+			case Types.NULL:
+				return DataTypes.createStructField(getName(), DataTypes.NullType, true);
+			default:
+				return DataTypes.createStructField(getName(), DataTypes.NullType, true);
+		}
+	}
+
 }
```
mzweben: Yes. Good suggestion. Can encapsulate in SparkUtils or as an abstract class possibly. Will do.

jleach4:
I would lose the struct large case statement and go back to the DataValueDescriptor through the DataTypeDescriptor to get the struct.

jleach4:
Closing do to time delay, please re-open when the code is ready.

PR #234, opened 2016-10-28
SPLICE-966 Fixing With Clause omissions

jleach4:
Handles the "(" and ")" properly so with clauses can allow order bys...


graziergeek:
+1.  Looks good. Nice test.

jleach4:
I added a couple of other tests relating to duplicative bugs SPLICE-979 and SPLICE-980.

PR #236, opened 2016-10-28
SPLICE-1072: propogate table referece map from subquery(for 2.0)

jyuanca:
During subquery fattening, reference table map of the subquery was not propagated upwards. This information is needed for hash join with a function applied to join column. Missing this information will cause the result column of join  out of order.


jleach4:
+1 Nice one Jun...

genedavis:
+1

PR #241, opened 2016-10-31
SPLICE-1082: Correcting dependency/statement cache mismatch

scottfines:
This corrects a dependency-statement cache mismatch which lead to some imports using a pre-cached prepared statement when they should have invalidated and recompiled the insert. Also added an IT to confirm (although it's byzantine, in order to reproduce correctly).


jleach4:
+1

genedavis:
+1

PR #249, opened 2016-11-02
SPLICE-1082: Correcting dependency/statement cache mismatch

scottfines:
2.0.x version of dependency statement fix. Identical fix.


genedavis:
+1 except if possible change reference to SPLICE-1087 (from 1082) where possible

PR #250, opened 2016-11-03
SPLICE-1101: default fallback cardinality factor to 10

jyuanca:


jfilali:
what is fall Back Cardinality ? need java doc

PR #252, opened 2016-11-04
SPLICE-159 Supporting Hadoop Compliant File Systems

jleach4:
Small Fix to File System abstraction to allow for access to Hadoop Compliant File Systems.  Added a small test for S3.


genedavis:
+1

PR #253, opened 2016-11-04
SPLICE-1109 Make VTI operations work with joins

dgomezferro:
Implement getExecRowDefinition() for VTI operations, added tests


jleach4:
+1

PR #255, opened 2016-11-04
SPLICE-159 Supporting Hadoop Compliant File Systems

jleach4:
Supporting hadoop compliant filesystems.  Small change to file system abstraction.  Wrote a simple S3 test...

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/S3IT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/S3IT.java
@@ -1,0 +1,100 @@
+package com.splicemachine.derby.impl.sql.execute.operations;
+
+import com.splicemachine.derby.test.framework.SpliceSchemaWatcher;
+import com.splicemachine.derby.test.framework.SpliceTableWatcher;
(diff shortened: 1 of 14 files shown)


hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/S3IT.java
```diff
@@ -0,0 +1,100 @@
...
+                            "%d," +    // max bad records
+                            "'%s'," +  // bad record dir
+                            "null," +  // has one line records
+                            "null)",   // char set
+                    spliceSchemaWatcher.schemaName, "NATION2",
+                    LOCATION,
+                    0,
+                    BADDIR.getCanonicalPath()));
+            ps.execute();
+        ResultSet rs = methodWatcher.executeQuery("select * from nation2");
```
snowspeeder: might want to add order, not sure if order is always guarunteed

jleach4: The validation is unordered.

genedavis:
+1

jfilali:
+1 some methods miss java doc though

PR #256, opened 2016-11-04
SPLICE-1101: set cardinality to 0 when there is no statistics

jyuanca:
Default cardinality to row count will result in a very low selectivity for nested loop join predicate. Thus nestedloop join will be more likely picked than other join strategies. Reset cardinality to 0, the selectivity of join predicate will be 0.1. This was the behavior before John's code change to statistics.


jfilali:
+1

jleach4:
+1

genedavis:
+1

PR #258, opened 2016-11-07
SPLICE 969 -

jfilali:
Fixed for TPC-DS  nested aggregates in queries: 12,20,47,53,57,63,89,98 with Window Functions


jleach4:
Nice Job Jebril...  Maybe add an empty test for aggregate window functions?  We have always had problems when there are no results with items, just a thought.  +1

PR #260, opened 2016-11-08
DB-5671 & DB-5672

jyuanca:
Fixes two customer bugs:
DB-5671: correct the result column order for hash join whose join column is applied by a function. We used to rely on table number and table reference map to determine if a column is from left or right child result set, but these information can be missing. It's more reliable to lookup source result set of a result column from left/right subtree to determine whether the column is from left/right child result set.

DB-5672: an optimization to extract out duplicate predicates from where clause does not work in general. The query that can be optimized is rarely seen, but it can cause problem if there are duplicate predicates. Disable it for now.


genedavis:
+1

PR #264, opened 2016-11-10
SPLICE-1138 Null Fraction Computational Error

jleach4:
Careless mistake when computing the percentage.  Added test for coverage.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsFunctions.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsFunctions.java
@@ -42,9 +42,9 @@ public class StatisticsFunctions {
 
     public static float STATS_NULL_FRACTION(ColumnStatisticsImpl itemStatistics){
         if(itemStatistics==null) return 0;
-        if (itemStatistics.notNullCount() == 0)
+        if (itemStatistics.notNullCount() + itemStatistics.nullCount() == 0) // Divide by null check
             return 0;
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsFunctions.java
```diff
@@ -44,7 +44,7 @@ public static float STATS_NULL_FRACTION(ColumnStatisticsImpl itemStatistics){
         if(itemStatistics==null) return 0;
         if (itemStatistics.notNullCount() == 0)
             return 0;
```
genedavis: Should return 1

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsFunctions.java
```diff
@@ -44,7 +44,7 @@ public static float STATS_NULL_FRACTION(ColumnStatisticsImpl itemStatistics){
...
-        return itemStatistics.nullCount()/itemStatistics.notNullCount();
+        return ((float) itemStatistics.nullCount()/ (float) itemStatistics.notNullCount());
```
jleach4: Nice catch.  Will fix and re-push.  Thanks Gene.

jleach4: Fixed...

genedavis:
+1

PR #267, opened 2016-11-10
SPLICE-1130: Using a VTI to import data and specifying the statusDire…

edriggers:
…ctory splice property causes a null pointer exception

SPLICE-1130: Using a VTI to import data and specifying the statusDirectory splice property causes a null pointer exception

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
@@ -28,7 +28,6 @@ import com.splicemachine.derby.stream.iapi.OperationContext;
 import com.splicemachine.si.api.txn.TxnView;
 import com.splicemachine.derby.stream.control.BadRecordsRecorder;
 import com.splicemachine.stream.accumulator.BadRecordsAccumulator;
-
 import org.apache.log4j.Logger;
 import org.apache.spark.Accumulable;
 import org.apache.spark.Accumulator;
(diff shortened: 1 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
```diff
@@ -455,6 +455,7 @@ public boolean isFailed(){
     public void setPermissive(String statusDirectory, String importFileName, long badRecordThreshold){
         this.permissive=true;
         this.badRecordThreshold = badRecordThreshold;
+	if(importFileName == null)importFileName="unspecified_" + System.currentTimeMillis();
```
jfilali: use a static contant for both reference to "unspecified_" outside of the method

jyuanca:
-1

edriggers:
jyuanca - The problem is that in some cases with VTI there is no filename - for example, if I am using the SpliceJDBCVTI there is no filename, or if you create a custom one that is being called through spark streaming.  I think it needs to handle those cases.

jyuanca:
Understood.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
```diff
@@ -455,6 +456,7 @@ public boolean isFailed(){
     public void setPermissive(String statusDirectory, String importFileName, long badRecordThreshold){
         this.permissive=true;
         this.badRecordThreshold = badRecordThreshold;
+	if(importFileName == null)importFileName=BAD_FILENAME + System.currentTimeMillis();
```
scottfines: Be careful here, to make sure that this file doesn't already exist; it's possible for system clocks to go backwards in time, and under enough load that would result in duplicates. I know, I know, it's highly unlikely, but it would also be an extremely hard bug to track down if it DID happen.

genedavis:
+1 - is this going into branch-2.0 and master branch?

edriggers:
I wanted to put it in both branches.  I was waiting to put it in the master branch until the change was approved in the 2.0 branch.

PR #270, opened 2016-11-11
SPLICE-938 : Splice Machine 3.0 on a cluster with Spark 2.0.0

ryanwoodsmall:
Point 0: this is not (yet) to merge, just want to ask you all to take a look at it.

Fellas, heres a branch with some pom changes to get MapR 5.x.x to start up. SELECT works in master at least to HBase. Still don't have Spark starting up and a "CALL SYSCS_UTIL.SYSCS_GET_VERSION_INFO();" peters out as well.

I bumped some Netty dep versions to their latest 3.x and 4.x "final" releases, shaded and included them in the assembly.

Updating the JIRA issue shortly.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -89,6 +89,8 @@
                                 kryo-serializers,
                                 lucene-core,
                                 memory,
+                                netty,
+                                netty-all,
                                 opencsv,
                                 protobuf-java,
                                 sketches-core,
--- a/hbase_sql/pom.xml
+++ b/hbase_sql/pom.xml
@@ -494,6 +494,8 @@
                     <artifactSet>
                         <includes>
                             <include>io.netty:netty-all</include>
+                            <include>io.netty:netty</include>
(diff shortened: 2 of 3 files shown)


jfilali:
+1

jleach4:
Tough to see, I would suggest adding the dependency tree mods in the future.  Netty is a very tough one...

ammolitor:
@jleach4 what do you mean by add the dependency tree mods?

ammolitor:
in attempting to fix MapR are we breaking any other platforms?

jleach4:
@ammolitor By dependency tree, I mean what are you trying to exclude?  I would also show the appropriate dependency:tree sections from Maven.

ammolitor:
@jleach4, I think i get it. So said another way, a before and after dependency:tree output?  

Seems like a challenge to capture that (well) in a pull request, maybe put that detail information on the JIRA?

PR #271, opened 2016-11-11
SPLICE-1141 Allow for empty directories for ORC and Parquet external …

jleach4:
…tables.

Returning empty set when there are no files in the directory for PARQUET and ORC.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
@@ -34,6 +34,7 @@ import org.apache.log4j.Logger;
 import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.function.Function;
+import org.apache.spark.sql.AnalysisException;
 import org.apache.spark.sql.Column;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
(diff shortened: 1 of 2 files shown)


hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -33,6 +35,14 @@
...
     public static void cleanoutDirectory() {
+        try {
+            File file = new File(getExternalResourceDirectory());
+            if (file.exists())
+                FileUtils.deleteDirectory(new File(getExternalResourceDirectory()));
+            file.mkdir();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
     }
```
jfilali: why are checking for the existence of the file after you catch the exception ? why not checking before you try to read or create a empty parquet file ? Is this for performance reason ?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -306,6 +313,27 @@ public Partitioner getPartitioner(DataSet<LocatedRow> dataSet, ExecRow template,
...
+        // Cannot Infer Schema, Argh
+        if (e instanceof AnalysisException && e.getMessage() != null && e.getMessage().startsWith("Unable to infer schema")) {
+            // Lets check if there are existing files...
+            FileInfo fileInfo = ImportUtils.getImportFileInfo(location);
+            if (fileInfo.fileCount() == 0) // Handle Empty Directory
+                return getEmpty();
+        }
+        throw e;
+    }
+
```
jfilali: why are you checking for the existence of the file after you catch the exception ? why not checking before you try to read or create a empty parquet file ? Is this for performance reason ?

jleach4: I figured an empty table would be a special case.  I do not want to recurse the files unless I have to (exception).

genedavis:
+1

PR #272, opened 2016-11-11
SPLICE-1145 Adding logging required for parquet reading on a cluster.

jleach4:
Adding for assembly so Parquet can run on parcels.  We thought this was provided by Hadoop and we were wrong...


genedavis:
+1

PR #277, opened 2016-11-14
SPLICE-1129 Rollback only rows with WW-conflicts

dgomezferro:
When a row had a write write conflict we would fail all mutations on the
bulkwrite


scottfines:
Why do this? If you get a WWConflict, then we need to error out all the way to the user (since there isn't anything that we can do to fix it).

Seems like maybe the problem is that importing multiple files where the PK violation occurs accross files? If that's the case, that's what AdditiveWriteConflicts are for: we need to ensure that the transactions for each file are additive siblings of one another. Then, an AdditiveWriteConflict can be explicitely checked for and dealt with as a specific case. At least, that's how we solved this problem in 1.5.x

dgomezferro:
You could have a write write conflict with a totally different transaction (import &  streaming updates or interactive users, for instance). You can't guarantee transactions will always be siblings, right? 

Are you suggesting in that case we should fail the whole import process?

scottfines:
ah, I see your point—you want to continue the import even in the event of a WW conflict. That seems contrary to expected behavior in a lot of cases, but in imports that seems reasonable enough. 

However, in the event of conflicting with 2 siblings, you really aren’t in conflict. If you get a WWConflict with a sibling transaction (that’s not a user-level Txn), then you are actually getting a different type of error(like PK or Unique violation) and the error message should reflect that. That’s what I mean about the AdditiveWriteConflict stuff.

Sorry for the confusion there.

dgomezferro:
Yeah, it's probably my fault for not creating a separate Jira for the ww conflict stuff. You are right about the PKs, Jun is addressing that. 
I'll open  a new Jira and resubmit to clear up the confusion, sorry.

PR #278, opened 2016-11-14
SPLICE-1129 Rollback only rows with WW-conflicts

dgomezferro:
When a row had a write write conflict we would fail all mutations on the
bulkwrite

Backport 2.0


dgomezferro:
Resubmitting with fresh Jira

PR #281, opened 2016-11-14
Added Spark type to ResultColumnDescriptor and now resultSetToSF uses…

mzweben:
… new Tungsten Row interface on ExecRow

... Modified to use Uses getNull() on a DTD to get DVD which has a StructField


genedavis:
+1

PR #282, opened 2016-11-15
SPLICE-1129: make import transaction additive

jyuanca:
Import transaction should be additive, so that write-write conflicts between sibling transaction can be handled as primary key/unique constraints violations.
I intended to add ITs similar to Daniel's except some more assertions. Since Daniel will check in IT, I will modify after his checkin.


dgomezferro:
This change still doesn't report all failed rows, it seems there's still something missing.

Imported rows:
> cat /Users/dgomezferro/multiconflicts/*                                                                                                                                               
1
2
3
1
5
6
1
8
9
2
3
10
11

There's 4 duplicates: {1, 1, 2, 3}

Call:
>splice> call SYSCS_UTIL.IMPORT_DATA ('splice', 'a', null, '/Users/dgomezferro/multiconflicts', ',', null, null, null, null, -1, '/tmp/bad5', true, null);
rowsImported        |failedRows          |files      |dataSize            |failedLog
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
13                  |2                   |4          |28                  |/tmp/bad5/multiconflicts.bad

Only 2 failed rows:

>% cat /tmp/bad5/*                                                                                                                                                                      
 PRIMARYKEY: { 1 }
 PRIMARYKEY: { 1 }

PR #285, opened 2016-11-15
SPLICE-1162 Don't publish reference before finishing booting

dgomezferro:
We risk an NPE otherwise

Backport to 2.0


genedavis:
+1

PR #287, opened 2016-11-15
SPLICE-1129: make import transaction additive

jyuanca:
backport to 2.0


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperation.java
```diff
@@ -248,8 +249,18 @@ public void writeExternal(ObjectOutput out) throws IOException{
 
         operationContext.pushScope();
         try{
-            if(statusDirectory!=null)
+            if(statusDirectory!=null) {
                 dsp.setSchedulerPool("import");
+                if (txn instanceof ActiveWriteTxn) {
+                    ActiveWriteTxn activeWriteTxn = (ActiveWriteTxn) txn;
+                    activeWriteTxn.setAdditive(true);
```
scottfines: This doesn't actually make the transaction additive, it just changes what we think the current representation of that transaction is. It needs to be made additive when the transaction is created.

genedavis:
+1

PR #288, opened 2016-11-16
Add snappy compression in DRDA layer (server side)

snowspeeder:
odbc server side snappy changes from server side.

### DIFF
--- a/db-drda/pom.xml
+++ b/db-drda/pom.xml
@@ -42,6 +42,11 @@
             <artifactId>db-tools-i18n</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.xerial.snappy</groupId>
+            <artifactId>snappy-java</artifactId>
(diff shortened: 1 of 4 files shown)


jyuanca:
One question: where the data gets decompressed?

snowspeeder:
in the odbc driver which is another pull request

scottfines:
What about the JDBC side?

snowspeeder:
it is right now for odbc improvements, jdbc was not the focus, may be later
we can add that too.

genedavis:
+1

jleach4:
+1

jleach4:
Can we get this in?

snowspeeder:
yep, its on my radar for this week merge, I wanted to get all external
table issues in first due to priority.

jleach4:
Makes sense…

I am close to getting a patch in that will speed up our ITs.  It is a bit rough right now.

John

PR #289, opened 2016-11-17
SPLICE-1143: extract repetive predicate from DNF

jyuanca:
Extract repetitive predicate from a DNF to enable hashable join strategy.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/ast/RepeatedPredicateVisitor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/ast/RepeatedPredicateVisitor.java
@@ -28,11 +28,8 @@ package com.splicemachine.db.impl.ast;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.compile.C_NodeTypes;
 import com.splicemachine.db.iapi.sql.compile.Visitable;
-import com.splicemachine.db.impl.sql.compile.AndNode;
-import com.splicemachine.db.impl.sql.compile.BinaryListOperatorNode;
-import com.splicemachine.db.impl.sql.compile.BinaryRelationalOperatorNode;
-import com.splicemachine.db.impl.sql.compile.OrNode;
-import com.splicemachine.db.impl.sql.compile.ValueNode;
+import com.splicemachine.db.impl.sql.compile.*;
+
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


dgomezferro:
I'd like to have extensive unit tests covering this transformation, would that be too difficult?

db-engine/src/main/java/com/splicemachine/db/impl/ast/RepeatedPredicateVisitor.java
```diff
@@ -153,16 +160,25 @@ public Visitable defaultVisit(Visitable node) throws StandardException {
         // DB-5672: Disable repeated predicate elimination because it does not work in general. Much
```
dgomezferro: This comment no longer applies.

genedavis:
Daniel's comment is a good one - we will need more tests here, not having enough coverage is what got us in trouble in the first place.  At a minimum we can create a JIRA to add a bunch of tests

ammolitor:
@genedavis, how many JIRA's to "add more tests" do we have in our backlog? Shouldn't we do the right thing here and actually finish the work before committing, or are we establishing (er.. continuing) the "we'll test it later" development pattern?

genedavis:
@ammolitor - fair point - there are some IT's on this checkin already, but the key words here are "extensive" from Daniel and "at a minimum" from me.  I'll talk to Jun.

dgomezferro:
If not extensive, some unit tests would be needed. I'm not sure the included ITs actually test something related to the transformation itself.

jyuanca:
These tests are related to where clause transformation. If the repetitive predicates are recognized correctly, a hashable join strategy will be picked.

jyuanca:
Add a few more test cases to make sure 
1) predicates that repeat in all sub-clauses are all extracted as hashable join predicates
2) repeated predicate that's under NOT node is not subject to this optimization because it cannot be  hash join predicate
3)  where clause that's not a DNF won't be optimized accidentally.
4) predicate that's not repeated in every sub-clause won't be extracted. 

This should be sufficient for test coverage.

dgomezferro:
Jun, my problem with those ITs is that they don't test the correctness of the transformation, they just test whether the transformation is applied or not. At least run the queries to make sure they return the correct results. 

I still think this is a good case for Unit Tests rather than ITs. If unit testing this kind of transformation is not trivial we should build support for it.

jyuanca:
Daniel, it is sufficient to look at the transformed predicate from explain plan. This is not suitable for unit test, because it requires many contexts from derby, context manager, language connection context, node factory, etc..

I added more tests to check predicates, please take another look.

PR #295, opened 2016-11-21
SPLICE-1019: fix the nullability of the not-null columns with rollup …

snowspeeder:
Rollup columns are nullable by nature, but if the schema is defined with not-null columns and rollup used with the sub-queries, the nullability of those columns were not getting set causing the issue in the DRDA layer later. This fix adds the check and sets the nullability of such columns during the binding process where all the result columns should be set.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/MultiGroupGroupedAggregateOperationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/MultiGroupGroupedAggregateOperationIT.java
@@ -16,6 +16,7 @@
 
 package com.splicemachine.derby.impl.sql.execute.operations;
 
+import com.splicemachine.homeless.TestUtils;
 import org.spark_project.guava.collect.Lists;
 import com.splicemachine.derby.test.framework.SpliceDataWatcher;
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/MultiGroupGroupedAggregateOperationIT.java
```diff
@@ -488,7 +519,159 @@ public void testRollupAllOperations() throws Exception{
         }
         Assert.assertEquals("Not all groups found!", pairStats.size()+unameStats.size()+1,row);
     }
-	
+
+    @Test
+    public void testRollupWithNotNull_1() throws Exception {
```
scottfines: What are the differences in these tests? Can those differences be reflected in the test name, rather than as _1,_2, etc?

snowspeeder: There are subtle differences, idea is to make compiler walk through various permutaions. I'll reflect that in one line comment in each test

jyuanca:
Does this need to backport to 2.0?

snowspeeder:
It could be, not sure what is the back-port criteria. Currently this issue
is assigned for Janus

PR #296, opened 2016-11-22
SPLICE-1189 Parallelize Control Side Unions

jleach4:
Parallelize Union Operations on the control side.


genedavis:
Is there a test we can add?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MultiProbeTableScanOperation.java
```diff
@@ -231,34 +232,41 @@ public String toString() {
...
+                        .baseColumnMap(baseColumnMap)
+                        .optionalProbeValue(probeValues[i])
+                        .buildDataSet(this);
+                dataSet = dataSet.union(ds);
+                i++;
+            }
+            return dataSet;
         }
-        return dataSet;
+        catch (Exception e) {
```
dgomezferro: Is this needed?

dgomezferro:
+1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MultiProbeTableScanOperation.java
jleach4: Sorry for the delay, yes it is needed since clone can throw exceptions other than StandardException.  Clone is required so the activation is not shared when parallelized.

PR #300, opened 2016-11-22
SPLICE-1192 Removing restrictive checks for prospect

jleach4:
Splice Machine prospect would like to remove restrictions on the column transformation on top of a union.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
@@ -2176,13 +2176,13 @@ public class ResultColumnList extends QueryTreeNodeVector<ResultColumn>{
 			** Check type compatability.
 			*/
             ClassFactory cf=getClassFactory();
-            if(!unionCompatible(thisExpr,otherExpr)){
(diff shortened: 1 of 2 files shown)


dgomezferro:
What was the example that required this change? I think removing the check completely is wrong, we might want to make it less restrictive though.

What's the reasoning behind allowing the query in the added test case?

jleach4:
The rationale for this is that a splice user would like the normal type conversion to occur vs. a more restrictive type structure for unions.  The union type conversion is much more restrictive than type conversion for example for an insert.

dgomezferro:
I don't think we should be relaxing type safety so much for unions, it would be almost worthless. Isn't there any other option casting one side of the union to the other type, for example? What do other databases implement? 

I'm worried about inconsistencies if we allow this. For instance, in this case:

`create table a (i int);
create table b (i varchar(20));`

This statement 
`insert into a select * from b;`
would fail, but this one
`insert into a select * from a union all select * from b;`
would work (only if all values in b are integers...).

scottfines:
I'm inclined to agree with Daniel on this: Do we know the reason why Derby chose to be more restrictive here? I have a sinking feeling that it's because of relational algebra/type system constraints, and we'll be hating ourselves for opening up the system.

jfilali:
Should we close that pull request for now ? 
@jleach4

PR #301, opened 2016-11-24
SPLICE-998 Move SpliceTransaction into SI module

dgomezferro:
Added some tests around savepoints


jleach4:
+1, Thanks Daniel...  Very helpful for future work.

PR #303, opened 2016-11-28
SPLICE 1139 : CREATE EXTERNAL FILE AND REFRESH SYSTEM PROCEDURE

jfilali:
This feature will do 2 things :
1) Create a empty file for external table at the location provided if no file exist there.
2) Add a new system procedure CALL SYSCS_UTIL.SYSCS_REFRESH_EXTERNAL_TABLE('SPLICE','FOO20'); to refresh the external table in the case that file is modified outside of Splice.


jleach4:
+1

PR #307, opened 2016-12-01
SPLICE-1218 External Table Issues on Cluster: ended up being an issue…

jleach4:
… with scala-lang libraries in assembly


genedavis:
+1

PR #309, opened 2016-12-02
Added Compression for Extenal Table. GZIP or SNAPPY

jfilali:
example : 
create external table compressed_orc (col1 int, col2 varchar(24)) COMPRESSED WITH ZLIB STORED AS ORC LOCATION /somewhere


jleach4:
This has already been approved, asked Jebril to recommit...

PR #312, opened 2016-12-06
SPLICE-1234: error out well for multiple distinct aggregates

snowspeeder:
Do not give wrong answers for multiple distinct aggregates instead error out well until we add the support for it.


jleach4:
+1

PR #314, opened 2016-12-07
SPLICE-1219 Fixes the inner join with restriction and adds test

jleach4:
This fixes the inner join piece.  This will address Bob's immediate need.


genedavis:
+1

PR #315, opened 2016-12-07
SPLICE-1216 Block Coding Configurability

jleach4:
Pull Request which was reverted due to build issues that have been fixed.


genedavis:
Question - it appears that our default block encoding is changing?  Does this impact migrations to this version of the code (if cluster is on previous version) or should there be no issue?

jleach4:
No issue.  Only new tables will be encoded in fast_diff by default.

John

PR #319, opened 2016-12-09
SPLICE-1270 Fixing Dictionary Index Lookup Logic

jleach4:
Inside the dictionary (not query stack), we currently do our index lookups from an index scan serially .  This will perform a single batch multi-get to reduce RPC calls.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
@@ -16,6 +16,8 @@
 package com.splicemachine.derby.impl.store.access.base;
 
 import com.carrotsearch.hppc.BitSet;
+import com.google.common.base.Function;
+import com.google.common.collect.Lists;
 import com.splicemachine.access.api.PartitionFactory;
 import com.splicemachine.db.iapi.error.StandardException;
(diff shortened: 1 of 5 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
```diff
@@ -167,6 +185,51 @@ public boolean fetch(RowLocation loc,DataValueDescriptor[] destRow,FormatableBit
...
+                row.setRowArray(destRow);
+                row.resetRowArray();
+                DataCell keyValue=result.userData();
+                rowDecoder.set(keyValue.valueArray(),keyValue.valueOffset(),keyValue.valueLength());
+                rowDecoder.decode(row);
+                i++;
+            }
+            return true;
+        }catch(Exception e){
+            e.printStackTrace();
```
scottfines: lose the extra printStackTrace() please

jleach4: Oops...

jleach4: Will fix and push, thanks.

jleach4: Fixed

PR #320, opened 2016-12-09
SPLICE-1180 Fixing Time, Small int, and Clob data types

jleach4:
Fixes places where the DataValueDescriptor#getObject() does not allign to Spark types.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTime.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTime.java
@@ -1180,7 +1180,7 @@ public final class SQLTime extends DataType
 			setToNull();
 		else {
 			isNull = false;
-			encodedTime = computeEncodedTime(row.getDate(ordinal));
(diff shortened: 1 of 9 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTime.java
```diff
@@ -1180,7 +1180,7 @@ public void read(Row row, int ordinal) throws StandardException {
 			setToNull();
 		else {
 			isNull = false;
-			encodedTime = computeEncodedTime(row.getDate(ordinal));
+			encodedTime = computeEncodedTime(row.getTimestamp(ordinal));
```
scottfines: This looks weird: we really get a timestamp from the ordinal so that we can get a time value? Not really sure if it's a bug or not, but it looks weird.

jleach4: Time translates to timestamp in Spark world.  I screwed up and had it as a date.

PR #323, opened 2016-12-09
SPLICE-1243 Adding Other FileSystem Support, QA to test

jleach4:
Switching to the correct file system API vs. the API that only gets HBase's FileSystem.

Will ask QA to test S3 since I do not want to open up write on S3.


jfilali:
+1

PR #328, opened 2016-12-12
SPLICE-1177: Correct failed rows for import

jyuanca:
Fix two issues that caused import to report wrong failed rows
1) Take a write row lock(instead of read lock) before writing. A read lock may fail to detect duplicate rows
2) For ControlDatasetWriter, report failed rows after flushing write buffer.

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/storage/HLock.java
+++ b/hbase_storage/src/main/java/com/splicemachine/storage/HLock.java
@@ -28,18 +28,6 @@ import java.util.concurrent.locks.Lock;
  *         Date: 12/15/15
  */
 public class HLock implements Lock{
-    protected static boolean HAS_READ_LOCK_IMPL = false;
-
-    static {
-        try {
-            Class rowLockContextClass = Class.forName("org.apache.hadoop.hbase.regionserver.HRegion$RowLockContext");
-            rowLockContextClass.getDeclaredMethod("newReadLock", new Class[]{});
-            HAS_READ_LOCK_IMPL = true;
-        } catch (Exception e) {
-
-        }
-    }
-
 
     private final byte[] key;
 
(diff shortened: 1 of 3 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/load/HdfsImportIT.java
```diff
@@ -338,8 +338,8 @@ public void testImportMultiFilesPKViolations() throws Exception {
                 assertTrue(rs.next());
 
                 // TODO SPLICE-1177 check for exact number of failed rows
```
dgomezferro: Remove the TODO comments

splice_machine/src/test/java/com/splicemachine/derby/impl/load/HdfsImportIT.java
```diff
@@ -349,7 +349,7 @@ public void testImportMultiFilesPKViolations() throws Exception {
                     badLines.addAll(Files.readAllLines((new File(BADDIR, badFile)).toPath(), Charset.defaultCharset()));
                 }
                 // TODO SPLICE-1177 expect exactly 4 lines
-                assertTrue("Expected some lines in bad files "+badFiles, badLines.size() > 0);
+                assertTrue("Expected some lines in bad files "+badFiles, badLines.size()==4);
```
dgomezferro: Use assertEquals

hbase_storage/src/main/java/com/splicemachine/storage/HLock.java
```diff
@@ -73,7 +73,7 @@ public void lockInterruptibly() throws InterruptedException{
     @Override
     public boolean tryLock(){
         try{
-            delegate = region.getRowLock(key,HAS_READ_LOCK_IMPL?true:false); // Null Lock Delegate means not run...
```
dgomezferro: Remove the HAS_READ... field and the static intialization.

hbase_storage/src/main/java/com/splicemachine/storage/HLock.java
```diff
@@ -28,13 +28,11 @@
  *         Date: 12/15/15
  */
 public class HLock implements Lock{
-    protected static boolean HAS_READ_LOCK_IMPL = false;
 
     static {
```
dgomezferro: Remove the whole static initialization block, it's not used for anything other than HAS_READ_LOCK_IMPL, right?

jfilali:
+1

PR #331, opened 2016-12-12
Force the user to specify the partition data type for external tables…

jfilali:
… and force to use a scalar data type


dgomezferro:
Do we have a list somewhere of the supported types for partitioned columns? Why are blobs OK but dates aren't?

Also, please tag the commit with the JIRA number so it's easier to identify looking at the log

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -8964,6 +8964,19 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
 
+           <msg>
+               <name>EXT21</name>
+               <text>'{0} is not defined in the table definition'</text>
+               <arg>columnName</arg>
+           </msg>
+
+           <msg>
+               <name>EXT22</name>
+               <text>'Partition column {0} need to be numerical '</text>
```
scottfines: 'numerical' isn't the same thing as 'scalar'; a float is numerical but is not a scalar.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -11749,11 +11750,25 @@ tableDefinition() throws StandardException :
...
+                if (isExternal && partitionedResultColumns != null){
+                     List<String> columnsPartitions = Arrays.asList(partitionedResultColumns.getColumnNames());
+
+                     for(String columnPartititon : columnsPartitions){
+                        ColumnDefinitionNode definition = tableElementList.findColumnDefinition(columnPartititon);
+                        if(definition ==  null){
+                                throw StandardException.newException(SQLState.EXTERNAL_TABLES_PARTITIONS_REQUIRED, columnPartititon);
+                            }
+                        if(!definition.getType().getTypeId().isScalar()){
+                                throw StandardException.newException(SQLState.EXTERNAL_TABLES_SCALAR_CONSTRAINT, columnPartititon);
```
scottfines: why do we limit the column types to scalar only?

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
jfilali: yes true, I need instead to test if id either numerical or scalar, there is a confusion there

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
jfilali: Currently, numeric data types and string type are supported in Spark

PR #332, opened 2016-12-13
SPLICE-1288 Retry a few times when status not available for OLAP job

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/olap/AsyncOlapNIOLayer.java
```diff
@@ -53,6 +53,7 @@
 public class AsyncOlapNIOLayer implements JobExecutor{
     private static final Logger LOG=Logger.getLogger(AsyncOlapNIOLayer.class);
 
+    private static final int MAX_RETRIES = 10;
```
scottfines: can we make this configurable? Not critical, but would be convenient

jfilali: What the conclusion on that ?

PR #333, opened 2016-12-13
SPLICE-1288 Retry a few times when status not available for OLAP job

dgomezferro:
Backport to 2.0

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/olap/AsyncOlapNIOLayer.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/AsyncOlapNIOLayer.java
@@ -54,6 +54,7 @@ import java.util.concurrent.locks.ReentrantLock;
 public class AsyncOlapNIOLayer implements JobExecutor{
     private static final Logger LOG=Logger.getLogger(AsyncOlapNIOLayer.class);
 
+    private static final int MAX_RETRIES = 10;
     private final ChannelPool channelPool;
     private final ScheduledExecutorService executorService;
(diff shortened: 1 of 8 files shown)


hbase_sql/src/main/java/com/splicemachine/olap/AsyncOlapNIOLayer.java
```diff
@@ -440,16 +442,24 @@ public void operationComplete(Future<Void> future) throws Exception{
...
+            if (LOG.isTraceEnabled()) {
+                LOG.trace("Received " + or);
+            }
             //TODO -sf- deal with a OlapServer failover here (i.e. a move to NOT_SUBMITTED from any other state
             if(or instanceof SubmittedResult) {
                 future.tickTimeNanos = TimeUnit.MILLISECONDS.toNanos(((SubmittedResult) or).getTickTime());
                 future.lastStatus = System.currentTimeMillis();
             } else if(future.submitted && !future.isDone() && or instanceof NotSubmittedResult) {
-                // The job is no longer submitted, assume aborted
+                // Server says the job is no longer submitted, give it a couple of tries in case messages are out of order
```
jleach4: What causes them to be out of order?

dgomezferro: We request the status a fixed time after the previous request, not after the previous answer. Furthermore we use a different channel for each request, so I'm not sure we can guarantee in order processing of the requests.

jleach4: Yuck.  +1

jleach4:
+1

PR #339, opened 2016-12-14
SPLICE 1238 : Force the user to define partitions keys

jfilali:
Force the user to specify the partition data type for external tables


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TableElementList.java
```diff
@@ -1581,5 +1581,9 @@ public boolean containsColumnName(String colName)
 	{
         return findColumnDefinition(colName) != null;
 	}
+
+	public TableDescriptor getTd() {
```
dgomezferro: This doesn't seem to be used, remove if so.

PR #344, opened 2016-12-15
SPLICE-1163 : Make sure we have the correct permission for external tables

jfilali:
Make sure we have the correct permission for external tables:

1) When someone create a table, make sure if the file doesn't exist to verify if the parent file( aka the directory where you want to store that file ) is writable .
2) When inserting something, make sure the file is writable
3) Found some bug related to compression that came up with those changes


dgomezferro:
Is it possible to add a test for the compression problem?

jleach4:
+1

jfilali:
@dgomezferro  i caught the bug with the current tests. but sure i'm going to add more to make sure everything is alright

dgomezferro:
@jfilali That's fine then, I thought it was an unrelated problem.

jfilali:
@dgomezferro : adding a jira to remind myself that : 

https://splice.atlassian.net/browse/SPLICE-1310

I will merge it and add more tests after next release

PR #346, opened 2016-12-16
SPLICE-1290 Adding Batch Writes to the dictionary

jleach4:
SPLICE-1290: Again with Daniel's suggestions added.


splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
```diff
@@ -221,6 +201,7 @@ public boolean batchFetch(List<RowLocation> locations, List<ExecRow> destRows,Fo
             }
             return true;
         }catch(Exception e){
+            e.printStackTrace();
```
dgomezferro: Remove extra printStackTrace()

dgomezferro:
@jleach4 Please remove the .printStackTrace() so I can merge. Thanks!

jleach4:
Cleaned up printStack

PR #347, opened 2016-12-20
Splice 976 v1

snowspeeder:
Added a check for odbc driver so that only splice odbc driver is allowed to proceed. Tested with our odbc and jdbc app, and odbc app with Easysoft derby odbc driver which would now fail to connect.
**Note: the only new change is in DRDAConnThread


jfilali:
+1

PR #353, opened 2017-01-05
SPLICE-1208: fix mergeSortJoin overestimate by 3x

snowspeeder:
It seems since 1.5 release we were multiplying the mergeSortJoin cost by3x since it was being replicated three times. It does not make sense now to do so since that is not the case, so adjusting the cost by removing that factor.
Tested the plans with the tpcds workload query plan for sanity.


dgomezferro:
This made some tests fail, please revisit:
http://206.225.8.98:8080/job/spliceengine-master/job/spliceengine-platform/412/

PR #355, opened 2017-01-06
SPLICE-1294: remove the legacy costing that involvs hbase temp table …

snowspeeder:
Removing the legacy costing which involves the hbase temp table model


jleach4:
+1

PR #356, opened 2017-01-06
SPLICE-1207: remove the static partitions count and rowOrdering(null)…

snowspeeder:
remove the static partition count from legacy temp hbase temp table approach, keep the row ordering in consistent with all other joinStrategies use.


dgomezferro:
-1
This change makes the rowOrdering of the MergeSort join equal to the rowOrdering of the underlying operation. MergeSort join doesn't guarantee the same ordering since it's a hash join, so this change makes it possible to end up with a MergeJoin on top of a MergeSortJoin which shouldn't be possible.

The duplicate line in this case and the one we should remove would be the first one:
95 innerCost.setRowOrdering(outerCost.getRowOrdering());

snowspeeder:
Thanks Daniel. I’ll remove the first dup line then.

On Mon, Jan 9, 2017 at 2:40 AM, Daniel Gómez Ferro <notifications@github.com

PR #360, opened 2017-01-09
SPLICE-1122:	Deleting a table needs to remove the pin for the table.

jfilali:
- Added a IS_PINNED column in sys tables descriptor.
- Added a job to un cache the pin table
- Remove the pin table from memory when drop table is the table is pinned.

It is difficult to add news IT because DROP TABLE doesn't return any meaningful information and it is difficult to do that in the scope of that jira. I will create a JIRA to add more meaningful informations which will help to add ITs.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreatePinConstantOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreatePinConstantOperation.java
@@ -21,9 +21,12 @@ import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.services.io.FormatableBitSet;
 import com.splicemachine.db.iapi.sql.Activation;
 import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
+import com.splicemachine.db.iapi.sql.depend.DependencyManager;
 import com.splicemachine.db.iapi.sql.dictionary.*;
 import com.splicemachine.db.iapi.sql.execute.ConstantAction;
 import com.splicemachine.db.iapi.store.access.TransactionController;
+import com.splicemachine.db.impl.services.uuid.BasicUUID;
+import com.splicemachine.ddl.DDLMessage;
 import com.splicemachine.derby.ddl.DDLUtils;
 import com.splicemachine.derby.impl.sql.execute.operations.LocatedRow;
 import com.splicemachine.derby.impl.sql.execute.operations.ScanOperation;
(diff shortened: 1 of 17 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/pin/DropPinJob.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        DistributedDataSetProcessor dsp = EngineDriver.driver().processorFactory().distributedProcessor();
+        dsp.setSchedulerPool("admin");
+        dsp.setJobGroup(request.jobGroup, "");
+        String scope = request.scope;
+        DataSet<LocatedRow> dataSet = request.scanSetBuilder.buildDataSet(request.prefix);
+        dataSet.pin(request.scanSetBuilder.getTemplate(),conglomID);
+        jobStatus.markCompleted(new SuccessfulOlapResult());
+        return null;
```
jyuanca: This creates a pin instead of dropping a pin.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropTableConstantOperation.java
```diff
@@ -175,6 +177,12 @@ public void executeConstantAction(Activation activation) throws StandardExceptio
 
             /* Drop the store element at last, to prevent dangling reference for open cursor, beetle 4393. */
             tc.dropConglomerate(heapId);
+
+            /* is the table pinned ? , if yes we need to drop it */
+            if(td.isPined()){
+                EngineDriver.driver().processorFactory().distributedProcessor().dropPinnedTable(td.getHeapConglomerateId());
+            }
```
jyuanca: Transaction semantics may break if a pin was not dropped but the table got dropped. Throw an exception to fail the transaction in this case.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreatePinConstantOperation.java
```diff
@@ -111,6 +117,54 @@ public void executeConstantAction( Activation activation ) throws StandardExcept
...
+        dd.startWriting(lcc);
+        // Drop the table and then recreate it with the pin marked.
+        try {
+            dd.dropTableDescriptor(td,sd,userTransaction);
+        } catch (StandardException e) {
+            if (ErrorState.WRITE_WRITE_CONFLICT.getSqlState().equals(e.getSQLState())) {
+                throw ErrorState.DDL_ACTIVE_TRANSACTIONS.newException("Add Pin ()",
+                        e.getMessage());
+            }
+            throw e;
```
jyuanca: I think you don't need to drop the table descriptor.

PR #362, opened 2017-01-10
SPLICE-1324 Support Tony Danza or Tiny Dancer or TinyInt

jleach4:
Adding support for TinyInt.  It is already there in the code but was left out of the parser since DB2 once upon a time did not support it.


jfilali:
+1

PR #364, opened 2017-01-11
FIX FOR DB-5745 - 'Cannot resolve column name' - …

jfilali:
Error when selecting against ORC table

Now look at the dataframe schema and get the column name based on the index.
I create a test case with a pre existing external file.


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -398,14 +404,92 @@ public void dropPinnedTable(long conglomerateId) throws StandardException {
...
+                        break;
+                    case DataType.ORDER_OP_GREATERTHAN:
+                        orCol = q.negateCompareResult() ? orCol.leq(value) : orCol.gt(value);
+                        break;
+                    case DataType.ORDER_OP_GREATEROREQUALS:
+                        orCol = q.negateCompareResult() ? orCol.lt(value) : orCol.geq(value);
+                        break;
+                    case DataType.ORDER_OP_EQUALS:
+                        orCol = q.negateCompareResult() ? orCol.notEqual(value) : orCol.equalTo(value);
+                        break;
```
snowspeeder: please add default. btw, is there not_equal operator

PR #365, opened 2017-01-12
DB-5744: remove check for collecting schema level stats for external …

snowspeeder:
Removing the check that was not enabling stats collection from schema level for external table, added more tests.


jfilali:
+1

PR #370, opened 2017-01-12
Splice 1208: fix mergeSortJoin overestimate by 3x

snowspeeder:
already merged-reverted change, just updating an IT after review and tpcds workload sanity run.


dgomezferro:
@snowspeeder There's still some failing test cases:
http://206.225.8.98:8080/job/spliceengine-master/job/spliceengine-platform/427/
http://206.225.8.98:8080/job/spliceengine-master/job/spliceengine-platform/428/

They fail a bit randomly, it's not restricted to a single platform. Can you look at it?

snowspeeder:
looks like it is being non-deterministic based on memory, I’ll update ITs
accordingly.

PR #371, opened 2017-01-13
Splice 1325

OlegMazurov:
A global thread pool is created. The default maximum number of threads in the pool is 256.
The pool is used in NLJoinFunction to avoid new thread pool creation on every invocation.

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/api/SConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/api/SConfiguration.java
@@ -101,6 +101,8 @@ public interface SConfiguration {
     // OperationConfiguration
     int getSequenceBlockSize();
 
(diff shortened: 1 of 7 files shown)


jleach4:
Make sure your thread pool is in a state engine.  You create it but I did not see you shut it down.  If you have thread pools open, I believe HBase will not shutdown gracefully (I believe we had that problem before).

OlegMazurov:
I'd introduce a shutdown() or unloadDriver() method to EngineDriver then.
The thread pool could be shutdown there.

OlegMazurov:
Pool threads stuck in I/O (waiting for server response) will prevent the
JVM from exiting. Creating all pool threads as daemon.
Assuming the service is stopped asynchronously, I believe we can shutdown
the pool more aggressively.

jleach4:
+1

PR #372, opened 2017-01-13
DB-5744: remove check for collecting schema level stats for external table

snowspeeder:
We missed this one fix between master and 2.5 branch, so back-porting to 2.5 branch. (Ignore the noise by cherry-pick on xml and shell script)


dgomezferro:
Why should we ignore the script/xml changes? Are they needed or not? 
If not you should remove them.

snowspeeder:
Well, script/xml changes are already there in the 2.5 branch, looks like my
jira branch was before this change and so cherry-pick is showing that in
the diff here again.

dgomezferro:
I see, I think that's weird behaviour from GitHub. Thanks for the clarification!

PR #373, opened 2017-01-16
SPLICE-976: do not allow odbc and jdbc requests other than splice driver

snowspeeder:
Already reviewed/reverted change. Sending new pull request for merge. Updated failing IT and squashed all changes in one commits to track.

### DIFF
--- a/splice_machine/pom.xml
+++ b/splice_machine/pom.xml
@@ -131,6 +131,11 @@
             <type>jar</type>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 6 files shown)


splice_machine/pom.xml
```diff
@@ -131,6 +131,11 @@
             <type>jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
OlegMazurov: Shouldn't the scope be test only?

PR #374, opened 2017-01-16
SPLICE-1335 Upgrade to Spark 2.1.0

jleach4:
This upgrade will unblock Jebril's work with null ordering.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/serializer/SpliceKryoSerializerInstance.java
+++ b/hbase_sql/src/main/java/com/splicemachine/serializer/SpliceKryoSerializerInstance.java
@@ -26,7 +26,7 @@ import org.apache.spark.serializer.KryoSerializerInstance;
 public class SpliceKryoSerializerInstance extends KryoSerializerInstance {
 
     public SpliceKryoSerializerInstance(KryoSerializer ks) {
-        super(ks);
+        super(ks,false);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


hbase_sql/src/main/java/com/splicemachine/serializer/SpliceKryoSerializerInstance.java
```diff
@@ -26,7 +26,7 @@
 public class SpliceKryoSerializerInstance extends KryoSerializerInstance {
 
     public SpliceKryoSerializerInstance(KryoSerializer ks) {
-        super(ks);
+        super(ks,false);
```
dgomezferro: Why not use Unsafe? Does it fail or is it slower?

jfilali: @jleach4 : Please review this comment, so wen can proceed to the merge after that clear out.

jleach4: Sorry for the delay, it fails with an underflow.  I think it is caused by some optimization in not encoding but I did not have time to chase it down.

jleach4: Meant to say int encoding

PR #379, opened 2017-01-20
SPLICE-1349: serialize and initialize BatchOnceOperation correctly

jyuanca:
serialize column positions for correlated subquery, so that BatchOnceOperation can initialize correctly.


jleach4:
+1

PR #386, opened 2017-01-24
SPLICE-1323: Fix NPE for in-list caluse on spark side

snowspeeder:
Need to fence the populateQualifier with more null checks on the probeValue. Tested with various tests manually on multiple size tables/indexes with IN clause combination.


jfilali:
+1

PR #387, opened 2017-01-24
  DB-5753 :

jfilali:
Added check constraint during the table definition.
- Now check the number of attributes to match what we have in the external file
- Now check the datatypes to match the external file

clean up tests

Add Error Handling for ORC and Parquet Files where the definition does not match the data in file.

WIll create a 2.5 commit after this one is in master


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -337,6 +338,26 @@ public Partitioner getPartitioner(DataSet<LocatedRow> dataSet, ExecRow template,
...
+            }
+            if (storedAs.toLowerCase().equals("t")) {
+                schema =  SpliceSpark.getSession().read().csv(location).schema();
+            }
+        }
+
+        return schema;
+    }
+
+    @Override
```
jleach4: I know I created this case statement all over the code where we check "pot".  OK, no laughing.  Would you mind encapsulating this a bit so we can lose the checks all over the code?  Maybe an external file interface which implements getExternalFileSchema for example.  I think I have these in a couple of places in the code.

jfilali: Can you create a seperate JIRA for that and I will clean it up. I would rather have that in master first so it is done and work on the clean up later on.

wtang16: Done. I created https://splicemachine.atlassian.net/browse/DB-5782.

PR #388, opened 2017-01-24
DB-5743: fix table number to allow predicate push down

jyuanca:
Correct table number of a predicate's column reference, so that predicate can be pushed down from  JoinNode to base table.


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnReference.java
```diff
@@ -985,6 +985,10 @@ public ValueNode remapColumnReferencesToExpressions()
 					// column, so fall back on column name, which is unique
 					// then.
 					ftRC = rcl.getResultColumn(columnName);
+					tableNumber = ft.getTableNumber();
+				}
+				else {
+					tableNumber = ftRC.getTableNumber();
 				}
```
jleach4: Love the tests Jun, anyway you can add a comment here explaining the issue?  Maybe link to JIRA?

PR #389, opened 2017-01-24
Splice-1329 Memory leak in SpliceObserverInstructions

OlegMazurov:
The proposed fix moves pushContext(statementContext) from SpliceObserverInstructions.ActivationContext.populateActivation() and the matching popContext() from QueryJob.call() to ActivationHolder so that they are either invoked or not invoked together.
The assumption is that ActivationHolder.reinitialize() must be matched by ActivationHolder.close() unlike ActivationHolder.init(), which closes underlying SpliceTransactionResourceImpl, and thus pops all contexts and context managers, itself.


OlegMazurov:
I did run platform ITs (cdh5.7.2). Now can confirm that the fix is good with ITs and mem as well.

PR #390, opened 2017-01-24
DB-5743: fix table number to allow predicate push down

jyuanca:
Correct table number of a predicate's column reference, so that predicate can be pushed down from JoinNode to base table.


jfilali:
+1

PR #409, opened 2017-01-25
SPLICE-1360 Adding SQL Array Data Type: Basic Serde Functions

jleach4:
Adding Array Serialization/Deserialization for our current data types.  This commit is a pre-cursor commit to full SQLArray support.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
@@ -41,8 +41,10 @@ import org.apache.hadoop.hbase.util.Order;
 import org.apache.hadoop.hbase.util.OrderedBytes;
 import org.apache.hadoop.hbase.util.PositionedByteRange;
 import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.UnsafeArrayData;
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow;
 import org.apache.spark.sql.catalyst.expressions.codegen.BufferHolder;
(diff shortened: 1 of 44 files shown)


ryanwoodsmall:
+1 - exciting additions!

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
```diff
@@ -1584,6 +1586,49 @@ public void write(UnsafeRowWriter unsafeRowWriter, int ordinal) {
...
+	 * @param unsafeArrayWriter
+	 * @param ordinal
+	 * @throws StandardException
+     */
+	@Override
+	public void writeArray(UnsafeArrayWriter unsafeArrayWriter, int ordinal) throws StandardException {
+		if (isNull())
+			unsafeArrayWriter.setNull(ordinal);
+		else {
+			byte[] data = new byte[12];
```
snowspeeder: putting values at higher offset than 12 in the data array?

jleach4: Timestamp is serialized as 3 ints (4 bytes).  Are you asking me to put the magic 12 in a static variable somewhere?

snowspeeder: No, I was curious about how the data is being written by the next statements like Patform.puInt(data,16,...) Isn't 16 the offset, which I thought, 16th byte of data-array it would start writing? pardon my ignorance, I am not too familiar with spark's Platform APIs.

jleach4: Platform API's are a bit different.  They go against the actual memory.  The first 16 bytes of memory is held for the array object instance and references in java.

PR #410, opened 2017-01-25

ryanwoodsmall:
…e 2.5/2.6

fall back on hbase.master.keytab.file for spark.yarn.keytab if we don't set it from hbase.regionserver.keytab.file. Need both spark.yarn.keytab and spark.yarn.principal for Kerberized Spark on YARN.

Successfully tested on Kerberized CDH 5.8.0 and Splice Machine master branch.


ammolitor:
+1

dgomezferro:
Same, use .equals() or .isEmpty()

PR #411, opened 2017-01-25

ryanwoodsmall:
…e 2.5/2.6

fall back on hbase.master.keytab.file for spark.yarn.keytab if we don't set it from hbase.regionserver.keytab.file. Need both spark.yarn.keytab and spark.yarn.principal for Kerberized Spark on YARN.

Successfully tested on CDH 5.8.0 and Splice Machine master branch.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
@@ -149,6 +149,15 @@ public class SpliceSpark {
             conf.set("spark.yarn.keytab", HConfiguration.unwrapDelegate().get("hbase.regionserver.keytab.file"));
         }
 
+        // fallback on hbase.master.keytab.file property if we don't yet have a keytab
+        if(conf.get("spark.yarn.principal", "") != ""){
+            if(conf.get("spark.yarn.keytab", "") == ""){
+                if(HConfiguration.unwrapDelegate().get("hbase.master.keytab.file") != null){
+                    conf.set("spark.yarn.keytab", HConfiguration.unwrapDelegate().get("hbase.master.keytab.file"));
+                }
+                // likely need an else that does a conf.remove("spark.yarn.principal"); if we don't have a keytab, and vice versa
+            }
+        }
 
         // set all spark props that start with "splice.".  overrides are set below.
         for (Object sysPropertyKey : System.getProperties().keySet()) {


hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -149,6 +149,15 @@ private static SparkSession initializeSparkSession() {
             conf.set("spark.yarn.keytab", HConfiguration.unwrapDelegate().get("hbase.regionserver.keytab.file"));
         }
 
+        // fallback on hbase.master.keytab.file property if we don't yet have a keytab
+        if(conf.get("spark.yarn.principal", "") != ""){
```
dgomezferro: I don't think it's a problem here, but in general we shouldn't compare strings with the (in)equality operators, use .isEmpty() or .equals() instead.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -149,6 +149,15 @@ private static SparkSession initializeSparkSession() {
...
+            if(conf.get("spark.yarn.keytab", "") == ""){
```
dgomezferro: Same

dgomezferro:
Apart from the string comparison, lgtm.

ryanwoodsmall:
@dgomezferro - I _think_ this is cleaner than my initial Spark conf.get stuff, as it only relies on Hadoop config. Comments on doing this instead of my initial patch?

```
// pull out Kerberos and set Yarn properties
if((HConfiguration.unwrapDelegate().get("hbase.master.kerberos.principal") != null) ||
   (HConfiguration.unwrapDelegate().get("hbase.regionserver.kerberos.principal") != null)){
    if(HConfiguration.unwrapDelegate().get("hbase.master.kerberos.principal") != null){
        conf.set("spark.yarn.principal", HConfiguration.unwrapDelegate().get("hbase.master.kerberos.principal"));
    } else if(HConfiguration.unwrapDelegate().get("hbase.regionserver.kerberos.principal") != null){
        conf.set("spark.yarn.principal", HConfiguration.unwrapDelegate().get("hbase.regionserver.kerberos.principal"));
    }
}

if((HConfiguration.unwrapDelegate().get("hbase.master.keytab.file") != null) ||
   (HConfiguration.unwrapDelegate().get("hbase.regionserver.keytab.file") != null)){
    if(HConfiguration.unwrapDelegate().get("hbase.master.keytab.file") != null){
        conf.set("spark.yarn.keytab", HConfiguration.unwrapDelegate().get("hbase.master.keytab.file"));
    } else if(HConfiguration.unwrapDelegate().get("hbase.regionserver.keytab.file") != null){
        conf.set("spark.yarn.keytab", HConfiguration.unwrapDelegate().get("hbase.regionserver.keytab.file"));
    }
}
```

dgomezferro:
That's fine with me.

PR #413, opened 2017-01-26
SPLICE-1169 : Fix External Table with Csv - Text FIles

jfilali:
- re-implemented readTextFile in SparkDataSetProcessor to be consistent with the others methods of the class. 
I have a jira to clean that class and have that merged with SpliceFileVTI

- Added to test to make sure data and small datatype work with csv


ryanwoodsmall:
+1

jleach4:
Can you import it as well?  +1

jfilali:
@jleach4 : Would you give more explanation, what are you referring to ?

PR #414, opened 2017-01-26

ryanwoodsmall:
…e 2.5/2.6

Second attempt. Use HBase Master principal/keytab settings and fall back to RegionServer.


jyuanca:
+1 if it is tested on a Kerberos cluster.

ryanwoodsmall:
tested, working on your Kerberized cluster, @jyuanca

PR #420, opened 2017-01-26
SPLICE-1362 IndexOOB exceptions loading TPCC data

OlegMazurov:
Synchronize access to internalConnection's contextManager.


jfilali:
there is the annotation @NotThreadSafe ? in the file 
I think that @dgomezferro  should review that, so basically ActionHolder is only used during serialization ?

PR #427, opened 2017-01-27
DB-5789 redo nested connection on Spark fix

OlegMazurov:
I undo all changes implemented for SPLICE-1329 (memory leak in internalConnection) and SPLICE-1362 (concurrent access to internalConnection) and provide a new fix for the original problem: nested connections are not created on Spark.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
@@ -15,9 +15,8 @@
 package com.splicemachine.derby.stream;
 
 import com.splicemachine.EngineDriver;
-import com.splicemachine.db.iapi.services.context.Context;
-import com.splicemachine.db.iapi.services.context.ContextManager;
 import com.splicemachine.db.impl.jdbc.EmbedConnection;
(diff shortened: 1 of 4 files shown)


OlegMazurov:
Please review.

splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
```diff
@@ -199,12 +198,9 @@ public void reinitialize(TxnView otherTxn, boolean reinit) {
...
             activation = soi.getActivation(this, impl.getLcc());
 
-            Context statementContext = activation.getLanguageConnectionContext().getStatementContext();
+            // Push internal connection to the current context manager
             EmbedConnection internalConnection = (EmbedConnection)EngineDriver.driver().getInternalConnection();
-            ContextManager cm = internalConnection.getContextManager();
-            synchronized (cm) {
-                cm.pushContext(statementContext);
-            }
+            new EmbedConnectionContext(impl.getLcc().getContextManager(), internalConnection);
```
OlegMazurov: This context is pushed to the current ContextManager created in impl.marshallTransaction(). All such contexts are popped in impl.close().

PR #428, opened 2017-01-27
DB-5792: avoid bad file naming collision

jyuanca:
Change bad file name postfix from a hash code to a UUID, because hash code can collide for large number of import tasks.


wtang16:
+1

ammolitor:
+1

PR #432, opened 2017-01-27
SPLICE-1365 :ORDER BY: Nulls first does not work for pinned table

jfilali:
Fixed correct ordering for null first or last

Backport to 2.5


jleach4:
Please rename this...  It has nothing to do with pinned tables.  It is whether spark Window functions support nulls first and nulls last.

PR #433, opened 2017-01-30
SPLICE-1211: open and close latency calculated twice for nested Loop …

snowspeeder:
adjusting the cost for NLJ


dgomezferro:
@snowspeeder Where was the open/close latency calculated twice?

snowspeeder:
in the base cost function..generateCost()

dgomezferro:
If I understand correctly, innerCost.getRemoteCost() is where the open/close latency is computed, right? The problem I see is that we are dividing that cost by innerCost.getEstimatedRowCount(), and that doesn't make sense to me. The open/close cost is independent of the number of rows in the inner table, right?

PR #439, opened 2017-01-31
SPLICE-1370 ; INSERT, UPDATE, DELETE error message for pin tables

jfilali:
Fixed the issues :

### DIFF
--- a/db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
+++ b/db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
@@ -8977,8 +8977,6 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
 
 
 
-
-
            <msg>
                <name>EXT21</name>
                <text>'{0} is not defined in the table definition'</text>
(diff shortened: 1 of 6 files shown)


db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -9000,6 +9000,21 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
+           </msg>
+
+           <msg>
+               <name>EXT26</name>
+               <text>UPDATE a PIN TABLE is not allowed</text>
+           </msg>
+
+           <msg>
+               <name>EXT27</name>
+               <text>DELETE in a PIN TABLE is not allowed</text>
```
snowspeeder: what about truncate, should we add that too.

jfilali: Juste verified and TRUNCATE doesn't not support splice-properties . The user get an error message whenever he will try to do that so we are good.

snowspeeder: good, thanks for checking

PR #440, opened 2017-01-31
SPLICE-1320 Add Support For Array Data Type

jleach4:
SPLICE-1320 Adding Array Type Support

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
@@ -351,6 +351,7 @@ public class DataTypeDescriptor implements Formatable{
 
     private TypeDescriptorImpl typeDescriptor;
     private TypeId typeId;
+    private DataTypeDescriptor[] children;
 
     /**
      * Derivation of this type. All catalog types are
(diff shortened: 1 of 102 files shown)


dgomezferro:
I created a crucible review in case anyone prefers it: https://codereview.splicedev.com:8443/cru/CR-676
I'll add my comments here though.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
```diff
@@ -788,6 +793,10 @@ public DataTypeDescriptor getDominantType(DataTypeDescriptor otherDTS,ClassFacto
             scale=higherType.getScale();
         }
 
+        lowerType.getChildren();
```
dgomezferro: Extra line

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ArrayConstantNode.java
```diff
@@ -0,0 +1,407 @@
...
+import java.util.List;
+
+/**
+ * This node represents an Array Initialization Node.  It supports arrays in the format
+ * [1,1,1] and will set them in the corresponding row to a SQLArray
+ *
+ *
+ */
+
+public class ArrayConstantNode extends ValueNode {
```
dgomezferro: This class can be cleaned up a little bit, remove references to coalesce from comments if possible.

snowspeeder: I suggest John could address these in next commit for minor cleanup, we need master back to green and this commit is reqd soon today, so I will merge it.

PR #442, opened 2017-02-01
SPLICE-1375 Fix concurrency issues reporting failedRows

dgomezferro:
Create ThreadLocal decoder and synchronize closes/writes to the log
file.


wtang16:
This fix needs to go to master and 2.0 branches as well. Thanks.

PR #447, opened 2017-02-01
SPLICE-1379 The number of threads in the HBase priority executor is low

OlegMazurov:
I'm changing two configuration properties in standalone setup:
hbase.regionserver.handler.count = 100 - number of threads (handlers) in the default executor;
hbase.regionserver.metahandler.count = 100 - number of threads in the priority executor.


jfilali:
+1

PR #448, opened 2017-02-02
SPLICE-1285 - backport to 2.5 already approved

jfilali:
Partially fixed the issue when the table wasn't pined to begin with. This will help to understand the underlying issue that causes the problem by removing one case bug.


wtang16:
Can you change all variable names from "*Pined" to "*Pinned"?

jfilali:
ok, i ll do that tmw .

2017-02-02 19:49 GMT-08:00 wtang16 <notifications@github.com>:

dgomezferro:
@jfilali ping me when ready, thanks!

PR #451, opened 2017-02-03
SPLICE-1370 ; INSERT, UPDATE, DELETE error message for pin tables

jfilali:
Fixed the issues :

### DIFF
--- a/db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
+++ b/db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
@@ -9022,6 +9022,22 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
            </msg>
 
 
+           <msg>
+               <name>EXT29</name>
+               <text>INSERT in a PIN TABLE is not allowed</text>
+           </msg>
+
+           <msg>
+               <name>EXT30</name>
+               <text>UPDATE a PIN TABLE is not allowed</text>
+           </msg>
+
+           <msg>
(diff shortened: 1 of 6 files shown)


db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -9022,6 +9022,22 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
 
 
+           <msg>
+               <name>EXT29</name>
+               <text>INSERT in a PIN TABLE is not allowed</text>
+           </msg>
+
+           <msg>
+               <name>EXT30</name>
+               <text>UPDATE a PIN TABLE is not allowed</text>
```
snowspeeder: slight grammar corrections, should read:
UPDATE on a PIN TABLE is not allowed

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -9022,6 +9022,22 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
+           </msg>
+
+           <msg>
+               <name>EXT31</name>
+               <text>DELETE in a PIN TABLE is not allowed</text>
```
snowspeeder: should read: DELETE from a PIN TABLE is not allowed

jfilali:
@snowspeeder I'm going to change it on master directly, along with others typos . I would rather have it in master first.

PR #452, opened 2017-02-03
SPLICE-1374: bad file in S3

jyuanca:
make bad file work in S3.


jleach4:
How are we going to test S3?  QA?

jyuanca:
@jleach4 Good question. Can't store S3 credentials in public repository. I tested manually.

PR #455, opened 2017-02-04
SPLICE-1369: store external table on S3

jyuanca:


splice_machine/src/main/java/com/splicemachine/procedures/external/CreateExternalTableJob.java
```diff
@@ -63,9 +64,10 @@ public Void call() throws Exception {
         dsp.setJobGroup(request.getJobGroup(), "");
 
         // look at the file, if it doesn't exist create it.
-
-        if(!SIDriver.driver().fileSystem().getInfo(request.getLocation()).exists()){
```
dgomezferro: Does it make sense to remove the SIDriver.driver().fileSystem() API? Is there a legitimate usage or do we always want to use SIDriver.driver().getSIEnvironment().fileSystem() ?

PR #459, opened 2017-02-04
SPLICE-1353: export to S3

jyuanca:
backport to 2.0


jleach4:
Jun, did you see the patch I attached to one of the S3 bugs?  Long story short, I would suggest you add a test for export to S3 with invalid credentials.  I am concerned with our error handling.  It throws a runtime exception for issues with S3, so it will end up calling plainwrapexception which will break the connection.  Would you mind adding this test?

jyuanca:
@jleach4 OK, I will add a negative test.

dgomezferro:
@jyuanca please fix mapr distributions in all 3 branches. If you can do that quickly I won't revert, otherwise let me know.

PR #462, opened 2017-02-05
DB-5773 Remove sortedness from NLJ

dgomezferro:
Backport 2.0


jleach4:
Should we set it explicitly to null?

dgomezferro:
Yes, you are right. I'll do that and add a test that catches this issue, good catch.

PR #463, opened 2017-02-06
Splice 1391 Refactor derby's ContextService for clarity, robustness, and performance

OlegMazurov:
There are several issues with the current ContextService/ContextManager implementation:
- access to allContexts is synchronized inconsistently. It's a CopyOnWriteArraySet (unlike derby's original HashSet), which may be expensive on updates, yet some methods use 'synchronized' to access it.
- a list of nested ContextManager's is implemented as java.lang.Stack, whose methods are 'synchronized' (inherited from Vector).
- forceRemoveContext removes ThreadLocal, which causes its re-creation on next access. Its only use in SpliceTransactionResourceImpl is incoherent.
- many "optimizations" implemented in ContextSevice are not worth their complexity, which already caused some obscure problems with undetected memory leaks and sporadic failures. In Splice environment most of those micro-optimizations are not visible.

IT tests pass for mem/cdh5.7.2.


jfilali:
@dgomezferro  Are we merging this one ?

jleach4:
What is the story here?

OlegMazurov:
I created this pull request some time ago. Major motivation: eliminate TLS
trashing and excessive synchronization around ContextManagers (as in global
allContext and thread local threadContextList) - both quite visible in
profiles.

db-engine/src/main/java/com/splicemachine/db/iapi/services/context/ContextService.java
```diff
@@ -423,73 +368,57 @@ public void resetCurrentContextManager(ContextManager cm){
...
-
-            if(SanityManager.DEBUG_ON("memoryLeakTrace")){
-
-                if(stack.size()>10)
-                    System.out.println("memoryLeakTrace:ContextService:threadLocal "+stack.size());
+            if(SanityManager.DEBUG_ON("memoryLeakTrace")) {
+                int size = 0;
+                for (; link != null; link = link.next) ++size;
+                if (size > 10)
+                    System.out.println("memoryLeakTrace:ContextService:threadLocal " + size);
```
jleach4: Log vs. System.out

OlegMazurov:
I replaced SanityManager with Logger. Note that the memory leak detection has never been used in our testing (we enable SanityManager.DEBUG but not SanityManager.DEBUG_ON("memoryLeakTrace")). Had it been used it would have generated lots of false positives during TPC-C tests, for example. I increased the threshold for allContext, which is now consistent with the current TPC-C parameters but may have to be revisited when we increase the number of terminals/connections for TPC-C.
Synced up the branch with master. IT/mem is good, IT/cdh5.7.2 will finish in an hour.

wtang16:
I'd like to close the loop here. Oleg, we need to have a full regression and benchmark run on your branch (after you rebasing to latest master). Please coordinate with QA.

OlegMazurov:
Two more platforms have been tested in Jenkins:
HDP 2.5.0 http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/176/default/
MAPR 5.2.0 http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/175/default/

PR #464, opened 2017-02-06
SPLICE-1370 ; INSERT, UPDATE, DELETE error message for pin tables -approved

jfilali:
Fixed the issues :


jfilali:
already approved

PR #470, opened 2017-02-07
SPLICE-1211: open and close latency calculated twice for NLJ

snowspeeder:
just taking out open/close latency cost which is already factored in transfer cost.


jleach4:
+1 I am fine for now.  We need to build a suite to run the different joins and index lookups to get better timing metrics.

PR #475, opened 2017-02-07
SPLICE-1285 ; please merged - already approved

jfilali:
Partially fixed the issue when the table wasn't pined to begin with. This will help to understand the underlying issue that causes the problem by removing one case bug.


dgomezferro:
@jfilali this has a conflict with branch-2.5, please resolve. Sorry about that.

jfilali:
@dgomezferro done,

PR #480, opened 2017-02-09
SPLICE-1424 Removing Unneeded Visitor from FromTable

jleach4:
This visitor does a lot of work especially on queries with many joins at the FromTable.  Not needed anymore.


jleach4:
Please look at it again @jyuanca

PR #484, opened 2017-02-10
SPLICE-1411: resolve over clause expr with alias from inner query 

snowspeeder:
When inner query has aliases and used in the window function, over clause could not resolve that expression well from the cache map causing NPE. We check again if the match is present when the equivalent alias is present for it.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
@@ -495,6 +495,35 @@ public class WindowResultSetNode extends SingleChildResultSetNode {
                         // Cache the ref
                         pulledUp.put(expKey, new ColumnRefPosition(bottomRC, childRCL.size()));
                     }
+                } else {
+                    // Check again if we find match, if not then add it in cache ref if we have the equivalent match
+                    if (pulledUp.get(expKey) == null) {
+                        ResultColumn srcRC = findOrPullUpRC(expRC,
+                                ((SingleChildResultSetNode) childResult).getChildResult(),
+                                rCtoCRfactory);
+                        if (srcRC != null) {
+                            // We found the RC referenced by this column from the over clause. Create RC->CR pairs
+                            // in our two RCLs (our projection -- this RCL -- and the PR we created as a staging
+                            // below us. Then cache the PR's RC so that we can reference repeated columns to the
+                            // same RC.
+                            String baseName = exp.getColumnName();
+
+                            // create RC->CR->srcRC for bottom PR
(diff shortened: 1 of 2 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
```diff
@@ -495,6 +495,35 @@ private void addOverColumns(Map<String, ColumnRefPosition> pulledUp) throws Stan
                         // Cache the ref
                         pulledUp.put(expKey, new ColumnRefPosition(bottomRC, childRCL.size()));
                     }
+                } else {
+                    // Check again if we find match, if not then add it in cache ref if we have the equivalent match
+                    if (pulledUp.get(expKey) == null) {
```
dgomezferro: This check will always be true

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
```diff
@@ -495,6 +495,35 @@ private void addOverColumns(Map<String, ColumnRefPosition> pulledUp) throws Stan
...
+                        ResultColumn srcRC = findOrPullUpRC(expRC,
+                                ((SingleChildResultSetNode) childResult).getChildResult(),
+                                rCtoCRfactory);
```
dgomezferro: This whole branch is exactly the same as the first half of " if (! contains(expRC, winRCL)) { ", could you consolidate it there?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
snowspeeder: no, it won't be always true; in-fact it should be mostly false especially when contains() returns true, it is supposed to find the entry in the map but due to aliases it does not, and so we need to add it with alias.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
snowspeeder: yes, let me refactor it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowResultSetNode.java
dgomezferro: I don't understand. To get here in line 444 "(pulledUp.get(expKey) == null)" evaluated to true, then in line 445 "(! contains(expRC, winRCL))" evaluated to false. Now we are evaluating  "(pulledUp.get(expKey) == null)" again, did either of pulledUp or expKey change in the meantime?

snowspeeder:
I think I understand what you’re saying now, I forgot there is else

PR #489, opened 2017-02-13
SPLICE-1222 Implement in-memory subtransactions

dgomezferro:
I've closed the previous PR due to the amount of changes.

This merge breaks a few ITs on the mem profile, I created SPLICE-1436 to track that.

This passed a preliminary regression build.

It also fixes SPLICE-1168

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
@@ -108,7 +108,7 @@ public class HBaseConfiguration implements ConfigurationDefault {
      * Defaults to 8192
      */
     public static final String TIMESTAMP_BLOCK_SIZE = "splice.timestamp_server.blocksize";
-    protected static final int DEFAULT_TIMESTAMP_BLOCK_SIZE = 8192;
+    protected static final int DEFAULT_TIMESTAMP_BLOCK_SIZE = 32768;
 
     public static final int DEFAULT_JMX_BIND_PORT = 10102;
 
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
@@ -105,7 +105,7 @@ public class HBaseConfiguration implements ConfigurationDefault {
 
     /**
      * The number of timestamps to 'reserve' at a time in the Timestamp Server.
-     * Defaults to 8192
+     * Defaults to 32768
      */
     public static final String TIMESTAMP_BLOCK_SIZE = "splice.timestamp_server.blocksize";
     protected static final int DEFAULT_TIMESTAMP_BLOCK_SIZE = 32768;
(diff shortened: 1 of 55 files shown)


splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
```diff
@@ -108,7 +108,7 @@
      * Defaults to 8192
      */
     public static final String TIMESTAMP_BLOCK_SIZE = "splice.timestamp_server.blocksize";
-    protected static final int DEFAULT_TIMESTAMP_BLOCK_SIZE = 8192;
+    protected static final int DEFAULT_TIMESTAMP_BLOCK_SIZE = 32768;
```
wtang16: Please change the comment to reflect the default TS block size change from 8k to 32k. But why do you propose the change?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -614,6 +614,8 @@ private void elevateTxnForDictionaryOperations(LanguageConnectionContext lcc) th
             return;
         assert rawTransaction instanceof SpliceTransaction:
                 "Programmer Error: Cannot perform a data dictionary write with a non-SpliceTransaction";
+        // No subtransactions from here on
+        ((SpliceTransaction) rawTransaction).getTxn().forbidSubtransactions();
         /*
```
jyuanca: Can you explain why no sub-transactions are allowed?

splice_si_api/src/main/java/com/splicemachine/si/impl/data/StripedTxnLifecycleStore.java
```diff
@@ -136,6 +137,30 @@ public void rollbackTransaction(long txnId) throws IOException{
...
+    public void rollbackSubtransactions(long txnId, long[] subIds) throws IOException {
+        long beginTS = txnId & SIConstants.TRANSANCTION_ID_MASK;
+
+        Lock lock=lockStriper.get(beginTS).writeLock();
+        acquireLock(lock);
+        try{
+            Txn.State state=baseStore.getState(beginTS);
+            if(state==null){
+                return;
+            }
```
jyuanca: In what case, a transaction has a null state?

splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
dgomezferro: Updated comment. We increase the block size in accordance with the increase in the timestamp step increment.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
dgomezferro: Added a longer comment.

splice_si_api/src/main/java/com/splicemachine/si/impl/data/StripedTxnLifecycleStore.java
dgomezferro: I'm pretty sure this will never happen, the only chance is if txnId refers to a readonly transaction that hasn't been registered on the Txn table. But I'd rather change this in a follow up JIRA just in case it breaks tests or other stuff.

PR #490, opened 2017-02-14
SPLICE-1222 Implement in-memory subtransactions (master)

dgomezferro:
Ported to master.

### DIFF
--- a/splice_si_api/src/main/java/com/splicemachine/si/impl/TransactionImpl.java
+++ b/splice_si_api/src/main/java/com/splicemachine/si/impl/TransactionImpl.java
@@ -60,58 +60,20 @@ public class TransactionImpl extends BaseTransaction {
             ignoreSavePoints=true;
         }
 
-        if(LOG.isDebugEnabled())
-            SpliceLogUtils.debug(LOG,"Before setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
+//        if(LOG.isDebugEnabled())
+//            SpliceLogUtils.debug(LOG,"Before setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
         setActiveState(false,false,null); //make sure that we are active
         Txn currentTxn=getTxn();
         Txn child=lifecycleManager.beginChildTransaction(currentTxn,null);
         txnStack.push(Pair.newPair(name,child));
-        if(LOG.isDebugEnabled())
-            SpliceLogUtils.debug(LOG,"After setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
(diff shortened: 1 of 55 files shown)


splice_si_api/src/main/java/com/splicemachine/si/impl/txn/WritableTxn.java
```diff
@@ -129,46 +131,66 @@ public State getState(){
...
-                    throw exceptionFactory.cannotCommit(txnId,state);
+                    throw exceptionFactory.cannotCommit(txnId, state);
+            }
+            synchronized (this) {
+                //double-checked locking for efficiency--usually not needed
+                switch (state) {
+                    case COMMITTED:
+                        return;
+                    case ROLLEDBACK:
+                        throw exceptionFactory.cannotCommit(txnId, state);
```
snowspeeder: Nice to add default to switch stmt or convert to if-else if there are only 2 states

splice_si_api/src/main/java/com/splicemachine/si/impl/TransactionImpl.java
```diff
@@ -60,58 +60,20 @@ public int setSavePoint(String name,Object kindOfSavepoint) throws IOException{
...
+//        if(LOG.isDebugEnabled())
+//            SpliceLogUtils.debug(LOG,"Before setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
         setActiveState(false,false,null); //make sure that we are active
         Txn currentTxn=getTxn();
         Txn child=lifecycleManager.beginChildTransaction(currentTxn,null);
         txnStack.push(Pair.newPair(name,child));
-        if(LOG.isDebugEnabled())
-            SpliceLogUtils.debug(LOG,"After setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
+//        if(LOG.isDebugEnabled())
+//            SpliceLogUtils.debug(LOG,"After setSavePoint: name=%s, savePointStack=\n%s",name,getSavePointStackString());
```
snowspeeder: should these debug messages be uncommented (also on #63,64) or removed?

splice_si_api/src/main/java/com/splicemachine/si/impl/txn/WritableTxn.java
dgomezferro: Done

splice_si_api/src/main/java/com/splicemachine/si/impl/TransactionImpl.java
dgomezferro: I removed them, now we end up with huge stacks that take a long time to serialize into a String.

PR #493, opened 2017-02-14
SPLICE-1062: retry if region location cannot be found

jyuanca:


hbase_storage/src/main/java/com/splicemachine/si/data/HExceptionFactory.java
```diff
@@ -165,7 +165,16 @@ public IOException processRemoteException(Throwable e){
         }else if(e instanceof CallTimeoutException){
             return new HCallTimeout(e.getMessage());
         }else if(e instanceof IOException) {
-            return (IOException)e;
+            /**
+             * Rpc client may get an IOException with an error message containing "HRegionInfo was null"
+             * when preparing rpc invocation, because a region is in a transitional state. If that happens,
+             * return a retriable exception
+             *
```
snowspeeder: how many times would it retry, and should we add sleep for few miliseconds and retry.

dgomezferro: That should be handled by the current failure handling in BulkWriteAction, search for "busy".

PR #497, opened 2017-02-15
DB-5806: allow inner table of broadcast join to be any FromTable

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/BroadcastJoinStrategy.java
```diff
@@ -157,16 +132,28 @@ public void estimateCost(Optimizable innerTable,
...
+            innerCost.setLocalCostPerPartition(joinCost);
+            innerCost.setRemoteCost(SelectivityUtil.getTotalRemoteCost(innerCost, outerCost, totalOutputRows));
+            innerCost.setRowOrdering(outerCost.getRowOrdering());
+            innerCost.setRowCount(totalOutputRows);
+            innerCost.setEstimatedHeapSize((long) SelectivityUtil.getTotalHeapSize(innerCost, outerCost, totalOutputRows));
+        }
+        else {
+            // Set cost to max to rule out broadcast join
+            innerCost.setLocalCost(Double.MAX_VALUE);
+            innerCost.setRemoteCost(Double.MAX_VALUE);
```
dgomezferro: What happens if we hint to use a Bcast but the rowCount is bigger than the threshold? Does setting the cost to MAX_VALUE screw up the rest of the plan?

jyuanca: Good point. I'll think about it.

jyuanca: If it is hinted to broadcast, make it feasible and return actual cost. Otherwise, bump up the cost to rule it out. 

@dgomezferro take another look

dgomezferro: +1

PR #500, opened 2017-02-15
SPLICE-1436 Fix mem profile after SPLICE-1222

dgomezferro:
Added IT for in mem sub transactions


dgomezferro:
Previous PR got 2 +1s, merging

PR #502, opened 2017-02-15
DB-5471 Allow more packages to be loaded from user code

dgomezferro:
I've inspected the packages we are allowing to be loaded with this change and couldn't find any obvious attack vectors, but I'm no security expert, please have a look.


jfilali:
+1

PR #503, opened 2017-02-15
DB-5471 Allow more packages to be loaded from user code (branch 2.5)

dgomezferro:
Backport


wtang16:
This needs to be backported to 2.0 as well.

PR #512, opened 2017-02-17
SPLICE-1433 Fix drop pinned table case (master)

dgomezferro:
Follow up on SPLICE-1433, missed the case when we try to drop a table that's currently pinned.

Protect unsafe API SpliceSpark.getSession/Context against misuse, add
explicit Unsafe versions for testing and handful of legitimate cases

I ignored DataFrameIT, because it was converting a resultset to a Dataframe using a local SparkContext, outside the OlapServer. I think that kind of operation should be performed on the OlapServer, we could provide better APIs to make it easier to do so. The IT broke with the added protections against misuse of the getSession() APIs.

The added protections could break some user code accessing the SpliceSpark.getSession/Context APIs, for instance from a VTI or Procedure. Should we deprecate them instead of raising an exception?


dgomezferro:
I'll break up these changes in two jiras

PR #513, opened 2017-02-17
SPLICE-1433 Fix drop pinned table case (2.5)

dgomezferro:
Protect unsafe API SpliceSpark.getSession/Context against misuse, add
explicit Unsafe versions for testing and handful of legitimate cases

(backport to 2.5)


dgomezferro:
I'll break up these changes in two jiras

PR #514, opened 2017-02-17
SPLICE-1448 Make sure SpliceSpark.getContext/Session isn't misused (master)

dgomezferro:
Ignore DataFrameIT for now since it's creating a local SparkContext to convert a
resultset into a DF, and it breaks with these protections.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
@@ -56,6 +56,18 @@ public class SpliceSpark {
 
     // Sets both ctx and session
     public static synchronized SparkSession getSession() {
+        String threadName = Thread.currentThread().getName();
+        if (!threadName.startsWith("olap-worker-")) {
+             // Not running on the Olap Server... raise exception. Use getSessionUnsafe() if you know what you are doing.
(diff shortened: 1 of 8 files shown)


dgomezferro:
I think we should create these protections at least in master. There's a chance we break some users' code if they are accessing the Spark Context from a VTI or user Procedure, but in that case that VTI or Procedure would only be correct when running on Spark, not control side. We have to provide better APIs for those use cases.

snowspeeder:
no backports?

dgomezferro:
@snowspeeder I was hoping for some discussion on
1) whether it makes sense to do this
2) since this could break user's code, whether to backport it or not

Having it on master at least would help us developers not misuse the APIs, since we'll always be developing against master, so I think we should do it.

For 2.5 it might make more sense deprecating the methods so users are warned they shouldn't be using those APIs, rather than raising exceptions outright. But maybe it's better to force them to avoid these dangerous APIs.

wtang16:
Daniel, once latest master branch passes regression and benchmarks, please have a run with your branch (rebasing to latest master). We'll go from there. You could also organize a review meeting on this change.

jfilali:
+1

jfilali:
Are we merging this in master ?

jfilali:
@dgomezferro  : Are we good with this one ?

dgomezferro:
@jfilali Yes, please merge it.

PR #517, opened 2017-02-17
SPLICE-1330 Speeding up Join Planning and Predicate Pull Up

jleach4:
Before this patch, each join node would do the following from bottom to top of the tree.

1.  Grab all ColumnReference's below the node. 
2.  For each ColumnReference, recurse to the bottom of the tree until you terminate.
3.  Construct a large HashMap...
4.  Lookup each ColumnReference in a map to set the source for the ColumnReference on the join.

After this patch, the first place where we split the tree from the top does the following

1.  Grab all ColumnReference's below the top node.
2.  For each ColumnReferences, recurse one level down if possible to the next column reference.
3.  Construct a Child Parent tree.
4.  Each Join Node after the tree split, will iteratively probe the tree until it gets to the appropriate level.


genedavis:
This is a big enough change to where I think we need to add IT's.  Code looks good but would be good to see IT's for many-join performance and any functional edge cases as appropriate

wtang16:
It would be interested to know the impact of using com.carrotsearch.hppc data structures. The project seems to have stopped active development. I understand the need for speed. But Java Collection Framework will continue to be improved and maintained whereas com.carrotsearch.hppc is not.

jleach4:
@wtang16 The hpcc model does not wrap each entry in the collection with an object and allows for primitive collections.  I think this is the right approach for a data structure that technically is unbounded.

jleach4:
@genedavis What specific coverage would you like added?  We have all the join strategies and subqueries covered by other tests.

genedavis:
I retract my request for added IT's.  We'll need to make sure QA adds tests to cover this case. +1 - let's rock.

PR #537, opened 2017-02-24
SPLICE-1464: dump csv schema for external table

jyuanca:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -353,20 +357,43 @@ public StructType getExternalFileSchema(String storedAs, String location){
...
+                 schema = getCsvSchema(location);
             }
         }
 
         return schema;
     }
 
+
+    private StructType getCsvSchema(String location) throws IOException {
+        InputStream is = null;
```
dgomezferro: Are we assuming all csv files will have a "schema" file? What happens when they don't?
Couldn't we just disable checking the types when it's a CSV file?

jyuanca: Agree, the csv file may not be produced by splice with schema information. Let's bypass schema checking for csv file

PR #539, opened 2017-02-24
SPLICE-1467 Accidental System.out.println inclusion

jleach4:
This is what is slowing down explain plan generation.


wtang16:
+1

PR #540, opened 2017-02-24
SPLICE-1467 Removing System.out.println inclusion

jleach4:
This is what is slowing down explain plan generation.


wtang16:
+1

wtang16:
This is the reason why we see longer explain time.

PR #542, opened 2017-02-24
SPLICE-1464: bypass schema checking for csv file

jyuanca:


jfilali:
Why you would do that ? csv has a schema that you need to match

jyuanca:
csv's schema is all string type, so it cannot be matched.

jfilali:
you can infer schema for csv : it will scan the file and infer the schema automatically. it works by scanning the data but i don't think that any user will have a huge textfile anyway. In my opinion we would rather make sure we have a correct schema than just ignoring . We got a lot of issues with inconsistent schema in the POCS. 


Please circle back with @jleach4 or @wtang16 . But in my opinion we should pay that extra scan when we create the table. 

look at inferschema in https://github.com/databricks/spark-csv/blob/master/README.md

wtang16:
Thanks Jebril for the comment. My recommendation is that we should not infer data types at table creation time for csv text files. The reasons are as follows:
1. Type inference can be wrong (for example, "True" or "1" can be boolean values or string values. In the case of "True", it could be a legit user conversation response that should be captured as string as boolean).
2. There is little advance by us auto-inferring the data types while the user familiar with the data can come up with a desired database table schema fast.
3. If we infer automatically, users would lose the flexibility to "force" interpret certain data fields (see case I mentioned in 1 above). The data file will be unusable until we give them a software patch to fit their requirement.

I think @jyuanca's change is benign enough and it solves the problem at hand.

@jleach4 can weigh in as well.

jfilali:
@wtang16  : Ok, though the user might get into ClassCastException if he try to convert data that Spark is unable to convert.

wtang16:
Valid point. I think we don't need to over-engineer this too much. User data can have invalid entries as well. When this happens, users will have to fix data as well or use a more accommodating data type definition when creating a table.

+1

PR #551, opened 2017-02-28
SPLICE-1353: Fix IT to check a more generic error

jyuanca:


dgomezferro:
I'm not sure I agree with the change. I understand it sucks that the ITs fail if S3 is down, but that shouldn't be a common occurrence, it's as if our internal network suddenly fails and builds are red (probably less common even!).

jyuanca:
@dgomezferro We should check SQLState rather than the text messages. Text message can change and cause test maintenance problems.

dgomezferro:
@jyuanca You are right, +1

PR #556, opened 2017-03-01
SPLICE-1470 Make sure user transaction rollbacks on Spark failure (2.0)

dgomezferro:
SPLICE-1417 Don't rely on the statusDirectory hint for Spark execution,
explicitly set useSpark=true for imports

SPLICE-116 Unignore test


jyuanca:
This is a fix to SPLICE-1417, but contains an IT for SPLICE-1470

dgomezferro:
@jyuanca the change is primarily for SPLICE-1470: changes around computeModifiedRows(), modifiedRowCount(), openCore() and the IT

Changing openCore() broke the logic for sending imports to Spark, so I fixed that (fixing SPLICE-1417 in turn)

That's why the changes are a bit intermingled, sorry about that.

PR #566, opened 2017-03-03
SPLICE-1489 Make Predicate Pushdown defaulted for ORC

jleach4:
Make Predicate Pushdown default for ORC.


ammolitor:
should you do the same for parquet? Perhaps in a separate commit?
```
spark.sql.parquet.filterPushdown true
```

jfilali:
+1

jfilali:
@jleach4  : Do we need one for 2.5 ?

PR #567, opened 2017-03-03
SPLICE-1490 Bringing Derby Style Forward

jleach4:
Running automated routines against the db-engine to bring the style forward.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/error/StandardException.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/error/StandardException.java
@@ -712,14 +712,13 @@ public class StandardException extends Exception
 	*/
 	public String toString() {
 		String msg = getMessage();
-		StringBuffer buf = new StringBuffer(256);
-		buf.append("ERROR ").append(getSQLState()).append(": ").append(msg).append("\n");
+		String buf = "ERROR " + getSQLState() + ": " + msg + "\n" +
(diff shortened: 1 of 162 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/error/StandardException.java
```diff
@@ -712,14 +712,13 @@ public String getErrorProperty(String type) {
 	*/
 	public String toString() {
 		String msg = getMessage();
-		StringBuffer buf = new StringBuffer(256);
-		buf.append("ERROR ").append(getSQLState()).append(": ").append(msg).append("\n");
+		String buf = "ERROR " + getSQLState() + ": " + msg + "\n" +
+				"Splice Machine Release: " + System.getProperty(Property.SPLICE_RELEASE) + "\n" +
+				"Splice Machine Version Hash: " + System.getProperty(Property.SPLICE_VERSION_HASH) + "\n" +
+				"Splice Machine Build Time: " + System.getProperty(Property.SPLICE_BUILD_TIME);
 		// Include the Splice Machine version/release information.
```
jleach4: String concat is converted to a StringBuilder op (non-synchronized) and supposedly easier to read.

jyuanca: It's ok. I'm not concerned about performance of string concatenation. The automated routine is inconsistent. Sometimes it converts string concatenation to stringbuffer, sometime the other way.

PR #572, opened 2017-03-08
SPLICE-1395: add a generic error message for import failure from S3

jyuanca:
Add a generic error message to report exception from S3. Removed some duplicates message caused by previous bad merge.


jfilali:
@jyuanca  : Do you have one for 2.5 , 2.0 ?

jyuanca:
@jfilali Do we need to backport every single bug fix to all branches?

wtang16:
Backport it to 2.5 only please. 2.5 will be the base for some cloud trials (S3 issues will need to be better handled).

PR #575, opened 2017-03-09
SPLICE-79: generate correct insert statement to import char for bit column

jyuanca:
Generated bulk insert statement has a syntactic error for char for bit column - the column size was not included.


jfilali:
+1

PR #578, opened 2017-03-10
SPLICE-1098: prevent nonnull selectivity from being 0

jyuanca:
For a column with all null values, let its nullSelectivity slightly less than 1 to prevent nonnull selectivity from being 0.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
@@ -219,8 +219,18 @@ public class StoreCostControllerImpl implements StoreCostController {
 
     @Override
     public double nullSelectivity(int columnNumber) {
-        return  tableStatistics.rowCount() ==0?0.0d:((double) (tableStatistics.rowCount() - tableStatistics.notNullCount(columnNumber-1)))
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
```diff
@@ -219,8 +219,16 @@ public double nonNullCount(int columnNumber) {
...
     @Override
     public double nullSelectivity(int columnNumber) {
-        return  tableStatistics.rowCount() ==0?0.0d:((double) (tableStatistics.rowCount() - tableStatistics.notNullCount(columnNumber-1)))
-                / ((double)tableStatistics.rowCount());
+        long rowCount = tableStatistics.rowCount();
+        long nonNullCount = tableStatistics.notNullCount(columnNumber-1);
+        if (rowCount == 0)
+            return 0.0d;
+        else if (nonNullCount == 0)
+            return (double)(rowCount - 1) / (double)rowCount;
```
dgomezferro: Please add a comment here for future reference.

jleach4: +1

jyuanca:
@dgomezferro Added some comments

PR #579, opened 2017-03-10
SPLICE-1006: parse delimiter character for FileVTI

jyuanca:
backport to 2.5


splice_machine/src/main/java/com/splicemachine/derby/impl/load/ImportUtils.java
```diff
@@ -103,4 +103,72 @@ public static DistributedFileSystem getFileSystem(String path) throws StandardEx
...
+            } else if (ch == '\\') {
+                hadBackslash = true;
+                continue;
+            } else if (ch == '^') {
+                hadControl = true;
+                continue;
+            }
+        }
+        return c;
+    }
```
dgomezferro: This code is the very similar to HdfsImport.unescape() Please consolidate them into a single method if possible.

PR #582, opened 2017-03-13
SPLICE-77: order by column in subquery not projected should not be resolved

snowspeeder:
Order by column in subquery being not projected in the subquery select list could get resolved during bind time for the outer query, which however is not in scope for the outer query and so it should be errored out. Compared the semantics with mysql which supports order by in the subquery, which does the similar error-out.
(Ran full cdh profile regression suite)


jfilali:
+1

jfilali:
@snowspeeder  no 2.5 version ?

PR #583, opened 2017-03-13
SPLICE-1430 and SPLICE-1487 ( branch 2.5 )

jfilali:
-Removed pins from dictionary
-Return a error message if the user try to pin a table already pinned.
-Corrected the error message when the user try to delete a pin.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/TableDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/TableDescriptor.java
@@ -160,6 +160,8 @@ public class TableDescriptor extends TupleDescriptor implements UniqueSQLObjectD
     private String storedAs;
     private String location;
     private String compression;
+    // NOT USED ANYMORE, for backward compatibility only
+    @Deprecated
     private boolean isPinned;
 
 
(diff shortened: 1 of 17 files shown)


dgomezferro:
Have you tested the system table change doesn't affect upgradeability?

jfilali:
@dgomezferro : This is why I want to merge that into 2.5. Pin table in the dictionary wasn't in 2.0. Only in 2.5 and 2.6. So this is reverting back to what we had originally.

genedavis:
@jfilali I interpreted @dgomezferro 's question differently.  We have people on v2.5 already.  If this change is checked into v2.5, can those people safely just deploy the new parcel/etc?  Or will they encounter problems?

jfilali:
@genedavis ok then no , those people won't be able to deploy that safely. This changes directly the system table. If we want to roll out that patch for 2.5 I need to make a special version and keep the system table unchanged.
Let me knows what you think. If you requires  to have that bug fixed for 2.5 , then I will go ahead and change it. ( basically if you stop and restart the cluster, all pins table will be in a inconsistent state )

dgomezferro:
@jfilali We could leave the flag in the system table and ignore it, that would maintain backwards compatibility and we could still fix the issue, right?

jfilali:
@dgomezferro  , yes

jfilali:
@dgomezferro : Can you take a quick look to see again.

PR #586, opened 2017-03-14
SPLICE-1516 Enable compression for WritePipeline

dgomezferro:


jleach4:
+1

jleach4:
I still would like to see object - > compressed byte stream vs. object -> byte[] -> compressed byte stream.  I will take a shot at this if I ever get back to encoding...

PR #589, opened 2017-03-15
SPLICE-1497 Add flag for inserts to skip conflict detection 

dgomezferro:
Limitations:
- It only skips conflict detection on base table, support for indexes will be added later
- It skips PK check constraints

Added a new import procedure that uses this flag: IMPORT_DATA_UNSAFE() with the same parameters as the existing one.


wtang16:
+1

jfilali:
+1

PR #616, opened 2017-04-03
SPLICE-1541,1543: fix IN-list issues with dynamic bindings and char column

snowspeeder:
Two fixes related to IN-list and multi-probe index scan:
- with dynamic binding prepared statement, the duplicates in the IN-list were not eliminated,
- with char column, there was no correct match happening, it was missing the right space padding in the probevalue list causing the wrong answer.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/InListOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/InListOperatorNode.java
@@ -142,11 +142,19 @@ public final class InListOperatorNode extends BinaryListOperatorNode
 		 */
 		if (rightOperandList.size() == 1)
 		{
-			BinaryComparisonOperatorNode equal = 
+			ValueNode value = (ValueNode)rightOperandList.elementAt(0);
+            if ((value instanceof CharConstantNode) && (leftOperand instanceof ColumnReference)) {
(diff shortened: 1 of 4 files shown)


snowspeeder:
Dynamic binding fix is only in MultiProbeTableScanOperation.java. By the
time dynamic bindings happen, all probevalues are populated at this point
and sortRequired is marked tobe sorted. So, that is what we do in this
case, we sort and elimiante dupes and all probevalues there. To compare and
constrast with static binding case, dupes elimination and sorting would
happen much earlier so sorting is not done again, we just pass the probe
values to the operator.
Hope this helps clear bit.

dgomezferro:
@snowspeeder does this also solve the case where the IN-list contains non-constants? (2nd case from SPLICE-1543)

If so could you add a test case that covers it? Thanks
+1

yxia92:
The wrong result with passing value (6) to the prepared statement "select * from t1 where i in (2,5,?)" in SPLICE-1543 seems a different issue as there is no duplicates in the inlist. If it is not covered by the current fix, please track it in a separate issue. Code change looks good.

snowspeeder:
Thanks Yi, I missed the other case with non-duplicate dynamic binding which unrelated to the current fix anyway. Looks like only one time it gets bind correctly and then keeps giving the same result regardless of different binding values for IN-list. We could track the other issue separately, this is good to merge anyway.

PR #631, opened 2017-04-10
SPLICE-1541,1543: fix IN-list issues with dynamic bindings and char column (2.5)

snowspeeder:
backport to 2.5, already merged in master


snowspeeder:
Well, it does take care one case of 1543, I’ve made a note in the issue.
Other option is to split the issue for 3 separate cases, but I think it is
ok to keep 1543 open to track it.

yxia92:
Sure Mandar. That is fine with me.

Thanks,
Yi Xia

PR #636, opened 2017-04-13
SPLICE-1559: bulkImportDirectory is case sensitive(2.0)

jyuanca:
splice property bulkImportDirectory should be case sensitive.


splice_machine/src/test/java/com/splicemachine/derby/impl/load/HBaseBulkLoadIT.java
```diff
@@ -63,17 +63,17 @@
     public SpliceWatcher methodWatcher = new SpliceWatcher(SCHEMA_NAME);
 
     @BeforeClass
-    public static void loadData() throws Exception {
+    public static void loaddata() throws Exception {
```
snowspeeder: I think you should keep the camel-case convention for methods.

PR #639, opened 2017-04-13
SPLICE-1482: fix system procedure to split table and index

jyuanca:
System procedures to compute split key and split table/index was not checked into master.


jfilali:
+1

PR #641, opened 2017-04-14
SPLICE-1561 Allowing Clients to turn off cache and lazily execute, te…

jleach4:
…sts are coming in the spark_adapter commit


jyuanca:
Where SpliceClient.isClient is set to true?

jleach4:
The setting will be used by the spark adapter commit comming after this...  I was trying to chop it up because there were around 10 items I fixed/implemented to get Spark working correctly from scala.

PR #647, opened 2017-04-17
SPLICE-1575 ValueRow.hashCode() value should be cached

OlegMazurov:
Cache hashCode() value for ValueRow. Invalidate on any change to the object. Include in serde.


jleach4:
It will be interesting to see if this works...  @dgomezferro is there any set by reference items in the activation that would bypass this?

dgomezferro:
@jleach4 I don't think so, although it is a risk. AFAIK we only use the .hashcode() in ValueRows that act as "keys", and we shouldn't be modifying those. I'm OK with merging and fixing any issues that may come up.

jleach4:
+1 and lets see how it shakes out.

dgomezferro:
@OlegMazurov could we get a run on jenkins prior to merging?

PR #653, opened 2017-04-19
SPLICE-865 : Check if enterprise version is activated and if the user  try to use column privileges.

jfilali:
Straightforward, when the user try to do a GRANT on a column, we first check to see which version the user used and return a error message in case of the open source version.

### DIFF
--- a/db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
+++ b/db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
@@ -1987,5 +1987,6 @@ public interface SQLState {
 	String DIRECTORY_REQUIRED			    						= "EXT33";
 	String ACCESSING_S3_EXCEPTION                                   = "EXT34";
 
+
 }
 
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
@@ -14,7 +14,10 @@
 
 package com.splicemachine.derby.impl.sql.execute.actions;
 
+import com.splicemachine.EngineDriver;
+import com.splicemachine.colperms.ColPermsManager;
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.sql.Activation;
 import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
 import com.splicemachine.db.iapi.sql.dictionary.*;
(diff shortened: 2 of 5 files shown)


db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
```diff
@@ -1986,6 +1986,7 @@ The message identifier (ie. the value that matches a field in this class)
 	String TABLE_NOT_PINNED				    						= "EXT32";
 	String DIRECTORY_REQUIRED			    						= "EXT33";
 	String ACCESSING_S3_EXCEPTION                                   = "EXT34";
+	String COLS_PERMS_EXCEPTION                                     = "EXT35";
```
jyuanca: EXT* is for external table related exception. You may want to put it close to authorization/authentication excpetions

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
```diff
@@ -55,23 +59,38 @@ public void executeConstantAction( Activation activation ) throws StandardExcept
         LanguageConnectionContext lcc = activation.getLanguageConnectionContext();
         TransactionController tc = lcc.getTransactionExecute();
 
-		List <PermissionsDescriptor> permissionsDescriptors = privileges.executeGrantRevoke( activation, grant, grantees);
+        Manager manager = EngineDriver.driver().manager();
```
jyuanca: Try to get ColPermsManager from Manager. If enterprise code is not enabled, an exception will be thrown, and you don't need to add another exception.

db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
jfilali: good catch

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
```diff
@@ -55,23 +59,38 @@ public void executeConstantAction( Activation activation ) throws StandardExcept
...
+        if (privileges instanceof TablePrivilegeInfo) {
+            TablePrivilegeInfo info = (TablePrivilegeInfo) privileges;
+            if (!manager.isEnabled() && info.hasColumns()) {
+                throw StandardException.newException(SQLState.COLS_PERMS_EXCEPTION);
+            }
+        }
```
jyuanca: If the column is not present, you should throw a different exception

jfilali: this for perms only

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
```diff
@@ -55,23 +59,38 @@ public void executeConstantAction( Activation activation ) throws StandardExcept
...
             return ProtoUtil.createRevokeSchemaPrivilege(txnId, schemaPermsDescriptor);
         }
         else if (permissionsDescriptor instanceof TablePermsDescriptor) {
             TablePermsDescriptor tablePermsDescriptor = (TablePermsDescriptor) permissionsDescriptor;
+
             return ProtoUtil.createRevokeTablePrivilege(txnId, tablePermsDescriptor);
         }
         else if (permissionsDescriptor instanceof ColPermsDescriptor) {
+
+            Manager manager = EngineDriver.driver().manager();
```
jyuanca: Same problem here

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/GrantRevokeConstantOperation.java
jfilali: well no this need to happen only when the user use colum permission so i need to check myself

jfilali:
@jyuanca  done, please review

jfilali:
@snowspeeder  : one more +1

jfilali:
@yxia92 : One more +1 ?

PR #654, opened 2017-04-19
DB-5912: fix incremental backup hang(2.5)

jyuanca:
Make sure coprocessor hooks set and clear flags for flushing and split.

Please also review pull request on ee branch.


jfilali:
+1

PR #660, opened 2017-04-24
SPLICE-1584 fix IndexOutOfBound exception when not all column stats a…

yxia92:
…re collected and we try to access column stats for estimation.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/stats/PartitionStatisticsImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/stats/PartitionStatisticsImpl.java
@@ -35,6 +35,7 @@ import com.splicemachine.db.iapi.sql.dictionary.PartitionStatisticsDescriptor;
 
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
 
 /**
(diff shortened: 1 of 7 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/stats/PartitionStatisticsImpl.java
```diff
@@ -43,16 +44,26 @@
...
 
     }
 
-    public PartitionStatisticsImpl(PartitionStatisticsDescriptor partitionStatistics) {
+    public PartitionStatisticsImpl(PartitionStatisticsDescriptor partitionStatistics,
+                                   double fallbackNullFraction,
+                                   double extraQualifierMultiplier) {
        this.partitionStatistics = partitionStatistics;
         for (ColumnStatisticsDescriptor columnStatisticsDescriptor : partitionStatistics.getColumnStatsDescriptors()) {
             itemStatistics.add(columnStatisticsDescriptor.getStats());
```
jyuanca: I guess this is the root cause. Should add with an index to it.

yxia92:
@jyuanca, thanks for the review. Yes, root cause is correct. The logic made the wrong assumption that either stats are not collected or all columns have stats. So every column has an entry in itemStatistics array and that they are stored in the same order as the columnId. 
I’ve made the change as suggested. I’m going to withdraw this pull request and make a new one with the latest fix shortly.

Thanks,
Yi Xia

yxia92:
Based on comments above, I changed the fix and created a separate pull request, so closing this one.

PR #661, opened 2017-04-24
DB-5910: fix hash join column ordering

jyuanca:


jyuanca:
@yxia92 

When some transformations are applied, such as outer join to inner join, the result column still refers to the original outer join result set, which does not exist in the final plan. In this case, it cannot tell whether the join column is from left or right side. 

fromLeftChild is used when reconstructing join result column list if join predicates has a function applied to it.

yxia92:
@jyuanca, thanks for the explanation. Could you take a look at the other part of my comment:
In rebuildJoinNodeRCL(), we are adding a new result column to the result column list. Could you set this new column’s fromLeftchild based on whether it comes from left or right source? I understand that this change is not needed to fix DB-5910’s problem, but since fromLeftChild is now a field of ResultColumn, I would like its value to be set correctly in case in the future someone else will use it.

Thanks.

jfilali:
+1

PR #669, opened 2017-04-26
SPLICE-1425 :  Fix data type inconsistencies with unary functions and  external table based on TEXT data format

jfilali:
Spark use String Type by default for csv. This fix make sure what we convert the columns in the right format by using the execRow schema and the baseColumnMap
passed in readTextFIle

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
@@ -36,7 +36,7 @@ import com.splicemachine.si.impl.driver.SIDriver;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.ql.io.orc.OrcNewInputFormat;
 import org.apache.hadoop.hive.serde2.ColumnProjectionUtils;
-import org.apache.hadoop.hive.serde2.objectinspector.StructField;
+
 import org.apache.hadoop.io.LongWritable;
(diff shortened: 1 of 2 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -514,9 +515,17 @@ private String intArrayToString(int[] ints) {
...
+                int fieldIndex = 0;
+                for( int index = 0; index< baseColumnMap.length; index++){
+                    if(baseColumnMap[index]!=-1){
+                        StructField ft = table.schema().fields()[index];
+                        Column cl = new Column(ft.name()).cast(execRow.schema().fields()[baseColumnMap[index]].dataType());
+                        table = table.withColumn(ft.name(),cl);
+                    }
+
+                }
+
```
jyuanca: Remove correctSchemaSelection?

jfilali: yes that was not so correct after all.

jfilali: @jyuanca : this is fixing a defect on the logic , +1 ?

jyuanca: correctSchemaSelection is no longer used, remove it?

PR #670, opened 2017-04-27
SPLICE-1602 readResolve fetches transactions for a no-op readResolver

OlegMazurov:
Splice currently uses NoOpReadResolver by default. SimpleTxnFilter.readResolve() fetches transactions and checks their state before submitting tasks for a generic ReadResolver thus imposing significant overhead for an empty operation. This check is removed from SimpleTxnFilter. If considered necessary or helpful, it should be pushed down to ReadResolver implementation.

### DIFF
--- a/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
+++ b/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
@@ -161,13 +161,8 @@ public class SimpleTxnFilter implements TxnFilter{
             return;
         }
 
-        TxnView t=fetchTransaction(ts);
-        assert t!=null:"Could not find a transaction for id "+ts;
-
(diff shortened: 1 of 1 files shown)


dgomezferro:
I'm not sure I agree with the change, we are going to fetch the transaction anyway a bit later and if we eventually re-enable the readResolve we might end up overloading it more with uncommitted transactions.

Did you measure a performance difference with and without the change?

OlegMazurov:
Moving the cost to where it's relevant is a good thing as potentially we have more control over it there. I seem to have almost 2x performance improvement in the steady state - what we really care about but not exactly what we get in Jenkins. I'll be doing more testing/profiling before merge, if at all.
Anyway, I have a failing unit test, which I need to analyze. The proposed fix does not affect TPCC.

OlegMazurov:
Profiling shows that practically all the cost is moved to checkVisibility() called right after readResolve(). The change as proposed is not justified.

PR #671, opened 2017-04-27
SPLICE-1597 : Fix issue with cache dictionary when  SYSCS_UTIL.SYSCS_UPDATE_SCHEMA_OWNER is called .

jfilali:
Issue : If as splice i create a schema call schema1, then use the command CALL SYSCS_UTIL.SYSCS_UPDATE_SCHEMA_OWNER( 'SCHEMA1', 'bob'); the user bob should now be able to create tables, etc. in schema1.  This is not the case. It says bob cannot create tables in schema1.

Solution : clear the cache after SYSCS_UPDATE_SCHEMA_OWNER get called.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
@@ -869,6 +869,7 @@ public class SpliceAdmin extends BaseAdminProcedures{
                 throw StandardException.newException(String.format("User '%s' does not exist.", ownerName));
             }
             ((DataDictionaryImpl)dd).updateSchemaAuth(schemaName, ownerName, tc);
+            dd.clearCaches();
         } catch (StandardException se) {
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -869,6 +869,7 @@ public static void SYSCS_UPDATE_SCHEMA_OWNER(String schemaName, String ownerName
                 throw StandardException.newException(String.format("User '%s' does not exist.", ownerName));
             }
             ((DataDictionaryImpl)dd).updateSchemaAuth(schemaName, ownerName, tc);
+            dd.clearCaches();
         } catch (StandardException se) {
             throw PublicAPI.wrapStandardException(se);
         } catch (Exception e) {
```
jyuanca: This does not update dictionary caches in other nodes.

PR #679, opened 2017-04-30
DB-5908 prune query blocks based on unsatisfiable conditions

yxia92:


wtang16:
Is this fix going to 2.0 and 2.5 as well?

db-engine/src/main/java/com/splicemachine/db/iapi/services/property/PropertyUtil.java
```diff
@@ -92,7 +92,8 @@
 		"derby.language.logQueryPlan",
 		Property.LOCKS_ESCALATION_THRESHOLD,
 		Property.DATABASE_PROPERTIES_ONLY,
-		Property.DEFAULT_CONNECTION_MODE_PROPERTY
+		Property.DEFAULT_CONNECTION_MODE_PROPERTY,
+		"derby.language.unSatTreePruningOff"
 	};
```
jyuanca: It would be nicer to define a constant for the new property

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLStatementNode.java
```diff
@@ -145,6 +148,13 @@ public void optimizeStatement() throws StandardException {
         // selectivity 1.0.)
         accept(new ConstantExpressionVisitor());
 
+        // prune tree based on unsat condition
+       GenericLanguageConnectionContext lcc = (GenericLanguageConnectionContext)
+                ContextService.getContext(LanguageConnectionContext.CONTEXT_ID);
+
+        if (lcc == null || !lcc.getUnSatTreePruningOff())
+            accept(new TreePruningVisitor());
+
```
jyuanca: Can lcc be null?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HalfOuterJoinNode.java
```diff
@@ -797,6 +797,22 @@ public JBitSet LOJgetRPReferencedTables(int numTables)
...
+
+        if (leftResultSet.isUnsatisfiable()) {
+            sat = Satisfiability.UNSAT;
+            return true;
+        }
+
+        sat = Satisfiability.NEITHER;
+        return false;
+    }
+
```
jyuanca: What's the difference between UNKNOWN and NEITHER?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TreePruningVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class TreePruningVisitor implements Visitor {
+
+    @Override
+    public Visitable visit(Visitable node, QueryTreeNode parent) throws StandardException {
+        if (node instanceof SelectNode) {
+            node = ((SelectNode) node).unsatTreePruning();
+        } /*else if (node instanceof ProjectRestrictNode) {
+            node = ((ProjectRestrictNode) node).unsatTreePruning();
+        } */
```
jyuanca: remove commented out code

yxia92:
Thanks @jyuanca

Please see my in-line [Response] to your comments.

 
> On Apr 30, 2017, at 10:28 PM, jyuanca <notifications@github.com> wrote:
> 
> @jyuanca commented on this pull request.
> 
> Can you add some Unit tests or ITs?
> 
> 

[Response]  The unit tests are in UnsatTreePruningIT.java.

> 
> In db-engine/src/main/java/com/splicemachine/db/iapi/services/property/PropertyUtil.java <https://github.com/splicemachine/spliceengine/pull/679#discussion_r114098865>:
> 
> > @@ -92,7 +92,8 @@
>  		"derby.language.logQueryPlan",
>  		Property.LOCKS_ESCALATION_THRESHOLD,
>  		Property.DATABASE_PROPERTIES_ONLY,
> -		Property.DEFAULT_CONNECTION_MODE_PROPERTY
> +		Property.DEFAULT_CONNECTION_MODE_PROPERTY,
> +		"derby.language.unSatTreePruningOff"
>  	};
> It would be nicer to define a constant for the new property
> 
[Response] Will do.

> 
> In db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLStatementNode.java <https://github.com/splicemachine/spliceengine/pull/679#discussion_r114098894>:
> 
> > @@ -145,6 +148,13 @@ public void optimizeStatement() throws StandardException {
>          // selectivity 1.0.)
>          accept(new ConstantExpressionVisitor());
>  
> +        // prune tree based on unsat condition
> +       GenericLanguageConnectionContext lcc = (GenericLanguageConnectionContext)
> +                ContextService.getContext(LanguageConnectionContext.CONTEXT_ID);
> +
> +        if (lcc == null || !lcc.getUnSatTreePruningOff())
> +            accept(new TreePruningVisitor());
> +
> Can lcc be null?
> 
[Response] Not really, I just want to be conservative and that except for the case that this rewrite is explicitly turned off, it should be triggered. I can remove the check of “lcc == null”, it is causing confusion.


> 
> In db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HalfOuterJoinNode.java <https://github.com/splicemachine/spliceengine/pull/679#discussion_r114099450>:
> 
> > +    public boolean isUnsatisfiable() {
> +        if (sat == Satisfiability.UNSAT)
> +            return true;
> +
> +        if (sat != Satisfiability.UNKNOWN)
> +            return false;
> +
> +        if (leftResultSet.isUnsatisfiable()) {
> +            sat = Satisfiability.UNSAT;
> +            return true;
> +        }
> +
> +        sat = Satisfiability.NEITHER;
> +        return false;
> +    }
> +
> What's the difference between UNKNOWN and NEITHER?
> 
[Response] UNKNOWN means we haven’t gone through the exercise to evaluate whether the corresponding node is satisfiable, unsatisfiable or neither. Once we have gone through the exercise to evaluate, we will arrive at 3 results: satisfiable, unsatisfiable or neither. I’m using the variable sat to store the evaluation result. Since the rewrite is done through a traversal of the query tree top down, this could avoid repeated evaluation of the same node. 


> 
> In db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TreePruningVisitor.java <https://github.com/splicemachine/spliceengine/pull/679#discussion_r114099696>:
> 
> > +
> +/**
> + * <p>
> + * This visitor rewrites a SelectNode with unsatisfiable condition
> + * to a simple "select * from values (resultcolumnlist) where 1=0".
> + * </p>
> + */
> +public class TreePruningVisitor implements Visitor {
> +
> +    @Override
> +    public Visitable visit(Visitable node, QueryTreeNode parent) throws StandardException {
> +        if (node instanceof SelectNode) {
> +            node = ((SelectNode) node).unsatTreePruning();
> +        } /*else if (node instanceof ProjectRestrictNode) {
> +            node = ((ProjectRestrictNode) node).unsatTreePruning();
> +        } */
> remove commented out code
> 
> 
[Response] will do.

db-engine/src/main/java/com/splicemachine/db/iapi/services/property/PropertyUtil.java
jleach4: This would create another dimension to testing.  It would have to be always on for the commit to be allowed.  We are already stretched with control vs. Spark.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TreePruningVisitor.java
jleach4: instead of instanceof constructs, implement a method/function on Visitable.  It will do nothing except for when a select node is supplied.

yxia92:
@jleach4, do you mean not to allow this rewrite to be turned off? My original thinking is to have a way to temporarily turn off the rewrite so that we have a workaround at emergency. By default it is on. 
I can take this functionality out if it is the convention not to offer.

Thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TreePruningVisitor.java
yxia92: @jleach4 ,  TreePruningVisitor implements Visitor, not ISpliceVisitor, it does not have an abstract function for each visitable. I looked at the classes that implement the ISpliceVisitor (for example, JoinConditionVisitor), they need to be wrapped in another adaptor class SpliceDerbyVisitorAdapter of type Visitor to be invoked during tree traversal. So I didn’t choose it.
Let me know if mis-understand your comment.

yxia92:
Update the branch to accommodate review comments:
- remove the code to allow the rewrite to be turned off. So it is always enabled.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2339,4 +2336,169 @@ public Integer apply(SelectNode input) {
...
+
+        sat = Satisfiability.NEITHER;
+        return false;
+    }
+
+    /* the non-aggregate part is unsatisfiable if:
+       1. its wherePredicate contains an unsat condition (1=0) (assume CNF form); or
+       2. its fromList contains a node that is unsat
+     */
+
```
snowspeeder: Please add a comment here summarizing about this optimization/rewrite

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultSetNode.java
```diff
@@ -76,6 +76,10 @@
     // query. Subclasses will set this value where appropriate.
     protected CostEstimate finalCostEstimate;
 
+    protected enum Satisfiability {UNKNOWN, UNSAT, SAT, NEITHER}
```
snowspeeder: would be helpful to put a comment for each value what it means.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
yxia92: updated branch with comments.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultSetNode.java
yxia92: updated branch with comments.

yxia92:
@jleach4 @jyuanca @snowspeeder , I updated the pull request to patch a small fix for update operation with UNSAT condition. 
For example, 
update t1 set a1=1 where 'xxxx'='y' and b1 in (select b2 from t2);

In UpdateNode.makeConstantAction(), there is an assumption that the target table (t1 here) always participates in planning and that its accesspath will be set. With the UNSAT rewrite, this is no longer true, and it could lead to NullPointerException. the fix takes care of it. Also updated IT with test cases for DML operations. Files changed are:
UpdateNode.java
UnsatTreePruningIT

PR #682, opened 2017-05-03
DB-5875 Implement Kerberos JDBC support

dgomezferro:
Follow DRDA kerberos protocol


db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -1944,8 +1949,53 @@ private int parseACCSEC() throws  DRDAProtocolException
...
+											}
+											return null;
+										}
+									});
+
+									if (exception != null) {
+										throw exception;
+									}
+								} catch (Exception e) {
+									println2Log(null, session.drdaID, e.getMessage());
```
snowspeeder: which log file is println2log logs going to?

dgomezferro: I'm not sure, actually I think it's disabled, I just followed the same pattern as the DRDA code. But we should look into it...

snowspeeder:
Please also review code in spliceengine-ee branch

PR #688, opened 2017-05-04
DB-5934: do not throw exception for transaction absent in transaction table  because this could happen for restored data(2.0)

jyuanca:
Fix a couple problems with restore

Restored data may be written by a transaction that's absent from transaction table. This happens when backup is running concurrently with write workload. The transaction table is copied first before write workload starts, and later data written by the workload get copied. When the backup is restored and data is read and tested for visibility, do not return an error. Ignore them as if they are not visible.

Only do roll back transaction once for a series restores.

code cleanup.

Please also review code changes in ee branch.


snowspeeder:
please resolve conflicts, thanks.

PR #689, opened 2017-05-04
DB-5934: do not throw exception for transaction absent in transaction table  because this could happen for restored data(2.5)

jyuanca:


snowspeeder:
please resolve conflicts

PR #691, opened 2017-05-08
Splice 1479: iterator based stats collection (master)

snowspeeder:
Iterator based stats collection + reduce the number of items being collected for each column. Due to this there are some estimates in IT have changed, so ITs are re-factored.


jyuanca:
It changes data dictionary, so requires future upgrade changes. Please help keep track of this by filing a Jira

wtang16:
We need this to be backported to 2.5 as well.

snowspeeder:
noticed after the branch mere with master, stat job collection was failing due to getSession()'s new restriction, had to add exclusion meantime. Please review the only change in SpliceSpark.java:getSession()

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -60,7 +60,8 @@
     // Sets both ctx and session
     public static synchronized SparkSession getSession() {
         String threadName = Thread.currentThread().getName();
-        if (!SpliceClient.isClient && !threadName.startsWith("olap-worker-")) {
+        if (!SpliceClient.isClient && !threadName.startsWith("olap-worker-")
+                && !threadName.startsWith("DRDAConnThread_")) { // stat collection job
              // Not running on the Olap Server... raise exception. Use getSessionUnsafe() if you know what you are doing.
```
jyuanca: This was caused by removing StatsCollectionJob. You may consider to add it back.

snowspeeder:
StatsCollectionJob was collection based which is now replaced with
StatisticsOperation class that is iterator based. I did not see any issue
when I ran ITs on the branch build until I merged with master. In master,
Daniel had added changes to make the getSession more restrictive and so we
need to allow this job to run need to add this exclusion check.

jyuanca:
@snowspeeder You need to create a branch from ee with the same name as that of spliceengine branch for branch build to pick up your changes and run ITs properly.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsOperation.java
```diff
@@ -0,0 +1,155 @@
...
+        return "statistics " + scope;
+    }
+
+    @Override
+    public List<SpliceOperation> getSubOperations() {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public void openCore() throws StandardException{
```
dgomezferro: @snowspeeder This should submit the StatisticsOperation to the master, is that not happening?

PR #700, opened 2017-05-10
SPLICE-1622 Return only latest version for sequences

dgomezferro:
Increase default allocation size for sequences


genedavis:
Will we need a master version as well?

wtang16:
Daniel, do we need to port this fix to 2.5 and master?

dgomezferro:
Yes, I'll do that.

wtang16:
Thanks Daniel!

PR #706, opened 2017-05-12
SPLICE-1582 Apply memory limit on consecutive broadcast joins (2.5)

yxia92:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2405,4 +2415,38 @@ protected long getMinTimeout() {
...
+            AccessPath ap;
+
+            if (checkSortAvoidancePlan)
+                ap = optimizableList.getOptimizable(proposedJoinOrder[i]).getBestSortAvoidancePath();
+            else
+                ap = optimizableList.getOptimizable(proposedJoinOrder[i]).getBestAccessPath();
+
+            if (ap == null || ap.getJoinStrategy().getJoinStrategyType() != JoinStrategy.JoinStrategyType.BROADCAST)
+                break;
+            memoryAlreadyConsumed += ap.getCostEstimate().getBase().getEstimatedHeapSize();
```
dgomezferro: Is this getting the estimated heap size of the join or the inner table? Because it should only take into account the size of the inner table

dgomezferro:
Would it be possible to perform the check at the BroadcastJoinStrategy level, rather than having to deal with access paths etc.?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
yxia92: It is the inner table's heap size. The base table's heap size is stored in an CostEstimate's baseCost. In the above, ap.getCostEstimate().getBase() is used to retrieve the baseCost.

yxia92:
@dgomezferro, I thought about putting the check directly in the cost of BroadcastJoinStrategy.  However, according to the current planning logic, for a given joinOrder path (which is the order a set of conglomerates are to be joined), optimizer tracks two plans: the best plan and the best sort avoidance plan. The physical binary join plans chosen on the two paths may be different. They are saved separately in each conglomerates' field bestAccessPath and bestSortAvoidancePath. 
For a given conglomerate, in the context of best plan, its broadcast join may be in a sequence of consecutive broadcast joins, but in the context of best sort avoidance plan, it may not be in a sequence of consecutive broadcast joins.
The cost of broadcast join for a given conglomerate is used to update both bestAccessPath and bestSortAvoidancePath, so I cannot apply the memory limit in BroadcastJoinStrategy based on the context of best plan only.

dgomezferro:
@yxia92 that makes sense, thanks for the thorough explanation!

PR #707, opened 2017-05-15
SPLICE-1452 correct cardinality estimation when there is missing part…

yxia92:
…ition stats

Port the change to master branch.

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2012 - 2017 Splice Machine, Inc.
+ *
+ * This file is part of Splice Machine.
+ * Splice Machine is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Affero General Public License as published by the Free Software Foundation, either
+ * version 3, or (at your option) any later version.
+ * Splice Machine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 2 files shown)


snowspeeder:
There is a missing IT, which I believe you're fixing it. I'd suggest port both the changes together to master then.

yxia92:
The IT hasn’t been fixed yet and the test case is currently ignored, do you still want me to port to master?

Thanks.

snowspeeder:
I’d wait until IT is fixed unless this is very urgent change reqd.

yxia92:
Sure. I’ll see if I can re-design the IT that can give consistent behavior.

yxia92:
@snowspeeder, I've fixed the unit test case to give consistent result.

snowspeeder:
Thanks @yxia92

OlegMazurov:
@yxia92, in PR#696 for 2.5 you changed StatsUsageIT.java. Here you add CostEstimationIT.java but leave StatsUsageIT.java intact. Changes to StoreCostControllerImpl.java are identical in both PRs. Why such inconsistency?

yxia92:
@Oleg,

The fix of SPLICE-1452 was first shipped to 2.5, but then its unit test which was added to StatsUsageIT.java was commented out due to non-deterministic behavior, also this test case fails under the mem platform(The test case needs a table split call which does not work for mem database). In the pull request for 2.6, I put the test case into a separate file which will only be called for hbase platform, and also modify the unit test case so that it will behave consistently for repeated runs.
I’m making a pull request to ship the test case to 2.5 and remove the original test case from StatsUsageIT.
Sorry for the confusion.

OlegMazurov:
@yxia92 thanks for the explanation

PR #709, opened 2017-05-16
DB-5946 Fixing S3 File System Implementation

OlegMazurov:
Rolling back a performance fix for preventing early aborts when reading ORC files from S3.
Increasing io.file.buffer.size from default 4K to 64K.


wtang16:
Oleg, let's open a new PR for the added the IT. If it doesn't work for now, we can ignore the test at the moment until it's resolved.

PR #713, opened 2017-05-17
SPLICE-1023 add more info in the message for data import error

yxia92:
This is a fix that Brandon provided. Thanks Brandon.


yxia92:
Sorry, it seems that the fix breaks some regression test cases maybe due to different error messages thrown. close the request for now to fix these issues and will re-open later.

PR #714, opened 2017-05-17
SPLICE-1023 add more info in the message for data import error

yxia92:
Port fix contributed by Brandon with minor correction.


wtang16:
Need it for 2.5

PR #715, opened 2017-05-17
SPLICE-1023 add more info in the message for data import error(2.5)

yxia92:


splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
```diff
@@ -124,20 +123,34 @@ private void writeNullableUTF(ObjectOutput out, String source) throws IOExceptio
...
+        int numofColumnsinTable = 0;
+        int numofColumnsinFile = 0;
+        boolean columnnumbermistmatch = false;
+
         operationContext.recordRead();
         try {
             ExecRow returnRow = execRow.getClone();
-            if(values==null)
+            if (values == null) {
+                columnnumbermistmatch = true;
```
snowspeeder: why assign to value to local variable before throwing exception?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
```diff
@@ -124,20 +123,34 @@ private void writeNullableUTF(ObjectOutput out, String source) throws IOExceptio
...
                 throw StandardException.newException(SQLState.COLUMN_NUMBER_MISMATCH, execRow.nColumns(), 0);
+            }
+            else if (values.size() < returnRow.nColumns()) {
+                columnnumbermistmatch = true;
```
snowspeeder: same here

splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
yxia92: The variable columnnumbermismatch is used later in processing the exception in the catch part.

snowspeeder: I see, exception is being caught in the same function body later

PR #719, opened 2017-05-19
SPLICE-1637 Enable compression for HFile gen in bulk loader (2.5)

dgomezferro:


jyuanca:
Need to port to 2.0 and master. Thanks Daniel!

PR #720, opened 2017-05-19
SPLICE-1637 Enable compression for HFile gen in bulk loader

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/derby/stream/function/HFileGenerationFunction.java
```diff
@@ -137,7 +140,7 @@ private void init(Long conglomerateId, byte[] key) throws IOException{
 
     private StoreFile.Writer getNewWriter(Configuration conf, Path familyPath)
             throws IOException {
-        Compression.Algorithm compression = Compression.Algorithm.NONE;
+        Compression.Algorithm compression = Compression.getCompressionAlgorithmByName(compressionAlgorithm);
```
snowspeeder: if the string is empty, would this set to NONE compression by default, or we should add some check here, just in -case

PR #726, opened 2017-05-21
SPLICE-1640 apply memory limit check for consecutive outer broadcast …

yxia92:
…join and derived tables:

1. Introduce a new variable accumulatedMemory in the class SimpleCostEstimate to record the accumulated memory usage by a sequence of consecutive broadcast join. This variable is set as part of the outermostCostEstimate when we start planning the joins of a new set of tables, and need to carry the cost from the outer block or the outer tables of an outer join.
2. Add method getAccumulatedMemory() to the interface Optimizer. Given the best join sequence planned by the current optimizer, if the join sequence ends with consecutive broadcast joins, return the accumulated memory usage by these broadcast joins.
3. In the class OptimizerImpl that implements the interface Optimizer, add the actual implementation of the method getAccumulatedMemory().
4. Also in the class OptimizerImpl, apply the memory check in considerCost(),  similar to the check SPLICE-1582 added in the function costBasedCostOptimizable(). The difference between considerCost() and costBasedCostOptimizable() is that, costBasedCostOptimizable()decides for an optimizable corresponding to a base table whether the current cost could be picked up as best cost, while considerCost() decides for an optimizable corresponing to a derived table or some intermedidate result whether the current cost could be picked as best cost.
5. Add method getMemoryUsage4BCJ() to the interface Optimizable, as well as all its class implementations. The implemenation for JoinNode will take care of the consecutive broadcast joins through outer joins or joins of the form table1 (left/inner) join table2 on joincond1 (left/inner) join table3 on joincond2....

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CostEstimate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CostEstimate.java
@@ -271,4 +271,8 @@ public interface CostEstimate extends StoreCostResult {
     double localCostPerPartition();
 
     void setLocalCostPerPartition(double localCostPerPartition);
+
+    double getAccumulatedMemory();
+
+    void setAccumulatedMemory(double memorySize);
 }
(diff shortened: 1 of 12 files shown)


dgomezferro:
Same, please rename method, other than that +1

Thanks!

yxia92:
@dgomezferro , I updated the pull request to refactor the function name getMemoryUsage4BCJ() to getMemoryUsage4BroadcastJoin().

PR #728, opened 2017-05-22
SPLICE-1452 fix unit test case in 2.5

yxia92:


OlegMazurov:
CostEstimationIT.testCardinalityAfterTableSplit failed:
http://206.225.8.98:8080/job/spliceengine-2.5/job/spliceengine-platform/platform=mapr5.2.0/212/

wtang16:
Yi Xia is looking at it.

yxia92:
I reproduce the failure on MapR on my local system, and confirmed that split_table could sometimes could return after a timeout period of 10 seconds without the split taking place. (Debugging shows that split doesnot happen because of the exception "Can't connect to ZooKeeper”).
I've made separate pull request #738 and #739  to fix the test case.

PR #729, opened 2017-05-23
SPLICE-1640 apply memory limit check for consecutive outer broadcast …(2.5)

yxia92:
…join and derived tables

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
@@ -428,4 +428,9 @@ public interface Optimizable {
 	 * @exception StandardException		Thrown on error
 	 */
 	double uniqueJoin(OptimizablePredicateList predList) throws StandardException;
+
+	/**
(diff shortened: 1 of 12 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
```diff
@@ -428,4 +428,9 @@ boolean feasibleJoinStrategy(OptimizablePredicateList predList,
 	 * @exception StandardException		Thrown on error
 	 */
 	double uniqueJoin(OptimizablePredicateList predList) throws StandardException;
+
+	/**
+	 * get the current optimizable's memory usage if its best join plan is a broadcast join
+	 */
+	double getMemoryUsage4BCJ();
```
dgomezferro: I'd rather have a complete name, rather than the abbreviature. I wouldn't mind on a local variable but for the method name I think it's better to spell it out.

yxia92:
@dgomezferro , I refactored the function name getMemoryUsage4BCJ() to getMemoryUsage4BroadcastJoin() based on your comment and updated the full request.

PR #749, opened 2017-06-01
SPLICE-1586 Prevent NPE when Spark job fails

dgomezferro:
Bail out as soon as we get a Failure message to prevent a NPE


jyuanca:
Is this needed for master branch?

dgomezferro:
It was already on master

PR #762, opened 2017-06-05
SPLICE-1669: Support bulk delete.

jyuanca:
Added a hint bulkDeleteDirectory for delete statement to generate and load HFiles.


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLModStatementNode.java
```diff
@@ -1557,8 +1557,6 @@ public void optimizeStatement() throws StandardException
 			if (colBitSet != null)
 			{
 				for (int i = 0; i < cols.length; i++) {
-                    // Do not Add Primary Keys, we have them in the rowKey (BaseTable) or in the RowLocation (IndexTable)
-                    if (!ArrayUtils.contains(primaryKeyColumns,cols[i]))
```
dgomezferro: This change basically reverts https://splicemachine.atlassian.net/browse/DB-3905 , right? Wouldn't that hurt us on TPCC?
Pinging @jleach4  since he made that change.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkBulkIDeleteTableWriterBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.derby.stream.iapi.DataSet;
+import com.splicemachine.derby.stream.output.BulkDeleteDataSetWriterBuilder;
+import com.splicemachine.derby.stream.output.DataSetWriter;
+import com.splicemachine.derby.stream.output.delete.DeleteTableWriterBuilder;
+
+/**
+ * Created by jyuan on 5/31/17.
+ */
+public class SparkBulkIDeleteTableWriterBuilder extends DeleteTableWriterBuilder implements BulkDeleteDataSetWriterBuilder{
```
dgomezferro: Typo? It should be SparkBulkDelete... (extra 'I')

jyuanca: Good catch, thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLModStatementNode.java
jyuanca: I'll only do this for bulk delete, and leave other DMLs as it is.

PR #763, opened 2017-06-05
SPLICE-1591: Support bulk delete.(2.5)

jyuanca:
Backport to 2.5


splice_machine/src/test/java/com/splicemachine/derby/impl/load/HBaseBulkLoadIT.java
```diff
@@ -97,9 +97,44 @@ public static void loaddata() throws Exception {
...
+        spliceClassWatcher.execute(sql);
+        ResultSet rs = spliceClassWatcher.executeQuery(String.format("select count(*) from %s", tableName));
+        rs.next();
+        int count = rs.getInt(1);
+        Assert.assertTrue(count==0);
+    }
+
+    private static void countUsingIndex(String tableName, String indexName) throws Exception {
+        ResultSet rs = spliceClassWatcher.executeQuery(
+                String.format("select count(*) from %s --splcie-properties index=%s", tableName, indexName));
```
snowspeeder: typo: splice-properties

jyuanca: Good catch, Thanks!

PR #765, opened 2017-06-07
SPLICE-1671: Table/Schema snapshot

jyuanca:
Check Splice-1671 for description of the design.
https://splice.atlassian.net/browse/SPLICE-1671


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
```diff
@@ -68,7 +68,7 @@
     /** The conglomerate id of the first user table */
     // NOTE: JC - this constant is also defined in (splice) EnvUtils. When adding a new sys table, this
     // number will need to be increased in BOTH places.
-    long FIRST_USER_TABLE_NUMBER = 1458;
+    long FIRST_USER_TABLE_NUMBER = 1568;
```
yxia92: What is the increase of this number for?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -519,22 +542,36 @@ private void upgradeIfNecessary(TransactionController tc) throws StandardExcepti
...
+            createSnapshotTable(tc);
+            SpliceLogUtils.info(LOG, "Catalog upgraded: added SYS.SYSSNAPSHOTS table.");
+        }
+        createOrUpdateAllSystemProcedures(tc);
+    }
+
+    private void addNewColumToSystables(TransactionController tc) throws StandardException {
         SchemaDescriptor sd = getSystemSchemaDescriptor();
-        TableDescriptor td = getTableDescriptor("SYSTABLES", sd, tc);
+        TableDescriptor td = getTableDescriptor(SYSTABLEPERMSRowFactory.TABLENAME_STRING, sd, tc);
```
yxia92: Why do we need to change "SYSTABLEs" to "SYSTABLEPERMS"?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -315,6 +335,9 @@ protected void createDictionaryTables(Properties params,
         //create the Statistics tables
         createStatisticsTables(tc);
 
+        // TODO - this needs to be included into an upgrade script (JY)
```
yxia92: is //TODO comment here still needed, I see that there is logic added to upgradeSystables() for the new system table created?

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -915,4 +916,245 @@ public static void SET_PURGE_DELETED_ROWS (String schemaName, String tableName,
...
+                        "where s.schemaid=t.schemaid and " +
+                        "c.tableid = t.tableid and " +
+                        "c.schemaid=s.schemaid and " +
+                        "s.schemaname='%s' and " +
+                        "isindex=false and " +
+                        "isconstraint=false", schemaName);
+
+        String sql2 = String.format(
+                "select conglomeratename, conglomeratenumber " +
+                        "from sys.sysschemas s, sys.systables t, sys.sysconglomerates c " +
```
yxia92: Why systables is needed in this query?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
jyuanca: This should be increased earlier. For anyone who introduces new system tables, bump up this value.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
jyuanca: Yes.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
jyuanca: Good catch, thanks!

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
jyuanca: To make sure the indexes are associated to the table or schema.

yxia92:
@jyuanca Thanks for the explanation, I don't have further comments.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -915,4 +916,245 @@ public static void SET_PURGE_DELETED_ROWS (String schemaName, String tableName,
...
+    {
+        ensureSnapshot(snapshotName, true);
+        LanguageConnectionContext lcc = ConnectionUtil.getCurrentLCC();
+        TransactionController tc  = lcc.getTransactionExecute();
+        DataDictionary dd = lcc.getDataDictionary();
+        dd.startWriting(lcc);
+
+        PartitionAdmin admin=SIDriver.driver().getTableFactory().getAdmin();
+        String sql = String.format(
+                "select conglomeratenumber from sys.syssnapshots where snapshotname='%s'", snapshotName);
```
dgomezferro: Could we use PreparedStatement parameters rather than String.format()? I don't think this would be exploitable but I think we should do it the right way on production code.
Also for the other places the idiom is used.

jyuanca:
Addressed Daniel's commit.

PR #768, opened 2017-06-07
DB-5832: fix stats collection on external table

snowspeeder:
Fix for stats collection where it was missing partition by info for orc file type, for text file the index was zero based but for stats collection it has tobe one based.
(2.5 back port to follow)

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
@@ -519,9 +519,9 @@ public class SparkDataSetProcessor implements DistributedDataSetProcessor, Seria
             try {
                 table = SpliceSpark.getSession().read().csv(location);
 
-                if (op == null && baseColumnMap[0] == 1) {
+                if (op == null) {
                     // stats collection scan
-                    for(int index = 0; index< baseColumnMap.length; index++) {
+                    for(int index = 0; index < execRow.schema().fields().length; index++) {
                         StructField ft = table.schema().fields()[index];
                         Column cl = new Column(ft.name()).cast(execRow.schema().fields()[index].dataType());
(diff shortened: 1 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -513,19 +513,27 @@ private String intArrayToString(int[] ints) {
...
+        assert baseColumnMap != null:"baseColumnMap Null";
         try {
             Dataset<Row> table = null;
             try {
                 table = SpliceSpark.getSession().read().csv(location);
 
-
-                for( int index = 0; index< baseColumnMap.length; index++){
-                    if(baseColumnMap[index]!=-1){
+                if (op == null && baseColumnMap[0] == 1) {
```
yxia92: My understanding is that baseColumnMap parameter tracks the columns that need to collect stats. So baseColumnMap[0] may not always be 1.  Is the check "baseColumnMap[0]==1" appropriate? If collect stats is disabled for the first column, will your fix still be triggered?

snowspeeder:
Right, added that redundant check there, I’ve tested without it, we just
need to make sure that it is stats collection scan job and the index is all
one based

yxia92:
I mean the additional check may not be redundant and it could return false even for StatsOperation and restrict the fix from being kicked in when it should.
For example, if external table have 3 columns (a1, b1, c1), and stats collection on a1 is disabled, in that case, would baseColumnMap[0] be 2 instead of 1? If so, the fix won’t be triggered. Would that be a problem?

Thanks.

snowspeeder:
I am taking that check out, it is in-correct..testing more cases, will
update the PR soon

snowspeeder:
Please take a look at new changes, tested stat collection on textfile external table with the some columns disabled from the schema also.

PR #776, opened 2017-06-13
SPLICE-1675 merge partition stats at the stats collection time

yxia92:


db-engine/src/main/java/com/splicemachine/db/iapi/stats/TableStatisticsImpl.java
```diff
@@ -69,6 +70,13 @@ public TableStatisticsImpl(String tableId,
         assert partitionStatistics.size() > 0:"Partition statistics are 0";
         this.fallbackNullFraction = fallbackNullFraction;
         this.extraQualifierMultiplier = extraQualifierMultiplier;
+        isMergedStats = false;
+        if (partitionStatistics.size() > 0) {
+            if (partitionStatistics.get(0).getPartitionStatistics() != null) {
+                int statsType = partitionStatistics.get(0).getPartitionStatistics().getStatsType();
+                isMergedStats = statsType == 2 || statsType == 3;
+            }
```
jyuanca: Define an enum to be more readable

dgomezferro:
The merging should happen on Spark, otherwise we might run out of memory on the RegionServer.

I ran it by @jleach4 and he'd like to have both per region statistics plus the merged statistics stored in the catalog, so that later on we can recompute the merged statistics when a single region changes and for display purposes. The optimizer should only ever use the merged statistics.

yxia92:
@jyuanca , instead of using enum for the statsType, I defined a constant for each statsType as follows in SYSTABLESTATISTICSRowFactory.java:
/* Following are values supported in statsType */
 +    public static final int REGULAR_NONMERGED_STATS = 0;
 +    public static final int SAMPLE_NONMERGED_STATS = 1;
 +    public static final int REGULAR_MERGED_STATS = 2;
 +    public static final int SAMPLE_MERGED_STATS = 3;
And updated all files that originally used directly the value 0,1,2,3 with the constant name. This way, I avoid the conversion between enum and int type.

yxia92:
added upgrade logic in SpliceDataDictionary.java for the view definition change of the system view systablestatistics, which now can work on both merged and non-merged partition stats.

yxia92:
@dgomezferro @jleach4 , with regard to your comment "he'd like to have both per region statistics plus the merged statistics stored in the catalog, so that later on we can recompute the merged statistics when a single region changes and for display purposes. The optimizer should only ever use the merged statistics.", I would like to address them in a separate JIRA issue. Here are a few things to do or to clarify to meet this requirement:
1. With the fix of the current pull request, we allow either merged or non-merged stats in the dictionary, but not both at the same time. With both merged and non-merged stats in the dictionary, we need to change the logic that pulls stats from dictionary by applying a filtering condition "statsType=REGULAR_MERGED_STATS  or statsType=SAMPLE_MERGED_STATS";
2. What do we want the system view definition systablestatistics to display? Should it be computed based on the merged stats entry, or non-merged stats entries or should it return two records, one based on merged stats and one based on non-merged stats? 
3. I understand the purpose of keeping partition level(non-merged) stats, which is to support incremental stats refresh, if so, do we need some logic to keep the two versions of stats in-sync? 
Should we introduce at least an internal system procedure to merge stats from the partition level stats?
4. Currently, when we collect stats, we first drop the stats and then collect, what should the behavior be when we have two versions of the stats in the dictionary? Simple way is, always drop both versions and re-collect both versions. Once incremental stats is in place, an update of the partition-level stats should trigger the drop and recreation of merged stats. 

Please let me know if you are fine with the proposal of a separate JIRA to do the above. Thanks.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -712,6 +715,24 @@ private boolean needToUpgrade(Splice_DD_Version catalogVersion){
         return false;
     }
 
+
+    private void updateSysTableStatsView1(TransactionController tc) throws StandardException{
```
dgomezferro: I'm going to introduce the change we discussed earlier, adding the 'sprint number' to the version as part of the upgrade feature. Do you want to use that mechanism instead of comparing the view definition?

yxia92:
Yes, I want to use this mechanism. Long-time wise, it is the way to go. Thanks.

PR #777, opened 2017-06-13
DB-6012: avoid deleting a nonexist snapshot

jyuanca:
Add a hbase snapshot to a list after it is taken successfully. The list can be used to delete the snapshot if the txn needs to roll back.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
@@ -1180,7 +1180,6 @@ public class SpliceAdmin extends BaseAdminProcedures{
             String objectName = rs.getString(1);
             long conglomerateNumber = rs.getLong(2);
             String sname = snapshotName + "_" + conglomerateNumber;
-            snapshotList.add(sname);
             DateTime creationTime = new DateTime(System.currentTimeMillis());
             if (LOG.isDebugEnabled())
             {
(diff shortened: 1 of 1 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -1190,6 +1189,7 @@ private static void snapshot(ResultSet rs, String snapshotName, String schemaNam
                     new SnapshotDescriptor(snapshotName, schemaName, objectName, conglomerateNumber,creationTime, null);
             admin.snapshot(sname, "splice:" + conglomerateNumber);
             dd.addSnapshot(descriptor, tc);
+            snapshotList.add(sname);
```
snowspeeder: not sure how it helps, list add() is only moved down in the same loop, there is no condition

yxia92: My understanding is that if "admin.snapshot()" or "dd.addSnapshot()" fails, we should abort/return early and we won't reach the line "snapshotList.add(sname)". Later on, deleteSnapshots() in the below code won't delete a snapshot that fails to be created:

        try {
            dd.startWriting(lcc);

            ResultSet rs = getResultSet(sql1, schemaName, tableName);
            snapshot(rs, snapshotName, schemaName, dd, tc, snapshotList);

            rs = getResultSet(sql2, schemaName, tableName);
            snapshot(rs, snapshotName, schemaName, dd, tc, snapshotList);
        }
        catch (Exception e)
        {
            deleteSnapshots(snapshotList);
            throw e;
        }

snowspeeder:
Makes sense. Thanks!

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
jyuanca: Yes

PR #779, opened 2017-06-14
SPLICE-1681 Introduce query hint "skipStats" after a table identifier…

yxia92:
… to bypass fetching real stats from dictionary tables.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/StatsUsageIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/StatsUsageIT.java
@@ -78,6 +78,44 @@ public class StatsUsageIT extends SpliceUnitTest {
                 "call SYSCS_UTIL.COLLECT_SCHEMA_STATISTICS('%s',false)",
                 schemaName));
 
+        new TableCreator(conn)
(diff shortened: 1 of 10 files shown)


splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/StatsUsageIT.java
```diff
@@ -110,4 +148,99 @@ public void testSelectivityWithStats() throws Exception {
...
+        try(ResultSet resultSet = methodWatcher.executeQuery(query)){
+            int i = 1;
+            double outputRowWithStats = 0;
+            double outputRowWithoutStats = 0;
+            while(resultSet.next()){
+                if (i == 4) {
+                    /* line 4 scan table Y, which uses real stats, so outputRows should be 40960 */
+                    outputRowWithStats = parseOutputRows(resultSet.getString(1));
+                    Assert.assertEquals(outputRowWithStats, 40960, 10);
+                } if (i == 5) {
```
snowspeeder: minor: should be else if ()

yxia92: Thanks. I will correct.

yxia92: push the change to update the IT as suggested.

PR #784, opened 2017-06-19
DB-6001: update the error message when partial record is found (2.5)

snowspeeder:


snowspeeder:
Updated to print partial record, note that in the csv file there is ^M char which is making that field partial record, but not seen in the web git version here.

wtang16:
Please merge this to master as well.

PR #789, opened 2017-06-19
DB-6008: suppress false constraint violation during retry

jyuanca:


dgomezferro:
Would it be possible to add some tests? Or is it too hard to reproduce?

jyuanca:
@dgomezferro This is hard to reproduce.

pipeline_api/src/main/java/com/splicemachine/pipeline/constraint/UniqueConstraintChecker.java
```diff
@@ -48,7 +49,18 @@ public UniqueConstraintChecker(boolean isPrimaryKey,
     @Override
     public MutationStatus checkConstraint(KVPair mutation, DataResult existingRow) throws IOException {
         // There is an existing row, if this is an insert then fail.
```
dgomezferro: Update this comment with something similar to what's in UniqueConstraint.java please

jyuanca:
@dgomezferro added comments.

PR #791, opened 2017-06-20
SPLICE-774 Support upgrade from K2 (2.5)

dgomezferro:
Support new conglomerate format, store  V1 transaction threshold in ZK and use it when resolving transactions.


dgomezferro:
For the 2.0 part of it shouldn't be a problem. For the final 2.5 upgrade some changes might be required, good point.

PR #803, opened 2017-06-21
DB-6037: correct a query to find indexes of a table(2.5)

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -999,7 +999,8 @@ public static void SNAPSHOT_TABLE(String schemaName, String tableName, String sn
                         "isindex=false and " +
                         "isconstraint=false and " +
                         "(s.schemaname<>'SYS' or t.tablename<>'SYSSNAPSHOTS') and " +
-                        "t.tablename=?";
+                        "t.tablename=? and " +
+                        "t.tableid=c.tableid";
```
yxia92: line 1003 is a duplicate, there is already the same condition at line 996:
                        "c.tableid = t.tableid and " +

jyuanca:
removed

PR #807, opened 2017-06-22
SPLICE-1707 Fix Tail Recursion Issues

jleach4:
Remove Tail Recursion in Hot Paths.


jyuanca:
@jleach4 Can you explain why you want to eliminate tail recursion?

jleach4:
Good ?  Tail recursions have always been an issue with the java jvm.  The jit struggles to optimize underneath them.  In general, it is a structure we should avoid.

Regards,
John

PR #812, opened 2017-06-22
DB-6018: keep the column indexes zero based for new orc stats collection job

snowspeeder:
This is an issue in the master branch where new orc input format is being used. The basecolumn map indexes needed to be zero based than the one based indexes to match the expectation in the reader job. Other external table formats were also tested with the last column as a partitioning column, it was not having this array bounds problem while stats collection.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsOperation.java
@@ -93,8 +93,13 @@ public class StatisticsOperation extends SpliceBaseOperation {
                     statsDataSet = dsp.readTextFile(null, builder.getLocation(), builder.getDelimited(), null, builder.getColumnPositionMap(), null, null, null, builder.getTemplate(), useSample, sampleFraction);
                 else if (storedAs.equals("P"))
                     statsDataSet = dsp.readParquetFile(builder.getColumnPositionMap(), builder.getPartitionByColumnMap() , builder.getLocation(), null, null, null, builder.getTemplate(), useSample, sampleFraction);
-                else if (storedAs.equals("O"))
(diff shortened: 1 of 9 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsOperation.java
```diff
@@ -93,8 +93,13 @@ public StatisticsOperation(ScanSetBuilder scanSetBuilder, boolean useSample, dou
                     statsDataSet = dsp.readTextFile(null, builder.getLocation(), builder.getDelimited(), null, builder.getColumnPositionMap(), null, null, null, builder.getTemplate(), useSample, sampleFraction);
                 else if (storedAs.equals("P"))
                     statsDataSet = dsp.readParquetFile(builder.getColumnPositionMap(), builder.getPartitionByColumnMap() , builder.getLocation(), null, null, null, builder.getTemplate(), useSample, sampleFraction);
-                else if (storedAs.equals("O"))
-                    statsDataSet = dsp.readORCFile(builder.getColumnPositionMap(), builder.getPartitionByColumnMap(), builder.getLocation(), null, null, null, builder.getTemplate(), useSample, sampleFraction);
+                else if (storedAs.equals("O")) {
+                    int[] baseColumnMap = builder.getColumnPositionMap();
+                    for (int i = 0; i < baseColumnMap.length; i++) {
+                        baseColumnMap[i] = baseColumnMap[i] - 1;  //make it zero based
```
yxia92: There are places in code that use (baseColumnMap[I] == -1) to check if it is a valid column id or to terminate a loop over the baseColumnMap array. Unconditionally subtracting every element by 1 may cause confusion as we will see -2 in the cells.  Could you change the logic to only update "if (baseColumnMap[I] >0)".

snowspeeder:
makes sense, will update.

snowspeeder:
added a check before making zero based columnmap

snowspeeder:
Instead of passing adjusted columnmap, pass the type of job to orcNewInutFormat reader so it can adjust the index accordingly. Saw an issue with earlier approach in PartitionStatisticsImpl when we pass zero based columnmap. For stats collection, we need to keep that map one based.

PR #814, opened 2017-06-23
SPLICE-1690 Merge statistics on Spark (2.5)

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/stream/function/ReturnStatisticsFlatMapFunction.java
```diff
@@ -0,0 +1,35 @@
...
+public class ReturnStatisticsFlatMapFunction
+    extends SpliceFlatMapFunction<StatisticsOperation, Iterator<MergeStatisticsHolder>, LocatedRow> {
+
+    public ReturnStatisticsFlatMapFunction() {
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Iterator<LocatedRow> call(Iterator<MergeStatisticsHolder> holders) throws Exception {
+        return holders.next().toList();
```
yxia92: In the original logic, when we finish merging the column stats from all partitions, we call builder.getValue().terminate(), where builder is of type ColumnStatisticsMerge, but I don't see this function called in the new code. Is it no longer needed? I thought we still need it. Please correct me if I'm wrong.

yxia92: Please ignore this comment, I see that terminate() is called in ColumnStatisticsMerge.toExecRow().

dgomezferro:
@yxia92 Thanks for the review, good points. I've addressed your comments.

PR #816, opened 2017-06-23
DBAAS-608

jleach4:
Incorrect Configuration Setting.  This is supposed to sparingly used configuration when a customer migrates from LDAP to Native A and A.  Unfortunately, it is set to true by default.

This allows the first user to login to a region server to become the system schema and database owner.  [CRITICAL FIX]

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/AuthenticationConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/AuthenticationConfiguration.java
@@ -54,7 +54,7 @@ public class AuthenticationConfiguration implements ConfigurationDefault {
     // In this particular instance, there are Splice beta customers with AnA disabled and they want to
     // switch to using native AnA.  So we allow a manual override here.  See DB-2088 for more details.
     public static final String AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = "splice.authentication.native.create.credentials.database";
-    private static final boolean DEFAULT_AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = true;
+    private static final boolean DEFAULT_AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = false;
 
     public static final String AUTHENTICATION_CUSTOM_PROVIDER = "splice.authentication.custom.provider";
     public static final String DEFAULT_AUTHENTICATION_CUSTOM_PROVIDER = "com.splicemachine.derby.authentication.SpliceUserAuthentication";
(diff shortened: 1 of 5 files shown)


splice_access_api/src/main/java/com/splicemachine/access/configuration/AuthenticationConfiguration.java
```diff
@@ -54,7 +54,7 @@
     // In this particular instance, there are Splice beta customers with AnA disabled and they want to
     // switch to using native AnA.  So we allow a manual override here.  See DB-2088 for more details.
     public static final String AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = "splice.authentication.native.create.credentials.database";
-    private static final boolean DEFAULT_AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = true;
+    private static final boolean DEFAULT_AUTHENTICATION_NATIVE_CREATE_CREDENTIALS_DATABASE = false;
```
snowspeeder: There is also an xml property associated: splice.authentication.native.create.credentials.database, do we need to change that too as a default?

wtang16:
@jleach4 Following up on @snowspeeder's comment, here are the places that have the property set to be true (or in test expecting "true"):
**/Users/weitang/IdeaProjects/spliceengine/hbase_storage/src/test/resources/splice-site.xml**:

```
  <property>
    <name>splice.authentication.native.create.credentials.database</name>
    <value>true</value>
  </property>
```

**/Users/weitang/IdeaProjects/spliceengine/assembly/mapr5.1.0/src/main/resources/examples/conf/hbase-site.xml**:

`<property><name>splice.authentication.native.create.credentials.database</name><value>true</value></property>`

**/Users/weitang/IdeaProjects/spliceengine/assembly/mapr5.2.0/src/main/resources/examples/conf/hbase-site.xml**:
`<property><name>splice.authentication.native.create.credentials.database</name><value>true</value></property>`

**/Users/weitang/IdeaProjects/spliceengine/hbase_storage/src/test/java/com/splicemachine/access/HConfigurationTest.java**:
```
    key = "splice.authentication.native.create.credentials.database";
        value = auths.get(key);
        assertNotNull(value);
        assertEquals("true", value);
```

jleach4:
No, we want to store the username and password in database for now as default.

This configuration and naming leaves a lot to be desired.  Very confusing.  

JOHN

jleach4:
Oops, you are absolutely correct.

yxia92:
I updated the PR with the changes for the 4 files listed above. Since this is a blocker, I'll go ahead and merge the pull request.

jleach4:
Thanks yi...

PR #819, opened 2017-06-26
SPLICE-1716 Reduce Olap job latency for short tasks

dgomezferro:
Block on the server the first time we request the status of a job, if
it's short it will be available before we return to the client, lowering
latency for short jobs.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/olap/OlapStatusHandler.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/OlapStatusHandler.java
@@ -50,6 +50,8 @@ public class OlapStatusHandler extends AbstractOlapHandler{
                 case SUBMITTED:
                 case RUNNING:
                     status.wait(waitTime, TimeUnit.MILLISECONDS);
+                default:
+                    // fall-through, send response without blocking
             }
         }
 
--- a/hbase_sql/src/main/java/com/splicemachine/olap/OlapStatusHandler.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/OlapStatusHandler.java
@@ -14,10 +14,9 @@
 
 package com.splicemachine.olap;
 
-import com.splicemachine.olap.OlapMessage;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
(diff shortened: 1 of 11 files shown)


dgomezferro:
@tgildersleeve , @jleach4  could you try this fix and see if you can still reproduce the timeout/hanging issue?  I wasn't able to reproduce it on my laptop, but I think this should fix it for you. Please let me know. Thanks!

jleach4:
@tilgersleeve, can you validate?

snowspeeder:
@tgildersleeve did you get chance to validate this?

tgildersleeve:
@snowspeeder SPLICE-1716-2 still did not work on my computer, but it and the original works intermittently on John's

dgomezferro:
I'm giving up on @tgildersleeve 's laptop, sorry. I removed the timeouts from the tests and they should complete, albeit slowly.

I've upgraded the OlapServer to Netty 4 (client was using 4 and server 3) but it didn't make a difference.

snowspeeder:
@dgomezferro Do we still want this PR in or close it?

dgomezferro:
Yes, I'd like to merge it.

hbase_sql/src/main/java/com/splicemachine/olap/OlapStatusHandler.java
```diff
@@ -31,22 +33,32 @@ public OlapStatusHandler(OlapJobRegistry registry){
...
         OlapJobStatus status = jobRegistry.getStatus(cmd.getUniqueName());
-        writeResponse(e,cmd.getUniqueName(),status);
+
+        OlapMessage.Status statusMsg = cmd.getExtension(OlapMessage.Status.command);
+        if (statusMsg.hasWaitTimeMillis() && status != null) {
+            long waitTime = statusMsg.getWaitTimeMillis();
+            switch (status.checkState()) {
+                case SUBMITTED:
+                case RUNNING:
+                    status.wait(waitTime, TimeUnit.MILLISECONDS);
```
snowspeeder: need default case?

OlegMazurov:
See http://206.225.8.98:8080/job/spliceengine-master/job/spliceengine-platform/834/ for compilation errors.

PR #820, opened 2017-06-26
SPLICE-1692 Perform (anti)tombstone txn resolution only when needed

dgomezferro:
We only care about the most recent (anti)tombstone, so we shouldn't waste time performing txn resolution for older tombstones when we already have one that we can see.

### DIFF
--- a/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
+++ b/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
@@ -14,7 +14,6 @@
 
 package com.splicemachine.si.impl;
 
-import com.carrotsearch.hppc.LongArrayList;
 import com.carrotsearch.hppc.LongOpenHashSet;
 import com.splicemachine.si.api.filter.RowAccumulator;
 import com.splicemachine.si.api.filter.TxnFilter;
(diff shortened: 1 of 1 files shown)


splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -227,25 +227,22 @@ private TxnView fetchTransaction(long txnId) throws IOException{
...
-        if(isVisible(txnId)){
-			/*
-			 * We can see this anti-tombstone, hooray!
-			 */
-            if(!tombstonedTxnRows.contains(txnId))
-                antiTombstonedTxnRows.add(txnId);
+        /*
+         * Check if we can see this anti-tombstone
+         */
+        if(antiTombstonedTxnRows.isEmpty() && !tombstonedTxnRows.contains(txnId) && isVisible(txnId)){
```
OlegMazurov: Checking for antiTombstonedTxnRows.isEmpty() is new logic here. Effectively, we can have only one value in antiTombstonedTxnRows. Is this intended?

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -227,25 +227,22 @@ private TxnView fetchTransaction(long txnId) throws IOException{
...
+        long txnId=data.version();
 		/*
 		 * Only add a tombstone to our list if it's actually visible,
 		 * otherwise there's no point, since we can't see it anyway.
 		 */
-        if(isVisible(txnId)){
-            if(!antiTombstonedTxnRows.contains(txnId))
-                tombstonedTxnRows.add(txnId);
-        }
+        if(tombstonedTxnRows.isEmpty() && !antiTombstonedTxnRows.contains(txnId) && isVisible(txnId))
```
OlegMazurov: Ditto.

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
dgomezferro: Yes, that was intentional. I simplified the code as you suggested, thanks!

PR #832, opened 2017-06-29
SPLICE-1728 Fix Cache Miss

jleach4:
DIctionary scans go crazy during DDL.


jleach4:
Closing, moving to a different approach.

PR #833, opened 2017-06-29
SPLICE-1728 Fix Cache Miss

jleach4:
DIctionary scans go crazy during DDL.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
@@ -270,7 +270,7 @@ public class DataDictionaryCache {
     }
 
     public void conglomerateCacheAdd(Long conglomId, Conglomerate conglomerate,TransactionController xactMgr) throws StandardException {
-        if (!dd.canUseCache(xactMgr))
+        if (!dd.canUseCache(xactMgr)  && conglomId>=DataDictionary.FIRST_USER_TABLE_NUMBER)
             return;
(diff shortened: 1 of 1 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -270,7 +270,7 @@ public Conglomerate conglomerateCacheFind(Long conglomId) throws StandardExcepti
     }
 
     public void conglomerateCacheAdd(Long conglomId, Conglomerate conglomerate,TransactionController xactMgr) throws StandardException {
-        if (!dd.canUseCache(xactMgr))
+        if (!dd.canUseCache(xactMgr)  && conglomId>=DataDictionary.FIRST_USER_TABLE_NUMBER)
```
dgomezferro: We can't rely on DataDictionary.FIRST_USER_TABLE_NUMBER to separate user from system conglomerates. When upgrading we sometimes add new system tables and that would make user tables fall inside the "system tables" category. I would add a method that checks a conglomId against the coreInfo and nonCoreInfo arrays.

jleach4: Ugh...  Not good.

jleach4: EnvUtils#getTableType has the same issue.

jleach4:
Closing...  Opening another pull request with a different approach.

PR #835, opened 2017-06-29
SPLICE-1729 Handle 'drop table table_name if exists'

tgildersleeve:
We currently handle 'drop table if exists table_name'
This commit allows for 'drop table table_name if exists' to work as well

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/DropTableIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/DropTableIT.java
@@ -1,0 +1,117 @@
+package com.splicemachine.derby.impl.sql.execute.operations;
+
+
+import com.splicemachine.derby.test.framework.SpliceWatcher;
+import org.junit.*;
+import com.splicemachine.homeless.TestUtils;
+import java.sql.*;
+
+/**
+ * Created by tgildersleeve on 6/29/17.
+ * SPLICE-1729
+ *
+ * Tests that the following queries execute successfully, both when a table exists and does not exist:
+ *
+ * drop table table_name
+ * drop table if exists table_name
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/DropTableIT.java
```diff
@@ -0,0 +1,117 @@
...
+ * Tests that the following queries execute successfully, both when a table exists and does not exist:
+ *
+ * drop table table_name
+ * drop table if exists table_name
+ * drop table table_name if exists
+ *
+ * _true tests test existing tables; _false tests test nonexistent tables
+ */
+public class DropTableIT {
+
```
jleach4: Can you run the test twice?

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/DropTableIT.java
```diff
@@ -0,0 +1,117 @@
...
+    public SpliceWatcher methodWatcher = new SpliceWatcher();
+
```
jleach4: Need to create a schema referencing the class.  Other tests could step on you.

mzweben:
Would be helpful to do this with views too

tgildersleeve:
That's next up! SPLICE-398 :)

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/DropTableIT.java
tgildersleeve: yes, given that they all pass, which they do.

genedavis:
Any new files need our Copyright header at the top of the file

PR #839, opened 2017-06-29
Splice 1732 - IDENTITY_VAL_LOCAL() test does not throw error: Works in petclinic

Ben-Epstein:


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/sequence/SpliceSequenceIT.java
```diff
@@ -79,13 +79,15 @@ public void testShortSequence() throws Exception {
...
+    @Test
     public void testIdentityValLocal() throws Exception{
         methodWatcher.executeUpdate(String.format("create table t1(c1 int generated always as identity, c2 int)"));
         methodWatcher.executeUpdate(String.format("insert into t1(c2) values (8)"));
         methodWatcher.executeUpdate(String.format("insert into t1(c2) values (IDENTITY_VAL_LOCAL())"));
 
-
+        methodWatcher.executeUpdate(String.format("create table t2(c1 int generated always as identity(start with 1, increment by 1), c2 int)"));
+        methodWatcher.executeUpdate(String.format("insert into t2(c2) values (8)"));
+        methodWatcher.executeUpdate(String.format("insert into t2(c2) values (IDENTITY_VAL_LOCAL())"));
```
snowspeeder: please add more data and check the results, see: https://db.apache.org/derby/docs/10.8/ref/rrefidentityvallocal.html

dgomezferro:
This probably needs some more work for it to work in Spark, I don't think it will be trivial to support.

jleach4:
Search for tests where --splice-properties useSpark=true

snowspeeder:
@Ben-Epstein could you re-open this pull request after spark problem fixed, please close PR in the meantime.

PR #841, opened 2017-06-29
SPLICE-1733 Support Type Conversion

Ben-Epstein:
Varchar to Int type supported, Int to varchar not supported: Tests created and @ignored

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperationIT.java
@@ -14,6 +14,7 @@
 
 package com.splicemachine.derby.impl.sql.execute.operations;
 
+import com.splicemachine.db.iapi.reference.SQLState;
 import org.spark_project.guava.collect.Maps;
 import com.splicemachine.derby.test.framework.SpliceSchemaWatcher;
 import com.splicemachine.derby.test.framework.SpliceWatcher;
@@ -30,6 +31,7 @@ import java.util.*;
 
 import static com.splicemachine.derby.test.framework.SpliceUnitTest.getBaseDirectory;
 import static com.splicemachine.derby.test.framework.SpliceUnitTest.getResourceDirectory;
+import static com.splicemachine.derby.test.framework.SpliceUnitTest.rowsContainsQuery;
 
 public class InsertOperationIT {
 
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperationIT.java
```diff
@@ -88,6 +89,74 @@ public static void createSharedTables() throws Exception {
...
+        classWatcher.executeUpdate("insert into varcharTab (c1, c2) select col1, col2 from intTab");
+
+        //int to char
+        classWatcher.executeUpdate("insert into charTab (c1, c2) select col1, col2 from intTab");
+
+        //bigint to varchar
+        classWatcher.executeUpdate("insert into varcharTab (c1, c2) select col1, col2 from bigintTab");
+
+        //bigint to char
+        classWatcher.executeUpdate("insert into charTab (c1, c2) select col1, col2 from bigintTab");
```
snowspeeder: should verify values inserted into the char fields?

dgomezferro:
@jleach4 why do we want this? I think it's better to have an explicit cast rather than an implicit conversion...

jleach4:
@dgomezferro Unfortunately we are blocking unica support since neteeza and oracle support this implicit casting.  We are blocked on a POC.

wtang16:
If it's for a POC, I suggest that this goes to a dev branch instead of master.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperationIT.java
yxia92: Can you add a negative test case to insert a varchar with value "abc" to int? I expect that runtime error will be reported as we cannot convert "abc" to an integer, so the query should be aborted gracefully.

Ben-Epstein: I ran the negative case and received java.sql.SQLDataException: Invalid character string format for type int.

splice1732.branch
```diff
@@ -0,0 +1,45 @@
+diff --git a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```
snowspeeder: added wrong file

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperationIT.java
snowspeeder: what happens if you insert too bigint value to a char(1) or char(2) col, does it get error or truncated, not sure what the semantics are but can compare with other database.

jleach4:
@wtang16 It handles your case with a SQL exception.  @Ben-Epstein created a test case to handle it.  This is an incremental commit that supports more implicit conversion on writes (we already support it for filters, etc.).

dgomezferro:
@wtang16  asked me to chime in here with my worries.

My main concern with implicit type conversion is that it will be surprising at one point or another. The prime example is Javascript with its implicit conversions: http://www.informit.com/articles/article.aspx?p=1997934&seqNum=3 
https://gist.github.com/MichalZalecki/c964192f830360ce6361

I agree that some amount of implicit conversions is desirable and user friendly, but we should be very careful when expanding the amount of conversions we make and I don't feel we were careful at all in this case. 

Here some examples with implicit conversions and surprising results in SQL:
https://mysoftwarenotes.wordpress.com/2010/03/19/the-danger-of-implicit-data-type-conversions-in-sql-server/
https://www.mssqltips.com/sqlservertip/1734/convert-implicit-and-the-related-performance-issues-with-sql-server/

PR #845, opened 2017-06-30
SPLICE-1728 Fixing Data Dictionary Caching

jleach4:
Based on Daniel's recommendation to lose the less than range concept on dictionary conglomerates since that will not be applicable in the case of a system upgrade.  Very good catch.  I implicitly check the conglomerates now in the dictionary.  Also, added a field to memDatabase to speed those tests up a bit with regards to caching.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
@@ -31,6 +31,8 @@
 
 package com.splicemachine.db.impl.sql.catalog;
 
+import com.carrotsearch.hppc.LongOpenHashSet;
+import com.splicemachine.db.impl.db.BasicDatabase;
 import com.splicemachine.db.impl.sql.execute.ValueRow;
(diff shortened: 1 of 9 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -6160,7 +6163,7 @@ public void run(){
                 // to allow it to be garbage-collected. The idea
                 // is that a running database might never need to
                 // reference a non-core table after it was created.
-                clearNoncoreTable(noncoreCtr);
```
dgomezferro: Could you remove the comment and the function too?

jleach4: Removed

dgomezferro:
It looks like on a reboot the nonCoreInfo array is not eagerly populated, so we'd mislabel a lot of system tables as user tables. Should we eagerly populate it before caching the conglomerates in the new structure?

jleach4:
Ugh...  I hate the dictionary.

PR #848, opened 2017-06-30
SPLICE-1728 Data Dictionary Cache Misses

jleach4:


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -7339,6 +7342,8 @@ protected TabInfoImpl getNonCoreTIByNumber(int catalogNumber) throws StandardExc
             }
             initSystemIndexVariables(retval);
             noncoreInfo[nonCoreNum]=retval;
+            // Lazy Path for database re-boot
+            populateSystemConglomerates(new TabInfoImpl[]{retval});
```
dgomezferro: I don't think this is enough, at least for Vacuum. For vacuum we access the list of HBase conglomerates and then we'd use BasicDatabase.isSystemConglomerate() API, but some of these conglomerates might not have been lazily loaded.

dgomezferro: I don't think this is enough, at least for Vacuum. For vacuum we access the list of HBase conglomerates and then we'd use BasicDatabase.isSystemConglomerate() API, but some of these conglomerates might not have been lazily loaded.

jleach4: Good catch.  Sounds like lazy load has to go.

PR #849, opened 2017-06-30
DB-1721 Performance updates and code cleanup

Ben-Epstein:


OlegMazurov:
The issue is SPLICE-1721, not DB-1721. It will be confusing in history and comments

snowspeeder:
I just merged; no Jira was mentioned in the commen., Ben you should put the
Jira ticket number before the checkin always. Please update the Jira with
the checkin in master. Thanks

PR #855, opened 2017-07-03
SPLICE-1739: Added CREATE SCHEMA IF NOT EXISTS functionality

Ben-Epstein:
You can now run CREATE SCHEMA <schemaname> IF NOT EXISTS as well as 
CREATE SCHEMA IF NOT EXISTS <schemaname> 
A warning will throw if the schema exists, else it will create the schema.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -11492,6 +11492,13 @@ schemaDefinition() throws StandardException :
 {
 	String	schemaName = null;
 	String	authName = null;
+	Token ifTok = null;
+	Token notTok = null;
+    Token existsTok = null;
+    Token ifTok2 = null;
+    Token notTok2 = null;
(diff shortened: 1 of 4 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -15391,8 +15403,8 @@ dropPinStatement() throws StandardException :
 											C_NodeTypes.DROP_PIN_NODE,
 											tableName,
 											(ifExists ?
-											  new Integer(StatementType.DROP_IF_EXISTS) :
-											  new Integer(StatementType.DROP_DEFAULT)),
+											  new Integer(StatementType.CREATE_IF_NOT_EXISTS) :
+											  new Integer(StatementType.CREATE_DEFAULT)),
```
dgomezferro: This shouldn't have been changed, right?

Ben-Epstein: Yes, updated

PR #862, opened 2017-07-05
SPLICE-1732: identity_val_local working with and without useSpark

Ben-Epstein:


tgildersleeve:
looks good

snowspeeder:
Ben, the test coverage for this function looks very poor, could you take a look at the derby test below and add something equivalent in our IT, with and without spark. 
.//db-testing/src/test/java/com/splicemachine/dbTesting/functionTests/tests/lang/AutoIncrementTest.java

dgomezferro:
Following up on @snowspeeder's comment, you should check the return value of the IDENTITY_VAL_LOCAL function. On the current tests I think the insert running on Spark won't behave as expected (i.e., it won't insert the value generated on the previous insert running in control mode).

Another case running in Spark would be to have the first insert in Spark and then using the IDENTITY_VAL_LOCAL function in control mode, which again I don't think will behave as expected.

Ben-Epstein:
Test cases added using suggestions, @dgomezferro @snowspeeder with useSpark and checking return values

PR #863, opened 2017-07-06
DB-5860 Resubmit to Spark if we consume too many resources in control

dgomezferro:
We add ControlExecutionLimiter that keeps count of how many rows are buffered in memory during 'control' execution. When the configured limit is surpassed an exception is raised to limit the memory impact on the RegionServer. If possible, we resubmit the query to Spark. The only case where this is not possible is when we have already returned some rows to the user.

We have to keep track of buffered rows mainly in 3 places:
- ControlDataSet 
- ControlPairDataSet
- BroadcastJoin

This mechanism is disabled when running in Spark, because sometimes we run in 'control-mode' from Spark itself (for instance, when executing a nestedloop join in Spark)


genedavis:
What is the configured limit and how do I change it?  I'm nervous about certain queries that SHOULD go through control that might get kicked out (read: Winnie)

snowspeeder:
Looks like default  DEFAULT_CONTROL_EXECUTION_ROWS_LIMIT = 2000000;
But not sure, if/how we could change dynamically.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
```diff
@@ -428,10 +459,24 @@ public void reopenCore() throws StandardException{
...
+                try {
+                    locatedRow = execRowIterator.next();
+                    if (LOG.isTraceEnabled())
+                        SpliceLogUtils.trace(LOG, "getNextRowCore %s locatedRow=%s", this, locatedRow);
+                    ExecRow result = locatedRow;
+                    returnedRows = true;
+                    return result;
+                } catch (ResubmitDistributedException e) {
+                    if (!returnedRows) {
+                        LOG.warn("The query consumed too many resources running in control mode, resubmitting in Spark");
```
snowspeeder: Is this case possible: query returned few rows from control and then it resubmitted the job to spark, which got all of the rows, but that would make more rows than correct number of rows.

dgomezferro: If I understand your question correctly, no, it's not possible: as soon as the control query returns some values we can no longer resubmit to Spark, we raise an exception and ask the user to submit to Spark himself. That's guarded by the `returnedRows` field.

dgomezferro:
@genedavis , as @snowspeeder said it is controlled by the `CONTROL_EXECUTION_ROWS_LIMIT` parameter. I haven't added a dynamic way to control it (via query hints or otherwise), do you think that's necessary? 

The resubmission should only kick in when lots of rows have been buffered in memory, in that case running it in control mode doesn't seem correct. Currently we don't take into account "unbuffered" rows, which could trigger the resubmission even if not many rows are buffered at one point, but I think that case shouldn't be too common. I can also make sure we take into account unbuffered rows to reduce the false positives.

genedavis:
I was just hoping that worst case we can tweak that parameter up or down if needed without having to get a new build.  Just as a Splice variable might make sense.  Generally good practice with magic numbers?

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -3121,6 +3121,11 @@ Guide.
             </msg>
 
             <msg>
+                <name>42ZD0</name>
+		<text>Surpassed limit of proces rows in control mode, plase resubmit as Spark query</text>
```
jyuanca: typos error message

splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
```diff
@@ -210,6 +210,9 @@
     public static final String NESTEDLOOPJOIN_BATCH_SIZE = "splice.nestedLoopJoin.batchSize";
     private static final int DEFAULT_NESTEDLOOPJOIN_BATCH_SIZE = 10;
 
+    public static final String CONTROL_EXECUTION_ROWS_LIMIT = "splice.controlExecution.rowsLimit";
+    private static final int DEFAULT_CONTROL_EXECUTION_ROWS_LIMIT = 2000000;
+
```
jyuanca: The threshold to execute on spark is 20K rows. DEFAULT_CONTROL_EXECUTION_ROWS_LIMIT is 100X larger, is it too big?

dgomezferro: When we start to run in control mode I think it makes sense to continue doing so until we start having memory issues. This size is in line with the maximum broadcast size, so I think it's ok. What size would you suggest?

snowspeeder:
changing CONTROL_EXECUTION_ROWS_LIMIT dynamically would help, should add it.

dgomezferro:
@snowspeeder you mean dynamically with a splice-property ? If you are using splice-properties you might as well choose control or spark execution yourself and don't rely on this mechanism at all. This is not intended to improve on the CostChoosing data set processor, it's just a protection mechanism that takes care of rogue queries, I'm can't think of an use case where someone would be playing with the threshold.

snowspeeder:
Not using  splice property per query, but I was thinking with some
configurable mechanism may be through hbase configuration, if there is a
need to change such threshold.

dgomezferro:
Oh, I see, that's already possible via the splice.controlExecution.rowsLimit property. I misunderstood what you were referring to, sorry.

snowspeeder:
np, if there is a way to tweak this limit, then its cool.
+1

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
```diff
@@ -1339,5 +1339,16 @@ void setReferencedColumnMap(TableDescriptor td,
 	 */
```
jleach4: LCC is a nice place to put this...

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -308,8 +327,8 @@ public long count() {
 
     @SuppressWarnings({ "unchecked", "rawtypes" })
     @Override
-    public DataSet< V> intersect(DataSet< V> dataSet) {
-        return intersect(dataSet,"Intersect Operator",null,false,null);
+    public DataSet< V> intersect(DataSet<V> dataSet, OperationContext context) {
```
jleach4: Nice cleanup of passing in the context.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
jleach4: Hmm, my only concern is we have zip for resource management on the hbase side.  I am fine either way, my vote would favor stability (smaller size).

dgomezferro: Makes sense to have a smaller size. What would you suggest? 100K rows? I was afraid some of the OLTP queries would end up rejected by this mechanism. Should we disable this if the user forces control execution?

dgomezferro:
@jleach4 I'm counting the buffered rows, not the read rows. In theory a straight scan in control-mode shouldn't be too hard on the region server even if it's big, and it also takes care of the joins etc., yes.

I do not mark the activation, good point. I'll do that. Please hold on merging until these 2 issues are solved.

dgomezferro:
I'll address the issues in a follow up PR

OlegMazurov:
Oops, I didn't notice the last comment. Marking the  PR with 'changes requested' would be better.
Let's go with a follow up PR.

OlegMazurov:
This PR broke the build: ExecutorService is not defined in ControlDataSet. I'm reverting it.
@dgomezferro, you can fix it in the new PR you are preparing.

PR #868, opened 2017-07-07
SPLICE-1763 Fix select from partitioned orc table error

tgildersleeve:
previously you couldn't select from an orc table that was partitioned by any column other than the last column.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
@@ -111,6 +111,8 @@ public interface ExecRow extends Row, KeyableRow, org.apache.spark.sql.Row, Comp
 
 	StructType createStructType();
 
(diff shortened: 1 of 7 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
```diff
@@ -111,6 +111,8 @@
 
 	StructType createStructType();
```
jleach4: Why is this required vs. createStructType?  Not clear from comments or code.

tgildersleeve: Depending on which columns you select from an orc table, createStructType can return the incorrect columns because it only has access to the total number N of columns selected, and returns the first N columns in the table rather than the correct N. Passing in ColumnBaseMap to createStructTypeCorrected allows the correct N columns to be returned.

PR #869, opened 2017-07-10
SPLICE-1749: fix delete over nestedloop join

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
```diff
@@ -108,8 +109,9 @@ private void loadBatch() throws StandardException {
                 currentOperationContext = result.getFirst();
                 rightSideNLJIterator = result.getSecond();
                 leftRow = currentOperationContext.getOperation().getLeftOperation().getCurrentRow();
-                leftRowLocation = currentOperationContext.getOperation().getLeftOperation().getCurrentRowLocation();
+                leftRowLocation = new HBaseRowLocation(leftRow.getKey());
                 operationContext.getOperation().getLeftOperation().setCurrentRow(getLeftLocatedRow());
+                operationContext.getOperation().getLeftOperation().setCurrentRowLocation(getLeftRowLocation());
```
yxia92: Since we now get current row location directly from the current row,  is this line a must for the fix of this wrong result? Or is this just a proactive action for other scenarios that would access the currentRowLocation?

jyuanca: It's for correctness of delete over nested loop join. row location is still needed and needs to be populated.

yxia92: Got it. Thanks.

PR #875, opened 2017-07-10
SPLICE-1621 Fix select from partitioned orc table error

tgildersleeve:
previously, selecting columns from a partitioned orc table would throw an error if it was partitioned by any column other than the last. in ValueRow, createStructType() could return columns that weren't selected. OrcRecordReader had some index issues where the partitioned column would be listed last in the query results, and the true last column would be chopped off.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
@@ -109,9 +109,7 @@ public interface ExecRow extends Row, KeyableRow, org.apache.spark.sql.Row, Comp
 	 */
 	void getNewObjectArray();
 
-	StructType createStructType();
-
-	StructType createStructTypeCorrected(int[] baseColumnMap);
+	StructType createStructType(int[] baseColumnMap);
 
 	org.apache.spark.sql.Row getSparkRow();
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
@@ -111,6 +111,8 @@ public interface ExecRow extends Row, KeyableRow, org.apache.spark.sql.Row, Comp
 
 	StructType createStructType();
 
+	StructType createStructTypeCorrected(int[] baseColumnMap);
+
 	org.apache.spark.sql.Row getSparkRow();
 
 	ExecRow fromSparkRow(org.apache.spark.sql.Row row);
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/ValueRow.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/ValueRow.java
@@ -45,6 +45,7 @@ import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.services.io.FormatableBitSet;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.types.DataValueDescriptor;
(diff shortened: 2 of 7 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/ValueRow.java
```diff
@@ -623,6 +623,21 @@ public StructType createStructType() {
...
+	@Override // SPLICE-1621
+	public StructType createStructTypeCorrected(int[] baseColumnMap) {
+		StructField[] fields = new StructField[length()];
+		int j = 0;
+		for(int i = 0; i < baseColumnMap.length; i++){
+			if (baseColumnMap[i] != -1){
+				// put the selected columns, designated in baseColumnMap, in field
+				fields[j] = getColumn(j+1).getStructField(getNamedColumn(i));
+				j++;
+			}
```
snowspeeder: what happens to the remaining array for fields if the numcols mismatch between baseColumnMap and fields[], should it to get initialized to -1 or something similar.

tgildersleeve: fields[] always becomes completely filled. the baseColumnMap columns that = -1 are ignored

tgildersleeve: so the number of columns in baseColumnMap that are not -1 is the same as length() and fields.length

snowspeeder:
Ok, then should you add an assert check there just so that there are no
array bound issues.

hbase_sql/src/main/java/com/splicemachine/orc/OrcRecordReader.java
```diff
@@ -460,10 +446,21 @@ private void advanceToNextStripe()
     {
         List<StreamDescriptor> streamDescriptors = createStreamDescriptor("", "", 0, types, orcDataSource).getNestedStreams();
         OrcType rowType = types.get(0);
-        StreamReader[] streamReaders = new StreamReader[rowType.getFieldCount()];
-        for (int columnId = 0; columnId < rowType.getFieldCount(); columnId++) {
+        StreamReader[] streamReaders = new StreamReader[rowType.getFieldCount() + 1]; // + 1 is to compensate for a missing partitioned column
```
dgomezferro: What happens with multiple partitioned columns?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ExecRow.java
```diff
@@ -111,6 +111,8 @@
 
 	StructType createStructType();
 
+	StructType createStructTypeCorrected(int[] baseColumnMap);
```
dgomezferro: createStructType() isn't used anywhere else, right? You should just overwrite it, not add a new method.

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTablePartitionIT.java
```diff
@@ -0,0 +1,667 @@
...
+            Assert.assertEquals("COL2 |COL3 |\n" +
+                    "------------\n" +
+                    "  2  | AAA |\n" +
+                    "  4  | BBB |\n" +
+                    "  6  | CCC |",TestUtils.FormattedResult.ResultFactory.toString(rs4));
+        } catch (SQLException e) {
+            Assert.fail("An exception should not be thrown");
+        }
+    }
+
```
dgomezferro: Add test with multiple partitioned columns.

tgildersleeve: oops!! thank you so much

PR #876, opened 2017-07-10
SPLICE-1773 Unifying the thread pools

jleach4:
Unifies Thread Pool Execution.


splice_machine/src/main/java/com/splicemachine/EngineDriver.java
```diff
@@ -105,6 +105,8 @@ public void close(SpliceSequence entity) throws Exception{
                     return t;
                 },
                 new ThreadPoolExecutor.CallerRunsPolicy());
+        ((ThreadPoolExecutor)threadPool).allowCoreThreadTimeOut(false);
```
OlegMazurov: This redundantly reenforces the default.

PR #877, opened 2017-07-11
SPLICE-1764 Fix select from partitioned parquet and avro tables

tgildersleeve:
previously, selecting from partitioned parquet and avro tables could throw an error if the tables were partitioned by a column other than the last column. This was because the schema containing the list of selected columns would return out of order, with the partitioned column always coming last.


tgildersleeve:
Testing for this is in SPLICE-1621- fix select from partitioned orc tables

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreateTableConstantOperation.java
```diff
@@ -458,12 +458,27 @@ protected void createTable(Activation activation) throws StandardException {
...
+                                for (int j = fixSchema.length - 1; j > i; j--){
+                                    fixSchema[j] = fixSchema[j-1];
+                                }
+                                fixSchema[i] = partitionCol;
+                                break;
+                            }
+                        }
+                    }
+
                     // test types equivalence. Make sure that the type defined correspond
```
jyuanca: This seems to be duplicate code too.

tgildersleeve: It's in a different file, is it okay to leave as is?

tgildersleeve:
-need to allow for multiple partitioned columns, new commit coming in asap

tgildersleeve:
-one more commit coming in

PR #882, opened 2017-07-12
DB-5934: restore a chain of backup

jyuanca:
- Given an incremental backup Id, find all of it parent backups and restore all of them.
- Added more logging
- pass hbase.mapreduce.bulkload.max.hfiles.perRegion.perFamily to restore job
- ignore unfound txn during compaction and txn filtering. From a restored database, some data may be wriiten by a txn that cannot be found in txn table.

Please also review ee branch.


db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
```diff
@@ -1955,6 +1955,7 @@ The message identifier (ie. the value that matches a field in this class)
 	String PARENT_BACKUP_MISSING                                   = "BR013";
 	String EXISTS_CONCURRENT_BACKUP                                = "BR014";
 	String BACKUP_TIMEOUT                                          = "BR016";
+	String INCOMPATIBLE_BACKUP                                     = "BR017";
```
snowspeeder: where are we throwing this exception?

PR #883, opened 2017-07-12
DB-5934: restore a chain of backup(2.5)

jyuanca:
- Given an incremental backup Id, find all of it parent backups and restore all of them.
- Added more logging
- pass hbase.mapreduce.bulkload.max.hfiles.perRegion.perFamily to restore job
- ignore unfound txn during compaction and txn filtering. From a restored database, some data may be wriiten by a txn that cannot be found in txn table.

Please also review ee branch.


jyuanca:
@pushed

PR #891, opened 2017-07-13
SPLICE-1788 Mem Database should traverse the same cache as HBase

jleach4:
Without this, memory database is worthless in understanding cache misses.


jyuanca:
@jleach4 backport to 2.5?

PR #895, opened 2017-07-14
SPLICE-1372 List and kill running operations

dgomezferro:
This adds an OperationManager, when an operation is opened it registers itself on the OperationManager and unregisters when it is closed.

There's two new system procedures:
- SYSCS_UTIL.SYSCS_GET_RUNNING_OPERATIONS(): lists all running operations for the current user (or for all users if it is the db owner)
- SYSCS_UTIL.SYSCS_KILL_OPERATION(): kills an operation (either control-mode or spark)

The killing works by closing the operation and interrupting the DRDAConnThread associated with it.

The procedures are local to the region server, they only work on operations running on the given region server.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/BasicNoPutResultSetImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/BasicNoPutResultSetImpl.java
@@ -601,6 +601,11 @@ implements NoPutResultSet
 	    return ( ! isOpen );
 	}
 
+	@Override
+	public boolean isKilled() {
+		return false;
+	}
+
 	public void	finish() throws StandardException
 	{
 		finishAndRTS();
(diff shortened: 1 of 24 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/BasicNoPutResultSetImpl.java
```diff
@@ -601,6 +601,11 @@ public boolean	isClosed() {
 	    return ( ! isOpen );
 	}
 
+	@Override
+	public boolean isKilled() {
+		return false;
+	}
```
Ben-Epstein: Why does isKilled() return false every time? What is its benefit?

jyuanca:
It would be better if the user can view all running operations for all nodes. This can be an enhancement as a follow-up.

PR #902, opened 2017-07-14
SPLICE-1718 Fixes TCP/IP Timeout and Sets Default to 60 seconds

jleach4:
Fixes TPC/IP Timeouts from JDBC Client.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
@@ -86,7 +86,11 @@ public class SpliceDatabase extends BasicDatabase{
     public void boot(boolean create,Properties startParams) throws StandardException{
         Configuration.setConfiguration(null);
         SConfiguration config = SIDriver.driver().getConfiguration();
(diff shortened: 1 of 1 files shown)


MurrayBrown:
Doesn't this setting translate to the 'socketTimeout' that is set on the sockets which our server opens? I think that means we'll never be able to run any query that takes longer than 60 seconds to return date.

If we *have* to set a value to improve the setting, until we have keepalives, could i suggest something longer, like 1 hour?  I recognize that a long setting means we will leak sockets on the server side, but at present we've got queries on jdbc and inside zeppelin notebooks that are failing because the server hangs up on the socket (before results are returned).

jleach4:
This basically retests a connection when we have not heard anything from it in 60 seconds (Handles hard disconnects).  On the query execution front, my understanding is that we need tcp keep alives.   Here is a document on their setup for AWS and different operating systems.  http://docs.aws.amazon.com/redshift/latest/mgmt/connecting-firewall-guidance.html

jleach4:
This setting allows for TCP keep alives but java defers to the OS level settings.

wtang16:
Also need this in 2.6.

PR #903, opened 2017-07-15
SPLICE-1784 Fixing Serial Cutpoint Generation

jleach4:
Makes cutpoint generation a regular coprocessor vs. a serial coprocessor call from cached regions.  Also, I found a resource leak where the HFile.Reader is not closed.


genedavis:
+1 very cool

mzweben:
+1

PR #906, opened 2017-07-17
DB-6021: enable JDBC client like sqlshell.sh for SSL/TLS

snowspeeder:
Added ssl mode args for sqlshell.sh and added a new ssl server profile for standalone server startup, please see SPLICE-1607 and DB-6021 comments for usage.


snowspeeder:
I need to see more into spark adapter library, I presume its a jdbc app.
Ideally, as long as we compile that app through following properies in the
example, that jdpc app should work.

java -classpath ".:./db-client-2.6.0.1726-SNAPSHOT.jar" \

-Djavax.net.ssl.keyStore=/Users/mpimpale/vault/ClientKeyStore \

-Djavax.net.ssl.keyStorePassword=bigdata4u \

-Djavax.net.ssl.trustStore=/Users/mpimpale/vault/ClientTrustStore \

-Djavax.netDjavax.net.ssl.trustStore.ssl.trustStorePassword=secretClientTrustStorePassword
\

SampleJDBC


Where do we include our jdbc jar in the spark adapter library? maybe there
we should add these four jvm args.

jleach4:
Do we have a comparable setting for ODBC?

snowspeeder:
I do not believe SSL would work through ODBC driver untill we add support
for SSL sockets to it, will check with TimW and open a seaprate ticket.

PR #907, opened 2017-07-18
SPLICE-1798 Parallel Queries can fail on SPS Descriptor Update...

jleach4:
Swallow update errors on SPSDescriptor...


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/SPSDescriptor.java
```diff
@@ -911,12 +911,18 @@ private void updateSYSSTATEMENTS(LanguageConnectionContext lcc, int mode, Transa
...
+        try {
+            dd.updateSPS(this,
+                    tc,
+                    recompile,
+                    updateSYSCOLUMNS,
+                    firstCompilation);
+
+        } catch (StandardException se) {
+            // This can fail since multiple threads/clusters can recompile sps descriptors
+            // especially after collecting stats on sys tables.
```
dgomezferro: Could you check it is a WW conflict and continue raising the exception otherwise?

jleach4:
Let me check…

PR #909, opened 2017-07-18
Splice-1578 Upgrade from 2.5 to 2.6

jyuanca:
Please refer to https://splice.atlassian.net/browse/SPLICE-1578

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/services/io/StoredFormatIds.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/services/io/StoredFormatIds.java
@@ -1511,6 +1511,9 @@ public interface StoredFormatIds {
 
     int ACCESS_B2I_V5_ID =
             (MIN_ID_2 + 470);
+
+    public static final int ACCESS_B2I_V6_ID =
+            (MIN_ID_2 + 479);
     /******************************************************************
     **
     ** PropertyConglomerate
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
@@ -18,10 +18,10 @@ import java.util.Comparator;
 import java.util.NavigableSet;
 import java.util.TreeMap;
 
+import com.splicemachine.access.api.SConfiguration;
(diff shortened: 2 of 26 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -61,22 +62,16 @@ else if(v1==v2)
         scripts=new TreeMap<>(ddComparator);
         scripts.put(new Splice_DD_Version(sdd,1,0,0),new UpgradeScriptForFuji(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,1,1,1),new LassenUpgradeScript(sdd,tc));
-        scripts.put(new Splice_DD_Version(sdd,2,6,0),new SourceCodeUpgradeScript(sdd,tc));
+        scripts.put(new Splice_DD_Version(sdd,2,6,0, 9999),new UpgradeScriptFor260(sdd,tc));
```
yxia92: should the version threshold be 2.5.0.9999 instead of 2.6.0.9999?

db-engine/src/main/java/com/splicemachine/db/iapi/services/io/StoredFormatIds.java
```diff
@@ -1511,6 +1511,9 @@ and another may restore (read) the object. The implication of this
 
     int ACCESS_B2I_V5_ID =
             (MIN_ID_2 + 470);
+
+    public static final int ACCESS_B2I_V6_ID =
+            (MIN_ID_2 + 479);
```
jyuanca: No, it is actually no longer used.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
jyuanca: Should be 2.6.0. If the catalog version is less than 2.6.0, then run this upgrade script.

yxia92: What if the catalog version is 2.6.0.3 for example, we should not trigger the upgrade script right? But it looks that we will trigger as it is less than 2.6.0.9999.

PR #914, opened 2017-07-19
SPLICE-1621 Fix misordering of columns when selecting from partitioned external tables

tgildersleeve:
Fix the misordering of columns when selecting from partitioned parquet, orc, and avro external tables. Passed jenkins spliceengine-branch, build 327.


hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -621,7 +621,7 @@ public void testReadSmallIntFromCsv() throws Exception {
 
 
     @Test
-    @Ignore
+    @Ignore // this throws an error because we use spark to infer the datatype of partitioned columns, and it incorrectly determines float is a double.
```
jleach4: JIRA #?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -668,6 +670,34 @@ private static Column createFilterCondition(Dataset dataset,String[] allColIdInS
...
+                partitions.put(partitionColumnMap[i], fixSchema[schemaColumnIndex]);
+                schemaColumnIndex--;
+            }
+            List<Integer> partitionColumns = new ArrayList(partitions.keySet());
+            Collections.sort(partitionColumns);
+
+            // sort the partitioned columns back into their correct respective indexes
+            for (Integer partitionColumn : partitionColumns) {
+                StructField partitionColumnInfo = partitions.get(partitionColumn);
+                for (int i = fixSchema.length - 1; i > partitionColumn; i--) {
```
yxia92: It seems to me that the for loop is not needed. Can you remove?

yxia92: Please ignore my previous comment, the code is needed.

PR #915, opened 2017-07-19
SPLICE-1701_v2.5: JXM mbean server for cache and enginedriver exec service

Ben-Epstein:


splice_machine/src/main/java/com/splicemachine/EngineDriver.java
```diff
@@ -162,4 +200,176 @@ public OlapClient getOlapClient() {
...
+                        dvds[0].setValue(connections.get(i).getFirst());
+                        dvds[1].setValue(mBean.getSize());
+                        dvds[2].setValue(mBean.getMissCount());
+                        dvds[3].setValue(mBean.getMissRate());
+                        dvds[4].setValue(mBean.getHitCount());
+                        dvds[5].setValue(mBean.getHitRate());
+                    } catch (StandardException se) {
+                        throw PublicAPI.wrapStandardException(se);
+                    }
+                    i++;
```
snowspeeder: is 'i' being used here?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/TotalManagedCache.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+    @Override public long getMissCount(){
+        long missCount = 0;
+        for(ManagedCache<K, V> mc : managedCache){
+            missCount += mc.getMissCount();
+        }
+        return missCount;
+    }
+    @Override public double getHitRate(){
+        return (getHitCount()*1.0)/getRequestCount();
```
snowspeeder: divide by zero possible, can it be safeguarded

splice_machine/src/main/java/com/splicemachine/EngineDriver.java
Ben-Epstein: Yes, line 350, dvds[0].setValue(connections.get(i).getFirst());

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/TotalManagedCache.java
Ben-Epstein: resolved

splice_machine/src/main/java/com/splicemachine/EngineDriver.java
```diff
@@ -96,7 +130,7 @@ public void close(SpliceSequence entity) throws Exception{
 
         /* Create a general purpose thread pool */
         final AtomicLong count = new AtomicLong(0);
-        this.threadPool = new ThreadPoolExecutor(0, config.getThreadPoolMaxSize(),
+        ThreadPoolExecutor tpe = new ThreadPoolExecutor(0, config.getThreadPoolMaxSize(),
```
OlegMazurov: This is different from master for no reason (corePoolSize). Code below doesn't make sense if the number of core threads is 0.

Ben-Epstein: This value was set to 0 in branch 2.5 without my edit (as seen in the removed line)

OlegMazurov: That's right. The change was not backported to 2.5. Should be done now.
Again, tpe.prestartAllCoreThreads() doesn't make sense if the number of core threads is 0.

splice_machine/src/main/java/com/splicemachine/EngineDriver.java
```diff
@@ -14,30 +14,64 @@
...
+
+import javax.management.MalformedObjectNameException;
+import javax.management.remote.JMXConnector;
 
 /**
  * @author Scott Fines
  *         Date: 1/6/16
  */
-public class EngineDriver{
+public class EngineDriver extends BaseAdminProcedures{
```
OlegMazurov: Make it similar to your branch for master

PR #916, opened 2017-07-19
SPLICE-1617 Support for exclude null keys and exclude default keys fo…

jleach4:
…r indexes

New Syntax for index creation...

	[ unique = unique() ] <INDEX>
		indexName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH) <ON> tableName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH)
				<LEFT_PAREN> indexColumnList(indexColumnList) <RIGHT_PAREN>
		[ properties = propertyList(false) <CHECK_PROPERTIES>]
		[ <EXCLUDE>
		    [ <NULL> <KEYS> { excludeNulls = true;} ]
		    [ <_DEFAULT> <KEYS> { excludeDefaults = true;} ]
		]


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExcludedIndexColumnVisitor.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * The visitor collects tables found in a branch of the tree, this method
+     * return true if the ResultColumn's source table has been found by the Visitor.
+     * It will return false if the source table has not been found (i.e. it exists in another
+     * area of the plan, such as the opposite side of the join.
+     *
+     * @param rc
+     * @return
```
yxia92: These comments should be removed, it does not seem to be relevant to the code.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -3639,4 +3640,37 @@ public boolean isUnsatisfiable() {
...
+        if (size == 0)
+            return false;
+        int[] baseColumnPositions = cd.getIndexDescriptor().baseColumnPositions();
+        ExcludedIndexColumnVisitor excludedIndexColumnVisitor = new ExcludedIndexColumnVisitor(0,baseColumnPositions[0],excludeNulls,excludeNulls?null:
+                td.getDefaultValue(cd.getIndexDescriptor().baseColumnPositions()[0]));
+        boolean canSupportExcludedColumns = false;
+        for(int index=0;index<size;index++){
+            Predicate pred=elementAt(index);
+            if(pred.isOrList()){
+                return false;
```
yxia92: Why do we need to return false here? I think we can just skip the OR condition and let the loop continue. For example:

Example:
create table food3 (col1 int, col2 int);
create index food3_ix on food3(col1) exclude null keys;

We should be able to use the index for the following query:

select * from food3 f --splice-properties index=food3_ix
where (col2 > 2 or col2< 0) and col1 is not null;

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -3639,4 +3640,37 @@ public boolean isUnsatisfiable() {
...
+            }else {
+                pred.accept(excludedIndexColumnVisitor);
+                if (excludedIndexColumnVisitor.isValid)
+                    canSupportExcludedColumns = true;
```
yxia92: should we break out from the loop once canSupportExcludedColumns returns true?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExcludedIndexColumnVisitor.java
```diff
@@ -0,0 +1,119 @@
...
+                if (parent != null &&
+                        parent.getNodeType() == C_NodeTypes.IS_NULL_NODE &&
+                        nullsExcluded) {
+                    // Predicate does not validate a restriction on the index.
+                } else if (parent != null &&
+                        parent instanceof BinaryRelationalOperatorNode && // BRO
+                        ((BinaryRelationalOperatorNode) parent).getRightOperand() instanceof ConstantNode) {
+                    DataValueDescriptor compare = ((ConstantNode) ((BinaryRelationalOperatorNode) parent).getRightOperand()).getValue();
+                    if (compare != null && compare.equals(defaultValue)) {
+                        // Predicate Does not restrict since it is doing a predicate comparison
```
yxia92: The check here seems to be too loose. If we have binary relation operator other than equality, checking whether the right constant value is equal to the default value is not enough. For example:
create table food4 (col1 int, col2 int default  5);
create index food4_ix on food4(col2) exclude default keys;
insert into food4(col1) values 1,2,3;

The following query will return wrong result:
select * from food4 f --splice-properties index=food4_ix
where col2 > 3;

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -12088,6 +12093,11 @@ indexDefinition() throws StandardException :
 		indexName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH) <ON> tableName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH)
 				<LEFT_PAREN> indexColumnList(indexColumnList) <RIGHT_PAREN>
 		[ properties = propertyList(false) <CHECK_PROPERTIES>]
+		[ <EXCLUDE>
+		    [ <NULL> <KEYS> { excludeNulls = true;} ]
```
jleach4: That is correct.  We only care if the first column in the index is null.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExcludedIndexColumnVisitor.java
jleach4: Will add a test case based on this simple query...  Thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExcludedIndexColumnVisitor.java
jleach4: Will do.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
jleach4: I was not going to perform or validation for this first release.  If you have or predicates we will not use the index.  It is a bit restrictive but handles our customers use case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
jleach4: makes sense.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -3639,4 +3640,37 @@ public boolean isUnsatisfiable() {
...
+    public boolean canSupportIndexExcludedDefaults(int tableNumber, ConglomerateDescriptor cd, TableDescriptor td) throws StandardException {
+        return exclusionSupport(tableNumber,cd,td,false);
+    }
+
+    private boolean exclusionSupport(int tableNumber, ConglomerateDescriptor cd, TableDescriptor td, boolean excludeNulls) throws StandardException {
+        int size = size();
+        if (size == 0)
+            return false;
+        int[] baseColumnPositions = cd.getIndexDescriptor().baseColumnPositions();
+        ExcludedIndexColumnVisitor excludedIndexColumnVisitor = new ExcludedIndexColumnVisitor(0,baseColumnPositions[0],excludeNulls,excludeNulls?null:
```
yxia92: With table number set to 0, we may not be able to use the null/default exclusive index in joins. 

For example:

create table t2 (a2 int, b2 int, c2 int);
insert into t1 values (1,1,1), (2,2,2);
insert into t2 values (null,1,1), (2,2,2);

create index t2_ix on t2(a2) exclude null keys;

explain select a1, a2 from t1, t2 --splice-properties index=t2_ix
where b1=b2 and a2=2;

This is because in the above case, t2's table number is 1. 

By switching the order of the two tables, the index can be picked up.

It would be considered this as an enhancement in the near future if it is not going to be addressed in the first cut.

wtang16:
Need to ship to 2.5/2.6 as well.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExcludedIndexColumnVisitor.java
dgomezferro: Another case having to do with null keys:

create table a ( v varchar(20));
insert into a values null;
create index a_ix on a(v) exclude null keys;
select v from a --splice-properties index=a_ix
where substr(v, 1, 1) is null;

It shouldn't be allowed because we aren't restricting v in any way.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkDataSetWriter.java
```diff
@@ -199,7 +199,7 @@ protected void getAllConglomerates(List<Long> allCongloms, ArrayList<DDLMessage.
                 DDLMessage.DDLChange ddlChange = ProtoUtil.createTentativeIndexChange(txn.getTxnId(),
                         activation.getLanguageConnectionContext(),
                         td.getHeapConglomerateId(), searchCD.getConglomerateNumber(),
-                        td, searchCD.getIndexDescriptor());
+                        td, searchCD.getIndexDescriptor(),null); // FIX JL TODO
```
dgomezferro: Are you planning on addressing this on a separate JIRA?

PR #922, opened 2017-07-21
DB-6106 Fix limit on multiple partitions on Spark

dgomezferro:


OlegMazurov:
Need a PR for branch-2.6 as well.

PR #926, opened 2017-07-21
SPLICE-1756 introduce database property collectIndexStatsOnly to spec…(master)

yxia92:
…ify the collect stats behavior

### DIFF
--- a/db-shared/src/main/java/com/splicemachine/db/iapi/reference/Property.java
+++ b/db-shared/src/main/java/com/splicemachine/db/iapi/reference/Property.java
@@ -1197,4 +1197,11 @@ public interface Property {
      */
     String SERVER_STARTED_FROM_CMD_LINE =
             "derby.__serverStartedFromCmdLine";
+
+	/**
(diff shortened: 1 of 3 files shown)


wtang16:
Code looks good. The only comment I have is that we need to document this in official documentation. @yxia92 Could you open a doc task for Gary on this? It's going to be for 2.5 and above.

Going forward, we shall think about having scoping for the database properties (system, session, user, etc.) Current way, the change is global for entire system. A user connected to the DB with no property set assumes Analyze applies to all columns. However, a user connected to the DB later may change the behavior.   Operationally, we shall teach users to use SYSCS_UTIL.SYSCS_GET_ALL_PROPERTIES() to know the latest settings on the system.

yxia92:
I added “Docs" label to SPLICE-1756, so Gary can pick it up. Currently we have JVM, Service, Database and App properties. I think Service property can be changed temporarily in a session. But there is no property bound to user to my best knowledge.

Thanks.

PR #929, opened 2017-07-24
Splice 1701: JXM mbean server for cache and enginedriver exec service

Ben-Epstein:


splice_machine/src/main/java/com/splicemachine/EngineDriver.java
```diff
@@ -14,31 +14,65 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.splicemachine.access.api.DatabaseVersion;
 import com.splicemachine.access.api.SConfiguration;
+import com.splicemachine.db.iapi.error.PublicAPI;
```
OlegMazurov: All these imports are not needed any more. Please clean up.

Ben-Epstein: Fixed

Ben-Epstein: Clean

PR #931, opened 2017-07-24
SPLICE-1737 fix value outside the range of the data type INTEGER erro…(2.6)

yxia92:
…r for analyze table statement.


wtang16:
Approved for 2.6.

PR #936, opened 2017-07-25
DB-6100 Make sure all ITs pass after K2 upgrade (2.5)

dgomezferro:


jyuanca:
Looks like a column(DELIMITED) was missing, but why it caused other problems(physical delete and analyze)?

dgomezferro:
Not just one column, all external tables columns were missing. When trying to access the purge_deleted column it wasn't where we expected it to be.

PR #939, opened 2017-07-25
DB-5860 Resubmit to Spark if we consume too many resources in control (2.6)

dgomezferro:
This has already been merged into master.


dgomezferro:
@OlegMazurov can you get this merged? I have to create a PR for SPLICE-1372 and it would be easier if this is merged (otherwise there'll be conflicts)

PR #942, opened 2017-07-26
SPLICE-1829 Breaking up hbase_sql into actual modules

jleach4:
Starting to isolate core functionality...


ammolitor:
:+1:

dgomezferro:
I think this is a very good reorganization of the code but I believe we should do the same for 2.5 and 2.6, otherwise backporting changes will be much harder and time consuming. Any change to the moved files will have to be reapplied manually (look at the conflicts here due to SPLICE-1824)

jyuanca:
This branch needs to rebase and resolve conflicts.

PR #956, opened 2017-07-26
DB-6057: replace _HOST with host name (2.5)

jyuanca:
HBase Kerberos principal is in the format of user/_HOST@realm, where _HOST will be replaced with local host name by HBase. However, this does not work for Yarn application master when it renews a Kerberos credential, because it does not replace _HOST with host name.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
@@ -15,12 +15,16 @@
 package com.splicemachine.derby.impl;
 
 import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 
 import com.splicemachine.db.catalog.types.RoutineAliasInfo;
 import com.splicemachine.db.iapi.sql.conn.StatementContext;
 import com.splicemachine.db.impl.jdbc.EmbedConnection;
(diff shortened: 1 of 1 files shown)


OlegMazurov:
Are you going to submit PRs for 2.6 and master as well?

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -174,13 +178,28 @@ private static SparkSession initializeSparkSession() {
...
-            if(HConfiguration.unwrapDelegate().get("hbase.master.kerberos.principal") != null){
-                conf.set("spark.yarn.principal", HConfiguration.unwrapDelegate().get("hbase.master.kerberos.principal"));
-            } else if(HConfiguration.unwrapDelegate().get("hbase.regionserver.kerberos.principal") != null){
-                conf.set("spark.yarn.principal", HConfiguration.unwrapDelegate().get("hbase.regionserver.kerberos.principal"));
+        try {
+            String principal = null;
+            String p = null;
+            String hostname = HConfiguration.unwrapDelegate().get("hbase.master.hostname");
+            if (hostname == null) {
+                hostname = InetAddress.getLocalHost().getHostName();
```
jleach4: Log a warning message here.  "Possible issues with multiple interfaces"...  Please set "hbase.master.hostname"...

PR #965, opened 2017-07-27
SPLICE-949 add ROUND(double) and ROUND(double,int) functions

Ben-Epstein:
beginning round to chosen decimal place function

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
@@ -17,6 +17,7 @@ package com.splicemachine.derby.impl.sql.execute.operations;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.Random;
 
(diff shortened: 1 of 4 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
```diff
@@ -85,6 +89,62 @@ public void testSinFunction() throws Exception{
...
+	 * @throws Exception
+	 */
+	@Test
+	public void testRound() throws Exception {
+		ResultSet rs;
+		//testing round to Long
+		for(double val : roundVals){
+			rs = methodWatcher.executeQuery("values ROUND("+val+")");
+			Assert.assertTrue(rs.next());
+			Assert.assertEquals(Math.round(val),rs.getDouble(1),0.0);
```
yxia92: shouldn't we use 
Assert.assertEquals(Math.round(val),rs.getLong(1));
instead here, as both Math.round() and ROUND() with 1 parameter returns bigint?

Ben-Epstein:
@yxia92 changes addressed and fixed

yxia92:
Thanks @Ben-Epstein

PR #969, opened 2017-07-27
SPLICE-1845 Tweek Kryo Serde for Missing Elements

jleach4:
Small Tweek for new pieces that were not registered with Kryo.


snowspeeder:
btw, do we need this on other branches?

PR #970, opened 2017-07-28
DB-5974 Reconnect to OlapServer in case of failure

dgomezferro:
Boot up all HMasters (active or backup) one at a time, so that they
don't step in each other's work


wtang16:
Need 2.5 and 2.6 shipment.

hbase_sql/pom.xml
```diff
@@ -60,6 +60,17 @@
...
+	<dependency>
+		<groupId>org.apache.curator</groupId>
+		<artifactId>curator-framework</artifactId>
+		<version>2.6.0</version>
+		<exclusions>
+			<exclusion>
+				<groupId>com.google.guava</groupId>
+				<artifactId>guava</artifactId>
+			</exclusion>
+		</exclusions>
```
snowspeeder: should check the right versions for hdp-2.5.5 and mapr 5.2.0, Jenkins showing problems on these platforms (mem and CDH platforms were ok).

PR #984, opened 2017-08-01
SPLICE-1850: choose dataset impl for broadcast join of RHS is an external table(master)

jyuanca:


jyuanca:
it is breaking scala tests

jyuanca:
pushed a fix

PR #989, opened 2017-08-03
SPLICE-1853 fix wrong result with query to partitioned ORC table with predicate

yxia92:
also add logic to support selection from Date partition column.

The fix is on top of  #914 for SPLICE-1621


yxia92:
rebase on the latest master branch and made a new PR #991 , so close this one.

PR #992, opened 2017-08-04
SPLICE-1853 fix wrong result with query to partitioned ORC table with…(2.6)

yxia92:
… predicate,

also add logic to support selection from Date partition column


snowspeeder:
@wtang16 Need approval for 2.6

PR #997, opened 2017-08-07
SPLICE-1858 fix wrong result for orc table where inequality predicate(2.6)

yxia92:


snowspeeder:
@wtang16 Need approval

PR #998, opened 2017-08-07
SPLICE-1714: Ignore 'should not give a splitkey which equals to startkey' exception

jyuanca:


db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -9136,6 +9136,12 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
                <text>Snapshot '{0}' does not exist </text>
                <arg>snapshotName</arg>
            </msg>
+
+           <msg>
+               <name>TS001</name>
+               <text>Should not give a splitkey '{0}' which equals to startkey</text>
```
snowspeeder: More better message may be: "Ignoring splitkey {0} since it is same as startkey"

PR #999, opened 2017-08-07
SPLICE-1714: Ignore 'should not give a splitkey which equals to startkey' exception(2.6)

jyuanca:


snowspeeder:
@wtang16 Need approval for 2.6

PR #1000, opened 2017-08-07
SPLICE-1714: Ignore 'should not give a splitkey which equals to startkey' exception(2.5)

jyuanca:


hbase_storage/src/main/java/com/splicemachine/access/hbase/H10PartitionAdmin.java
```diff
@@ -124,6 +141,18 @@ private void retriableSplit(TableName tableInfo, byte[] splitPoint) throws IOExc
                 } catch (InterruptedException ie) {
                     throw new InterruptedIOException();
                 }
+            } catch (IOException e) {
+                String errMsg = e.getMessage();
+                if (errMsg.compareTo("should not give a splitkey which equals to startkey!") == 0) {
```
snowspeeder: since you changed the message, this might need change

jyuanca: No, this is from HBase.

jyuanca: This is an HBase error message

snowspeeder:
Oh, ok, sorry got confused.

PR #1004, opened 2017-08-09
SPLICE-1854 fix wrong result with ORC table and predicate on Date column(2.6)

yxia92:
also, add missing logic in partition pruning for smallint as partition column


snowspeeder:
@wtang16 Need approval

yxia92:
This is an 2.6 regression by SPLICE-1517. I just updated JIRA with this info. Thanks.

PR #1013, opened 2017-08-15
DB-6143: Region Name mapping pre-Splice to and from post-Splice/hbase(2.5)

jyuanca:
Please refer to DB-6143(https://splicemachine.atlassian.net/browse/DB-6143) for a detailed explanation


splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -0,0 +1,408 @@
...
+        });
+
+        //create a list of start keys
+        byte[][] startKeys = new byte[partitionList.size()][];
+        for (int i = 0; i < partitionList.size(); ++i) {
+            Partition p = partitionList.get(i);
+            byte[] k = p.getStartKey();
+            startKeys[i] = k;
+        }
+        int index = binarySearchStartKey(startKeys, rowKey);
```
yxia92: It seems to me here we are doing more than we should. At this point, we've already conducted a sort (NlogN), then a full list scan (N) to prepare the startKeys list.  Then the binary search (logN). so complexity is dominant by the sort.
But We could have just got the region that we need by a simple scan of the partition list with complexity of N.

yxia92: I was wrong. please ignore this comment.

PR #1019, opened 2017-08-18
Add system procedures to list, compact, merge regions(2.5)

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -0,0 +1,683 @@
...
+        });
+
+        //create a list of start keys
+        byte[][] startKeys = new byte[partitionList.size()][];
+        for (int i = 0; i < partitionList.size(); ++i) {
+            Partition p = partitionList.get(i);
+            byte[] k = p.getStartKey();
+            startKeys[i] = k;
+        }
+        int index = binarySearchStartKey(startKeys, rowKey);
```
yxia92: It seems to me here we are doing more than needed. We could just scan all the regions once and keep the region with the largest startKey that is equal or smaller than the given rowkey. The complexity could be O(N) instead of O(NlogN).

jyuanca: Yes

PR #1026, opened 2017-08-22
SPLICE-1617 support for index exclude null/default keys

yxia92:
This PR include the code of PR #916, as well as code to address comments from PR #916 and some bug fixes. More details are in [SPLICE-1617](https://splice.atlassian.net/browse/SPLICE-1617).


db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/ValueRow.java
```diff
@@ -63,7 +64,7 @@
 	//	STATE
 	//
 	///////////////////////////////////////////////////////////////////////
-
+	private static final long serialVersionUID = 1626095185637654111l;
```
dgomezferro: This might cause serialization issues if upgrading, please check that's not the case or remove this.

yxia92:
Thanks @dgomezferro for the comment. I updated the pull request to get rid of the serialVersionUID in ValueRow.java, IndexDescriptorImpl.java and IndexRowGenerator.java that I previously introduced in the PR. It turns out that they are not needed, and like you said, adding them may lead to upgrade issue.

PR #1036, opened 2017-08-24
DB-6164 re-apply DB-6164's fix: honor Null ordering specification in …(2.6)

yxia92:
…Spark and Control path for Window function

by reverting commit 817a09dfe3560ac71b9f2cf74eec7d6782c699d8.


yxia92:
This PR has been reviewed and approved. It was backed-out due to the dependency on SPLICE-1867's fix. Merge now that SPLIcE-1867's fix #1024  is in.

PR #1037, opened 2017-08-24
DB-6164 re-apply DB-6164's fix: honor Null ordering specification in …(master)

yxia92:
…Spark and Control path for Window function

by reverting commit 758ac78deb592ac5bcba77377337fd3c8bfbe923.


yxia92:
This PR has been previously reviewed and approved. It was backed-out due to the dependency on SPLICE-1867's fix. Merge directly now that SPLICE-1867's fix #1025 is in.

PR #1041, opened 2017-08-25
SPLICE-1784: Fixing cutpoint generation

snowspeeder:
Earlier reverted due to tpc1g intermittent regressions. Now added the retry coprocessor while getting subsplits for the region during that interval. Tested with tpch1g 15 runs consecutively, and more manual testing.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/mrio/api/core/AbstractSMInputFormat.java
+++ b/hbase_sql/src/main/java/com/splicemachine/mrio/api/core/AbstractSMInputFormat.java
@@ -103,7 +103,22 @@ public abstract class AbstractSMInputFormat<K,V> extends InputFormat<K, V> imple
                 }
                 SubregionSplitter splitter = new HBaseSubregionSplitter();
 
-                return splitter.getSubSplits(table, splits, s.getStartRow(), s.getStopRow());
+                List<InputSplit> lss= splitter.getSubSplits(table, splits, s.getStartRow(), s.getStopRow());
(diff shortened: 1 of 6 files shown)


hbase_sql/src/main/java/com/splicemachine/mrio/api/core/AbstractSMInputFormat.java
```diff
@@ -103,7 +103,22 @@
                 }
                 SubregionSplitter splitter = new HBaseSubregionSplitter();
 
-                return splitter.getSubSplits(table, splits, s.getStartRow(), s.getStopRow());
+                List<InputSplit> lss= splitter.getSubSplits(table, splits, s.getStartRow(), s.getStopRow());
+                //check if split count changed in-between
+                List<Partition>  newSplits = clientPartition.subPartitions(s.getStartRow(), s.getStopRow(), refresh);
```
dgomezferro: If you do it this way you need to set refresh to true on the second call, otherwise you might pull the region info from the cache and miss the split. You were lucky while testing because the compaction caused by the split refreshes the cache, but that happens on a different thread and there's no guarantee that will always happen at the right time.

dgomezferro:
Other than the refresh change I'm +1

snowspeeder:
Thanks Daniel, incorporated your comment.

jyuanca:
@snowspeeder Did latest change pass tpch test?

snowspeeder:
Yes it did, ran multiple times.

PR #1042, opened 2017-08-25
SPLICE-1373 Extended query logging

dgomezferro:
Added more information to query logging:
1. Operation UUID, same as used int GET_RUNNING_OPERATIONS, unique for each statement execution
2. Execution mode: local vs Spark
3. Execution end time

Moved splice-derby.log to use log4j to allow log rotation, etc.


dgomezferro:
Feedback from the code review:
- Make it possible to have a log file with only statement logs
- Log to splice tables (to be addressed on another JIRA)

jyuanca:
+1

PR #1047, opened 2017-08-27
DB-6134 change resultant datatype to varchar if union of char types w…

yxia92:
…ith different length


wtang16:
Please ship this to shared-2.5 branch as well. Thanks.

PR #1051, opened 2017-08-29
SPLICE-1874 determine Spark path based on output row count computed u…(2.6)

yxia92:
…sing predicate against key value.


jyuanca:
+1

PR #1053, opened 2017-08-30
SPLICE-1879 Modify Key By function to look more like map()

jleach4:
The KeyBy piece is not correct.  I modified it to behave more like a map call to transform V to PairDataSet<K,V>.


wtang16:
@jleach4 What's the user impact of this error: performance/incorrect result? The Multiple Distinct Aggregate feature uses KeyBy. Some other operations use that too (MergeSortJoin/HalfMergeSortJoin/Sort/GroupedAggregate).

jleach4:
Non-lazy execution on the control side (performance).  No correctness issue for queries...

wtang16:
Cool. Thanks @jleach4!

PR #1055, opened 2017-08-31
DB-6047: create child persistent node to avoid ZOOKEEPER-2052

jyuanca:


dgomezferro:
@jyuanca  This change could leave persistent znodes blocking flushes after a restart, is that prevented somehow?

jyuanca:
@dgomezferro There is a timestamp in the root node for backup, which Backup updates periodically when it is active. The flushing thread can check the timestamp for timeout.

dgomezferro:
@jyuanca is that check implemented already?

jyuanca:
@dgomezferro Yes, it is checked in already.

dgomezferro:
@jyuanca cool, thanks!

PR #1060, opened 2017-08-31
DB-6203 honor Spark dataset processor once picked, and fix regression…(2.5)

yxia92:
… in SPLICE-1874


yxia92:
I opened [SPLICE-1883](https://splice.atlassian.net/browse/SPLICE-1883) to track Monte's comment.

PR #1062, opened 2017-09-06
DB-6124 fix conglomerate average row width estimation, also use local…(2.5)

yxia92:
…cost to prune join plan search space.

Analysis is in [DB-6124](https://splicemachine.atlassian.net/browse/DB-6124).


snowspeeder:
@wtang16  Need approval for 2.5

yxia92:
This will be replace with a new PR.

PR #1067, opened 2017-09-11
DB-6155 adjust selectivity estimation logic for range condition(2.5)

yxia92:
Enhance range selectivity estimation logic. More details, pease seee [DB-6155](https://splicemachine.atlassian.net/browse/DB-6155).


yxia92:
Based on the comment in the review meeting, I've updated ColumnStatisticsImpl.rangeSelectivity() to include a check of the range with min and max value. If the range falls beyond [min,max], return 0 selectivity directly.

PR #1068, opened 2017-09-11
DB-6155 adjust selectivity estimation logic for range condition

yxia92:


yxia92:
Based on the comment in the review meeting, I've updated ColumnStatisticsImpl.rangeSelectivity() to include a check of the range with min and max value. If the range falls beyond [min,max], return 0 selectivity directly.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -226,9 +231,45 @@ public long selectivity(DataValueDescriptor element) {
...
+        com.yahoo.sketches.frequencies.ItemsSketch.Row<DataValueDescriptor>[] items = frequenciesSketch.getFrequentItems(ErrorType.NO_FALSE_POSITIVES);
+        for (com.yahoo.sketches.frequencies.ItemsSketch.Row<DataValueDescriptor> row: items) {
+            DataValueDescriptor skewedValue = row.getItem();
+            try {
+                if (skewedValue != null &&
+                        (start == null || start.isNull() || skewedValue.compare(includeStart ? ORDER_OP_GREATEROREQUALS : ORDER_OP_GREATERTHAN, start, false, false)) &&
+                        (stop == null || stop.isNull() || skewedValue.compare(includeStop ? ORDER_OP_LESSOREQUALS : ORDER_OP_LESSTHAN, stop, false, false)))
+                    skewCount += row.getEstimate();
+            } catch (StandardException e) {
+                // cost estimation error does not need to fail the query
```
dgomezferro: Please log the error here

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -246,6 +287,36 @@ public long selectivity(DataValueDescriptor element) {
...
+            DataValueDescriptor maxValue = quantilesSketch.getMaxValue();
+            if (maxValue != null &&
+                    start != null && !start.isNull() && start.compare(includeStart ? ORDER_OP_GREATERTHAN : ORDER_OP_GREATEROREQUALS, maxValue, false, false))
+                return 0;
+            DataValueDescriptor minValue = quantilesSketch.getMinValue();
+            if (minValue != null &&
+                    stop != null && !stop.isNull() && stop.compare(includeStop ? ORDER_OP_LESSTHAN : ORDER_OP_LESSOREQUALS, minValue, false, false))
+                return 0;
+        } catch (StandardException e) {
+            // cost estimation error does not need to fail the query
```
dgomezferro: Log error

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
yxia92: Thanks @dgomezferro I added the code to log the error in the splice.log if debug log is turned on.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
yxia92: Same logic is added here too.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
dgomezferro: If we don't expect this exception to happen I would log it at WARN level, making it clear that something unexpected happened even though we don't fail the operation. 
If we expect this to happen often please add a comment explaining when that is.
And add the exception to the call so that we get the stacktrace on the log: LOG.level("msg", e);

yxia92:
Thanks @dgomezferro . We do not expect the exception to happen, but for some reason it happens, we don’t want to fail the query just because of this estimation error. Updated the code to use LOG.warn().

PR #1073, opened 2017-09-13
DB-6228 Enable Spark event log to debug jenkins runs (2.6)

dgomezferro:
Add sparkHistory profile that launches a History server pointing to
hbase_sql/target/SpliceTestYarnPlatform


dgomezferro:
This has already been approved for master

PR #1086, opened 2017-09-19
DB-6186 Add region name and id to compaction job description

dgomezferro:


snowspeeder:
@wtang16  Need approval for 2.5/2.6

PR #1090, opened 2017-09-20
DB-6225 GC Pressure

jleach4:
This still seems problematic.  Our serialization is not good.  I am going to try to fix as part of my work on changing serde.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSparkKryoRegistrator.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSparkKryoRegistrator.java
@@ -127,7 +127,7 @@ public class SpliceSparkKryoRegistrator implements KryoRegistrator, KryoPool.Kry
 
     @Override
     public void registerClasses(Kryo instance) {
-//    	instance.setReferences(false);
+    	instance.setReferences(false);
         instance.setRegistrationRequired(false); //TODO set to true
 
         instance.register(ValueRow.class, new SparkValueRowSerializer<ValueRow>(){
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/kryo/SparkValueRowSerializer.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/kryo/SparkValueRowSerializer.java
@@ -87,12 +87,12 @@ public abstract class SparkValueRowSerializer<T extends ExecRow> extends Seriali
                 if (formatId == StoredFormatIds.SQL_USERTYPE_ID_V3) {
                     Object o = dvds[i].getObject();
                     boolean useKryo = false;
-                    if (o != null && o instanceof UDTBase) {
+                    if (o != null && o instanceof UDTBase && 1==2) {
                         // This is a UDT or UDA, do not serialize using Kryo
                         output.writeBoolean(useKryo);
                         ByteArrayOutputStream outputBuffer = new ByteArrayOutputStream();
                         ObjectOutputStream objectOutputStream = new ObjectOutputStream(outputBuffer);
-                        objectOutputStream.writeObject(o);
+                        objectOutputStream.writeUnshared(o);
                         objectOutputStream.flush();
                         objectOutputStream.reset();
                         byte[] bytes = outputBuffer.toByteArray();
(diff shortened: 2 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/impl/kryo/SparkValueRowSerializer.java
```diff
@@ -87,12 +87,12 @@ public void write(Kryo kryo, Output output, T object) {
                 if (formatId == StoredFormatIds.SQL_USERTYPE_ID_V3) {
                     Object o = dvds[i].getObject();
                     boolean useKryo = false;
-                    if (o != null && o instanceof UDTBase) {
+                    if (o != null && o instanceof UDTBase && 1==2) {
```
wtang16: Why "1==2" here?

jleach4: mistake on my part.

wtang16: Ok. Thanks.

hbase_sql/src/main/java/com/splicemachine/derby/impl/kryo/SparkValueRowSerializer.java
```diff
@@ -92,7 +92,7 @@ public void write(Kryo kryo, Output output, T object) {
                         output.writeBoolean(useKryo);
                         ByteArrayOutputStream outputBuffer = new ByteArrayOutputStream();
                         ObjectOutputStream objectOutputStream = new ObjectOutputStream(outputBuffer);
-                        objectOutputStream.writeObject(o);
+                        objectOutputStream.writeUnshared(o);
```
OlegMazurov: I doubt this change is going to make any difference. objectOutputStream is local and becomes collectable once the method returns. There should be no accumulated effect.

wtang16: I thought the issue John was fixing a memory leak issue alleged with ObjectOutputStream. Using ObjectOutputStream.reset() along with writeUnshared()/readUnshared() typically solves the problem.

There are a few related articles on this topics:
https://www.securecoding.cert.org/confluence/display/java/FIO53-J.+Use+the+serialization+methods+writeUnshared%28%29+and+readUnshared%28%29+with+care
http://www.maultech.com/chrislott/blog/20100110_spacetime.html
https://www.reddit.com/r/learnprogramming/comments/66p4be/java_objectinputstream_why_am_i_getting_a_memory/
https://orenkishon.wordpress.com/2014/06/28/java-memory-leak-caused-by-objectinputstream/

OlegMazurov: Those links are OK by themselves but they don't seem to be relevant here as both ObjectOutputStream and UDTInputStream are used for one object only and are gone with all their cached links after serde (no cumulative effect). readUnshared creates more objects (potentially breaking functionality of the deserialized object), thus increasing GC pressure.
I struggle to see why the change could fix any memory leak. If I'm missing something, maybe more details could help here.

wtang16: I'll defer this to John/Yi who investigated the original use case.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSparkKryoRegistrator.java
```diff
@@ -127,7 +127,7 @@ public void register(Kryo instance) {
 
     @Override
     public void registerClasses(Kryo instance) {
-//    	instance.setReferences(false);
+    	instance.setReferences(false);
```
wtang16: The default behavior is True. Each appearance of an object in the reference graph after the first is stored as an integer ordinal. This allows multiple references to the same object and cyclic graphs to be serialized. It has a small amount of overhead and can be disabled to save space if it is not needed. Do we have any quantifiable overhead observed in profiler? It'll help us understand the benefits better.

hbase_sql/src/main/java/com/splicemachine/derby/impl/kryo/SparkValueRowSerializer.java
OlegMazurov: One thing to try here is to reuse objectOutputStream to avoid the cost of its initialization (quite high in the profile) and reduce the amount of generated garbage. Resetting it after use will be required in that case.

wtang16:
Closing this PR because we do not have enough support for it.

PR #1099, opened 2017-09-25
DB-6217: make table level select privileges to column level (2.5)

snowspeeder:
Add the column list from table as column level privileges so that users can revoke few columns later. Also see spliceengine-ee for ITs


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -946,5 +945,15 @@ public ColPermsDescriptor getColumnPermissions(UUID tableUUID,
         return manager.isEnabled()?manager.getColPermsManager().getColumnPermissions(this,tableUUID,privType,forGrant,authorizationId):null;
```
yxia92: A minor comment here, since we have the function isEEManagerEnabled() now, we could use this function instead of repeating the same code.

PR #1105, opened 2017-09-26
SPLICE-1898: column level privileges allowed with Enterprise Edition

snowspeeder:
Error out for non-EE case when attempted to use column level privileges.


snowspeeder:
Closing this PR since SPLICE-865 should take care of this case.

PR #1113, opened 2017-09-28
DB-6269: Handle transactions that are not in SPLICE_TXN

jyuanca:


splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -202,10 +206,13 @@ protected void doResolve(DataCell data,long ts){
 
     private boolean isVisible(long txnId) throws IOException{
         TxnView toCompare=fetchTransaction(txnId);
-        // If the database is restored from a backup, it may contain data that were written by a transaction which
-        // is not present in SPLICE_TXN table, because SPLICE_TXN table is copied before the transaction begins.
-        // However, the table written by the txn was copied
-        return toCompare != null ? myTxn.canSee(toCompare) : false;
+        // if we cannot find a txn, we conservatively assume it was committed
```
dgomezferro: I would only do that if some special parameter is set, splice.transactions.ignore.missing or something like that. I wouldn't feel comfortable silently ignoring transactions that aren't on the Txn table.

jyuanca: @dgomezferro Good point.

PR #1114, opened 2017-09-29
SPLICE-1302 Add minimum parallelism for Spark shuffles

dgomezferro:
It can be disabled setting splice.olap.shuffle.partitions to 0

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
@@ -101,13 +101,15 @@ public class SparkDataSet<V> implements DataSet<V> {
 
     public JavaRDD<V> rdd;
     private Map<String,String> attributes;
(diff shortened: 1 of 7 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -208,6 +210,7 @@ public int partitions() {
     public DataSet<V> distinct(String name, boolean isLast, OperationContext context, boolean pushScope, String scopeDetail) {
         pushScopeIfNeeded(context, pushScope, scopeDetail);
         try {
+            int numPartitions = SparkUtils.getPartitions(rdd, defaultPartitions);
```
yxia92: The variable numPartitions is not used, why introduced here?

dgomezferro: Thanks, removed.

PR #1116, opened 2017-10-02
DB-6275: exclude flatten and restart scripts (2.6)

snowspeeder:
keep master, 2.6 branch in sync with v2.0 and v2.5 for Cloudera certification


yxia92:
Could you add a link to DB-6042 in JIRA to indicate the relevance?

snowspeeder:
Added the link in Jira.

PR #1118, opened 2017-10-03
SPLICE-1728 Fixing Data Dictionary Caching

jleach4:
Makes sure we are caching data dictionary tables correctly.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/db/BasicDatabase.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/db/BasicDatabase.java
@@ -31,6 +31,7 @@
 
 package com.splicemachine.db.impl.db;
 
+import com.carrotsearch.hppc.LongOpenHashSet;
 import com.splicemachine.db.iapi.reference.*;
 import com.splicemachine.db.iapi.error.PublicAPI;
 import com.splicemachine.db.iapi.sql.compile.CompilerContext;
@@ -66,6 +67,7 @@ import com.splicemachine.db.iapi.services.property.PropertySetCallback;
 import com.splicemachine.db.iapi.store.access.TransactionController;
 import com.splicemachine.db.iapi.jdbc.AuthenticationService;
 import com.splicemachine.db.iapi.services.uuid.UUIDFactory;
+import com.splicemachine.db.impl.sql.catalog.TabInfoImpl;
 import com.splicemachine.db.impl.sql.execute.JarUtil;
 import com.splicemachine.db.io.StorageFile;
 import com.splicemachine.db.catalog.UUID;
@@ -96,6 +98,7 @@ import java.text.DateFormat;
 
 public class BasicDatabase implements ModuleControl, ModuleSupportable, PropertySetCallback, Database, JarReader
 {
+	public static LongOpenHashSet systemConglomerates;
 	protected boolean	active;
 	private AuthenticationService authenticationService;
 	protected AccessFactory af;
@@ -224,7 +227,23 @@ public class BasicDatabase implements ModuleControl, ModuleSupportable, Property
 
 		active = true;
 
+		// Populate System Conglomerates, used to tell if you are a system conglomerate.
+		systemConglomerates = new LongOpenHashSet();
+		populateSystemConglomerates(getDataDictionary().getCoreInfo());
+		populateSystemConglomerates(getDataDictionary().getNoncoreInfo());
+
     }
+
+	private void populateSystemConglomerates(TabInfoImpl[] tabInfos) {
+		for (TabInfoImpl tabInfo: tabInfos) {
+			if (tabInfo!=null) {
+				systemConglomerates.add(tabInfo.getHeapConglomerate());
+				for (int i = 0; i< tabInfo.getNumberOfIndexes();i++) {
+					systemConglomerates.add(tabInfo.getIndexConglomerate(i));
+				}
+			}
+		}
+	}
 
 	public void stop() {
         // The data dictionary is not available if this database has the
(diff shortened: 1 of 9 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -6886,9 +6889,71 @@ private void initializeCoreInfo() throws StandardException{
...
+        noncoreInfo[SYSBACKUPFILESET_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSBACKUPFILESETRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSBACKUPITEMS_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSBACKUPITEMSRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSBACKUPJOBS_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSBACKUPJOBSRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSCOLUMNSTATS_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSCOLUMNSTATISTICSRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSPHYSICALSTATS_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSPHYSICALSTATISTICSRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSTABLESTATS_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSTABLESTATISTICSRowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSSOURCECODE_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSSOURCECODERowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSDUMMY1_CATALOG_NUM-NUM_CORE]=new TabInfoImpl(new SYSDUMMY1RowFactory(luuidFactory,exFactory,dvf));
+        noncoreInfo[SYSSNAPSHOT_NUM-NUM_CORE]=new TabInfoImpl(new SYSSNAPSHOTSRowFactory(luuidFactory,exFactory,dvf));
```
snowspeeder: Are we sure, we're not missing any systables? when I run "show tables" on sqlshell, it gives me 33 rows of systables, it looks like there are 29 rows here. Just wanted to double-check

wtang16: @snowspeeder which ones are missing?

snowspeeder: I just did a diff below, 1st file is the list of tables derived form sqlshell.sh, the other is from this PR...NOTE the table names have also changed

/sqlshell.list /tmp/code.list 
8,10c8
< SYSCOLUMNS
< SYSCOLUMNSTATS
< SYSCONGLOMERATES
---
> SYSCOLUMNSTATISTICS
18c16
< SYSPHYSICALSTATS
---
> SYSPHYSICALSTATISTICS
23d20
< SYSSCHEMAS
29,30c26
< SYSTABLES
< SYSTABLESTATS
---
> SYSTABLESTATISTICS

snowspeeder: I guess, name change is ok since they are just the classnames here

snowspeeder: I see the missing in the core system tables, as well they are just not part of nonCore function there...
+1

db-engine/src/main/java/com/splicemachine/db/impl/db/BasicDatabase.java
```diff
@@ -224,8 +227,24 @@ public void boot(boolean create, Properties startParams)
 
 		active = true;
 
+		// Populate System Conglomerates, used to tell if you are a system conglomerate.
+		systemConglomerates = new LongOpenHashSet();
+		populateSystemConglomerates(getDataDictionary().getCoreInfo());
+		populateSystemConglomerates(getDataDictionary().getNoncoreInfo());
+
     }
```
jyuanca: Need to include conglomerate 16 as a system conglomerate

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -85,7 +86,7 @@ public void testVacuumDoesNotBreakStuff() throws Exception {
             Set<String> deletedTables=getDeletedTables(beforeTables,afterTables);
             for(String t : deletedTables){
                 long conglom=new Long(t);
-                Assert.assertTrue(conglom>=DataDictionary.FIRST_USER_TABLE_NUMBER);
+                Assert.assertTrue(BasicDatabase.isSystemConglomerate(conglom));
```
yxia92: we need to change to negation here: !BasicDatabase.isSystemConglomerate(conglom)

dgomezferro:
Apart from missing conglomerate 16, there's a race condition between populating the systemConglomerates structure and opening regions of existing user tables. In that window we might not load the Index coprocessor for user tables:

>
> select count(1) from e --splice-properties index=null
> ;
> 1
>   --------------------
> 6
> 
> 1 row selected
> select count(1) from e --splice-properties index=ei
> ;
> 1
>    --------------------
> 3
> 
> 1 row selected
> splice>

jyuanca:
@dgomezferro How did you reproduce the problem?

PR #1119, opened 2017-10-04
SPLICE-1728 Fixing Data Dictionary Caching(2.5)

jyuanca:


hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -85,7 +86,7 @@ public void testVacuumDoesNotBreakStuff() throws Exception {
             Set<String> deletedTables=getDeletedTables(beforeTables,afterTables);
             for(String t : deletedTables){
                 long conglom=new Long(t);
-                Assert.assertTrue(conglom>=DataDictionary.FIRST_USER_TABLE_NUMBER);
+                Assert.assertTrue(BasicDatabase.isSystemConglomerate(conglom));
```
yxia92: Should we do the negation here?

jyuanca: Good catch! Interesting, this test did not fail

PR #1120, opened 2017-10-05
SPLICE-1906 Make coprocessors throw only IOExceptions

dgomezferro:


dgomezferro:
This is a diff ignoring whitespace which might be easier to review: https://pastebin.com/9ZnSN1XB

PR #1121, opened 2017-10-05
SPLICE-1906 Make coprocessors throw only IOExceptions (2.5)

dgomezferro:


dgomezferro:
Diff ignoring whitespace: https://pastebin.com/G8L1JewD

wtang16:
approved for 2.5/2.6

PR #1124, opened 2017-10-05
DB-4300 enhance subquery processing logic to convert eligible where s…

yxia92:
…ubquery to from subquery

Please see more detailed solution description in [DB-4300](https://splicemachine.atlassian.net/browse/DB-4300).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
@@ -244,6 +244,16 @@ public interface CompilerContext extends Context
 	void setNumTables(int num);
 
 	/**
+	 * Some where subqueries can be converted to fromSubquery, so the number of tables could increase
+	 * during preprocess of optimization. maximalPossibleTableCount takes the where Subqueries into
+	 * considration as potentially the maximal possible table count.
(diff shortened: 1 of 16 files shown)


dgomezferro:
Nice one! 

The results you show on the JIRA seem to be for TPCH 1, right? You say it returns 186 rows but the current jmeter script expects a single row result. 

Can you also enable query 20 in the jmeter script? hbase_sql/src/test/jmeter/tpch1.jmx

yxia92:
@dgomezferro Thanks. Yes, the result is for TPCH1. I verified the 186 row result on a build without my fix by decomposing Q20 into 2 parts, 1) run the non-flattened non-correlated subquery directly and dump the result in a temp table. 2) Rewrite Q20 by changing this non-flattened IN subquery into an IN subquery that directly select from the temp table, this way, the IN subquery can be flattened, and the query could run. I compared this result with the result after the fix, they are identical.

I will enable Q20 in jmeter script.

wtang16:
+1

yxia92:
Updated fix to pick dataset broadcast join implementation if the right side of an inclusion join is complex (not a simple scan of the base table, but involves multiple joins or aggregate operation). For more detailed analysis, please see update in [DB-4300](https://splicemachine.atlassian.net/browse/DB-4300).

PR #1127, opened 2017-10-05
DB-1908: SELECT permission after granting of any other kind of permission (2.6)

snowspeeder:
Missed a check earlier, please also see EE for IT.


wtang16:
approved for 2.6

PR #1130, opened 2017-10-10
SPLICE-1915 Enable query logging by default (2.6)

dgomezferro:


wtang16:
Can we have a global configuration setting to enable/disable logging or levels of details to be logged?

PR #1131, opened 2017-10-10
DB-6274 Reduce logging on MemStoreAwareScanner

dgomezferro:
There is too much logging and that's creating issues on Jenkins

### DIFF
--- a/hbase_storage/hbase1.1.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
+++ b/hbase_storage/hbase1.1.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
@@ -70,11 +70,11 @@ public class MemStoreFlushAwareScanner extends StoreScanner {
     @Override
     public KeyValue peek() {
         if (didWeFlush()) {
-            if (LOG.isTraceEnabled())
-                SpliceLogUtils.trace(LOG, "already Flushed");
+//            if (LOG.isTraceEnabled())
+//                SpliceLogUtils.trace(LOG, "Peek: already Flushed");
             if (flushAlreadyReturned) {
-                if (LOG.isTraceEnabled())
-                    SpliceLogUtils.trace(LOG, "returning counter");
+//                if (LOG.isTraceEnabled())
+//                    SpliceLogUtils.trace(LOG, "returning counter " + counter);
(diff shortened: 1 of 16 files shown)


hbase_storage/hbase1.1.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -83,13 +83,16 @@ public KeyValue peek() {
                 return ClientRegionConstants.MEMSTORE_BEGIN_FLUSH;
             }
         }
-        if (beginRow)
+        if (beginRow) {
+            if (LOG.isTraceEnabled())
+                SpliceLogUtils.trace(LOG, "Peek: memstore begin");
```
yxia92: Why this change is not applied to 1.0.0.x and 1.0.0?

hbase_storage/hbase1.1.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -119,19 +122,27 @@ public boolean reseek(KeyValue kv) throws IOException {
     public boolean internalNext(List<Cell> outResult,ScannerContext scannerContext) throws IOException {
         if (beginRow) {
             beginRow = false;
+            if (LOG.isTraceEnabled())
+                SpliceLogUtils.trace(LOG, "Next: memstore begin");
```
yxia92: Why this change is not applied to 1.0.0.x and 1.0.0?

hbase_storage/hbase1.1.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -140,6 +151,8 @@ public boolean internalNext(List<Cell> outResult,ScannerContext scannerContext)
                 }
             } else {
                 flushAlreadyReturned = true;
+                if (LOG.isTraceEnabled())
+                    SpliceLogUtils.trace(LOG, "Next: returning begin flush ");
```
yxia92: Why this change is not applied to 1.0.0.x and 1.0.0?

dgomezferro: They weren't too important and the commits will have to be reverted or cleaned up after debugging the wrong result in jenkins, but I cleaned up this piece and added the missing logs. Thanks for the review!

PR #1135, opened 2017-10-13
DB-6274 Re-init SplitRegionScanner if MemstoreKVScanner returns null

dgomezferro:


hbase_storage/hbase1.0.0.x/src/main/java/com/splicemachine/access/client/MemstoreKeyValueScanner.java
```diff
@@ -64,9 +64,8 @@ public boolean nextResult() throws IOException{
             rows++;
             return true;
         }else{
-            cells=null;
-            peakKeyValue=null;
-            return false;
+            // This shouldn't happen, throw exception and re-init the scanner
+            throw new DoNotRetryIOException("Memstore scanner shouldn't end prematurely");
```
yxia92: @dgomezferro from your comment in JIRA "MemstoreKeyvalueScanner is reading a null row from the HBase scanner ... This shouldn't happen", why it cannot read a null row? How do we tell if we exhaust the rows to read? Thanks.

dgomezferro: Memstore scanners never close by themselves, when they reach the end they send a special value and then the SplitRegionScanner closes them when all scanners are exhausted. See MemStoreFlushAwareScanner.internalNext()/peek() and the endRowNeedsToBeReturned field.

yxia92: Thanks @dgomezferro for the explanation!

PR #1139, opened 2017-10-16
DB-6212 Let the user specify number of splits for a table

dgomezferro:
This feature adds a new splice-property for tables/indexes that specifies the requested number of splits for that scan. Splice tries to accommodate the request but it might not be able to if the scanned conglomerate is too small or resides entirely in the Memstore. It also can overshoot the request in some cases when there are few HBase splits (2 or 3).


hbase_sql/src/main/java/com/splicemachine/mrio/api/core/AbstractSMInputFormat.java
```diff
@@ -89,6 +90,7 @@
         Partition clientPartition = new ClientPartition(connection, table.getName(), table, clock, driver.getPartitionInfoCache());
         int retryCounter = 0;
         boolean refresh = false;
+        this.splits = conf.getInt(MRConstants.SPLICE_SPLITS_PER_TABLE, 0);
```
snowspeeder: should the default be 0?

dgomezferro: 0 means use the usual splitBlockSize heuristic

snowspeeder:
oh ok, thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
```diff
@@ -281,10 +282,10 @@ public int getScanArgs(
 		 * sorted, 3) the in-list column position in the index or primary key.
 		 */
         if (genInListVals) {
-            numArgs = 37;
+            numArgs = 38;
```
wtang16: How do we keep track of the numArgs increment?

PR #1145, opened 2017-10-19
DB-6274 Re-init SplitRegionScanner if MemstoreKVScanner returns null

dgomezferro:
Previous merge had some IT failures

Now we make sure we don't close the MemstoreFlushAwareScanner if it reaches the scan's stopRow


hbase_storage/hbase1.0.0/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -69,6 +70,7 @@ public MemStoreFlushAwareScanner(HRegion region, Store store, ScanInfo scanInfo,
 			this.memstoreAware = memstoreAware;
 			this.initialValue = initialValue;
 			this.region = region;
+			this.stopRow = Bytes.equals(scan.getStopRow(), HConstants.EMPTY_END_ROW) ? null : scan.getStopRow();
 		}
```
yxia92: Why for hbase1.0.0, we don't check "isStopRow(peek)" in addition to the check of "peek==null in peek()"?

dgomezferro: I missed that one, good eye!

PR #1146, opened 2017-10-19
DB-6267 Add parameter to ignore missing transactions from SPLICE_TXN

dgomezferro:


hbase_storage/src/main/java/com/splicemachine/si/impl/CoprocessorTxnStore.java
```diff
@@ -334,9 +337,22 @@ public long commitCount(){
      * **************************************************************************************************************
      */
         /*private helper methods*/
-    private TxnView decode(TxnMessage.Txn message) throws IOException{
+    private TxnView decode(long queryId, TxnMessage.Txn message) throws IOException{
         TxnMessage.TxnInfo info=message.getInfo();
```
jyuanca: Is queryId begin timestamp of a txn?

jleach4:
Oops that is backwards…

hbase_storage/src/main/java/com/splicemachine/si/impl/CoprocessorTxnStore.java
dgomezferro: Yes, the txnId/begin timestamp. When we create a mocked committed transaction it is as if the transaction committed at the same point in time it started.

dgomezferro:
What do you mean @jleach4 ?

wtang16:
@dgomezferro We need this backported to 2.5/2.6 too.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
```diff
@@ -50,6 +50,9 @@
     public static final String READ_RESOLVER_QUEUE_SIZE = "splice.txn.readresolver.queueSize";
     private static final int DEFAULT_READ_RESOLVER_QUEUE_SIZE=1<<16;
 
+    public static final String IGNORE_MISSING_TXN = "splice.ignore.missing.transactions";
```
snowspeeder: please document the details of this param in the JIRA too.

PR #1147, opened 2017-10-19
SPLICE-1925: Remove TXN Lookups from dictionary writes

jleach4:
Removes a transaction lookup for all dictionary writes (base tables and indexes)...


splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactorUtil.java
```diff
@@ -64,9 +65,9 @@
          * To be frank, this is only here to support legacy code without needing to rewrite everything under
```
wtang16: Is this comment still valid? What is the legacy code Scott refers to?

PR #1148, opened 2017-10-19
SPLICE-1926: HRegionUtil is leaking memory

jleach4:
Remove reader leak for cutpoint generation.


dgomezferro:
@jleach4 this change is what caused the regression back in the day: https://splice.atlassian.net/browse/SPLICE-1784?focusedCommentId=26602&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-26602

As far as I can see those readers are shared among everybody that access that storefile, and it should be closed properly when the whole region is closed.

jyuanca:
The store file reader was pre-created when a region is opened. getHFileReder() does not create extra one, so don't need to close it.

dgomezferro:
@jleach4 @wtang16  should we close this one?

PR #1149, opened 2017-10-20
DB-6315: keep alive backup

jyuanca:
Please also review code changes on ee branch.

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
@@ -127,6 +127,12 @@ public class HBaseConfiguration implements ConfigurationDefault {
 
     public static final String SPLICE_BACKUP_PARALLELISM = "splice.backup.parallelism";
     public static final int DEFAULT_SPLICE_BACKUP_PARALLELISM = 16;
+
+    public static final String SPLICE_BACKUP_KEEPALIVE_INTERVAL = "splice.backup.keepAliveIntervalMs";
+    public static final long DEFAULT_SPLICE_BACKUP_KEEPALIVE_INTERVAL = 3000L;
+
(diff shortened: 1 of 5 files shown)


splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
```diff
@@ -127,6 +127,12 @@
 
     public static final String SPLICE_BACKUP_PARALLELISM = "splice.backup.parallelism";
     public static final int DEFAULT_SPLICE_BACKUP_PARALLELISM = 16;
+
+    public static final String SPLICE_BACKUP_KEEPALIVE_INTERVAL = "splice.backup.keepAliveIntervalMs";
+    public static final long DEFAULT_SPLICE_BACKUP_KEEPALIVE_INTERVAL = 3000L;
+
+    public static final String SPLICE_BACKUP_TIMEOUT = "splice.backup.timeout";
+    public static final long DEFAULT_SPLICE_BACKUP_TIMEOUT = 10 * DEFAULT_SPLICE_BACKUP_KEEPALIVE_INTERVAL;
```
wtang16: Will 30 seconds be enough? What factors can impact the backup time?

jyuanca:
This needs to be merged with code changes on ee-branch.

splice_access_api/src/main/java/com/splicemachine/access/configuration/HBaseConfiguration.java
jyuanca: A keep alive thread will update timestamp every 3 seconds when backup is active. When backup completes or fails, backup remove znode /backup, and there should be no backup timeout issue.

If during backup, the region server fails and restarted. The node /backup will be still there and no keep alive thread updates the timestamp. In that case, after 30 seconds, we consider the backup timeout, and it is safe to remove /backup from zookeeper.

There are some code changes on ee branch, but no one reviewed yet.

PR #1150, opened 2017-10-21
DB-6311 skip merging stats from empty partitions to work around bug I…(2.5)

yxia92:
…n sketches library. For detailed analysis, please see [DB-6311](https://splicemachine.atlassian.net/browse/DB-6311)


db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsMerge.java
```diff
@@ -112,9 +110,12 @@ public void accumulate(ColumnStatisticsImpl value) throws StandardException {
     @Override
     public void merge(ColumnStatisticsMerge otherAggregator) {
         ColumnStatisticsMerge columnStatisticsMerge = (ColumnStatisticsMerge) otherAggregator;
-        quantilesSketchUnion.update(columnStatisticsMerge.quantilesSketchUnion.getResult());
-        frequenciesSketch.merge(columnStatisticsMerge.frequenciesSketch);
-        thetaSketchUnion.update(columnStatisticsMerge.thetaSketchUnion.getResult());
+        if (!columnStatisticsMerge.quantilesSketchUnion.getResult().isEmpty()) {
```
wtang16: Could the input param otherAggregator be null? Maybe we need to be more defensive here.

wtang16:
Also, if this problem doesn't exist in 2.6 due to sketches library version of 0.8.4, we can just backport https://splice.atlassian.net/browse/SPLICE-1351 to 2.5.

yxia92:
We want to upgrade the sketches-core library instead. So close this request.

PR #1152, opened 2017-10-23
DB-6286 flatten correlated Scalar Subquery in Select clause

yxia92:


wtang16:
We need to backport this to 2.6.

PR #1157, opened 2017-10-25
DB-5953 Allow first connection on Kerberized cluster

dgomezferro:
Please review also -ee PR


dgomezferro:
@wtang16 no, I don't think we have any kerberos ITs and it wouldn't be easy to do. I think it should be tested as part of the QA regression runs.

db-engine/src/main/java/com/splicemachine/db/iapi/services/property/PropertyUtil.java
```diff
@@ -92,7 +92,8 @@
 		"derby.language.logQueryPlan",
 		Property.LOCKS_ESCALATION_THRESHOLD,
 		Property.DATABASE_PROPERTIES_ONLY,
-		Property.DEFAULT_CONNECTION_MODE_PROPERTY
+		Property.DEFAULT_CONNECTION_MODE_PROPERTY,
+		Property.AUTHENTICATION_BUILTIN_ALGORITHM
```
yxia92: I don't quite understanding the relationship between adding this property and the main fix to allow first connection to go through without Kerberos authentication. Could you elaborate a bit?

dgomezferro: The first internal connection follows a different path from normal external connections. Usually the authentication algorithm is set on that other path, so for the first internal connection it wasn't set. By setting this here I make sure there's a suitable default for that first connection.

yxia92: Thanks Daniel for the explanation.

PR #1164, opened 2017-10-25
SPLICE-1928: decode region start/end key by fetching actual rowkey

jyuanca:
If a region start/end key cannot be decoded, fetch the first/last row of the region and decode it.


jyuanca:
@yxia92 You are right, thanks!

PR #1167, opened 2017-10-27
SPLICE-1802: create index by HFile loading(2.5)

jyuanca:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNode.java
```diff
@@ -957,6 +957,29 @@ public void init(Object arg1,
...
+                     Object arg10,
+                     Object arg11,
+                     Object arg12,
+                     Object arg13,
+                     Object arg14,
+                     Object arg15,
+                     Object arg16,
+                     Object arg17) throws StandardException{
+        if(SanityManager.DEBUG){
+            SanityManager.THROWASSERT("Fourteen-argument init() not implemented for "+getClass().getName());
```
yxia92: Error message should indicate "Seventeen-argement init()" instead of "Fourteen-argument init()".

PR #1168, opened 2017-10-27
DB-6299 Synchronize access to shared ArrayList

dgomezferro:
The main fix is synchronizing access to ArrayList from multiple threads in HBaseSubregionSplitter.

There were some issues with error handling too:
* An exception in the update() method in HBaseSubregionSplitter.getSubSplits() would cause missing splits
* An exception in the server side subsplit computation wouldn't be propagated correctly to the client, due to the hostname field being missing


dgomezferro:
You can use this link to view the diff ignoring whitespace, it's easier to see what changed: https://github.com/splicemachine/spliceengine/pull/1168/files?w=1

OlegMazurov:
Can you point out where that access to ArrayList from multiple threads occurred?

jyuanca:
@OlegMazurov callback function update() is executed by multiple threads

OlegMazurov:
@jyuanca In that case synchronizing around results.add() would be a cheaper solution not affecting all other operations on 'results' (get, size, etc.).
The performance impact is likely to be negligible anyway.

PR #1170, opened 2017-10-27
DB-6299 Synchronize access to shared ArrayList (2.6)

dgomezferro:
Improve error condition handling in subsplits computation


jyuanca:
+1

PR #1172, opened 2017-10-27
SPLICE-1934 WordUtils.wrap() from commons-lang3 3.5 is broken

OlegMazurov:


wtang16:
Need the fix in 2.5 also

OlegMazurov:
The fix is integrated in DB-6268-br2.5 (where we need it at the moment). Will be merged into 2.5 from that branch (but can be integrated separately if we want).

PR #1175, opened 2017-11-01
SPLICE-1937: add cdh5.12 platform (2.5)

snowspeeder:
Add the needed changes to add cdh 5.12. Parcel tested on cdh-5.12 cluster with tpch1g.
Also see EE repo PR.

NOTE to moderator:
This change is based on DB-6068 (should merge  6068 first and then apply this to avoid conflicts)


wtang16:
Please make sure flatten.sh is not included in the final binary package.

jleach4:
Do you have a way to flatten the database without it?

John

snowspeeder:
flatten script is not included in parcel package, we classified it as
support tool, as and when needed can be  pulled from our source tree and
applied.

PR #1179, opened 2017-11-07
DB-6352: integrate spark2.2 changes on branch-2.5

snowspeeder:
Now that Cloudera spark libraries are decoupled from the branch-2.5, it should be able tor run on the installed base of spark 2.2 or 2.1. Following changes picked from spark22-2.5 branch to be compatible on spark-2.2, which should also work on spark-2.1. Initial changes parcel tested on spark2.1 and spark-2.2 cluster. 
Testing done:
- ITs ran on standalone
- on cluster, ran tpch1g on both spark2.1 and spark2.2 installed


hbase_sql/src/main/java/com/splicemachine/stream/index/HTableOutputFormat.java
```diff
@@ -55,6 +54,9 @@ public Configuration getConf() {
     @Override
     public RecordWriter<byte[],Either<Exception, KVPair>> getRecordWriter(TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
         try {
+            if (outputCommitter == null)
+                getOutputCommitter(taskAttemptContext);
+            assert taskAttemptContext != null && taskAttemptContext.getConfiguration() != null:"configuration passed in is null";
```
yxia92: Shouldn't this assertion be moved before the call of getOutputCommitter()?

snowspeeder:
Right, will move ahead, thanks.

snowspeeder:
Updated the PR with review comments, please take a look. Thanks!

yxia92:
Thanks @snowspeeder !

PR #1186, opened 2017-11-08
DB-6346 allow more scenarios to qualify for CSSQ flattening

yxia92:
Please see [DB-6346](https://splicemachine.atlassian.net/browse/DB-6346) for the fix description.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
@@ -129,47 +129,54 @@ public abstract class NLJoinFunction <Op extends SpliceOperation, From, To> exte
         try {
             if (rightSideNLJIterator == null)
                 return false;
-            while (nLeftRows > 0 && !rightSideNLJIterator.hasNext()) {
+            while (true) {
+                while (nLeftRows > 0 && !rightSideNLJIterator.hasNext()) {
 
-                // We have consumed all rows from right side iterator, reclaim operation context
-                currentOperationContext.getOperation().close();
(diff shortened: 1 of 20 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
```diff
@@ -129,47 +129,54 @@ public boolean hasNext() {
...
-
-                if (leftSideIterator.hasNext()) {
-                    // If we haven't consumed left side iterator, submit a task to scan righ side
-                    ExecRow execRow = leftSideIterator.next();
-                    GetNLJoinIterator getNLJoinIterator = GetNLJoinIterator.makeGetNLJoinIterator(joinType,
-                            operationContextList.remove(0), execRow.getClone());
-                    completionService.submit(getNLJoinIterator);
-                    nLeftRows++;
-                }
+            while (true) {
```
dgomezferro: Why is this loop needed?

yxia92: When we get a match from the SSQ, I want to skip the remaining matching rows, and let  rightSideNJIterator iterate to the end.

dgomezferro: I see it now, thanks!

OlegMazurov:
Please, resolve conflicts.

yxia92:
Updated the PR by rebasing to latest.

wtang16:
Can we backport this to 2.6.1? Customer requires this version.

PR #1202, opened 2017-11-13
DB-6367: null checking for region operations

jyuanca:
If table name or split key is null, return a meaningful error message rather than a null pointer exception.

If schema name is null, use the current schema.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
@@ -340,6 +340,9 @@ public class SpliceRegionAdmin {
                                                String timeFormat,
                                                ResultSet[] results) throws Exception {
 
+        if(splitKey == null)
+            throw StandardException.newException(SQLState.SPLIT_KEY_CANNOT_BE_NULL);
+
         TableDescriptor td = getTableDescriptor(schemaName, tableName);
         ConglomerateDescriptor index = null;
         if (indexName != null) {
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
@@ -93,14 +93,10 @@ public class SpliceRegionAdmin {
                                    String timeFormat,
                                    ResultSet[] results) throws Exception {
 
-        // Check parameters
-        schemaName = schemaName.toUpperCase();
-        tableName = tableName.toUpperCase();
-
         TableDescriptor td = getTableDescriptor(schemaName, tableName);
         ConglomerateDescriptor index = null;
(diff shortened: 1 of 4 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -553,6 +537,20 @@ private static TableDescriptor getTableDescriptor(String schemaName, String tabl
         SpliceTransactionManager tc = (SpliceTransactionManager)lcc.getTransactionExecute();
         DataDictionary dd = lcc.getDataDictionary();
 
+        // Check parameters
+        if (schemaName == null) {
+            schemaName = lcc.getCurrentSchemaName();
+        } else {
+            schemaName = schemaName.toUpperCase();
```
dgomezferro: Do we handle case-sensitive schemas? 

create schema "thisIsCaseSensitive";

jyuanca: We should support quoted names in our system procedure. For now, I'll keep the case as it is. If the table name is case insensitive, the user needs to specify name in upper case. I'll remove toUpperCase().

wtang16: http://www.starlink.ac.uk/star/java/db/docs/html/ref/ref-single.html: An SQL92Identifier is a dictionary object identifier that conforms to the rules of SQL-92. SQL-92 states that identifiers for dictionary objects are limited to 128 characters and are case-insensitive (unless delimited by double quotes), because they are automatically translated into uppercase by the system. 

Let's be careful here and make sure we do not change user behavior on database object names. If the schema/index is created without quotes, users will expect the name to be case-insensitive. Don't we have a central module to deal with object name searching and matching?

jyuanca: We are lack of a module to normalize object names. Our behavior does not conform with SQL92 standard.

jyuanca: I will keep the current behavior as it is, but we should address the following problem in a separate Jira
1) If object name is not double quoted, it is case-insensitive, convert it to upper case
2) If object name is double-quoted, remove double quotes
3) if schema name is not specified, default to current schema, instead of SPLICE.

wtang16: I see. Please create a public improvement JIRA for the proposed change. Thanks @jyuanca. Copy @genedavis too.

wtang16: Also, this is not aligned with our documentation: https://doc.splicemachine.com/sqlref_identifiers_intro.html

PR #1212, opened 2017-11-15
DB-6016 avoid sort if some of the index columns are bound with constant.

yxia92:
Please see problem analysis and solution in [JIRA](https://splicemachine.atlassian.net/browse/DB-6016?focusedCommentId=51644&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-51644)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/RowOrdering.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/RowOrdering.java
@@ -93,7 +93,7 @@ public interface RowOrdering{
      * @return true means this ordering is ordered on the given column
      * in the given position.
      */
-    boolean orderedOnColumn(int direction,int orderPosition,int tableNumber,int columnNumber) throws StandardException;
+    int orderedOnColumn(int direction,int orderPosition,int tableNumber,int columnNumber) throws StandardException;
 
     /**
      * Tell whether this ordering is ordered on the given column.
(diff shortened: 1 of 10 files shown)


dgomezferro:
There's some code in FromBaseTable that claims to be handling a similar case as the one described in the JIRA: https://github.com/splicemachine/spliceengine/blob/e79767cdcda881418707d197cfe76eff4d5fb522/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java#L423-L443

Is that code obsolete, or does it handle a different case?

yxia92:
@dgomezferro This comment is obsolete. The original implementation of rowOdering.orderedOnColumn() contains a check if the column has single value,  treat it as already sorted, and not add it to the rowOrdering's ordering list. This has a side effect for the merge join qualification logic which compares the left table's rowOrdering with the inner table's index/primary key columns. DB-3819 reported that merge join could be qualified when it shouldn't, so removes the relevant logic in orderedOnColumn(). The problematic case I found with not adding single value column to ordering is that eligible merge join could not be picked. Here is an example:
     create table t1 (a1 int, b1 int, c1 int, primary key(a1,b1));
     create table t2 (a2 int, b2 int, c2 int, primary key(a2,b2));
     select * from t1 inner join t2 on a1=a2 and a1=1 order by a1,b1;
For this case, after planning t1, we have only b1 in the rowOrdering, when compare with the primary key of t2, which is (a2,b2), it fails to match in MergeJoinStrategy.mergeable(), and merge join cannot be picked.

dgomezferro:
I see, thanks for the explanation! It would be good to update the comment for future reference, by I'm +1 on the change. Thanks!

yxia92:
Thanks @dgomezferro I removed the obsolete comment to avoid confusion.

PR #1213, opened 2017-11-15
SPLICE-1948 Initialize Splice Spark context with user context

dgomezferro:
Add ITs for Spark streaming using this feature


dgomezferro:
@mzweben Yes, that's a good idea, the API seems more polished than the plain Streaming API too. Thanks for the review

PR #1218, opened 2017-11-16
DB-6373: set rowlocation for index lookup

jyuanca:


splice_machine/src/test/java/com/splicemachine/derby/impl/load/HBaseBulkLoadIT.java
```diff
@@ -101,22 +101,23 @@ public static void cleanup() throws Exception {
...
         countUsingIndex(LINEITEM, "L_SHIPDATE_IDX");
         countUsingIndex(ORDERS, "O_CUST_IDX");
         countUsingIndex(ORDERS, "O_DATE_PRI_KEY_IDX");
     }
 
-    private static void bulkDelete(String tableName) throws Exception {
-        String sql = String.format("delete from %s --splice-properties bulkDeleteDirectory='%s'", tableName, getResource("data"));
+    private static void bulkDelete(String tableName, String indexName) throws Exception {
+        String sql = String.format("delete from %s --splice-properties bulkDeleteDirectory='%s',index=%s", tableName, getResource("data"),
+                indexName!=null?indexName:"null");
```
snowspeeder: Could you also add IT for regular delete with index hint?

jyuanca: Regular delete was not impacted.

snowspeeder:
regular delete with index hint will not hit this issue?

splice_machine/src/test/java/com/splicemachine/derby/impl/load/HBaseBulkLoadIT.java
jyuanca: It only happens when using results from index lookup to create index using IndexTransformer. Bulk delete is the only case

snowspeeder:
I see, thanks for the clarification.

PR #1221, opened 2017-11-16
SPLICE-1948 Initialize Splice Spark context with user context (2.5)

dgomezferro:


assembly/pom.xml
```diff
@@ -30,6 +30,11 @@
             <artifactId>hbase_sql-${envClassifier}</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-core_${scala.binary.version}p</artifactId>
+            <version>${spark.version}</version>
```
snowspeeder: Would this add spark classes in splice uber jar?

dgomezferro: It didn't seem to, but it wasn't doing anything really so I removed it. Thanks

PR #1224, opened 2017-11-17
SPLICE-1948 Increase test timeout

dgomezferro:


dgomezferro:
This seems to make the build happy in jenkins: http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/default/598/

PR #1227, opened 2017-11-18
SPLICE-1961 Missing splice_spark module in pom.xml

OlegMazurov:
Added missing module to cdh5.8.3 and cdh5.12.0


wtang16:
Do we need to add this to other supported platform too (hdp2.5.5 and mapr5.2.0)?

PR #1240, opened 2017-11-28
DB-6411 fix column misalignment for join update through mergesort join(2.6.1)

yxia92:


snowspeeder:
@wtang16 need approval

PR #1241, opened 2017-11-28
DB-6411 fix column misalignment for join update through mergesort join

yxia92:
Note, compared to the commit to branch-2.5 and branch-2.6, the commit to the master's UpdateFromSubqueryIT has the merge join path commented out due to wrong result tracked in DB-6320.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
@@ -726,9 +726,12 @@ public class GenericStatement implements Statement{
     private void walkAST(LanguageConnectionContext lcc, Visitable queryTree, CompilationPhase phase) throws StandardException {
         ASTVisitor visitor = lcc.getASTVisitor();
         if (visitor != null) {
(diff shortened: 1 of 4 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -725,9 +725,14 @@ private void dumpOptimizedTree(LanguageConnectionContext lcc,StatementNode qt,bo
...
         ASTVisitor visitor = lcc.getASTVisitor();
-        if (visitor != null) {
-            visitor.begin(statementText, phase);
-            queryTree.accept(visitor);
+        try {
+            if (visitor != null) {
+                visitor.begin(statementText, phase);
+                queryTree.accept(visitor);
+            }
+        } catch (StandardException e) {
```
dgomezferro: You don't need the catch clause, unless you want to be explicit.

yxia92: Thanks @dgomezferro . I've updated the PR to remove the catch clause.

PR #1243, opened 2017-11-30
A couple of Backup fixes

jyuanca:


wtang16:
DB-6202 is part of this PR too, right? @jyuanca could you update the technical details to the included JIRAs (DB-6202, DB-6408, DB-6409) including root cause and solution summary?

dgomezferro:
I wouldn't merge DB-6408, I'm +1 on DB-6409

PR #1247, opened 2017-11-30
DB-6412 Bound the outer join row count by the left outer table's row count (2.5)

changli6:


snowspeeder:
@wtang16 Need approval for 2.5/2.6

jyuanca:
+1

PR #1248, opened 2017-12-01
 DB-6399 Broadcast Kerberos tokens to executors instead of keytabs

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -107,6 +125,51 @@ public static synchronized boolean isRunningOnSpark() {
...
+    private static UserGroupInformation applyCreds(Credentials credentials) throws IOException {
+        LOG.info("credentials:" + credentials);
+
+        logCredInformation(credentials);
+
+        UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+        ugi.addCredentials(credentials);
+        // specify that this is a proxy user
+        ugi.setAuthenticationMethod(UserGroupInformation.AuthenticationMethod.PROXY);
+        System.out.println("Set auth to token");
```
jleach4: Are you meaning System.out here?

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -107,6 +125,51 @@ public static synchronized boolean isRunningOnSpark() {
...
+
+        LOG.info("Applied credentials");
+
+        return ugi;
+    }
+
```
jleach4: I think this would fail a security audit.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
dgomezferro: Nope, good eye.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
dgomezferro: Why?

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -107,6 +125,50 @@ public static synchronized boolean isRunningOnSpark() {
...
+    public static void logCredInformation(Credentials credentials2) {
+        LOG.info("credentials:" + credentials2);
+        for (int i =0 ; i < credentials2.getAllSecretKeys().size(); i++) {
+            LOG.info("getAllSecretKeys:" + i + ":" + credentials2.getAllSecretKeys().get(i));
```
snowspeeder: Should we perhaps log this only for debug log? I'm not sure what details secretkeys do share though.

dgomezferro: You are certainly right, done

stpabhi:
@dgomezferro To deal with ticket renewal we can add ugi.checkTGTAndReloginFromKeytab() to the actions that need authentication. If the ticket is not close to expiration, then the method will not renew the ticket.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
wtang16: I think the comment is regarding writing the credential to log ("credentials:" + credentials). I do not know if the credential as is written to log will be interpreted and used. But we could log it under debug mode. BTW, we do see hbase logging certain kerberos access info in its log too. 
17/07/24 14:44:49 INFO security.AMCredentialRenewer: Attempting to login to KDC using principal: hbase/[HOST_NAME]@S[REALM_NAME]

wtang16: New change looks good.

PR #1264, opened 2017-12-08
DB-6445 Cannot import from an s3 bucket on HDP

OlegMazurov:
Upgrade joda-time to 2.8.2


wtang16:
@OlegMazurov joda-time for branches 2.6 and master are also using 2.3. Do you want to update it as well? 

Also, last joda-time change was in DB-6023. Please make sure it is Ok. 

Thanks.

PR #1272, opened 2017-12-11
SPLICE-1970 Exclude metrics jars from splice-uber jar to avoid class loader issues when using Spark Adapter(2.5)

stpabhi:


assembly/pom.xml
```diff
@@ -94,6 +94,11 @@
                                     <exclude>com.splicemachine:db-tools-ij</exclude>
                                     <exclude>com.splicemachine:db-shared</exclude>
                                     <exclude>com.splicemachine:db-drda</exclude>
+                                    <exclude>io.dropwizard.metrics:metrics-core</exclude>
+                                    <exclude>io.dropwizard.metrics:metrics-graphite</exclude>
+                                    <exclude>io.dropwizard.metrics:metrics-json</exclude>
+                                    <exclude>io.dropwizard.metrics:metrics-jvm</exclude>
+                                    <exclude>com.yammer.metrics:metrics-core</exclude>
```
OlegMazurov: Please use io.dropwizard.metrics:* and com.yammer.metrics:*

assembly/pom.xml
```diff
@@ -237,10 +242,13 @@
                                 db-shared,
                                 db-drda,
                                 javax.servlet-api,
-                                javax.ws.rs-api
+                                javax.ws.rs-api,
+                                metrics-core,
+                                metrics-graphite,
+                                metrics-json,
+                                metrics-jvm
```
OlegMazurov: Why are these needed?

assembly/pom.xml
stpabhi: Ok

assembly/pom.xml
stpabhi: Good question. These are not needed unless splice has any dependency on these versions. I wasn’t sure about it so added them in lib.

OlegMazurov: All dependencies on metrics are transitive only.

stpabhi: Ok, will remove them

stpabhi:
@OlegMazurov I have made the suggested changes.

PR #1275, opened 2017-12-11
DB-6146: delete region(2.5)

jyuanca:
To delete a region from hbase
1) Close the region
2) Purge HFiles from hdfs
3) Reopen the region 
4) Merge the region with its neighbor

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
@@ -49,6 +49,7 @@ import com.splicemachine.derby.utils.marshall.dvd.DescriptorSerializer;
 import com.splicemachine.derby.utils.marshall.dvd.VersionedSerializers;
 import com.splicemachine.si.api.data.TxnOperationFactory;
 import com.splicemachine.si.api.txn.TxnView;
+import com.splicemachine.si.constants.SIConstants;
 import com.splicemachine.si.impl.driver.SIDriver;
 import com.splicemachine.storage.DataResult;
 import com.splicemachine.storage.DataResultScanner;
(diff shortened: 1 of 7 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -81,6 +82,78 @@
...
+        Partition p = getPartition(td, index, encodedRegionName);
+        try {
+            // close the region
+            partitionName = p.getName();
+            regionClosed = true;
+            admin.closeRegion(p);
+            SpliceLogUtils.info(LOG, "Closed region %s", partitionName);
+
+            // move all store files to a temporary directory
+            fileMoved = true;
```
yxia92: should "fileMoved" to be set after the call of "moveToArchive()"?

jyuanca: Before it to be safe, just in case it fails when moving files. Then clean up by moving files back.

snowspeeder:
please add more logging at every critical stage so it helps debug in case something goes wrong.

jyuanca:
@snowspeeder It has logging for every step

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -693,6 +812,59 @@ private static long getFamilyModificationTime(Configuration conf,
...
+        Path familyArchive = new Path(regionArchive, SIConstants.DEFAULT_FAMILY_NAME);
+        FileSystem fs = region.getFileSystem(conf);
+        if (fs.exists(familyArchive)) {
+            FileStatus[] fileStatuses = fs.listStatus(familyArchive);
+            for (FileStatus fileStatus : fileStatuses) {
+                String name = fileStatus.getPath().getName();
+                Path src = fileStatus.getPath();
+                Path dest = new Path(family, name);
+                SpliceLogUtils.info(LOG, "Move %s to %s", src.toString(), dest.toString());
+                fs.rename(src, dest);
```
snowspeeder: should it check the return status if this api was successful or not?

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -693,6 +812,59 @@ private static long getFamilyModificationTime(Configuration conf,
...
+            }
+            fs.delete(familyArchive, true);
```
snowspeeder: same here

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -693,6 +812,59 @@ private static long getFamilyModificationTime(Configuration conf,
...
+        FileSystem fs = region.getFileSystem(conf);
+        if (fs.exists(familyArchive)) {
+            FileStatus[] fileStatuses = fs.listStatus(familyArchive);
+            for (FileStatus fileStatus : fileStatuses) {
+                String name = fileStatus.getPath().getName();
+                Path src = fileStatus.getPath();
+                Path dest = new Path(family, name);
+                SpliceLogUtils.info(LOG, "Move %s to %s", src.toString(), dest.toString());
+                fs.rename(src, dest);
+            }
```
snowspeeder: Log may help?

snowspeeder:
In general, every filesystem api should be fenced with return status check (if it provides, I assume it does):
 e.g. (!fs.rename(src,dest)) {
        SpliceLogUtils.info("Could not rename ...");
It would help catch hairy cases when it is unlucky.

jyuanca:
@snowspeeder Good suggestion!

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -81,6 +82,96 @@
...
+                    SpliceLogUtils.info(LOG, "Merged regions %s and %s", p.getEncodedName(), p2.getEncodedName());
+                }
+            }
+
+            // purge store files from file system
+            deleteTempDir(conf, conglomId, encodedRegionName);
+        }
+        catch (Exception e) {
+
+            SpliceLogUtils.error(LOG, getCallStack(e));
```
OlegMazurov: Use SpliceLogUtils.error(Logger logger, String message, Throwable error) here.

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -81,6 +82,96 @@
...
+                        regionClosed = true;
+                    }
+                    restoreFromArchive(conf, conglomId, encodedRegionName);
+                }
+
+                // If the region is closed, reopen it
+                if (regionClosed)
+                    admin.assign(p);
+            } catch (Exception ex) {
+                SpliceLogUtils.error(LOG, getCallStack(e));
```
OlegMazurov: Use SpliceLogUtils.error(Logger logger, String message, Throwable error) here.

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -728,4 +939,14 @@ private static void decodeNextRow(Partition partition, KeyHashDecoder decoder, E
             SpliceLogUtils.info(LOG, "Use %s %s as %s", Bytes.toHex(key), execRow, reverse ? "endKey" : "startKey");
         }
     }
+
+    public static  String getCallStack(Exception e) {
```
OlegMazurov: Not needed.

jyuanca:
@OlegMazurov I removed it.

PR #1254, opened 2017-12-05
DB-6356: forward port decoupling changes to master

snowspeeder:
Forward port the spark decoupling and related changes to the master by Oleg and me. Thanks to Oleg for resolving the parquet jar untangling issue.

PR #1276, opened 2017-12-11
SPLICE-1973 Exclude Kafka jars from splice-uber.jar for all platforms

stpabhi:


OlegMazurov:
This PR conflicts with #1254 already in the queue. Let us integrate that one first.

assembly/pom.xml
```diff
@@ -237,51 +238,69 @@
...
+            <properties>
+                <excludeSpark>org.apache.spark:*</excludeSpark>
+            </properties>
         </profile>
         <profile>
             <id>cdh5.8.3</id>
-            <properties><extraExclude>org.apache.spark:*</extraExclude></properties>
-        </profile>
-        <profile>
-            <id>cdh5.8.5</id>
```
OlegMazurov: Is 5.8.5 removed by mistake?
I'd rather see this whole part intact. All is needed is <exclude>org.apache.kafka:*</exclude> above.

stpabhi: @OlegMazurov must be my mistake. will revert this whole section

stpabhi:
@OlegMazurov Done. please review it.

PR #1277, opened 2017-12-11
SPLICE-1973 Exclude Kafka jars from splice-uber.jar for all platforms(2.5)

stpabhi:


assembly/pom.xml
```diff
@@ -94,7 +94,8 @@
                                     <exclude>com.splicemachine:db-tools-ij</exclude>
                                     <exclude>com.splicemachine:db-shared</exclude>
                                     <exclude>com.splicemachine:db-drda</exclude>
-                                    <exclude>${extraExclude}</exclude>
+                                    <exclude>${excludeSpark}</exclude>
+                                    <exclude>${excludeKafka}</exclude>
```
OlegMazurov: Shouldn't Kafka be excluded for all platforms, not just CDH?

stpabhi: Yes, will make changes

stpabhi:
@OlegMazurov excluded Kafka for all platforms

PR #1287, opened 2017-12-12
SPLICE-1920 Procedure to delete rows from dictionary tables (2.6)

dgomezferro:


db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
```diff
@@ -1942,6 +1942,8 @@ The message identifier (ie. the value that matches a field in this class)
 	String LANG_CANCELLATION_EXCEPTION							   = "SE008";
 	String LANG_INVALID_DAY										   = "SE022";
 	String ERROR_PARSING_EXCEPTION								   = "SE023";
+	String PARAMETER_CANNOT_BE_NULL								   = "SE024";
```
jyuanca: This is redundant with

String PARAMETER_CANNOT_BE_NULL									= "TS006";

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -8818,6 +8818,17 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
                <arg>message</arg>
            </msg>
 
+           <msg>
+               <name>SE024</name>
+	       <text>Parameter cannot be null: {0}</text>
+               <arg>parameterName</arg>
+           </msg>
```
jyuanca: redundant with

<msg>
               <name>TS006</name>
               <text>Parameter '{0}' cannot be null.</text>
               <arg>parameterName</arg>
           </msg>

jyuanca:
After deleting a row from base table, still need to check index to make sure index and base table are consistent. How can a user know rowid of an index?

db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
dgomezferro: I thought the "TS" prefix was specific to the region operations, that's why I moved it into the SE (Splice Engine) exception codes.

dgomezferro:
@jyuanca  This procedure shouldn't be used by end users, it's meant to fix a corrupted database. 

I'll add details to the JIRA, but in general to query the rowid of an index you can use the ROWID column in conjunction with the splice-properties index parameter:

>select ROWID from table --splice-properties index=YOUR_INDEX
> where ....;

To get the ROWID for the base table you'd do the same specifying index=null, otherwise you might get the ROWID of an index row.

jyuanca:
@dgomezferro This is confusing to me "In general to query the rowid of an index you can use the ROWID column in conjunction with the splice-properties index parameter" Should it still return rowid from base table?

dgomezferro:
@jyuanca the ROWID column returns the rowid of the conglomerate used for scanning, if you force it to use an index (index=IDEX_NAME) it returns the rowid for the index, if you force the base table (index=null) it returns the rowid of the base table. If you don't force either, then it depends on what the optimizer chooses.

jyuanca:
@dgomezferro This needs to be documented

PR #1294, opened 2017-12-14
SPLICE-1349: serialize and initialize BatchOnceOperation correctly(2.5)

yxia92:
This is a port of SpLICE-1349 to 2.5. The code is already in branch-2.0, branch-2.6 and master.
Also see [DB-6462](https://splicemachine.atlassian.net/browse/DB-6461).


wtang16:
I suspect it was due to the 2.5 branch cutoff from master in late Jan.

PR #1296, opened 2017-12-14
DB-6284: add ignore txn cache (2.5)

jyuanca:
Add an ignore txn cache to filter out txns that are committed during backup. The ignore txn cache entry is a pair of long [backup begin txnId, backup end TxnId]

### DIFF
--- a/hbase_storage/hbase1.0.0.x/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
+++ b/hbase_storage/hbase1.0.0.x/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
@@ -243,6 +243,17 @@ public class HBaseConnectionFactory{
         }
     }
 
+    public void createRestoreTable() {
+
+        try(Admin admin=connection.getAdmin()){
+            HTableDescriptor td=generateNonSITable(HConfiguration.IGNORE_TXN_TABLE_NAME);
+            admin.createTable(td);
+            SpliceLogUtils.info(LOG, HConfiguration.IGNORE_TXN_TABLE_NAME +" created");
+        }catch(Exception e) {
+            SpliceLogUtils.error(LOG, "Unable to create SPLICE_IGNORE_TXN Table", e);
+        }
+    }
+
     public static String escape(String first){
         // escape single quotes | compress multiple whitespace chars into one,
         // (replacing tab, newline, etc)
--- a/splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
+++ b/splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
@@ -1,0 +1,114 @@
+
+/*
+ * Copyright (c) 2012 - 2017 Splice Machine, Inc.
+ *
+ * This file is part of Splice Machine.
+ * Splice Machine is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Affero General Public License as published by the Free Software Foundation, either
+ * version 3, or (at your option) any later version.
+ * Splice Machine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Affero General Public License for more details.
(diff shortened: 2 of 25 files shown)


splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveIgnoreTxnCacheSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public boolean shouldIgnore(Long txnId) {
+
+        boolean ignore = false;
+        try {
+            ignoreTxnTableExists = delegate.ignoreTxnTableExists();
+            if (!ignoreTxnTableExists)
+                return false;
+            if (cache.isEmpty()) {
```
dgomezferro: I don't think a cache for  the transactions makes sense, the check is in memory and it should be inexpensive, unless we had many ignored ranges. In that case I would try to cache the minimum and maximum for all ranges and only perform the check if the Txn falls in that range.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveIgnoreTxnCacheSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+            for (Pair<Long, Long> range : cache) {
+                if (txnId > range.getFirst() && txnId < range.getSecond()) {
+                    ignore = true;
+                    break;
+                }
+            }
+
+        } catch (Exception e) {
+            SpliceLogUtils.warn(LOG, "shouldIgnore got an exception: %s", e);
```
dgomezferro: If we get an exception here we should raise it, otherwise we might treat data that we should ignore as committed.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveIgnoreTxnCacheSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+import com.splicemachine.utils.SpliceLogUtils;
+import org.apache.hadoop.hbase.util.Pair;
+import org.apache.log4j.Logger;
+
+import java.util.*;
+
+/**
+ * Created by jyuan on 10/30/17.
+ */
+public class ActiveIgnoreTxnCacheSupplier {
```
dgomezferro: I think the name of this class is confusing, I guess you made it symmetrical to the ActiveTxnCacheSupplier, but in this case it doesn't deal with Active transactions so I'd rename it.

hbase_storage/hbase1.0.0.x/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
```diff
@@ -243,6 +243,17 @@ public boolean createSpliceHBaseTables(){
         }
     }
 
+    public void createRestoreTable() {
```
dgomezferro: I would create the table unconditionally during upgrade, not during backup/restore.

jyuanca: It is only necessary for a restored database.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveIgnoreTxnCacheSupplier.java
jyuanca: The cache is for [min, max] of all ranges.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private DataResultScanner openScanner() throws IOException {
+        Partition table = partitionFactory.getTable(HBaseConfiguration.IGNORE_TXN_TABLE_NAME);
+        DataScan scan = txnOperationFactory.newDataScan(null);
+        return table.openResultScanner(scan);
+    }
+
+    private void init() throws IOException {
+        if (!initialized) {
```
OlegMazurov: 'initialized' needs to be declared 'volatile' to be used in double-checked locking.
Is lazy initialization justified here?

PR #1299, opened 2017-12-16
DB-6453 avoid sort for outer join when applicable(2.5)

yxia92:
and DB-6453 preserve left row order for batched nestedloop join. 

Summary of solution could be found in [DB-6453](https://splicemachine.atlassian.net/browse/DB-6453)


wtang16:
Approved for 2.5/2.6

PR #1301, opened 2017-12-16
DB-6453 avoid sort for outer join when applicable(2.6.1)

yxia92:
DB-6453 preserve left row order for batched nestedloop join


wtang16:
Approved for 2.5:2.6

PR #1310, opened 2017-12-20
SPLICE-1984 parallelize multi-probe and union operation(2.5)

jyuanca:
Execute Multi-probe and unions in parallel and do not materialize their iterators


jyuanca:
+1

PR #1312, opened 2017-12-21
SPLICE-1984: Parallelizes MultiProbeTableScan and Union Operations.  C…

jyuanca:
…ustomers with large number of in list elements or with numerous real-time union operations should see a reduction in execution time

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
@@ -14,6 +14,7 @@
 
 package com.splicemachine.derby.stream.spark;
 
+import com.splicemachine.EngineDriver;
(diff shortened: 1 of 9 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -300,6 +303,20 @@ public long count() {
         return union(dataSet, RDDName.UNION.displayName(), false, null);
     }
 
+
+    @Override
+    public DataSet<V> parallelProbe(List<DataSet<V>> dataSets) {
+        DataSet<V> toReturn = null;
+        int i = 0;
+        for (DataSet<V> dataSet: dataSets) {
```
dgomezferro: This function is not used, but the logic is wrong here, since we are not incrementing 'i'.

PR #1316, opened 2017-12-21
SPLICE-1905: Failed to run select query from partitioned external table in case of NULL presence.(2.6)

stpabhi:


stpabhi:
@jyuanca Could you please review and approve?

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -1928,6 +2020,17 @@ public void testPartitionBySmallIntOrcTable() throws Exception {
         resultString = TestUtils.FormattedResult.ResultFactory.toStringUnsorted(rs);
         assertEquals(expected, resultString);
         rs.close();
+
+        methodWatcher.executeUpdate("insert into t_partition_by_smallint values(NULL, 1, 'DDD')");
+        rs = methodWatcher.executeQuery("select a1, count(*) from t_partition_by_smallint group by a1 order by 1");
+
+        expected = "A1 |B1 |C1  |\n" +
+                "----------------\n" +
+                "   NULL    | 1 |";
```
yxia92: There are 3 rows inserted earlier before the row with null on the partition column is inserted, so the expected result is wrong. It should be:
        expected = "A1  | 2 |\n" +
                "----------\n" +
                "  1  | 2 |\n" +
                "  2  | 1 |\n" +
                "NULL | 1 |";

stpabhi: Thanks. Will fix it.

yxia92: Thanks. Change is only needed for branch-2.6. We have fixed the IT in the master branch.

stpabhi: @yxia92 Fixed

stpabhi:
@OlegMazurov Fixed in 2.6. Test passed.

PR #1317, opened 2017-12-28
DB-6429 Test coverage for Diablo

liprais:
added several configs to enable code coverage analysis using jacoco.
in order to generate a code coverage report, run 'mvn clean install -Dmaven.test.failure.ignore=true -Djacoco -Pmem,core,cdh5.12.0'


start-splice-cluster
```diff
@@ -188,9 +188,9 @@ echo "Starting Master and 1 Region Server. Log file is ${SPLICE_LOG}"
 ## (master + region server on 1527)
 
 if [ -z "${SERVER_SSLMODE}" ]; then
-  (${MVN} exec:exec -P${PROFILE},spliceFast ${SYSTEM_PROPS} -Dxml.plan.debug.path=${DEBUG_PATH} > ${SPLICE_LOG} 2>&1) &
+  (${MVN} exec:exec -P${PROFILE},spliceFast,jacoco  ${SYSTEM_PROPS} -Dxml.plan.debug.path=${DEBUG_PATH} > ${SPLICE_LOG} 2>&1) &
```
OlegMazurov: The 'jacoco' mode should not be used unconditionally in every standalone run.

start-splice-cluster
```diff
@@ -188,9 +188,9 @@ echo "Starting Master and 1 Region Server. Log file is ${SPLICE_LOG}"
...
 else
-  (${MVN} exec:exec -P${PROFILE},spliceFastSSL ${SYSTEM_PROPS} -Dxml.plan.debug.path=${DEBUG_PATH} > ${SPLICE_LOG} 2>&1) &
+  (${MVN} exec:exec -P${PROFILE},spliceFastSSL,jacoco ${SYSTEM_PROPS} -Dxml.plan.debug.path=${DEBUG_PATH} > ${SPLICE_LOG} 2>&1) &
```
OlegMazurov: Ditto.

start-splice-cluster
liprais: only when '-Djacoco' and -Pjacoco both provided ,the code coverage anlysis will be enabled ,so it won’t be used unconditionally in every standalone run, when '-Djacoco' is not provided, the jacocoeagent will be '-server',which is harmless I think

OlegMazurov: How are you going to provide '-Djacoco' for start-splice-cluster? 
Why do you need '-Pjacoco' in start-splice-cluster at all?

liprais: You are right,It is unnecessary  for code coverage I have removed it

PR #1318, opened 2018-01-03
SPLICE-1927 Amend pattern string for detecting splice machine ready to accept connections

changli6:
SPLICE-1927 Amend pattern string for detecting splice machine ready to accept connections


snowspeeder:
you can match this grep line similar to the CDH5.8.3 restart script (.../assembly/cdh5.8.3/src/main/resources/scripts), it has no classname.

changli6:
They are good suggestions, thanks @yxia92 @snowspeeder.

PR #1324, opened 2018-01-04
DB-6440 reduce lengthy lineage of transformation for MultiProbeTableS…

yxia92:
…can operation

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
@@ -14,7 +14,6 @@
 
 package com.splicemachine.derby.stream.spark;
 
-import com.splicemachine.EngineDriver;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.types.SQLLongint;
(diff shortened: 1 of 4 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -307,12 +297,27 @@ public long count() {
     @Override
     public DataSet<V> parallelProbe(List<DataSet<V>> dataSets) {
         DataSet<V> toReturn = null;
+        DataSet<V> branch = null;
         int i = 0;
         for (DataSet<V> dataSet: dataSets) {
```
dgomezferro: Would it make sense to return a balanced binary tree? Something like
return parallelProbe(firstHalf(dataSets)).union(parallelProbe(secondHalf(dataSets))

yxia92: Thanks @dgomezferro . My original thinking of the fix is to preserve the original implementation as much as possible. So for inlist of less than 100 (which may be most popular), it will behave the same as the original implementation.  

I also tried the fix to construct a more balanced union-all tree as follows:
{code}
    @Override
    public DataSet<V> parallelProbe(List<DataSet<V>> dataSets) {
        if (dataSets.isEmpty())
            return null;
        return parallelProbe(dataSets, 0, dataSets.size()-1);
    }

    private DataSet<V> parallelProbe(List<DataSet<V>> dataSets, int lo, int hi) {
        if (hi < lo)
            return null;

        if (hi == lo)
            return dataSets.get(hi);

        int mid = (hi + lo) / 2;
        DataSet<V> leftSet = parallelProbe(dataSets, lo, mid);
        DataSet<V> rightSet = parallelProbe(dataSets, mid+1, hi);
        return leftSet.union(rightSet);
    }
{code}

I don't notice any performance improvement. The more balanced union-all tree, in theory, can support inlist of a larger size, but it looks that before we get to an inlist of that large size, we are hitting limit imposed by "spark.driver.maxResultSize".

I see that you have approved the PR, so let me know if you want me to switch to the balanced binary tree implementation.

PR #1326, opened 2018-01-09
DB-6284: add ignore txn cache

jyuanca:


splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
```diff
@@ -0,0 +1,112 @@
...
+        if (!initialized) {
+            init();
+            initialized = true;
+        }
+        if (!cache.isEmpty()) {
+            for (Pair<Long, Long> range : cache) {
+                if (txnId > range.getFirst() && txnId < range.getSecond()) {
+                    ignore = true;
+                    break;
+                }
```
yxia92: My understanding is that the pairs will be in order in the cache. Would it be better to add 
else if (txnId <= range.getFirst()) {
     return ignore;
}
So that we can terminate early. I'm thinking if the system has gone through restore quite a few times, maybe this could help to reduce some performance impact?

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -189,6 +190,13 @@ protected void doResolve(DataCell data,long ts){
 		 * it matches, then we can see it.
 		 */
         long timestamp=data.version();//dataStore.getOpFactory().getTimestamp(data);
+        SIDriver driver = SIDriver.driver();
+        IgnoreTxnSupplier ignoreTxnSupplier = null;
+        if (driver != null) {
```
dgomezferro: This looks weird, when could the driver be null? I don't think this should be allowed.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
```diff
@@ -0,0 +1,112 @@
...
+        this.partitionFactory = partitionFactory;
+        this.txnOperationFactory = txnOperationFactory;
+        cache = new HashSet<>();
+    }
+
+    public boolean shouldIgnore(Long txnId) throws IOException{
+
+        boolean ignore = false;
+        if (!initialized) {
+            init();
```
dgomezferro: I wouldn't use lazy initialization here, it introduces too many issues for little gain. If you want to use lazy init you have to make sure the init() is properly synchronized, you'd need double checked locking, etc. etc. 

I'd just initialize the cache when we create it, it's not an expensive operation.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplier.java
dgomezferro: Building on top of that, I think at some point it was discussed having a global min/max pair to avoid checking the cache in the common case. For instance, if the pairs are (10, 20), (100, 120), the global min/max would be (10, 120). When we check txn 5 or 200, since they're outside the min/max we don't have to traverse the cache.

dgomezferro:
Argh, I meant to "request changes" not approve changes. The global min/max bit is up for discussion but the lazy initialization bit is faulty, this should be reverted or fixed.

PR #1330, opened 2018-01-09
SPLICE-1870 fix update through index lookup path

yxia92:


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/UpdateOperationIT.java
```diff
@@ -506,6 +519,114 @@ public void testUpdateUsingBoolDescIndex() throws Exception {
...
+                    " 8 |10000 | 8 |GGG |GGG |";
+            rs = methodWatcher.executeQuery("select * from t1");
+            assertEquals(expected, TestUtils.FormattedResult.ResultFactory.toString(rs));
+            rs.close();
+        }
+
+        // roll back the update
+        try {
+            conn.rollback();
+        } catch (Exception e) {} // Swallow for HFile Bit Running in Control
```
dgomezferro: Why do you have to swallow the exception?

yxia92: Thanks @dgomezferro for catching this. It is a careless copy-paste error. My purpose was just to rollback the update. I've updated the IT to correct the misleading code and comment.

PR #1332, opened 2018-01-09
SPLICE-1974 Splice derby timestamp format not compliant with Hadoop

changli6:
Modified the log4j default timestamp format for standalone version. The cluster version, the timestamp format of log file could be set in Cloudera Manager (CDH), Ambari (HDP) and file /opt/mapr/hbase/hbase-1.1.1/conf/log4j.properties (MapR in default).


wtang16:
@changli6 please prepare for 2.5 and 2.6 code shipment too. Some customers will want it.

PR #1342, opened 2018-01-12
DB-6429 test coverage for diablo

liprais:
fix coverage module's parent version problem.


dgomezferro:
As a suggestion, you should try to make the PRs as simple as possible, this one lists 23 commits for a simple change. I would have suggested using a separate branch for the version fix. It's not a big deal but it makes the review easier.

I think it makes sense to have multiple commits if there are several interrelated changes though.

PR #1343, opened 2018-01-12
DB-6343 Handle CannotCommit exception correctly

dgomezferro:


splice_machine/src/main/java/com/splicemachine/pipeline/Exceptions.java
```diff
@@ -63,7 +63,7 @@ public static StandardException parseException(Throwable e,PipelineExceptionFact
 
             ErrorState state = ErrorState.stateFor(e);
 
-            return state.newException(rootCause);
+            return state.newException(e);
```
yxia92: @dgomezferro In the JIRA comment, you said "In Exceptions.parseException we used the rootCause to select an ErrorState but then fed it the original exception", looking at the code and fix here, it seems the opposite: We use the original exception(e) to select an ErrorState. Should the comment in JIRA be updated or should the fix to change the line "ErrorState state = ErrorState.stateFor(e);" to "ErrorState state = ErrorState.stateFor(rootCause);" instead?

yxia92: Please ignore my previous comment. So it looks that we should use "e" to get the ErrorState and also to create the new Exception instead of "rootCause". I tried the other way around, it is returning unmeaningful error for the IT case. 
+1 for the code change.

dgomezferro: The code is a bit confusing, if the rootCause is not a StandarException or SQLException, we process the original exception using the ExceptionFactory, and this also extracts the rootCause in some cases.

PR #1346, opened 2018-01-12
DB-6463 Add JDBC timeout support

dgomezferro:
Fix non-terminating queries when they time out or are killed. Clear
resources in Spark after query has been killed/times out.


dgomezferro:
@wtang16 I looked at adding Timeout directives to those tests, but they don't do anything, I'm not sure why.

wtang16:
@dgomezferro How did you add them? Do you need to disable part of ITs then?

dgomezferro:
@wtang16 no need to disable them, the only problem would be if someday the tests break and run for longer they aren't terminated promptly by JUnit.

wtang16:
Sounds good. Thanks Daniel.

PR #1350, opened 2018-01-12
DB-6534 Close FutureIterator to avoid race condition

dgomezferro:


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/RowCountOperationIT.java
```diff
@@ -69,11 +69,15 @@
     private TestConnection conn;
     private String connectionString;
 
+    private static final String CONTROL = "jdbc:splice://localhost:1527/splicedb;create=true;user=splice;password=admin";
```
wtang16: To ensure the execution goes through CONTROL path, do we need to add explicit "useSpark=false"?

dgomezferro: Good point, I added it.

PR #1351, opened 2018-01-13
DB-6194 add system procedure to enable/disable all column statistics

yxia92:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceSystemProcedures.java
```diff
@@ -558,6 +558,26 @@ public SpliceSystemProcedures(DataDictionary dictionary) {
...
+                            .numOutputParams(0)
+                            .numResultSets(0)
+                            .modifiesSql()
+                            .varchar("schema",1024)
+                            .varchar("table",1024)
+                            .ownerClass(StatisticsAdmin.class.getCanonicalName())
+                            .build();
+                    procedures.add(disableStatsForAllColumns);
+
         			/*
```
jyuanca: schemaName and tableName are catalog types

yxia92: Thanks @jyuanca for the comment. updated the fix accordingly.

PR #1357, opened 2018-01-15
DB-6534 Close FutureIterator to avoid race condition (2.5)

dgomezferro:


snowspeeder:
@wtang16 need approval for 2.5/2.6

PR #1361, opened 2018-01-16
SPLICE-1900: Incorrect error message while reading data from Empty external table of AVRO file format.(2.6)

stpabhi:


jyuanca:
+1

PR #1370, opened 2018-01-22
DB-6558: cache more database properties

jyuanca:
I have tested on the cluster. The # of read requests for splice:16 does not increase if I keep running SQL queries


dgomezferro:
If I'm not mistaken, as long as we don't add the default the property should survive database restarts. Do we want it to survive or not?

jyuanca:
Service properties are not persisted. See PropertyConglomerate.saveProperty()

PR #1373, opened 2018-01-22
SPLICE-2006 TPC-C run fails to complete, Region Servers die

OlegMazurov:
Removing checking if already closed fixes memory leak.
It may cause multiple logging of statement processing finish, though.
I'm going to file another JIRA for redoing closing of SpliceBaseOperation.


OlegMazurov:
I can confirm that applying the proposed fix will cause multiple "Finished executing prepared statement" logs for the same statement. As suggested, a new JIRA will be filed for that.

OlegMazurov:
I tried to avoid multiple logging for finished statements by checking if closed only for that logging but that didn't work (each log was still doubled). I reverted that commit back to the originally proposed fix.

PR #1377, opened 2018-01-23
SPLICE-2012 HMaster doesn't exit after shutdown

OlegMazurov:
Shutdown the database lifecycle executor.

### DIFF
--- a/pipeline_api/src/main/java/com/splicemachine/lifecycle/DatabaseLifecycleManager.java
+++ b/pipeline_api/src/main/java/com/splicemachine/lifecycle/DatabaseLifecycleManager.java
@@ -75,7 +75,9 @@ public class DatabaseLifecycleManager{
     }
 
     public DatabaseLifecycleManager(){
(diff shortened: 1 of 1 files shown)


pipeline_api/src/main/java/com/splicemachine/lifecycle/DatabaseLifecycleManager.java
```diff
@@ -75,11 +75,7 @@ public static DatabaseLifecycleManager manager(){
     }
 
     public DatabaseLifecycleManager(){
-       this(Executors.newSingleThreadExecutor());
-    }
-
-    public DatabaseLifecycleManager(Executor lifecycleExecutor){
-        this.lifecycleExecutor = lifecycleExecutor;
+        this.lifecycleExecutor = Executors.newSingleThreadExecutor();
```
dgomezferro: Could you give the thread a name? It makes it easier to identify it in a thread dump

PR #1380, opened 2018-01-24
DB-6543: vacuum disabled table

jyuanca:


hbase_storage/src/main/java/com/splicemachine/access/hbase/H10PartitionAdmin.java
```diff
@@ -82,14 +82,23 @@ public PartitionCreator newPartition() throws IOException{
...
+
+        TableName tn = tableInfoFactory.getTableInfo(tableName);
         try{
-            admin.disableTable(tableInfoFactory.getTableInfo(tableName));
-            admin.deleteTable(tableInfoFactory.getTableInfo(tableName));
+            if (admin.isTableEnabled(tn)) {
+                admin.disableTable(tn);
+            }
+            admin.deleteTable(tn);
         }catch(TableNotFoundException ignored){
```
dgomezferro: Please log this exception, maybe with a WARN level

PR #1387, opened 2018-01-25
DB-6478 Add result columns SUBMITTED, ENGINE and JOBTYPE in SYSCS_GETRUNNING_OPERATIONS

changli6:


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -1566,6 +1581,27 @@ public static void SYSCS_GET_RUNNING_OPERATIONS_LOCAL(final ResultSet[] resultSe
         }
         resultSet[0] = new EmbedResultSet40(conn, resultsToWrap, false, null, true);
     }
+
+    private static String getElapsedTimeStr(Date begin, Date end)
```
wtang16: Starting Java8, we may just use between method in Duration class: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#between-java.time.temporal.Temporal-java.time.temporal.Temporal-. This may reduce the complexity of our code.

wtang16: Current implementation is fine nonetheless. 
+1

PR #1390, opened 2018-01-25
DB-6523 add hint useDefaultRowCount and defaultSelectivityFactor, als…

yxia92:
…o fix minr bugs in selectivity and join costing


wtang16:
Good catch Yi.

PR #1395, opened 2018-01-27
DB-6574: retry write workload on host unreachable exceptions

jyuanca:
Retry on some host unreachable exception. HBase will reassign the region soon.


pipeline_api/src/main/java/com/splicemachine/pipeline/client/WriteResult.java
```diff
@@ -85,7 +85,13 @@ public boolean shouldCancel() {
         return code.shouldCancel();
     }
     public boolean canRetry() {
```
wtang16: How do we control the maximum number of retries? What happens if max retries are exceeded.

jyuanca: It will be retired infinitely

PR #1399, opened 2018-01-30
SPLICE-175 External Olap server

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -0,0 +1,437 @@
...
+
+                // Set up the container launch context for the application master
+                ContainerLaunchContext amContainer =
+                        Records.newRecord(ContainerLaunchContext.class);
+                amContainer.setCommands(
+                        Collections.singletonList(prepareCommands(
+                                "$JAVA_HOME/bin/java",
+                                " -Xmx" + memory + "M " + outOfMemoryErrorArgument() + " " +
+                                        OlapServerMaster.class.getCanonicalName() +
+                                        " " + serverName.toString() + " " + olapPort +
```
wtang16: Do we need to worry about log rotation and retention here?

hbase_sql/src/main/resources/yarn-site.xml
```diff
@@ -105,12 +105,12 @@
   </property>
   <property>
     <name>yarn.nodemanager.resource.memory-mb</name>
-    <value>4096</value>
+    <value>8192</value>
```
wtang16: Are we requiring more memory than before now?

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -248,11 +249,11 @@ private static SparkSession initializeSparkSession() {
         try {
             String principal = null;
             String p = null;
-            String hostname = HConfiguration.unwrapDelegate().get("hbase.master.hostname");
+            String hostname = HConfiguration.unwrapDelegate().get("splice.olapServer.hostname");
```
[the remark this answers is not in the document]
dgomezferro: Good point , I'll address it.

hbase_sql/src/main/resources/yarn-site.xml
dgomezferro: On standalone we are starting one more process so we do require more memory. We won't be using the full 8Gb though, that's an upper limit for reservations, which could be made tighter if we wanted.

PR #1402, opened 2018-01-30
SPLICE-2008 Resolve transactions during flushes

dgomezferro:


hbase_storage/src/main/java/com/splicemachine/si/impl/server/SimpleCompactionContext.java
```diff
@@ -8,56 +8,99 @@
  * Splice Machine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Affero General Public License for more details.
- * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * You should have received a copy of the GNU Affero General Public License aprivate AtomicLong with Splice Machine.
```
wtang16: Please fix this comment line in all branches.

PR #1407, opened 2018-01-31
DB-6594 fix illegal merge join

yxia92:
Please see comment in [DB-6594](https://splicemachine.atlassian.net/browse/DB-6594?focusedCommentId=52851&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-52851) for a brief explanation.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
@@ -227,30 +227,30 @@ public class MergeJoinStrategy extends HashableJoinStrategy{
                 assert relop instanceof BinaryRelationalOperatorNode:
                         "Programmer error: RelationalOperator of type "+ relop.getClass()+" detected";
                 BinaryRelationalOperatorNode bron = (BinaryRelationalOperatorNode)relop;
-                ColumnReference innerColumn=relop.getColumnOperand(innerTable);
-                ColumnReference outerColumn=getOuterColumn(bron,innerColumn);
-                if (innerColumn == null || outerColumn == null) continue;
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
```diff
@@ -227,30 +227,30 @@ private boolean mergeable(RowOrdering outerRowOrdering,
...
-                if(innerColumnNumber==innerColumnPosition){
-                    innerColumns.set(i);
-                    int outerTableNum=outerColumn.getTableNumber();
-                    int outerColNum=outerColumn.getColumnNumber();
-                    if(ascending){
-                        int outerPos = outerRowOrdering.orderedPositionForColumn(RowOrdering.ASCENDING,outerTableNum,outerColNum);
-                        if(outerPos>=0){
-                            outerColumns.set(outerPos);
-                            innerToOuterJoinColumnMap[i] = outerPos;
+                if (bron.getOperator() == RelationalOperator.EQUALS_RELOP) {
```
dgomezferro: Are we forbidding any valid merge joins this way? If the last predicate (on the last index column) is an inequality we could still use merge join, right?

yxia92: Yes, @dgomezferro If all the leading index columns have either equality condition or are equal to a constant value, merge join can still be picked up even if the last index column has an inequality condition.
The 4th test case in MergeJoinIT.testMergeFeasiblityWithNonEqualityJoinConditionsOnKeys() is such an example. Thanks.

dgomezferro: @yxia92  I see, thanks!

PR #1409, opened 2018-01-31
SPLICE-1995: correct imported rows

jyuanca:


jyuanca:
This is a limitation of JDBC standard. see java.sql.Statement.executeUpdate()

yxia92:
I see. Thanks Jun!

PR #1412, opened 2018-02-01
DB-6582: improve export performance

jyuanca:
- Remove RDD cache
- Get written rows from task metrics.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkExportDataSetWriter.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkExportDataSetWriter.java
@@ -20,6 +20,7 @@ import java.io.ObjectOutput;
 import java.io.OutputStream;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.types.SQLLongint;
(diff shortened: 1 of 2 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkExportDataSetWriter.java
```diff
@@ -93,12 +98,25 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
         job.setOutputFormatClass(SparkDataSet.EOutputFormat.class);
         job.getConfiguration().set("mapred.output.dir",directory);
 
-        JavaRDD<V> cached=rdd.cache();
-        long writtenRows=cached.count();
+        SparkContext context = rdd.context();
+        List<Long> recordsWrittenByTasks = Lists.newArrayList();
```
dgomezferro: I think it would be cleaner to use an AtomicLong and not having to synchronize() on anything.

jyuanca: Good point. Thanks @dgomezferro

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkExportDataSetWriter.java
```diff
@@ -93,16 +98,23 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
             .setName(String.format("Export Directory: %s", directory))
             .saveAsNewAPIHadoopDataset(job.getConfiguration());
-        cached.unpersist();
 
+        long end = System.currentTimeMillis();
         ValueRow valueRow=new ValueRow(2);
-        valueRow.setColumn(1,new SQLLongint(writtenRows));
-        valueRow.setColumn(2,new SQLInteger(0));
+        valueRow.setColumn(1,new SQLLongint(writtenRows.get()));
+        valueRow.setColumn(2,new SQLLongint(end-start));
```
yxia92: If we change the second column to be the elapsed time here, do we need to change ControlExportDataSetWriter to make them consistent?

PR #1413, opened 2018-02-01
DB-6582: improve export performance(2.6)

jyuanca:


jyuanca:
Did't know export works on control too. Thanks @yxia92

PR #1421, opened 2018-02-02
DB-6603: prevent incremental backup from losing changes(2.5)

jyuanca:
During the first full backup, some incremental changes were produced right after a region finishes backup. SpliceHFileCleaner should keep incremental changes either it finds there was successful backup in the system, or there is an ongoing backup.


jyuanca:
Please also review ee branch

PR #1425, opened 2018-02-02
DB-6345 fix query failure with non-covering index as right table of a…(2.6.1)

yxia92:
… join

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/NonCoveringIndexIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/NonCoveringIndexIT.java
@@ -83,8 +83,7 @@ public class NonCoveringIndexIT extends SpliceUnitTest {
     }
 
     @Test
-    public void testJoinWithNonCoveringIndexInDT() throws Exception {
-        /* Q1 inner join */
+    public void testJoinWithNonCoveringIndexInDTAndInnerJoin() throws Exception {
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/NonCoveringIndexIT.java
```diff
@@ -0,0 +1,174 @@
...
+                .withRows(rows(row(3, 3, 3),
+                        row(4, 4, 4),
+                        row(4, 5, 5))).create();
+    }
+
+    @Test
+    public void testJoinWithNonCoveringIndexInDT() throws Exception {
+        /* Q1 inner join */
+        String sql = format("select * from --splice-properties joinOrder=fixed\n" +
+                "t1,\n" +
```
jleach4: Just a style thing for me, I like my methods to test a single concept and if it fails to be clear what the issue is for the developer.  Seems like you are testing 4 items.

jleach4: Good tests by the way.

yxia92: Thanks @jleach4 . The reason I grouped 4 cases into the same function is that they are variations of the same issue.  I can put them in separate functions if it makes the test more clear. I'll push a change shortly.

PR #1433, opened 2018-02-05
 SPLICE-2008 Resolve transactions during flushes

dgomezferro:
Take 2 with fix for restore (ignore missing transactions instead of raising NPE) and some ITs


dgomezferro:
Please hold on merging @yxia92 , the IT is failing after resolving the conflict.

PR #1441, opened 2018-02-06
DB-6262 Improve failure handling during initialization (2.6)

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/lifecycle/RegionServerLifecycle.java
```diff
@@ -61,14 +66,23 @@ public void distributedStart() throws IOException{
...
                     admin.createTable(desc);
-                }catch(InitializationCompleted ic){
+                    LOG.error("We could create SPLICE_INIT table, this means the SpliceMasterObserver hasn't loaded properly");
+                    try {
+                        admin.disableTable(spliceInit);
+                        admin.deleteTable(spliceInit);
+                    } catch (IOException e) {
+                        LOG.warn("Received exception while cleaning up SPLICE_INIT", e);
+                    }
+                    throw new DoNotRetryIOException("We could create SPLICE_INIT table, SpliceMasterObser not loaded");
```
wtang16: Shouldn't "SpliceMasterObser" be "SpliceMasterObserver"?

hbase_sql/src/main/java/com/splicemachine/lifecycle/RegionServerLifecycle.java
```diff
@@ -61,14 +66,23 @@ public void distributedStart() throws IOException{
         try(Admin admin=conn.getAdmin()){
             do{
                 onHold=false;
+                TableName spliceInit = TableName.valueOf(SpliceMasterObserver.INIT_TABLE);
                 try{
-                    HTableDescriptor desc=new HTableDescriptor(TableName.valueOf(SpliceMasterObserver.INIT_TABLE));
+                    HTableDescriptor desc=new HTableDescriptor(spliceInit);
                     desc.addFamily(new HColumnDescriptor(Bytes.toBytes("FOO")));
```
wtang16: Why do we use "FOO" as the description?

hbase_sql/src/main/java/com/splicemachine/lifecycle/RegionServerLifecycle.java
```diff
@@ -84,13 +98,22 @@ public void distributedStart() throws IOException{
...
                         ZkUtils.safeInitializeZooKeeper();
-                    }catch(InterruptedException e){
+                    } catch (InterruptedException e) {
                         throw new InterruptedIOException();
-                    }catch(KeeperException e){
+                    } catch (KeeperException e) {
                         throw new IOException(e);
                     }
+                } catch (TableExistsException e) {
+                    LOG.error("SPLICE_INIT already exists, this shouldn't happen. Remove it and retry");
```
wtang16: We may want to rewrite this to something like "SPLICE_INIT table exists. Removing it."

PR #1445, opened 2018-02-07
SPLICE-2032 Reenable OlapServerIT and increase robustness

dgomezferro:
Fix TPCH1G failure by increasing OlapServer memory


wtang16:
Talked to Daniel on doc. He will update our internal wiki first as part of the system architecture doc refresh.

PR #1450, opened 2018-02-09
DB-6617 Re-enable 7 passed cases and commented reasons to ignored cases

changli6:


db-engine/src/test/java/com/splicemachine/db/iapi/types/SQLDecimalTest.java
```diff
@@ -99,7 +99,7 @@ public void serdeNullValueData() throws Exception {
             }
         // TODO
         @Test
-        @Ignore
+        @Ignore //Tracked by SPLICE-2050
```
yxia92: It would be better to use the form @Ignore("Tracked by SPLICE-2050")

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -865,7 +865,7 @@ public void testWriteReadWithPreExistingAvroAndOr() throws Exception {
                 "  2  |",TestUtils.FormattedResult.ResultFactory.toString(rs));
     }
 
-    @Test @Ignore
+    @Test @Ignore //Tracked by SPLICE-2047
```
yxia92: It would be better to use the form @ignore("Tracked by SPLICE-2047")

PR #1461, opened 2018-02-13
DB-6541 User defined function does not work on spark

OlegMazurov:
SPLICE-1983 is undone to restore UDF functionality. ActivationHolder.close() is uncommented in SMRecordReaderImpl to avoid memory leaks. Trigger ITs are fixed by guarding against out-of-order  activationHolder.close() called from SMRecordReaderImpl and SMRecordWriter (violated nesting or push/pop semantics) whether sharing the same activationHolder or using different instances.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/services/context/ContextService.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/services/context/ContextService.java
@@ -362,8 +362,11 @@ public final class ContextService{
             }
         }
 
-        if(cm.activeCount!=-1){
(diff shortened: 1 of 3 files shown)


OlegMazurov:
A port to master requires additional work as ContextService/ContextManager were reworked there.
I'll create a PR for master later.

jyuanca:
can you add an IT to prevent regression?

OlegMazurov:
I filed DB-6651: Add an IT to cover user defined function regression, to offload that work from this PR.

jyuanca:
+1

PR #1468, opened 2018-02-15
SPLICE-1587 add modifyPriv for Schema level privilege

yxia92:
SPLICE-1587 add authorization check for PIN table statement
SPLICE-1587 add IT
SPLICE-1587 add upgrade logic

Fix description is in [SPLICE-1587](https://splice.atlassian.net/browse/SPLICE-1587?focusedCommentId=27119&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27119)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementSchemaPermission.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementSchemaPermission.java
@@ -33,9 +33,12 @@ package com.splicemachine.db.iapi.sql.dictionary;
 
 import com.splicemachine.db.catalog.UUID;
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.services.context.ContextManager;
 import com.splicemachine.db.iapi.sql.conn.Authorizer;
 import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
+import com.splicemachine.db.iapi.sql.depend.DependencyManager;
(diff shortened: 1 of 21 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementSchemaPermission.java
```diff
@@ -128,6 +151,87 @@ public void check( LanguageConnectionContext lcc,
...
+                        // another role has been set), we are able to
+                        // invalidate the ps or activation (the latter is used
+                        // if the current role changes).
+                        DependencyManager dm = dd.getDependencyManager();
+                        RoleGrantDescriptor rgd =
+                                dd.getRoleDefinitionDescriptor(role);
+                        ContextManager cm = lcc.getContextManager();
+
+                        dm.addDependency(ps, rgd, cm);
+                        dm.addDependency(activation, rgd, cm);
```
jleach4: The distributed part of dependency management is not functioning it looks like for roles...
https://splicemachine.atlassian.net/browse/DB-6670

yxia92: Thanks John for bringing this up to my attention. I'll take a look at DB-6670. But I think this is a general role grant/revoke issue, not specific to the modify schema privilege in this ticket, we can proceed with the review of this PR, right?

jyuanca:
@yxia92 How did you handle upgrade from build 1806 to 1807?

yxia92:
@jyuanca thanks for the question. The upgrade script will be triggered for any version upgrade before version 2.7.1. But triggering it multiple times won't cause problem, because we check the presence of the new column ModifyPriv. Only if it is not present, will we make dictionary change.

PR #1469, opened 2018-02-16
SPLICE-1587 add modifyPriv for Schema level privilege(2.5)

yxia92:
SPLICE-1587 add authorization check for PIN table statement
SPLICE-1587 add IT
SPLICE-1587 add upgrade logic


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -61,6 +61,7 @@ else if(v1==v2)
         scripts.put(new Splice_DD_Version(sdd,1,0,0),new UpgradeScriptForFuji(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,1,1,1),new LassenUpgradeScript(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,2,5,0),new K2UpgradeScript(sdd,tc));
+        scripts.put(new Splice_DD_Version(sdd,2,7,1), new UpgradeScriptForModifySchemaPermission(sdd,tc));
     }
```
jyuanca: We intended to add a script here to upgrade from previous version to 2.5

yxia92: I pushed a change only for branch-2.5 to change the version threshold from 2.7.1 to 2.5.1, to avoid confusion. Basically, version upgrade within 2.5 will trigger the upgrade script, but if the field ModifyPriv is already present, this upgrade script will exit without doing anything. Thanks.

PR #1470, opened 2018-02-16
DB-6541 User defined function does not work on spark

OlegMazurov:
Forward port from branch-2.5
Changes to ContextService.java are new as the content of the file was different from that in branch-2.5


jyuanca:
+1

PR #1471, opened 2018-02-19
DB-6662 Make JDBC timeout more robust

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
```diff
@@ -91,17 +102,20 @@ public void submit() throws Exception {
...
-                        continue;
+                    Future<Object> resultFuture = futures.remove();
+                    Object result = null;
+                    while (result == null) {
+                        if (jobStatus != null && !jobStatus.isRunning()) {
+                            throw new CancellationException("The olap job is no longer running, cancelling Spark job");
+                        }
+                        
+                        try {
+                            result = resultFuture.get(10, TimeUnit.SECONDS);;
```
yxia92: Please remove the redundant semi-colon.

hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
```diff
@@ -43,12 +55,12 @@
     private final int clientBatchSize;
     private final String host;
     private final JavaRDD<T> rdd;
-    private final ExecutorCompletionService<Object> completionService;
     private final ExecutorService executor;
     private final int clientBatches;
     private final UUID uuid;
     private final OperationContext<?> context;
     private OlapStatus jobStatus;
+    private BlockingQueue<Future<Object>> futures = new LinkedBlockingDeque<>();
```
yxia92: Is the use of LinkedBlockingDeque instead of LinkedBlockingQueue on purpose?

dgomezferro: No, initially I tried a Deque but I realized it wasn't needed.

dgomezferro:
@yxia92 The mechanism to cancel the running query was to cancel the job in QueryJob when we catch the CancellationException. This usually works except when the job takes a long time to be scheduled and the "cancelJobGroup()" call comes before the job exists. As far as I know this only happens at the beginning of the execution, when the Spark application isn't fully loaded yet. With this change we also cancel the job directly in StremableRDD if needed. I couldn't reproduce the original problem though, since it didn't happen at the beginning of the Spark application, but this change might also help in that case.

yxia92:
@dgomezferro thanks for the explanation.

PR #1478, opened 2018-02-21
SPLICE-2081 Excessive creation of Kryo instances

OlegMazurov:
Get a cached value when available.


jleach4:
Nice!!!  It will speed up Multi-probe and NLJ...  +1

PR #1485, opened 2018-02-21
SPLICE-1988: CDH 5.14.0 support

stpabhi:


stpabhi:
@jleach4 should I reopen against branch-2.5?

jleach4:
@stpabhi Yes it was merged last evening.  Sorry for the mongo commit, but I really feel like the install and deployments will be clearer without uber jars, etc.

stpabhi:
@jleach4 No problem. Yes, I think it will be nice.

PR #1487, opened 2018-02-22
DB-6685 invalidate dictionary cache after update schema owner

yxia92:
Please see problem description and solution in [DB-6685](https://splicemachine.atlassian.net/browse/DB-6685?focusedCommentId=53176&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-53176)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
@@ -385,6 +385,7 @@ public class DDLUtils {
         dd.getDataDictionaryCache().clearPermissionCache();
         // clear  TableDescriptor cache as it may reference the schema with an out-of-date authorization id
         dd.getDataDictionaryCache().clearOidTdCache();
+        dd.getDataDictionaryCache().clearNameTdCache();
     }
 
     public static void preCreateIndex(DDLMessage.DDLChange change, DataDictionary dd, DependencyManager dm) throws StandardException {
--- a/splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
@@ -14,27 +14,16 @@
 
 package com.splicemachine.derby.ddl;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
 import com.carrotsearch.hppc.BitSet;
-import com.splicemachine.db.iapi.services.loader.ClassFactory;
-import com.splicemachine.db.iapi.sql.dictionary.*;
-import org.apache.log4j.Logger;
-
 import com.splicemachine.access.api.SConfiguration;
 import com.splicemachine.concurrent.Clock;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.services.io.FormatableBitSet;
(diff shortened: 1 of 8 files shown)


splice_machine/src/test/java/com/splicemachine/derby/security/SchemaPrivilegeIT.java
```diff
@@ -1275,4 +1281,25 @@ public void testCreateAllPrivilegesToUserTableNotAllowed() throws Exception {
         assertFailed(user2Conn, query, SQLState.AUTH_NOT_OWNER);
 
     }
+
```
dgomezferro: Could you add some ITs dealing with multiple region servers, similar to those in ClusterDDLTestIT ? If you want to keep them in this class you can tag the test as @Category(HBaseTest) so that it doesn't run in the mem platform

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
```diff
@@ -382,6 +377,16 @@ public static void preDropSchema(DDLMessage.DDLChange change, DataDictionary dd,
...
     }
 
+    public static void preUpdateSchemaOwner(DDLMessage.DDLChange change, DataDictionary dd, DependencyManager dm) throws StandardException {
+        if (LOG.isDebugEnabled())
+            SpliceLogUtils.debug(LOG,"preUpdateSchemaOwner with change=%s",change);
+        dd.getDataDictionaryCache().schemaCacheRemove(change.getUpdateSchemaOwner().getSchemaName());
+        // clear permission cache as it has out-of-date permission info for the schema
+        dd.getDataDictionaryCache().clearPermissionCache();
+        // clear  TableDescriptor cache as it may reference the schema with an out-of-date authorization id
+        dd.getDataDictionaryCache().clearOidTdCache();
```
dgomezferro: Doesn't the normal dm.invalidateFor() work here?

splice_machine/src/test/java/com/splicemachine/derby/security/SchemaPrivilegeIT.java
yxia92: Thanks @dgomezferro I'll add the test for multiple region server case. I manually tested with connection to 1527 and 1528, but it is a good idea to put it in IT.

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
jyuanca: @dgomezferro that does not invalidate permission cache.

jyuanca: @yxia92 Does schema cache and nameTdCache need to be invalidated?

yxia92: @jyuanca Thanks for the comment. Yes, nameTdCache needs to be invalidated too. For schemaCache, I think we have invalidated the specific item in DDLUtils.preUpdateSchemaOwner() with the code 
{code}      dd.getDataDictionaryCache().schemaCacheRemove(change.getUpdateSchemaOwner().getSchemaName());
{code}
Is it not enough?

yxia92: I've added code to invalidate nameTdCache too.

splice_machine/src/test/java/com/splicemachine/derby/security/SchemaPrivilegeIT.java
yxia92: I've updated the IT to test two region servers too. Thanks.

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
jyuanca: @yxia92 I missed that part :-)

jyuanca:
+1 pending on test changes

PR #1491, opened 2018-02-22
SPLICE-2087 Fix HDP2.5.5 Build (2.5)

jleach4:
Fixes hdp2.5.5 build issues on jenkins.  Unfortunately, this does not use a HDP supplied Spark.


jleach4:
@OlegMazurov Was not claiming this fix allowed for installation.  I am still trying to figure out the installation...

PR #1494, opened 2018-02-22
DB-6685 invalidate dictionary cache after update schema owner(2.7)

yxia92:
DB-6685 address review comments by adding IT for multiple region servers and invalidate nameTdCache too


jyuanca:
+1

PR #1502, opened 2018-02-27
DB-6722 Fixing Log4j inclusion in hbase_sql jar

jleach4:
Fixes log4h inclusion which causes issues for log files on the cluster.


stpabhi:
@jleach4 That one liner 😞. Thanks for the help

stpabhi:
+1 I have tested this fix on CDH 5.14 and it's working. Logs are now properly routed to role logs instead of stdout.

PR #1503, opened 2018-02-28
SPLICE-2067: Fix for Client cannot authenticate via:[TOKEN, KERBEROS]

stpabhi:
@dgomezferro I'm able to fix the Client cannot authenticate via:[TOKEN, KERBEROS] issue. With this fix Olap Server no longer fails with the exception. Please review


OlegMazurov:
This fix also needs to be ported to 2.7/master

jyuanca:
@stpabhi Can you port to 2.7 and master brnaches

PR #1505, opened 2018-02-28
DB-6735 Fixing HDP2.6.3 build to support OLAP Server properly

jleach4:
Fixes install of HDP2.6.3 when OLAP Server is involved.

### DIFF
--- a/assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
+++ b/assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
@@ -226,8 +226,8 @@ class SPLICEMACHINE251ServiceAdvisor(service_advisor.ServiceAdvisor):
         "splice.txn.activeTxns.cacheSize" : "10240",
         "splice.txn.completedTxns.concurrency" : "128",
         "splice.txn.concurrencyLevel" : "4096",
-        "splice.olap.server.memory" : "8g",
-        "splice.olap_server.memoryOverhead" : "2g",
+        "splice.olap.server.memory" : "8192",
+        "splice.olap_server.memoryOverhead" : "2048",
         "splice.olap_server.virtualCores" : "2"
     }
     return hbase_site_desired_values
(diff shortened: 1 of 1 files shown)


assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
```diff
@@ -277,10 +262,10 @@ def getMasterDashDProperties(self):
         "splice.spark.yarn.am.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native",
         "splice.spark.yarn.am.waitTime=10s",
         "splice.spark.yarn.executor.memoryOverhead=2048",
-        "splice.spark.driver.extraJavaOptions=-Dlog4j.configuration=file:/etc/spark/conf/log4j.properties",
+        "splice.spark.yarn.am.extraJavaOptions=-Dhdp.version=2.6.3.0-235",
```
snowspeeder: do we need to provide precise version to the build level?  will be issue if somebody tried different version of 2.6.3.0?

jleach4: Yep and that would be a good thing since we have not tested it.

assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
```diff
@@ -244,15 +225,19 @@ def getHBaseSiteDesiredValues(self):
         "splice.timestamp_server.clientWaitTime" : "120000",
         "splice.txn.activeTxns.cacheSize" : "10240",
         "splice.txn.completedTxns.concurrency" : "128",
-        "splice.txn.concurrencyLevel" : "4096"
+        "splice.txn.concurrencyLevel" : "4096",
+        "splice.olap.server.memory" : "8g",
```
dgomezferro: This expects an Int, use 8192

assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
```diff
@@ -244,15 +225,19 @@ def getHBaseSiteDesiredValues(self):
...
+        "splice.olap_server.memoryOverhead" : "2g",
```
dgomezferro: Expects an Int, use 2048

PR #1510, opened 2018-03-02
Db 6619 2.5 urgent: CDH5.13.2 Support

jleach4:
5.13.2 Branch Support, Kafka version fix, and removal of dead spark_assembly_id.


assembly/cdh5.13.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+import os
+import re
+import sys
+import tarfile
+import time
+
+def _get_parcel_dirname(parcel_name):
+  """
+  Extract the required parcel directory name for a given parcel.
+  eg: CDH-5.0.0-el6.parcel -> CDH-5.0.0
```
MurrayBrown: update 'help message' to match base cdh 5.13.2 version?

genedavis:
+1

OlegMazurov:
Do we have any compile dependency on cdh5.13.2? In other words, are our bits going to be any different from those on cdh5.13.0?

jleach4:
Once you merge, @perryluopl can add 5.13.2 to Jenkins script for testing/etc.  Our actual source is the same for both but each run will compile and test against their respective cloudera versions.

pom.xml
```diff
@@ -894,9 +889,8 @@
                 <spark.major.version>spark2.1</spark.major.version>
                 <hive.version>1.1.0-cdh5.9.0</hive.version>
                 <zookeeper.version>3.4.5-cdh5.9.0</zookeeper.version>
-                <spark.version>2.1.0.cloudera1</spark.version>
-                <spark-assembly-id>spark-assembly-hadoop2.6.0-cdh5.9.0</spark-assembly-id>
-                <kafka.version>0.9.0-kafka-2.0.2</kafka.version>
+                <spark.version>2.2.0.cloudera2</spark.version>
```
OlegMazurov: Doesn't match spark.major.version

pom.xml
```diff
@@ -978,8 +997,7 @@
                 <spark.major.version>spark2.2</spark.major.version>
                 <zookeeper.version>3.4.5-cdh5.14.0</zookeeper.version>
                 <spark.version>2.2.0.cloudera1</spark.version>
```
OlegMazurov: If all cdh platforms are upgraded to 2.2.0.cloudera2, why leave this one at 2.2.0.cloudera1?
If 2.1.0.cloudera1, 2.2.0.cloudera1 are gone, hbase_sql/spark2.1.0.cloudera1, hbase_sql/spark2.2.0.cloudera1 should be removed too.

jleach4:
Fixed in a pull request that perry will push…

John

jleach4:
Comments cleaned up, nice catch.  I am holding off on the cleanup work.  Unclear whether we will have to support different spark versions in the future on cloudera.

PR #1514, opened 2018-03-02
SPLICE-2067: Fix for Client cannot authenticate via:[TOKEN, KERBEROS](2.6)

stpabhi:


stpabhi:
@OlegMazurov No more 2.6?

OlegMazurov:
@stpabhi  No. My understanding is that external customers will migrate 2.5 -> 2.7, with 2.6 becoming an internal release

PR #1522, opened 2018-03-03
SPLICE-2095 pad default values with extra space if it is shorter than…(2.5)

yxia92:
… the column size of fixed-length char type

Please see comment in [SPLICE-2095](https://splice.atlassian.net/browse/SPLICE-2095) for problem and solution description

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnDefinitionNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnDefinitionNode.java
@@ -31,37 +31,28 @@
 
 package com.splicemachine.db.impl.sql.compile;
 
-import com.splicemachine.db.iapi.services.loader.ClassInspector;
-
-import com.splicemachine.db.iapi.services.sanity.SanityManager;
-import com.splicemachine.db.iapi.services.io.StoredFormatIds;
-import com.splicemachine.db.iapi.reference.Limits;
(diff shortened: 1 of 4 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnDefinitionNode.java
```diff
@@ -669,6 +660,15 @@ void validateDefault(DataDictionary dd, TableDescriptor td)
...
 
+			// for fixed-length char type, if default value cannot fix in the column, for example,
+			// default of 'xxxxx' cannot fix in the type of CHAR(4), for such cases, we should error out.
+			// If default value after trimming the trailing space can fit in CHAR(4), it can still go
+			// through
+			if (columnTypeId.getTypeFormatId() == StoredFormatIds.CHAR_TYPE_ID && defaultValue != null) {
+				DataValueDescriptor newDefault = defaultValue.cloneValue(false);
+				newDefault.normalize(getType(), newDefault);
+			}
+
```
jyuanca: Why a clone is needed for normalization

yxia92: The code here is a test to see if the default value actually is compliant to the column definition or not. I wanted to keep the user input default value intact from the test, thus made a clone. Thanks.

PR #1523, opened 2018-03-03
SPLICE-2095 pad default values with extra space if it is shorter than…

yxia92:
… the column size of fixed-length char type


changli6:
+1

PR #1528, opened 2018-03-06
SPLICE-1883 Add rowcount threshold parameter to determine Spark or Control

changli6:
Test steps were in JIRA.


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerFactoryImpl.java
```diff
@@ -57,6 +57,8 @@
 	protected boolean noTimeout = false;
 	protected boolean useStatistics = true;
 	protected int maxMemoryPerTable = Integer.MAX_VALUE;//may need to be long, but this also can be defined in configuration
+	protected long determineSparkRowThreshold = 20000 ; // This value is not effected, the default value is defined in SQLConfiguration
+
```
jyuanca: Why use a hardcoded value here? Should this be tunable too?

changli6: The actual optimizer class is SpliceLevel2OptimizerFactoryImpl which inherited from OptimizerFactoryImpl. This assigned value here was just for initialization but not used. The value in use was gotten from SOLConfiguration. Thanks.

PR #1534, opened 2018-03-06
SPLICE-2096 Fix Cache Miss on all connections

jleach4:
Removes one sql scan on every connection.


jleach4:
@yxia92 This is a negative cache.  If the role descriptor exists, it caches it and if it does not it needs to cache that as well.  All connections to the database check to see if the user is a role.  This creates an extra scan per new connection.

The cache invalidation is like before where if you remove a role, the cache gets invalidated.

yxia92:
Thanks @jleach4 for the explanation.

The fix has a side effect for the following case, as it could give error incorrectly:

splice> connect 'jdbc:splice://localhost:1527/splicedb;user=xiayi;password=xiayi' as xiayi_con;
splice> set role zzz;
ERROR 0P000: Invalid role specification, role does not exist: 'ZZZ'.
splice> set connection splice;
splice> create role zzz;
splice> set connection xiayi_con;
splice> set role zzz;
ERROR 0P000: Invalid role specification, role does not exist: 'ZZZ'.

PR #1553, opened 2018-03-08
DB-6635: Support ldap group as authenticated user (2.5)

snowspeeder:
please also see EE repo PR.


snowspeeder:
returning an authenticated username other than the current username is
unique requirement for ldap,and that has made the refactoring messy for
other authentication methods, I will see if I can keep that return value
authenticateUser as boolean, and then make explicit call to fetch the new
authenticated username that will be set as the current user in the session.
Let me know if you have more thoughts @jleach

PR #1555, opened 2018-03-09
SPLICE-2103: Remove Explain Generation from Hot Path for real-time que…

jleach4:
Remove Regex compilation from the execution of each node in the operation tree for real time queries.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
@@ -188,7 +188,7 @@ public abstract class SpliceBaseOperation implements SpliceOperation, ScopeNamed
         // Right now this is only used by Spark UI, so don't change it
         // unless you want to change that UI.
         CompilerContext.DataSetProcessorType type = this.activation.getLanguageConnectionContext().getDataSetProcessorType();
-        if (type == CompilerContext.DataSetProcessorType.SPARK)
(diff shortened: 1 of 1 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
```diff
@@ -187,10 +187,9 @@ public void setExplainPlan(String plan){
         // No difference. We can change that later if needed.
         // Right now this is only used by Spark UI, so don't change it
         // unless you want to change that UI.
-
         CompilerContext.DataSetProcessorType type = this.activation.getLanguageConnectionContext().getDataSetProcessorType();
-        // JL-TODO Cannot do this in the hot path
-        explainPlan=(plan==null?"":plan.replace("n=","RS=").replace("->","").trim());
+        if (type == CompilerContext.DataSetProcessorType.SPARK)
```
yxia92: I think we should add FORCED_SPARK too.

jleach4: Nice catch Yi...

jleach4: Pushed.

jleach4:
"Hot Path" means that it runs for every operation in the tree for each execution.  The regex portions are not precompiled but have to be compiled each time as well.

wtang16:
Current Spark UI has some issue displaying the explain plan on the nodes in DAG. We'll need to open a new JIRA for that, which is unrelated to this PR.

PR #1558, opened 2018-03-09
SPLICE-2104 Dictionary Scans Should Use HBase Small Scans vs. the net…

jleach4:
…work heavy regular scan.


wtang16:
Do we know the scale of performance improvements?

jleach4:
@wtang16 It only effects dictionary scans.  This was a miss on the implementation.  These scans are small indexed scans and should use a small HBase scanner.  Usually around 30-40% faster based on prior characterizations we did while tuning TPCC.

PR #1559, opened 2018-03-10
SPLICE-2096 Remove Role Scan Hit

jleach4:
Fixing a scan per connection.  This should help IT's since they create a lot of connections.


wtang16:
Ship to 2.7/master? It can benefit as we will have more roles/privileges in the system as customers get more sophisticated.

PR #1560, opened 2018-03-10
SPLICE-2105 Throw exception if you attempt to delete or update a tabl…

jleach4:
…e without a primary key

This should give a pointer to people who try to update/delete from a table without a primary key.


wtang16:
Delete and update without PK are allowed in Splice. Are we selectively restricting CRUD operations if they are performed using Spark Adapter now? Granted that requests coming through spark adapters are usually from application developers/users. But may be we should still hold database user accountable for database modifications according DB roles and privileges? 

Maybe I didn't understand the intention or context of SPLICE-2105.

jleach4:
@wtang16 To update records using data from an underlying data frame, there would need to be a way to lookup the appropriate record for delete/update.

This fix makes sure there is a clear error for the developer.

wtang16:
Thanks @jleach4. So this is for practical reasons from the Spark Adapter code path. But are we disallowing developers to do range deletes and updates?

wtang16:
Got it. Thanks @jleach4

mzweben:
Glad this is done. Think you can close https://splicemachine.atlassian.net/browse/DB-6149 now as well .

wtang16:
Ship to 2.7/master as well?

jleach4:
@wtang16 My process right now is get the first commit in and then I will do the PR's for the other branches.  It should go in all branches.

jleach4:
@MurrayBrown Since this is scala, you **are** very rusty.  :)

PR #1567, opened 2018-03-12
SPLICE-137 allow default roles to be set automatically in a session

yxia92:
This is a port of SPLICE-137-2.5-1 to master. Some minor changes are made to resolve conflicts:
1. In the function insertIndexRowListImpl(), use batchInsert() to insert index rows instead of individual rows one by one.
2. In DataDictionaryCache(), replace canUseCache() with canReadCache() or canWriteCache() based on whether it is a read or write operation.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementColumnPermission.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementColumnPermission.java
@@ -31,16 +31,19 @@
 
 package com.splicemachine.db.iapi.sql.dictionary;
 
-import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.catalog.UUID;
-import com.splicemachine.db.iapi.sql.conn.Authorizer;
+import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.reference.SQLState;
-import com.splicemachine.db.iapi.services.io.FormatableBitSet;
-import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
-import com.splicemachine.db.iapi.sql.Activation;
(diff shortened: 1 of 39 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementColumnPermission.java
```diff
@@ -181,8 +184,9 @@ public void check( LanguageConnectionContext lcc,
 			if (rd == null) {
 				// we have lost the right to set this role, so we can't
 				// make use of any permission granted to it or its ancestors.
-				lcc.setCurrentRole(activation, null);
+				rolesToRemove.add(role);
```
wtang16: Combining the two "if (rd == null)" blocks into one?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -128,6 +129,7 @@ public void onRemoval(RemovalNotification<Object, Dependent> removalNotification
         aliasDescriptorCache = new ManagedCache<>(CacheBuilder.newBuilder().recordStats().maximumSize(1024).build());
         roleCache = new ManagedCache<>(CacheBuilder.newBuilder().recordStats().maximumSize(100).build());
         permissionsCache=new ManagedCache<>(CacheBuilder.newBuilder().recordStats().maximumSize(permissionsCacheSize).build());
+        defaultRoleCache = new ManagedCache<>(CacheBuilder.newBuilder().recordStats().maximumSize(1024).build());
         this.dd = dd;
```
wtang16: Maybe not in the scope of this PR. But can we have the sizes of various (dictionary) caches configurable?

PR #1569, opened 2018-03-13
SPLICE-2107 SnappyPipelineCompressor generates too much garbage

OlegMazurov:
Removed unused methods from the PipelineCompressor interface and its implementations.
Reimplemented SnappyPipelineCompressor with org.xerial.snappy.Snappy already used by Splice in db-drda.
Mean TPMC for TPC-C went up from 14675.4 to 15999 (practically the limit for the 200 rate used in the test). GC looks normal, SnappyPipelineCompressor is off the radar screen.


OlegMazurov:
2.5/2.7 PRs will be created later once this one is rigorously tested.
Another JIRA on org.xerial.snappy.Snappy configuration will be filed (it currently extracts the native library to /tmp, which should be avoided).

PR #1570, opened 2018-03-13
DB-6778 Fix null Dereferences

changli6:


db-engine/src/main/java/com/splicemachine/db/iapi/types/SqlXmlUtil.java
```diff
@@ -684,7 +684,7 @@ protected List evalXQExpression(XMLDataValue xmlContext,
          * XML_DOC_ANY (which means we can store it in a Derby
          * XML column).
          */
-        if ((itemRefs.size() == 1) && (itemRefs.get(0) instanceof Document))
```
dgomezferro: I think it'd be better to raise an exception in the default: case where we set itemRefs to null. With this change we silently continue processing after not knowing how to handle some XPathResult type.

changli6: Committed new one that thrown an exception for unknown XPath result type. Thanks.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SqlXmlUtil.java
```diff
@@ -670,12 +670,11 @@ protected List evalXQExpression(XMLDataValue xmlContext,
...
             default:
+                String msg = "Don't know how to handle XPath result type " + result.getResultType();
                 if (SanityManager.DEBUG) {
-                    SanityManager.THROWASSERT(
-                            "Don't know how to handle XPath result type " +
-                            result.getResultType());
+                    SanityManager.THROWASSERT(msg);
                 }
-                itemRefs = null;
+                throw new Exception(msg);
```
dgomezferro: Could you raise an IllegalArgumentException or UnsupportedOperationException, rather than a generic Exception ?

changli6: Right, throw an UnsupportedOperationException is better.

PR #1574, opened 2018-03-13
DB-6745: invalidate HBase connection cache on retry

jyuanca:


hbase_storage/hbase1.2.0/src/main/java/com/splicemachine/storage/Hbase10PartitionCache.java
```diff
@@ -40,6 +40,7 @@ public Hbase10PartitionCache(){ }
     public void invalidate(TableName tableName) throws IOException{
         partitionCache.invalidate(tableName);
         ((HConnection)HBaseConnectionFactory.getInstance(config).getConnection()).clearRegionCache(tableName);
+        ((HConnection)HBaseConnectionFactory.getInstance(config).getNoRetryConnection()).clearRegionCache(tableName);
```
dgomezferro: This has to be replicated in the Hbase10PartitionCache.java for the other platforms

jyuanca:
Pushed another commits with changes to other platforms as well.

PR #1580, opened 2018-03-15
DB-6431: Fix imported rows count and bad records count for IMPORT_DATA and HFILE_BULK_LOAD

wb14123:
This is the first patch to fix one of the bugs in DB-6431.

`AbstractPipelineWriter` already write the rows count to `operationContext` based on `writeStats`. But there are some problems:

1. The `writeStats` returned from `BulkWriteAction` is not correct.
2. `SparkOperationContext` record the written rows to `pipelineRowsWritten`, which seems to be not used in anywhere.
3. Because of 2, `AbstractPipelineWriter` didn't write the row count into `rowsWritten`. Thus some pipeline writer re-add these counters. And when they add these counters, the data is still in buffer which may fail in later writes.

This patch fix these issues.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
@@ -62,7 +62,6 @@ public class SparkOperationContext<Op extends SpliceOperation> implements Operat
     public LongAccumulator retryAttempts;
     public LongAccumulator regionTooBusyExceptions;
 
-    public LongAccumulator pipelineRowsWritten;
     public LongAccumulator thrownErrorsRows;
     public LongAccumulator retriedRows;
     public LongAccumulator partialRows;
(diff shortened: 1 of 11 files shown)


pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
```diff
@@ -322,7 +322,9 @@ else if(ctx.attemptCount>100 && ctx.attemptCount%50==0){
                                     ctx.refreshCache=ctx.refreshCache || isFailure;
                                     ctx.sleep=true; //always sleep
                                 }
-                                writtenCounter.add(currentBulkWrite.getSize() - writes.size());
+                                writtenCounter.add(currentBulkWrite.getSize() -
+                                        bulkWriteResult.getNotRunRows().size() -
+                                        bulkWriteResult.getFailedRows().size());
                                 partialRetriedRows.add(writes.size());
```
jyuanca: Good catch!

wb14123:
I've merged PR #1632 to here and add another commit to fix HFILE_BULK_LOAD with sampling. Please review the appended two commits. Thanks!

wb14123:
I've add some more changes: check `SparkOperationContext.op` is null before reset the accumulators and rewrite the last commit's description

splice_machine/src/test/test-data/rows_count.csv
```diff
@@ -0,0 +1,3 @@
+"abc","efg"
+"abc","def"
+"abd",null
```
jyuanca: can this small dataset reproduce the error count discrepancies for HFile import with sampling?

wb14123: Yes. With sampling the null row will be count twice.

wb14123:
@dgomezferro Currently `BadRecordsRecorder` only records the number of bad records so I don't see an easy way to reset the bad record files. May be we can fix this in a future issue?

jyuanca:
@dgomezferro Do you have any thoughts about Bin's response?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
```diff
@@ -221,6 +220,34 @@ public void prepare(){
     @Override
     public void reset(){
 //        impl.resetContextManager();
+        BadRecordsRecorder oldRecorder = this.badRecordsAccumulator.value();
+        BadRecordsRecorder badRecordsRecorder = new BadRecordsRecorder(
+                oldRecorder.getStatusDirectory(),
+                importFileName,
+                badRecordThreshold);
+        this.badRecordsAccumulator=SpliceSpark.getContext().accumulable(badRecordsRecorder,
```
dgomezferro: This should only be initialized if permissive is true

PR #1587, opened 2018-03-16
DB-6635: Support ldap group as authenticated user (2.5)

snowspeeder:
Key changes: fetch the authenticated group user name from ldap, and save it as a groupser. If the groupuser is same as as the database owner, permissions checks are passed.
Tested manually with the ldap server that with group user 'cn=splice' in the ldap with user, details in jira. (also see EE change)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementTablePermission.java
```diff
@@ -220,6 +220,10 @@ protected boolean hasPermissionOnTable(LanguageConnectionContext lcc,
 							Authorizer.PUBLIC_AUTHORIZATION_ID,
 							dbo);
 				}
+				if (rd == null && lcc.getCurrentGroupUser(activation) != null) {
```
yxia92: The current sequence of permission checks in the function hasPermissionOnTable() is the following:
1. whether PUBLIC has permission on this database object
2. if NONE/UNAUTHORIZED, whether the currentUser has permission on this database object
3 if NONE, whether PUBLIC has schema level permission for the schema this database object belongs to.
4. if NONE/UNAUTHORIZED, whether currentUser has schema level permission for the schema this database object belongs to.
5. If NONE, check the role permissions from PUBLIC/currentUser.

With the addition of CurrentGroupUser, In addition to the check for roles from the currentGroupUser, I think we also need to add check
2.5, If NONE, whether the currentGroupUser has permission on this database object
4.5 If NONE, whether currentGroupUser has schema level permission for the schema this database object belongs to.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementSchemaPermission.java
```diff
@@ -186,6 +186,10 @@ protected boolean hasPermissionOnSchema(LanguageConnectionContext lcc,
 							Authorizer.PUBLIC_AUTHORIZATION_ID,
 							dbo);
 				}
+				if (rd == null && lcc.getCurrentGroupUser(activation) != null) {
```
yxia92: Similar to StatementTablePermission.java, we need to add the following before we check role permissions:
If NONE, whether currentGroupUser has schema level permission for the schema this database object belongs to.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementPermission.java
```diff
@@ -159,6 +159,10 @@ public String getObjectType()
 					Authorizer.PUBLIC_AUTHORIZATION_ID,
 					dbo);
 			}
+			if (rd == null && lcc.getCurrentGroupUser(activation) != null) {
```
yxia92: Similar to StatementSchemaPermission.java, we need to add check 
		if( !isCorrectPermission( perm ) ) { perm = getPermissionDescriptor(lcc.getCurrentGroupUser(activation), dd ); }

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementColumnPermission.java
```diff
@@ -115,7 +115,7 @@ public void check( LanguageConnectionContext lcc,
 
         String currentUserId = lcc.getCurrentUserId(activation);
 
-		FormatableBitSet permittedColumns = null;
+			FormatableBitSet permittedColumns = null;
```
yxia92: When populating the permittedColumns, in addition to the call of addPermittedColumns() for public and currentUserId, I think we should add the call for currentGroupUser too.

snowspeeder:
Thanks Yi! I've added those checks for the groupuser.

PR #1588, opened 2018-03-16
SPLICE-2112 Configure org.xerial.snappy.Snappy to avoid extracting its native library to /tmp

OlegMazurov:
A wrapper, SpliceSnappy, is introduced. It sets "org.xerial.snappy.tempdir" to control Snappy native library extraction and provides robustness in case Snappy is not loaded.
Snappy version is downgraded to 1.0.4.1 to match the one provided in cluster environment.

### DIFF
--- a/db-drda/pom.xml
+++ b/db-drda/pom.xml
@@ -49,11 +49,6 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.xerial.snappy</groupId>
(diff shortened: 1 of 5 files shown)


wtang16:
Why do you need to downgrade the version? What would be the "right" version to use?

OlegMazurov:
At run time we get snappy-java-1.0.4.1.jar on all supported platforms. We should compile and test with that version to avoid potential runtime incompatibilities.

PR #1589, opened 2018-03-17
SPLICE-2108 Can't disable logging statements

OlegMazurov:


OlegMazurov:
Things are messier than initially thought. The database property, the config property and the system property do not interact in a meaningful way. More work is required.

PR #1592, opened 2018-03-19
DB-6670 cache role grant permissions(2.7)

yxia92:


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -130,6 +132,7 @@ public void onRemoval(RemovalNotification<Object, Dependent> removalNotification
         roleCache = new ManagedCache<>(CacheBuilder.newBuilder().recordStats().maximumSize(100).build());
```
wtang16: Do we need to make these cache sizes configurable? If not in this PR, maybe it's better to revisit all our settings on cache.

yxia92: @wtang16 Yes, currently, only 5 cache sizes are configurable, the remaining(8 caches) are not. I've opened SPLICE-2119 to make all of them configurable.

PR #1594, opened 2018-03-19
SPLCIE-2070:Default to HMaster principal/keytab for OlapServer when they are not explicitly set(2.7)

jyuanca:
…hey are not explicitly set


snowspeeder:
@wtang16 approval needed for 2.5/2.7

PR #1595, opened 2018-03-19
SPLCIE-2070:Default to HMaster principal/keytab for OlapServer when they are not explicitly set

jyuanca:


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -390,6 +409,22 @@ private String prepareCommands(String exec, String parameters) {
...
         }
+        // If user does not specify a kerberos keytab or principal, use HBase master's.
+        if (UserGroupInformation.isSecurityEnabled()) {
+            Configuration configuration = HConfiguration.unwrapDelegate();
+            String principal = System.getProperty(PRINCIPAL_KEY);
+            String keytab = System.getProperty(KEYTAB_KEY);
+            if (principal == null || keytab == null) {
+                principal = configuration.get(HBASE_MASTER_PRINCIPAL_KEY);
+                String hostname = InetAddress.getLocalHost().getHostName();
+                SpliceLogUtils.warn(LOG, "Trying to get local hostname. This could be problem for host with multiple interfaces.");
```
snowspeeder: Do we want to add some checks to ensure that the hostnames appear with multiple interfaces, if yes, print this warning (should reduce the warning noise when possible). What happens currently if we proceed with multiple interfaces, errors out or still works?

jyuanca: It will use wrong principal and error out

dgomezferro: We should use NetworkUtils.getHostname() here, it uses hbase.master.hostname if specified rather than relying on InetAddress.getLocalHost().getHostName(), which won't work properly with multiple interfaces

jyuanca:
@dgomezferro We are trying to get master server's host name which is local host name. NetworkUtil.getHostName may return olap server host name if it is configured.

dgomezferro:
@jyuanca The assumption in NetworkUtils.getHostname() is that each process will have the right options passed, please use that as an abstraction instead of having the same code in several places.

Also, in PRs please add new commits on top of what's there, rather than overwriting history, otherwise it's difficult to track changes for reviewers. Thanks.

jyuanca:
@dgomezferro if splice.olapServer.hostname is configured, will NetworkUtil.getHostName return olap server name instead of master's name, in this case local host name?

dgomezferro:
@jyuanca no, hbase.master.hostname takes precedence

jyuanca:
@dgomezferro hbase.master.hostname is usually not set.

PR #1596, opened 2018-03-20
SPLICE-2106: Fixing Transient Kryo Registrator and adding a few serde …

snowspeeder:
…tests


dgomezferro:
Do we have an IT that reproduces the original exception?

PR #1600, opened 2018-03-20
DB-6278: incremental backup for bulk import

jyuanca:


hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
```diff
@@ -56,6 +61,7 @@
     private FileSystem fs;
     Path rootDir;
     private AtomicBoolean preparing;
+    private Collection<StoreFile> storeFiles;
```
dgomezferro: Using this field for the existing storeFiles can cause issues, another concurrent bulkload could overwrite it and make you lose track of some hfiles.

I think you should use the paths in the familyPaths parameter to see which storefiles are new vs old

jyuanca: @dgomezferro familyPaths are paths before loading. After loading, the actual HFile name in the store is different.

Just realized that region.startRegionOperation() obtains a read lock. What we need here is a write lock.

PR #1603, opened 2018-03-21
DB-6330 updated cdh installation guide,change olapserver memory setting (2.5)

liprais:
updated cdh installation guide,change olapserver memory setting


wtang16:
@liprais Don't we need to set the java servlet library now (Install updated Java Servlet library)?

PR #1616, opened 2018-03-23
DB-6831: add checks for groupuser for grant operation

snowspeeder:


snowspeeder:
Added more operations, I hope this covers all, please take a quick look.

PR #1618, opened 2018-03-23
SPLICE-2120 update SYS.SYSROUTINEPERMS when upgrading system procedures (2.5)

yxia92:
also clean obsolete rows in sysroutineperms if any

Please see [SPLICE-2120](https://splice.atlassian.net/browse/SPLICE-2120) for analysis and solution.
Since the fix is in the upgrade logic, no automatic test is added to IT. I've manually tested with the upgrade from 2.5.1810T to 2.5.1812, 2.5.1810 to 2.5.1812.


yxia92:
I pushed a small fix to make the code to grant the list of system procedures defined in BaseDataDictionary.sysUtilProceduresWithPublicAccess to PUBLIC. On branch-2.7 and master, when database starts up, the execution privilege to the system procedure "SYCS_UTIL.SYSCS_MODIFY_PASSWORD" is granted to public by the function DataDictionaryImpl.grantPublicAccessToSystemRoutines(). Due to a bug on branch-2.5, this function fails to insert the corresponding permission row on branch-2.5.

PR #1627, opened 2018-03-23
DB-6735: Fixing HDP2.6.3 build to support OLAP Server properly 

snowspeeder:
Already in branch-2.5, porting to master/2.7


wtang16:
I noticed that the path to the script is assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py. Note "2.5.1" is used instead of "2.5.0", which is what we have for 2.5 release. Not a big deal though.

PR #1632, opened 2018-03-26
DB-6431 Fix imported rows count for HFILE_BULK_LOAD without sampling

wb14123:


wtang16:
Is this PR going to be combined with https://github.com/splicemachine/spliceengine/pull/1580? Also, I'd like this to be backported to 2.5/2.7.

Thanks.

wb14123:
Will combine this into #1580 , so close it

PR #1634, opened 2018-03-26
DB-6391: fix HFile import logging

jyuanca:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkDataSetWriter.java
```diff
@@ -348,7 +348,7 @@ protected void dumpCutPoints(List<Tuple2<Long, byte[][]>> cutPointsList, String
             Long conglomId = t._1;
             Double size = t._2._1;
             if (LOG.isDebugEnabled()) {
-                SpliceLogUtils.debug(LOG, "conglomerate=%d, size=%d", conglomId, size);
+                SpliceLogUtils.debug(LOG, "conglomerate=%d, size=%f", conglomId, size);
```
yxia92: In addition to these changes, we may want to change the print format at line 368 from using %s to using %f also:
```
            if (LOG.isDebugEnabled()) {
                SpliceLogUtils.debug(LOG, "totalSize=%s", totalSize);
            }
```

PR #1643, opened 2018-03-27
SPLICE-1625 Fix transactions for Spark self-inserts and triggers

dgomezferro:


yxia92:
@dgomezferro I think I don't fully understand the problem of SPLICE-1625. From the problem description, we are writing more rows to table. Are these rows visible or just taking more spaces.
Do we need to port the fix to 2.5 and 2.7?

dgomezferro:
Sorry @yxia92 , I focused the explanation on the IT breakage I had to fix. I updated the comment on SPLICE-1625 to explain why making the transaction additive fixes the original issue.

yxia92:
Thanks @dgomezferro  for the explanation.

PR #1654, opened 2018-03-29
DB-6875 properly compute the referencd columns for index lookup opera…

yxia92:
…tion

Please see analysis and solution in [DB-6875](https://splicemachine.atlassian.net/browse/DB-6875)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
@@ -1786,7 +1786,7 @@ public class FromBaseTable extends FromTable {
 			/* No BULK FETCH or requalification, XOR off the columns coming from the heap 
 			 * to get the columns coming from the index.
 			 */
-            indexReferencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,false,true);
+            indexReferencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,false,false);
             heapReferencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,true,false);
             if(heapReferencedCols!=null){
                 indexReferencedCols.xor(heapReferencedCols);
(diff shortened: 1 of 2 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1786,7 +1786,7 @@ else if(isOneRowResultSet()){
 			/* No BULK FETCH or requalification, XOR off the columns coming from the heap 
 			 * to get the columns coming from the index.
 			 */
-            indexReferencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,false,true);
+            indexReferencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,false,false);
```
dgomezferro: It looks like the isIndex parameter is always false, should we remove it?
It was added when storagePosition was added, but it seems that it's wrong to distinguish that case

yxia92: @dgomezferro This function is called in multiple places in FromBaseTable.changeAccessPath(). One place that requires isIndex to be true is when we compute the referenced columns for the scan of the index table:
{code}
        // Compact the RCL for the index scan down to the partial row.
        referencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,false,false,true);
        resultColumns=resultColumns.compactColumns(cursorTargetTable,false);
        resultColumns.setIndexRow(
                baseConglomerateDescriptor.getConglomerateNumber(),
                forUpdate());
{code}
 another place is to compute the referenced columns for a covering index:
{code}
		/* No need to go to the data page if this is a covering index */
		/* Derby-1087: use data page when returning an updatable resultset */
        if(ap.getCoveringIndexScan() && (!cursorTargetTable())){
....
			/* Compact RCL down to the partial row.  We always want a new
			 * RCL and FormatableBitSet because this is a covering index.  (This is 
			 * because we don't want the RID in the partial row returned
			 * by the store.)
			 */
            referencedCols=resultColumns.getReferencedFormatableBitSet(cursorTargetTable,true,false,true);
                        resultColumns=resultColumns.compactColumns(cursorTargetTable,true);

            resultColumns.setIndexRow(
                    baseConglomerateDescriptor.getConglomerateNumber(),
                    forUpdate());

            return this;
        }
{code}
I think it is correct to set isIndex to true for these two calls, as they are not reading from the base table but the index table.

dgomezferro: I see, thanks! Somehow I missed the other call sites in changeAccessPath()

PR #1659, opened 2018-03-30
DB-6431: Fix imported rows count and bad records count for IMPORT_DATA and HFILE_BULK_LOAD (2.7)

wb14123:


wb14123:
@jleach4 Tests added

PR #1660, opened 2018-03-30
DB 6825 using custom ambari service to install splicemachine

liprais:
DB 6825 using custom ambari service to install splicemachine


OlegMazurov:
Are the following files referred to anywhere in the documentation?
Splice_Machine_Installed.jpeg
custom_config_example.jpeg
custom_service_hive.jpeg

wtang16:
@OlegMazurov New model is that we are using the .md files in source tree (per platform) for installation docs. Official docs will be pointing to the github .md files for external users: http://docstest.splicemachine.com/onprem_install_intro.html

OlegMazurov:
@wtang16 I don't see those three files referred to from that page either. 
Why would we refer to custom_service_hive.jpeg anywhere in our docs?
The files look like temporary work and should be removed.

yxia92:
I'm merging the PR for now in the interest of time. @liprais please take a look at Oleg's comment, and confirm if there are unused files and move out when you are on-line. Thanks.

PR #1661, opened 2018-03-30
DB-6833 Fixing Index creation errors which where meta handlers not ha…

jleach4:
…ving resources to perform the scans

Makes sure the max threads for the write pipeline limiter are pointing to the correct configuration and also that 20% are reserved for meta scans.  For standard configuration (200), we should have 40 threads available.

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/PipelineConfiguration.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/PipelineConfiguration.java
@@ -146,8 +146,8 @@ public class PipelineConfiguration implements ConfigurationDefault {
     public static final String MAX_DEPENDENT_WRITES = "splice.client.write.maxDependentWrites";
     public static final int DEFAULT_MAX_DEPENDENT_WRITES = 40000;
 
-    public static final String IPC_THREADS="hbase.regionserver.handler.count";
-    public static final int DEFAULT_IPC_THREADS = 200;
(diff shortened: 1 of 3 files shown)


splice_access_api/src/main/java/com/splicemachine/access/configuration/PipelineConfiguration.java
```diff
@@ -146,8 +146,8 @@
     public static final String MAX_DEPENDENT_WRITES = "splice.client.write.maxDependentWrites";
     public static final int DEFAULT_MAX_DEPENDENT_WRITES = 40000;
 
-    public static final String IPC_THREADS="hbase.regionserver.handler.count";
-    public static final int DEFAULT_IPC_THREADS = 200;
+    public static final String IPC_THREADS="hbase.regionserver.metahandler.count";
+    public static final int DEFAULT_IPC_THREADS = 10;
```
OlegMazurov: CDH 5.12 is back to using hbase.regionserver.handler.count threads for RPCs (default thread pool).
We should not use priority thread pool's configuration (hbase.regionserver.metahandler.count) as basis for our configuration parameters.

wtang16: Adding a couple of related prior discussions on this subject:
https://splice.atlassian.net/browse/SPLICE-1379
https://splice.atlassian.net/browse/SPLICE-1392

PR #1666, opened 2018-04-02
DB-6833 Set Max Dependent and Independent write threads directly (2.5)

jleach4:
Set the configuration manually since there are issues between different HBase versions.  This should leave 50% of the threads open based on our default implementation.


wtang16:
The change looks good and much simpler than the previous PR. Thanks John.

John/Oleg, do you have any usage recommendation (for different types of platforms and different kinds of workloads)? Could you put those into DB-6833?

yxia92:
@jleach4 Could you prepare branch for branch-2.7 and master? Thanks.

PR #1667, opened 2018-04-03
SPLICE-2108 Remove splice.debug.logStatementContext and service level default value for derby.language.logStatementText

wb14123:
Details on https://splice.atlassian.net/browse/SPLICE-2108


wb14123:
@dgomezferro 

The statement log will still be enabled by default, see this code in `GenericLanguageConnectionContext` (line 355):

```
String logStatementProperty=PropertyUtil.getServiceProperty(getTransactionCompile(),"derby.language.logStatementText");
logStatementText=logStatementProperty == null || Boolean.valueOf(logStatementProperty); // log statements by default
```

@OlegMazurov 

Database property `derby.language.logStatementText` is not removed. You can still set it through `SYSCS_SET_DATABASE_PROPERTY` and the value will be persisted after restart. If I don't remove it from service properties, it will not persisted after restart and cannot be set through `hbase-site.xml` (I think this is the reason why we needed `splice.debug.logStatementContext` before).

PR #1671, opened 2018-04-04
SPLICE-2134 Fixing Column Sequencing for views, issue shows up with c…

jleach4:
…olumn level permissions on views

Customers should recreate views to incorporate these changes for permissions on views.


jleach4:
I think we punt on it and ask people to recreate views.  We are swamped.

wtang16:
We'll need to see if we can recreate system views automatically. We don't have to provide an update script for user views though. 

Also, we'll need to ship this to 2.7/master branches.

yxia92:
If I understand correctly, this bug only affects user views, the column description for system views (sys.systablestatistics, sys.syscolumnstatistics) are created correctly. So we just need to drop and rereate user views.

jleach4:
That is correct…

John

PR #1677, opened 2018-04-09
SPLICE-2119 Make dictionary cache size configurable

wb14123:


wb14123:
@jyuanca I've add the comment to JIRA issue and in the source code. I think these properties will be used internal only. Do I need to change anything else?

jyuanca:
@wb14123 You will need to handle token cache after DB-6727 is merged

yxia92:
I think adding new database property does not require any additional logic for upgrade consideration, but it would be good to just test upgrade to confirm.

wb14123:
@jyuanca Added token cache.

PR #1678, opened 2018-04-09
DB-6658: DDL timeout after transaction is rolled back

jyuanca:


splice_machine/src/test/java/com/splicemachine/derby/ddl/DDLWatchRefresherTest.java
```diff
@@ -225,7 +225,7 @@ public void testDDLTimesOut() throws Exception{
         TxnStore supplier = new TestingTxnStore(clock,new TestingTimestampSource(),null,100l);
```
dgomezferro: If you have time you can also remove the TickingClock clock = new IncrementingClock(0);	from these tests and update the comments that no longer make sense

PR #1687, opened 2018-04-10
SPLICE-2110 Fix date format that returned from connection

wb14123:


dgomezferro:
If you are connecting to a server with a different timezone, does this work?

PR #1693, opened 2018-04-11
SPLICE-2139 fix NPE for create index excluding defaults from hfile(2.5)

yxia92:
Please see [SPLICE-2139](https://splice.atlassian.net/browse/SPLICE-2139) for analysis.


hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkLoadKVPairFunction.java
```diff
@@ -35,10 +40,20 @@ public BulkLoadKVPairFunction(long conglomerateId) {
...
+        List<Tuple2<Long, Tuple2<byte[], byte[]>>> outList = new ArrayList<>();
+        while (kvPairIterator.hasNext()) {
+            KVPair kvPair = kvPairIterator.next();
+            if (kvPair == null)
+                continue;
+            Tuple2<byte[], byte[]> kv = new Tuple2(kvPair.getRowKey(), kvPair.getValue());
+            outList.add(new Tuple2(conglomerateId, kv));
+        }
+
+        return outList.iterator();
```
jleach4: Are you meaning to materialize the whole iterator into a list?  Is it not better to wrap the iterator vs. materializing the whole thing?

jleach4: Ok, looked at this.  I do not think you can do it this way.  This would take an iterator for the entire partition and materialize it into a list (expensive).

yxia92: Thanks @jleach4 for the comment. I pushed a commit to implement an iterator instead of materializing all rows into a list.

PR #1697, opened 2018-04-12
DB-6570 Set fs.s3a.impl in SpliceTestYarnPlatform

changli6:
Only need to set this property on standalone and HDP platform.

### DIFF
--- a/platforms/hdp2.5.5/docs/HDP-installation.md
+++ b/platforms/hdp2.5.5/docs/HDP-installation.md
@@ -169,7 +169,6 @@ To edit the HDFS configuration, click `HDFS` in the Ambari *Services* sidebar. T
 
    ````
    dfs.datanode.handler.count=20
-   fs.s3a.impl=com.splicemachine.fs.s3.PrestoS3FileSystem
    ````
 
 3. Save Changes
--- a/platforms/hdp2.5.5/docs/HDP-installation.md
+++ b/platforms/hdp2.5.5/docs/HDP-installation.md
@@ -169,6 +169,7 @@ To edit the HDFS configuration, click `HDFS` in the Ambari *Services* sidebar. T
 
    ````
    dfs.datanode.handler.count=20
+   fs.s3a.impl=com.splicemachine.fs.s3.PrestoS3FileSystem
    ````
 
 3. Save Changes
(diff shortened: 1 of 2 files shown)


platforms/hdp2.5.5/docs/HDP-installation.md
```diff
@@ -169,6 +169,7 @@ To edit the HDFS configuration, click `HDFS` in the Ambari *Services* sidebar. T
 
    ````
    dfs.datanode.handler.count=20
+   fs.s3a.impl=com.splicemachine.fs.s3.PrestoS3FileSystem
```
dgomezferro: I don't think this is the right place for the configuration, since this will affect any other cluster app that might not use Splice Machine.

changli6: Thanks @dgomezferro. This was a workaround solution for HDP, on CDH we won't set it. Need to spot the root cause. Anyone who familiar with configuration are welcome to comment.

wtang16: @dgomezferro This code change is for test. The installation guide doc change could be modified to say it's a workaround. And for 2.7 and master, we are dropping HDP2.5.5 support. HDP2.5.5 for Splice 2.5 release is still using the old uber jar installation approach. New HDP2.6.3 should use RPM+Ambari approach already.

wtang16: @changli6 Is HDP2.6.3 impacted by DB-6570?

wtang16: Here is my recommendation:
1. The change in SpliceTestYarnPlatform.java is applicable for standalone only, which can be merged.
2. The setting "fs.s3a.impl=com.splicemachine.fs.s3.PrestoS3FileSystem" to hdfs-site.xml or core-site.xml may impact other services running on the same cluster that may rely on a different implementation (for example org.apache.hadoop.fs.s3a.S3AFileSystem). Thus, it cannot be simply recommended as a general solution. For this, we can create a new JIRA to track the solution to the root packaging issue. We also need to ensure the solution works on all our supported platforms (HDP, CDH, MapR).

In summary, I am suggesting to merge the code to SpliceTestYarnPlatform.java to make standalone work with S3 import. 

For problems with S3 import on clusters, we are going to add a note in the troubleshooting doc (https://docstest.splicemachine.com/onprem_info_troubleshoot.html) with the following text:

* Importing data from Amazon S3
Currently, we need a custom setting of fs.s3a.impl in hdfs-site.xml. 
   ````
   fs.s3a.impl=com.splicemachine.fs.s3.PrestoS3FileSystem
   ````

   Please note: other non-splice services/applications may be impacted if they rely on a different implementation (e.g., org.apache.hadoop.fs.s3a.S3AFileSystem)".

wtang16: In the interest of time, moderator @yxia92 will help execute on the above recommendation. We need this in the 2.7 release.

PR #1707, opened 2018-04-17
SPLICE-2079 Wrong results in set query with covered index access in o…

msirek:
…ne branch and base table access in the other.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SetOpOperationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SetOpOperationIT.java
@@ -138,88 +138,80 @@ public class SetOpOperationIT extends SpliceUnitTest {
     @Test
     public void testSPLICE2079() throws Exception {
         ResultSet rs;
-/*        rs = methodWatcher.executeQuery(
-                format("select count(*) from \n" +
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SetOpOperationIT.java
```diff
@@ -123,6 +129,104 @@ public void testExcept() throws Exception {
...
+                format("select count(*) from \n" +
+                        "(select col1, col2 from foo4 --splice-properties index=foo4_ix\n" +
+                        "intersect\n" +
+                        "select col1, col2 from foo4 --splice-properties index=null\n" +
+                        ") dtab\n",useSpark));
+        Assert.assertTrue("One row expected.",rs.next());
+        Assert.assertEquals("SPLICE-2079 test 1 failed.", 3, rs.getInt(1));*/
+
+        rs = methodWatcher.executeQuery(
+                "select count(*) from \n" +
```
yxia92: This test can test both Spark and control path. By specifying another hint useSpark=%s in the query, and pass useSpark as a parameter like the example that you've commented out right above, you can cover both spark and control path. In fact, this ticket has wrong result only for control path, so it is better to explicitly test both paths and make sure both work fine.

yxia92:
Since it is a wrong result, it is better to port the fix to branch-2.5 and branch-2.7.

PR #1716, opened 2018-04-18
DB-6940 Text to Numeric Conversion in Splice(2.5)

jleach4:
Text to Numeric Conversion in Splice...

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/StringTypesToNumericTypeConversion.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/StringTypesToNumericTypeConversion.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright (c) 2012 - 2017 Splice Machine, Inc.
+ *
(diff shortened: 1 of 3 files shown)


wtang16:
Please prepare for 2.7/master shipment too.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/StringTypesToNumericTypeConversion.java
```diff
@@ -0,0 +1,209 @@
+/*
+ * Copyright (c) 2012 - 2017 Splice Machine, Inc.
```
snowspeeder: 2018
btw, this IT file does not end with IT.java like others, will unit test run pick this?

yxia92: I'm preparing PR for master and branch-2.7, I can add the suffix of IT to the file to be consistent.

PR #1717, opened 2018-04-19
SPLICE-1625 Fix transactions for Spark self-inserts and triggers (2.5)

dgomezferro:


dgomezferro:
This has already been merged into master on the previous sprint.

PR #1719, opened 2018-04-19
DB-6829 Add parameter to configure Olap Server yarn queue

dgomezferro:
Defaults to splice.spark.yarn.queue, or default if that's not set


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -130,6 +130,10 @@ public void run() {
             } else {
                 this.appStagingBaseDir = FileSystem.get(conf).getHomeDirectory();
             }
+            String yarnQueue = System.getProperty("splice.olapServer.yarn.queue");
```
wtang16: Is "splice.olapServer.yarn.queue" initialized anywhere (put in the property)? Or we are recommending to put olapServer to the default splice.spark.yarn.queue?

PR #1723, opened 2018-04-20
DB-6727 Coprocessor-level ACL checks (2.5)

dgomezferro:
I think it's best to use this link for reviewing https://github.com/splicemachine/spliceengine/pull/1723/commits/9a455f5291226d7564e731b1fb4593bd74c848ad

The other commit is purely about applying the same changes to all platforms and removing some that are not used anymore


dgomezferro:
http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/931/default/

Jenkins build was successful

PR #1728, opened 2018-04-20
Splice 221 - improvements to sqlshell

MurrayBrown:
SPLICE-221: add -P option to prompt for user password securely
SPLICE-2074: integrate disparate sqlshell implementations in a single common location
    combine mapr, hdp and cdh into a shared script
     support local mode via a fallback if the token is not replaced
SPLICE-2146: add kerberos options, cleanup help messaging
DB-6756: add check for jdk1.8
DB-6757: add an option to add ssl=basic as an optional security flag
    Handle splice lib directory for both installed and cloud tarball
DB-6899: add quiet mode for scripting
    add width, output file and URL as options to improve some scripting


wtang16:
@MurrayBrown can you prepare branch-2.7 shipment as well? Thanks.

MurrayBrown:
I'm hoping to cherry-pick one of these two branch commits (probably 2.5 branch) into a branch for 2.7, as that will be looots easier than re-creating these edits.

PR #1733, opened 2018-04-22
DB-6943: cache region name to avoid converting byte[] to string at co…

jyuanca:
…mpile time


hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/storage/RangedClientPartition.java
```diff
@@ -33,6 +33,7 @@
 public class RangedClientPartition extends ClientPartition implements Comparable<RangedClientPartition>{
     private final HRegionInfo regionInfo;
     private final PartitionServer owningServer;
+    private String regionName;
```
dgomezferro: It might be better to make this field volatile, since several threads could use the same cached partition without synchronization, but it might not be a big deal.

PR #1736, opened 2018-04-22
DB-6876: shut down olap server if splice spark application is killed

jyuanca:


dgomezferro:
@jyuanca is this just a workaround until we fix the root cause?

wtang16:
Reviews on 2.5 branch got 2 approvals. Please merge.

PR #1738, opened 2018-04-24
SPLICE-2149 Make Sort Operations and Sorting in Window Functions Use …

jleach4:
…Tungsten Encoding vs. ValueRowSerializer

Steps to reproduce experiment...

Load TPCH 1 LineItem table (6 Million Rows) on your laptop build of Splice Machine.

Run

select top 10 * from TPCH.LINEITEM order by L_PARTKEY, L_SUPPKEY;
With tungsten sort 69.693 Seconds.

Without tungsten sort ran 5 minutes and then crashed OOM.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
@@ -23,14 +23,7 @@ import com.splicemachine.derby.iapi.sql.execute.SpliceOperation;
 import com.splicemachine.derby.impl.sql.execute.operations.MultiProbeTableScanOperation;
 import com.splicemachine.derby.impl.sql.execute.operations.window.WindowContext;
 import com.splicemachine.derby.stream.control.output.ControlExportDataSetWriter;
-import com.splicemachine.derby.stream.function.KeyerFunction;
-import com.splicemachine.derby.stream.function.MergeWindowFunction;
-import com.splicemachine.derby.stream.function.SpliceFlatMapFunction;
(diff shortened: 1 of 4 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
```diff
@@ -114,6 +107,23 @@ public int partitions() {
...
+
+        //operationContext.pushScopeForOp(OperationContext.Scope.SHUFFLE);
+        PairDataSet sortedByKey=pair.sortByKey(new RowComparator(descColumns,nullsOrderedLow),
+                OperationContext.Scope.SORT.displayName(), operationContext);
+        //operationContext.popScope();
+
+        //operationContext.pushScopeForOp(OperationContext.Scope.READ_SORTED);
+        return sortedByKey.values(OperationContext.Scope.READ_SORTED.displayName());
+        //operationContext.popScope();
+    }
```
jyuanca: Remove code that were commented out.

jleach4: I left that in deliberately.  It is not clear to me why that is commented out, etc.

wtang16:
@jleach4 please set the target to master branch for this PR. @OlegMazurov and @MurrayBrown can validate the performance impact on colo and service using TPCH. After that we can backport it to 2.5 and 2.7. This way the risk to our existing 2.5 customers can be reduced.

Additionally do you know if there are existing ITs covering this PR?

Thanks.

PR #1753, opened 2018-04-27
DB-6966 Add required configs in HDP2.6.3 install doc

changli6:


jyuanca:
@changli6 Does hdp2.5.5 doc need to be updated?

changli6:
@jyuanca No, it had the configs in "Update these other configuration values" part.

changli6:
@dgomezferro Removed splice machine dir from yarn.application.classpath, set them by -Dsplice.spark.executor.extraClassPath in hbase-env .

dgomezferro:
@changli6 Sorry, I don't see where we are setting them on hbase-env , is it in a different PR ?

changli6:
@dgomezferro 
For HDP 2.5.5, it had been set in splice.spark.yarn.jars part. I didn't change it.

> 3. Under `Advanced hbase-env`, set the value of `hbase-env template` to the following:
> ...
> If you're using version 2.2
> ...
>  export HBASE_MASTER_OPTS=... -Dsplice.spark.yarn.jars=/opt/splice/default/lib/*

For HDP 2.6.3, it was set automatically by getMasterDashDProperties(self) in "service_advisor.py" . The splice.spark.yarn.jars didn't have "/var/lib/splicemachine/*" because we excluded Spark from uber jar, it was different with HDP 2.5.5 .

> splice.spark.executor.extraClassPath= ... /var/lib/splicemachine/*
> splice.spark.yarn.jars=/usr/hdp/2.6.3.0-235/spark2/jars/*

dgomezferro:
@changli6 so the only missing libraries to fix DB-6966 were the hadoop libraries, correct?

changli6:
@dgomezferro For HDP 2.6.3, we missed some hadoop related settings in yarn configurations, so I added this reminding.  

@jleach4 I agree with no-manually configurations. The current configuration method is "replace", I think "append" is better to avoid contamination.

wtang16:
@changli6 @dgomezferro @jleach4 Can we reach a conclusion on this?

changli6:
My opinion is we should not overwrite the "yarn.application.classpath" configurations by script,  since it may break other applications. 
I prefer to check classpath settings, if the demand directories were not in, then append them. I'm working on this way.

PR #1770, opened 2018-04-30
DB-6979 Automate Install Process: easier than all the possible issues…

jleach4:
… manually editing config files


MurrayBrown:
It seems like, with this PR, there are about 8 jpegs linked from the .md document, but 16 jpegs in the directory. Perhaps cleanup the extra 8 files now?

PR #1771, opened 2018-04-30
DB-6983: fix rpm ownership

MurrayBrown:
this should address the reported issue with rpm conflicts between splicemachine and hdp ambari-service rpms.


MurrayBrown:
oh doh, wrong base branch

PR #1772, opened 2018-04-30
DB-6983: do not claim parent directories in our rpm (2.5)

MurrayBrown:


wtang16:
Please ship to 2.7/master as well. Thanks.

PR #1773, opened 2018-04-30
DB-6984 Remove the close of the file system

jleach4:
LocalFileSystem can support close but a HDFS File System will fail after closed.


wtang16:
We'll need to ship to 2.7/master as well.

PR #1776, opened 2018-05-01
SPLICE-2161 Retry Meta Scanner Timeout Exceptions on Busy Clusters

jleach4:
This is a simple retry and log while loop.

### DIFF
--- a/hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/storage/ClientPartition.java
+++ b/hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/storage/ClientPartition.java
@@ -286,9 +286,14 @@ public class ClientPartition extends SkeletonHBaseClientPartition{
     private List<HRegionLocation> getAllRegionLocations(boolean refresh) throws IOException {
         if (refresh)
            ((HConnection) connection).clearRegionCache(tableName);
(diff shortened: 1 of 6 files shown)


hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/storage/ClientPartition.java
```diff
@@ -286,9 +286,14 @@ public Table unwrapDelegate(){
...
            ((HConnection) connection).clearRegionCache(tableName);
-        try(RegionLocator regionLocator=connection.getRegionLocator(tableName)){
-            return regionLocator.getAllRegionLocations();
+        while (true) {
+            try (RegionLocator regionLocator = connection.getRegionLocator(tableName)) {
+                return regionLocator.getAllRegionLocations();
+            } catch (Exception e) {
+                SpliceLogUtils.error(LOG,String.format("getAllRegionLocations for tablename=%s failing",tableName),e);
+            }
         }
```
jyuanca: why not wait a while and retry?

dgomezferro: In addition to the wait, I would suggest having a maximum number of retries. Having an unconditional loop here seems bound to bite us in the ass at some point.

PR #1781, opened 2018-05-02
DB-6889: support multiple groups for ldap user

snowspeeder:
Please see Db-6889 for details

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
@@ -118,14 +118,14 @@ public class SpliceDatabase extends BasicDatabase{
     }
 
     @Override
-    public LanguageConnectionContext setupConnection(ContextManager cm,String user, String groupuser, String drdaID,String dbname,
+    public LanguageConnectionContext setupConnection(ContextManager cm,String user, List<String> groupuserlist, String drdaID,String dbname,
                                                      CompilerContext.DataSetProcessorType dspt,
                                                      boolean skipStats,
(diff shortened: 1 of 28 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
```diff
@@ -153,15 +153,15 @@ public LanguageConnectionContext setupConnection(ContextManager cm,String user,
...
+    public LanguageConnectionContext generateLanguageConnectionContext(TxnView txn,ContextManager cm,String user, List<String> groupuserlist, String drdaID,String dbname,
                                                                        CompilerContext.DataSetProcessorType type,
                                                                        boolean skipStats,
                                                                        double defaultSelectivityFactor,
                                                                        String ipAddress) throws StandardException{
         TransactionController tc=((SpliceAccessManager)af).marshallTransaction(cm,txn);
         cm.setLocaleFinder(this);
         pushDbContext(cm);
-        LanguageConnectionContext lctx=lcf.newLanguageConnectionContext(cm,tc,lf,this,user,groupuser,drdaID,dbname,type,skipStats, defaultSelectivityFactor, ipAddress);
+        LanguageConnectionContext lctx=lcf.newLanguageConnectionContext(cm,tc,lf,this,user,groupuserlist,drdaID,dbname,type,skipStats, defaultSelectivityFactor, ipAddress);
```
yxia92: It seems that we allow groupuserList to be null, if so, I think passing null here is better than passing the same username as groupuserList. this can avoid duplicate permission check in the StatementSchema/Table/ColumnPermission modules (one for user, and one for grouper).

PR #1797, opened 2018-05-05
DB-6996 Close table on PipingCallBuffer

dgomezferro:
Add debugging parameters


pipeline_api/src/main/java/com/splicemachine/pipeline/callbuffer/PipingCallBuffer.java
```diff
@@ -308,6 +308,13 @@ public void close() throws Exception {
         }
         rebuildIfNecessary();
 
+        // Table
+        try {
+            table.close();
+        } catch (Exception e) {
+            LOG.warn("Exception while closing table", e);
```
changli6: Do we need to handle this exception or throw it out ?

PR #1807, opened 2018-05-07
SPLICE-2108 Can't disable logging statements redux

OlegMazurov:
Process startParameters always, not just when a new database is created.


yxia92:
Please provide PR for 2.5 and 2.7 also.

PR #1812, opened 2018-05-09
DB-6217 add support for grant all column permissions

yxia92:
Please see [DB-6217](https://splicemachine.atlassian.net/browse/DB-6217) for solution description.
IT is in EE branch.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
@@ -1766,11 +1766,11 @@ public interface DataDictionary{
      * @param perm    the permission to add/remove
      * @param grantee the name of the person to be granted
      * @param tc      the transaction to use
-     * @return True means revoke has removed a privilege from system
-     * table and hence the caller of this method should send invalidation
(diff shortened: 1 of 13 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
```diff
@@ -1766,11 +1766,11 @@ void addDescriptorArray(TupleDescriptor[] tuple,
...
      * @param tc      the transaction to use
-     * @return True means revoke has removed a privilege from system
-     * table and hence the caller of this method should send invalidation
-     * actions to PermssionDescriptor's dependents.
+     * @return 3 values: 0: no change, 1: add permission; -1: remove permission
+     *  hence the caller of this method know whether to send invalidation
+     * actions to PermssionDescriptor's dependents and dictionary cache.
      */
-    boolean addRemovePermissionsDescriptor(boolean add,
+    int addRemovePermissionsDescriptor(boolean add,
```
dgomezferro: I think it'd be better to return an enum here

yxia92:
Thanks @dgomezferro I pushed a commit to address your comment about making return type as enum.

PR #1834, opened 2018-05-15
DB-6997 fix fixed Char type column value for VTI

yxia92:
Please see comment in [DB-6997](https://splicemachine.atlassian.net/browse/DB-6997) for problem and solution description

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
@@ -20,10 +20,12 @@ import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.services.io.ArrayUtil;
 import com.splicemachine.db.iapi.services.io.StoredFormatIds;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
+import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.types.DataValueDescriptor;
 import com.splicemachine.db.iapi.types.DateTimeDataValue;
 import com.splicemachine.db.shared.common.reference.SQLState;
 import com.splicemachine.derby.iapi.sql.execute.SpliceOperation;
+import com.splicemachine.derby.impl.sql.execute.operations.VTIOperation;
 import com.splicemachine.derby.stream.iapi.OperationContext;
 import com.splicemachine.derby.stream.output.WriteReadUtils;
 import com.splicemachine.derby.stream.utils.BooleanList;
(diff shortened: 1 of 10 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
```diff
@@ -186,6 +195,14 @@ else if (values.size() < returnRow.nColumns()) {
...
                             dvd.setValue(SpliceDateFunctions.TO_TIMESTAMP(value, timestampFormat),calendar);
                         break;
+                    case StoredFormatIds.SQL_CHAR_ID:
+                    case StoredFormatIds.SQL_VARCHAR_ID:
+                        dvd.setValue(value);
+                        //normalize the char type
+                        if(dataTypeDescriptors != null && !dvd.isNull()){
+                            dvd.normalize(dataTypeDescriptors[i-1], dvd);
+                        }
+                        break;
```
msirek: If we want to catch truncation errors, perhaps other types like SQL_LONGVARCHAR_ID and SQL_CLOB_ID should be handled here as well.

Would it be worthwhile to build a NormalizeResultSetNode into the tree?  It seems like that would automatically handle all necessary normalizations for all data types.  It may add some processing overhead though.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/VTIOperationIT.java
```diff
@@ -180,18 +181,91 @@ public void testFileVTIExpectError() throws Exception {
...
+                    "-jzhang- | -How The West Won Texas-  | 34  |08-23X-2015 |2015-08-22 08:12:08.0 |11:08:08 |\n" +
+                    "-sfines- |-Senior Software Engineer- | 27  |08X-27-2015 |2015-08-27 08:08:08.0 |06:08:08 |";
+        assertEquals("\n"+sql+"\n", expected, TestUtils.FormattedResult.ResultFactory.toStringUnsorted(rs));
+        rs.close();
+
+        //test spark path
+        sql = String.format("select '-' || name || '-', '-' || title || '-', age, something, date_hired, clock from new com.splicemachine.derby.vti.SpliceFileVTI('%s','',',') as b" +
+                " (name varchar(10), title varchar(30), age int, something varchar(12), " +
+                "date_hired timestamp, clock time) --splice-properties useSpark=false\n" +
+                " where age < 40 and date_hired > TIMESTAMP('2015-08-21', '08:09:08') order" +
```
msirek: Comment says "test spark path" but it sets "useSpark=false".

yxia92: Thanks. I'll fix it.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
yxia92: I added the check for SQL_CLOB_ID.
After looking at SQLLongvarchar's case, I think we don't need to do normalization here. Different from the generic normalization operation which tries to convert from any type to any type, here we are just trying to convert String to the defined type. For SQLLongvarchar, it does not specify any column length, so putting string into the dvd of SQLLongvarchar type is not an issue.

PR #1843, opened 2018-05-15
SPLICE-2166: close result set if the operation is killed

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/OperationManagerImpl.java
```diff
@@ -75,6 +77,9 @@ public boolean killOperation(UUID uuid, String userId) throws StandardException
         unregisterOperation(uuid);
         op.getOperation().kill();
         op.getThread().interrupt();
+        ResultSet rs=activation.getResultSet();
+        rs.close();
+
```
msirek: Should rs be checked that it's not null before calling close?  I don't know if it's possible to be null, but I see other code that calls getResultSet() check for null, so it may be good to do the same.

jyuanca: Good idea. Thanks @msirek

PR #1852, opened 2018-05-17
DB-7012 fix First_Value ingoreNulls logic, fix First_value/Last_Value…

yxia92:
… with Frame clause for control path

Please see [DB-7012](https://splicemachine.atlassian.net/browse/DB-7012) for analysis.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
@@ -47,7 +47,7 @@ public class FirstLastValueFunction extends SpliceGenericWindowFunction {
     @Override
     protected void calculateOnAdd(WindowChunk chunk, DataValueDescriptor[] dvds) throws StandardException {
         DataValueDescriptor result = chunk.getResult();
-        if (result == null || result.isNull()) {
+        if (result == null || ignoreNulls && result.isNull()) {
             chunk.setResult(dvds[0].cloneValue(false));
         } else if (isLastValue) {   // We keep setting results if we're calc'ing LAST_VALUE. If FIRST_VALUE, stop on first result.
             DataValueDescriptor input = dvds[0];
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
```diff
@@ -60,7 +60,37 @@ protected void calculateOnAdd(WindowChunk chunk, DataValueDescriptor[] dvds) thr
 
     @Override
     protected void calculateOnRemove(WindowChunk chunk, DataValueDescriptor[] dvds) throws StandardException {
-       // nothing to do here we've already set all the results. For LAST_VALUE(), result is just the last item we added.
+       if (isLastValue) {
+           // we need to handle the scenario where the last value is the one that is removing, this could happen
+           // when ignoreNulls is true
+           DataValueDescriptor removedVal = dvds[0];
+           if (ignoreNulls && !removedVal.isNull() && chunk.getResult().equals(removedVal)) {
+               int position = chunk.last;
```
msirek: Is the equals() check a reliable method to determine if this is the removed value?  Could there be a case where chunk.getResult() happens to equal the removed value, but there are multiple rows with the same value, and it's not really the one being removed from the window?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
```diff
@@ -60,7 +60,37 @@ protected void calculateOnAdd(WindowChunk chunk, DataValueDescriptor[] dvds) thr
...
+                   position --;
+               }
+               chunk.setResult(result);
+           }
+       } else {// handle First_Value
+           int position = chunk.first;
+           DataValueDescriptor result = null;
+           while (position < chunk.last) {
+               result = chunk.get(position)[0];
+               if (result !=null && (!ignoreNulls || !result.isNull())) {
```
msirek: The loop to handle Last_Value decrements all the way back to chunk.first, because it uses 'position-1' as the index passed to chunk.get(), but the First_Value loop only iterates up to chunk.last-1, so doesn't include the last value in the window.  Is this by design, or should the while loop really use 'while (position **<=** chunk.last) {'   ?   

It's not clear if the value being removed is currently included in between chunk.first and chunk.last, or if the window has already been updated by the time we get in here, and the removed value is not in this range.  Knowing this might help answer if chunk.last should be included.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
yxia92: Thanks @msirek for the comment. the equals() check does not need to ensure chunk.getResult() to be the same object as the removed value. It just needs to rule out the cases where the chunk.getResult() is not possible to be the same as the removed value to save some processing time. For example, 
       original chunk content:  3, null, null, null
       chunk.getResult() is 3 which is last value of the chunk ignore nulls.
after removing the first value in the chunk, we have:
       removed value: 3
       chunk content: null, null, null
The above logic will be triggered, and the updated chunk.getResult() will be null.

Another example:
       original chunk content:  3, null, 3, null
       chunk.getResult() is also 3, but not the first value in the chunk
after removing the first value in the chunk, we have
       removed value: 3
       chunk content: null, 3, null
The above logic will still be triggered, but the chunk.getResult() will still point to the 3 currently in the chunk after update.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/window/function/FirstLastValueFunction.java
yxia92: The valid values in the chunk are from position first to last-1. 
The value being removed is not included in chunk.first and chunk.last-1. It may be clearer to look from the caller of calculateOnRemove() in SpliceGenericWindowFunction.remove().

PR #1861, opened 2018-05-18
SPLICE-2155 Add TaskId to transaction table (2.5)

dgomezferro:


OlegMazurov:
I had to fix com.splicemachine.pipeline.mem.DirectBulkWriter.java.write() to compile on mem when backported SPLICE-2155 to branch-2.5-1820a. Please make sure it compiles on mem.

OlegMazurov:
This change is good. My comment applies to DB-7035.

PR #1863, opened 2018-05-18
DB-7035 Fixes on ACL layer 

dgomezferro:


OlegMazurov:
com.splicemachine.pipeline.mem.DirectBulkWriter.java.write() needs to be fixed to compile on mem

jyuanca:
Why we need a partitionAdapterCache and can't use partitionCache?

dgomezferro:
@jyuanca if we use partitionCache it would get populated by AdapterPartitions and ClientPartitions, the latter don't know how to proxy the methods through the JDBC connection and they would fail when reused at a later point. It might be possible to make sure we only generate AdapterPartitions, but I found it easier to separate the two.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SConfigurationImpl.java
```diff
@@ -276,7 +277,10 @@ public boolean getAuthenticationTokenDebugConnections() {
     public int getAuthenticationTokenMaxConnections() {
         return authenticationTokenMaxConnections;
     }
-
+    @Override
+    public int getAuthenticationTokenPermissionCacheSize() {
```
yxia92: This configuration parameter does not seem to be used anywhere. Did I miss anything?

dgomezferro:
@yxia92 I forgot to create the -ee PR, thanks for noticing! Please review it here: https://github.com/splicemachine/spliceengine-ee/pull/218

splice_access_api/src/main/java/com/splicemachine/access/configuration/SConfigurationImpl.java
dgomezferro: @yxia92 It's used on the -ee side, I forgot to create the PR, sorry!

hbase_storage/hbase1.1.8/src/main/java/com/splicemachine/storage/Hbase10PartitionCache.java
```diff
@@ -63,5 +64,27 @@ public void configure(SConfiguration configuration){
     public void put(TableName tableName, List<Partition> partitions) throws IOException {
         partitionCache.put(tableName,partitions);
     }
+
+    @Override
+    public void invalidateAdapter(TableName tableName) throws IOException {
+        partitionAdapterCache.invalidate(tableName);
+        ((ClusterConnection)HBaseConnectionFactory.getInstance(config).getConnection()).clearRegionCache(tableName);
```
yxia92: Need to change the use of HConnection in the function invalidate() to ClusterConnection to be consistent with invalidateAdapter. We got into compilation error due to this for the special branch-2.5-1820c.

dgomezferro: @yxia92 I'm away from my computer, could you (or somebody else) fix it, if merging this is urgent? Thanks

yxia92: I can push a fix for the compilation issue.

PR #1864, opened 2018-05-18
SPLICE-2168 Race Condition Causes Significant Meta Refreshes during r…

jleach4:
…egion movements...

This should cutdown on the number of conflicting connection refreshes.

### DIFF
--- a/pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
+++ b/pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
@@ -33,7 +33,10 @@ import com.splicemachine.si.api.txn.WriteConflict;
 import com.splicemachine.utils.SpliceLogUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.log4j.Logger;
+import org.spark_project.guava.cache.Cache;
+import org.spark_project.guava.cache.CacheBuilder;
 import org.spark_project.guava.collect.Lists;
+import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
```diff
@@ -48,6 +51,11 @@
     private static final Logger LOG=Logger.getLogger(BulkWriteAction.class);
     private static final Logger RETRY_LOG=Logger.getLogger(BulkWriteAction.class.getName()+".retries");
     private static final AtomicLong idGen=new AtomicLong(0L);
+    public static final AtomicLong submitTimeCounter=new AtomicLong(0L);
+    public static long submitTime;
+    public static final Cache<byte[],Long> IGNORE_REFRESH = CacheBuilder.newBuilder().
+            concurrencyLevel(64).
+            expireAfterWrite(60,TimeUnit.SECONDS).initialCapacity(128).build();
```
dgomezferro: Is the .expireAfterWrite() needed? It doesn't seem to serve any function

jleach4: You can technically leak if you put things in the cache and never expire them.  It is a static variable.  The expire ejects them.

jyuanca:
Does this address "Not Functional Path" problem?

jleach4:
I think to fix the not functional path problem, I would suggest performing a retry where that exception is thrown.  I do think this is going to make the connection refreshing more sane.

pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
```diff
@@ -465,12 +473,48 @@ private Exception parseIntoException(BulkWriteResult response){
...
+     * |
+     * |                        s---------r (This would be allowed to invalidate cache if it is incorrect)...
+     * |
+     * ----------------------------------------------------------------------------------> t
+     *
+     *
+     * @param writerFactory
+     * @param tableName
+     */
+    public static synchronized void invalidateCache(BulkWriterFactory writerFactory, byte[] tableName, long submitTime) throws IOException {
```
dgomezferro: If we have multiple tables that need refreshing, we would serialize their invalidations because the latter ones would be blocked during the invalidation of the first one. This might not happen to frequently from a SparkExecutor but it might be common when forwarding index writes to several indexes. Could we make sure a thread only blocks if there's already an invalidation going on for the same table?

PR #1870, opened 2018-05-21
SPLICE-2172 Compaction always calls meta once for looking up the curr…

jleach4:
…ent host. Not needed.


dgomezferro:
@jyuanca it's already been approved and merged into 2.5 https://github.com/splicemachine/spliceengine/pull/1847

PR #1872, opened 2018-05-22
DB-6821 Cleanup statement logging and add more stats

wb14123:


db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedResultSet.java
```diff
@@ -616,6 +619,9 @@ public void close() throws SQLException	{
 
 			// we hang on to theResults and messenger
 			// in case more calls come in on this resultSet
+
+			LanguageConnectionContext lcc = getEmbedConnection().getLanguageConnection();
+			lcc.logEndFetching(getSQLText(), NumberofFetchedRows);
```
msirek: lcc is accessed here after the call to markClosed(), but getLanguageConnection() expects the connection to be open : 
	public final LanguageConnectionContext getLanguageConnection() {
		if (SanityManager.DEBUG)
			SanityManager.ASSERT(**!isClosed()** || isAborting(), "connection is closed");

... maybe it would be better to move the logEndFetching call into the finally block, above the call to markClosed() at line 612.

msirek: OK

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
```diff
@@ -1392,4 +1393,15 @@ void setReferencedColumnMap(TableDescriptor td,
 	double getDefaultSelectivityFactor();
 
 	String getClientIPAddress();
+
+	void logStartCompiling(String statement);
+	void logEndCompiling(String statement, long nanoTimeSpent);
+	void logErrorCompiling(String statement, Throwable t, long nanoTimeSpent);
+	void logCommit();
+	void logRollback();
+	void logStartFetching(String statement);
```
dgomezferro: Where is this used?

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -3639,4 +3620,91 @@ public double getDefaultSelectivityFactor() {
...
+                "XID=%s, SessionID=%s, Database=%s, DRDAID=%s, UserID=%s",
+                getTransactionExecute().getActiveStateTxIdString(),
+                getInstanceNumber(),
+                getDbname(),
+                getDrdaID(),
+                getSessionUserId());
+    }
+
+    private String formatLogStmt(String statement) {
+        if (statement == null) {
```
dgomezferro: Can you cache the formatted statement? we are calling this function over and over for logging purposes

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -3639,4 +3620,91 @@ public double getDefaultSelectivityFactor() {
...
+    }
+
+    @Override
+    public void logCommit() {
+        stmtLogger.info(String.format("Committing. %s", getLogHeader()));
+    }
+
+    @Override
+    public void logRollback() {
+        stmtLogger.info(String.format("Rolling back. %s", getLogHeader()));
```
dgomezferro: Please guard all logging with stmtLogger.isInfoEnabled() checks to avoid the formatting cost when logging is disabled.

wb14123: Thanks for the suggestion. I'll fix them in another PR.

PR #1882, opened 2018-05-23
SPLICE-2151 Convert out-of-range timestamps during IMPORT to the max …

msirek:
…or min timestamp.


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -440,6 +440,15 @@ private PreparedStatement prepMinion(LanguageConnectionContext lcc,
...
             cc.setProjectionPruningEnabled(!projectionPruningOptimizationDisabled);
 
+            String convertTimestampsString = PropertyUtil.getCachedDatabaseProperty(lcc.getTransactionCompile(),
+            Property.CONVERT_OUT_OF_RANGE_TIMESTAMPS);
+            // if database property is not set, treat it as false
+            cc.setConvertTimestampsEnabled(false);
+            if (convertTimestampsString != null)
+                cc.setConvertTimestampsEnabled(Boolean.valueOf(convertTimestampsString));
+
+
```
jyuanca: Is it possible to check this property at VTIOperation?

msirek: Yes, I could make the change.  Originally I was going to do something like this, but tried to follow the pattern of collecting cached properties in the CompileContext (cc), as what's done for projection pushing.

msirek: Hi @jyuanca ,  I tried moving the checking of the database property into the VTIOperation constructor as follows, but the property is always found to be null, even when set:

        LanguageConnectionContext lcc = activation.getLanguageConnectionContext();
        String convertTimestampsString = null;
        if (lcc != null) {
            PropertyUtil.getCachedDatabaseProperty(lcc.getTransactionCompile(),
                                                   Property.CONVERT_OUT_OF_RANGE_TIMESTAMPS);
        }
        // if database property is not set, treat it as false
		this.convertTimestamps = convertTimestampsString != null && Boolean.valueOf(convertTimestampsString);

Is the execution context different somehow, so the propery is not seen there?  Or do you see something wrong with the above code?

jyuanca: @msirek How did you set the property? If you call syscs_util.syscs_set_global_database_property, it should be persisted to splice:16, and be readable from anywhere.

msirek: Yes, that's what's in my script:

call syscs_util.syscs_set_global_database_property('derby.database.convertOutOfRangeTimeStamps', 'true');

msirek: @jyuanca  I found the problem.  Code is updated to check property cache in VTIOperation.

PR #1889, opened 2018-05-24
SPLICE-2180 Fix code that uses Spark JobProgressListener.

stpabhi:


jleach4:
@stpabhi Do you have a test case or at least a test scenario?  I would want someone to add something so we do not regress with new versions of spark for example.

stpabhi:
@jleach4 This change is backward compatible with Spark 2.2. I introduced this change so that splice doesn't break when the runtime is Spark 2.3. I can work on a test case.

hbase_sql/src/main/java/com/splicemachine/olap/CompactionJob.java
```diff
@@ -165,15 +166,15 @@ private boolean compactionRunning(List<Integer> jobIds) {
             return false;
```
dgomezferro: This function checks whether the compaction actually started (or completed) execution, with this change it returns true whenever it is scheduled, but before its tasks start running. 

To use the new API we'd need to call getStageInfo() which has the same activeTasks and completedTasks information we were checking with the old API. You can get the stageId from the jobInfo object (there's only one stage for compactions) and then pass it to statusTracker.getStageInfo()

stpabhi: Ok I misunderstood. Thanks for the clarification. Will make necessary changes.

stpabhi:
@dgomezferro Made changes. please review.

stpabhi:
@jyuanca Ok, will do.

stpabhi:
@jyuanca #1997 #1998. Please review and approve.

PR #1893, opened 2018-05-28
DB-6821 Check logger status before log and cache formatted statement

wb14123:
This is a follow up for PR #1872


db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -69,6 +69,7 @@
 import com.splicemachine.db.impl.sql.execute.*;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
+import org.apache.log4j.Priority;
```
OlegMazurov: Not used.

PR #1896, opened 2018-05-29
DB-6913 Fix NullPointerException in aggregate queries with UDFs.

msirek:
Please see comments in [DB-6913](https://splicemachine.atlassian.net/browse/DB-6913?focusedCommentId=54566&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-54566) for solution description.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
@@ -200,8 +200,10 @@ public class SparkOperationContext<Op extends SpliceOperation> implements Operat
             ActivationHolder ah = broadcastedActivation.getActivationHolder();
             activation=ah.getActivation();
 
-            if (ah.needsReinitialization())
+            if (ah.needsReinitialization()) {
+                ah.close();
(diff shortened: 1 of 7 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkOperationContext.java
```diff
@@ -197,7 +197,11 @@ public void readExternal(ObjectInput in)
         if(isOp){
             broadcastedActivation = (BroadcastedActivation)in.readObject();
             op=(Op)broadcastedActivation.getActivationHolder().getOperationsMap().get(in.readInt());
-            activation=broadcastedActivation.getActivationHolder().getActivation();
+            ActivationHolder ah = broadcastedActivation.getActivationHolder();
+            activation=ah.getActivation();
+
+            if (ah.needsReinitialization())
+                ah.reinitialize(null, false);
```
dgomezferro: When we call .reinitialize() on an ActivationHolder we have to call .close(). I haven't inspected the change in depth yet but at first glance it seems we are no closing it. Can you fix that or point me to where we are closing it? Thanks

PR #1898, opened 2018-05-29
SPLICE-1929: implement check_table system procedures

jyuanca:
check out SPLICE-1929 for a detailed description.


wtang16:
@jyuanca we need the feature in 2.7 as well.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -0,0 +1,269 @@
...
+            tableCountFuture = ((SparkPairDataSet) filteredTable).rdd.countAsync();
+            SpliceSpark.popScope();
+        }
+        // count number of rows in the index
+        SpliceSpark.pushScope(String.format("Count index %s.%s", schemaName, indexName));
+        JavaFutureAction<Long> indexCountFuture = ((SparkPairDataSet)index).rdd.countAsync();
+        SpliceSpark.popScope();
+
+        messages.addAll(checkMissingIndexes(filteredTable, index));
+        messages.addAll(checkInvalidIndexes(filteredTable, index));
```
yxia92: I think the call of checkInvalidIndexes() can be moved after getting the rowcount from table and index. As if rowcounts of table and index are the same and checkMissingIndexes() does not return any error, we don't need to call checkInvalidIndexes().

jyuanca: checkInvalidIndexes() always needs to run even if the counts matches and no missing or duplicate index.

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/CheckTableJob.java
```diff
@@ -299,19 +303,63 @@ private ExecRow getIndexExecRow(long conglomerateId) throws StandardException {
...
+        FormatableBitSet defaultValueMap = null;
+        int size = 0;
+        for (int i = 0; i < defaultValueRow.nColumns(); ++i) {
+            if (!defaultValueRow.getColumn(i+1).isNull())
+                size++;
+        }
+
+        int index = 0;
+        if (size > 0) {
+            defaultValueMap = new FormatableBitSet(size);
```
yxia92: I think the size of FormatableBitSet should be the same as the number of columns in defaultValueRow, and each bit that is set represent that the corresponding column's default value may be used to fill the row, however, the defaultMap set here does not seem so.

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/CheckTableJob.java
```diff
@@ -299,19 +303,63 @@ private ExecRow getIndexExecRow(long conglomerateId) throws StandardException {
...
+            for (int i = 0; i < defaultValueRow.nColumns(); ++i) {
+                if (!defaultValueRow.getColumn(i+1).isNull()) {
```
yxia92: We should check if defaultValueRow.getColumn(i+1) is null or not, as the column's value may not be populated.

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/CheckTableJob.java
```diff
@@ -309,19 +309,29 @@ private ExecRow getIndexExecRow(long conglomerateId) throws StandardException {
...
 
+    private FormatableBitSet getDefaultValueMap(ExecRow defaultValueRow) throws StandardException {
+        FormatableBitSet defaultValueMap = null;
+        int size = defaultValueRow.size();
         int index = 0;
         if (size > 0) {
             defaultValueMap = new FormatableBitSet(size);
             for (int i = 0; i < defaultValueRow.nColumns(); ++i) {
-                if (!defaultValueRow.getColumn(i+1).isNull()) {
+                if (defaultValueRow != null && !defaultValueRow.getColumn(i+1).isNull()) {
```
yxia92: I think we should check "defaultValueRow.getColumn(i+1) != null" instead.

PR #1903, opened 2018-05-30
DB-6941 Close NestedLoopJoins and MultiprobScans properly

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
```diff
@@ -61,6 +68,8 @@
     protected boolean isOneRowInnerJoin;
     protected boolean hasMatch;
     protected List<Future<Pair<OperationContext, Iterator<ExecRow>>>> futures;
+    protected BlockingQueue<OperationContext> activeContexts;
```
yxia92: My understanding of activeContexts is that it has a one-to-one mapping to the list of futures. However, I only see that activeContexts get populated in loadBatch() but not in hasNext(), also I only see the context corresponding to the first left row removed from activeContexts in loadBatch(), should we remove the remaining ones as the rows get consumed in hasNext()?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
```diff
@@ -95,6 +112,24 @@ private void initOperationContexts() throws StandardException {
...
+    public void close() {
+        StandardException se = null;
+        for (OperationContext ctx : activeContexts) {
+            try {
+                ctx.getOperation().close();
+            } catch (StandardException e) {
+                se = e;
+                LOG.error("Exception while closing operation", e);
+            }
+        }
```
yxia92: should we close currentOperationContext here too? From the code in loadBatch(), it could be removed from the activeContexts, so it may not be in the activeContexts when close() is called.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
dgomezferro: I think I overcomplicated the activeContexts a bit, I modified it to a list with all cloned contexts and closing all of them at the end.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
dgomezferro: With the new change, closing all contexts should take care of that.

PR #1905, opened 2018-05-30
DB-7073 Improve MultiprobeScan performance

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/TableScanOperation.java
```diff
@@ -358,6 +360,6 @@ public String toString(){
                 .location(location)
                 .partitionByColumns(getPartitionColumnMap())
                 .defaultRow(defaultRow,scanInformation.getDefaultValueMap())
-                .buildDataSet(this);
+                .buildDataSet(this).map(new SetCurrentLocatedRowAndRowKeyFunction<>(operationContext));
```
msirek: For operations which scan all rows in a table, it seems like saving the current row location and rowkey for every row would add some overhead that may affect performance.  What is the benefit of saving this information?

dgomezferro: What SetCurrentLocatedRowAndRowKeyFunction does is setting the current row in the activation, this is needed for some operations that try to access the current row down the stack (project restricts, for instance). It is true that it might add some overhead on simpler cases (select count(*) for instance) but this is the current behaviour, we can address that on other JIRAs. Previously this code was executed in TableScannerIterator.next() which I removed because that can be executed in parallel now when we have a MultiProbeScan

wtang16: @OlegMazurov may also chime in on the potential overhead. The overhead may not be visible in the performance profile.

msirek: OK, thanks.

PR #1913, opened 2018-05-30
Fixed anchor link

rlondner:
Fixed CentOS/Red Hat install link on master branch


rlondner:
You should also probably fix the https://s3.amazonaws.com/splice-releases/2.7.0.1815/standalone/splicemachine-2.7.0.1815.tar.gz link at the bottom of the page, because 1) it doesn't work and 2) it's outdated

PR #1921, opened 2018-06-04
DB-7104 fix size inconsistent for JBitSet in predicates

yxia92:
Please see analysis in [DB-7104](https://splicemachine.atlassian.net/browse/DB-7104)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -51,6 +51,8 @@
 
 import java.util.*;
 
+import java.util.*;
```
changli6: A duplicated line.

yxia92: thanks @changli6 , I removed the duplicate lines.

PR #1924, opened 2018-06-04
splice-2173 Fixing Compaction Locality

jleach4:
Fixes Compaction Locality:  Tested on srv91 with a single executor.  After compaction, a 13 region table had a locality of 1.0 for each region.

### DIFF
--- a/hbase_sql/hbase1.2.0/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
+++ b/hbase_sql/hbase1.2.0/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
@@ -18,60 +18,60 @@ import com.splicemachine.EngineDriver;
 import com.splicemachine.access.HConfiguration;
 import com.splicemachine.access.api.SConfiguration;
 import com.splicemachine.constants.EnvUtils;
-import com.splicemachine.db.catalog.UUID;
-import com.splicemachine.db.iapi.sql.conn.ConnectionUtil;
-import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
-import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
-import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
-import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
-import com.splicemachine.db.iapi.store.access.TransactionController;
-import com.splicemachine.db.impl.jdbc.EmbedConnection;
-import com.splicemachine.derby.impl.store.access.SpliceTransactionManager;
-import com.splicemachine.derby.jdbc.SpliceTransactionResourceImpl;
 import com.splicemachine.derby.stream.compaction.SparkCompactionFunction;
 import com.splicemachine.hbase.SICompactionScanner;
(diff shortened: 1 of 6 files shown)


jleach4:
@stpabhi This might be interesting to you guys as well.

stpabhi:
@jleach4 This is good. :) I worked on something that i call as Locality Favored HBase Balancer. It will move regions to the local node by getting the block information from NameNode. This tool is useful to get back the locality in case of rolling restarts and non local compactions.

wtang16:
@stpabhi please also add your comments/suggestions to SPLICE-2173 if you have any. Will @jleach4's PR remove the need for your script you currently use?

hbase_sql/hbase1.2.0/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -506,4 +500,209 @@ protected InternalScanner createScanner(Store store, List<StoreFileScanner> scan
...
+        try {
+            RegionServerServices rsServices = (RegionServerServices)FieldUtils.readField(((HStore) store).getHRegion(), "rsServices", true);
+            InetSocketAddress[] returnAddresses = (InetSocketAddress[]) MethodUtils.invokeMethod(rsServices,"getFavoredNodesForRegion",store.getRegionInfo().getEncodedName());
+            if ( (returnAddresses == null || returnAddresses.length == 0)
+                    && store.getFileSystem() instanceof HFileSystem
+                    && ((HFileSystem)store.getFileSystem()).getBackingFs() instanceof DistributedFileSystem) {
+                String[] txvr = conf.get("dfs.datanode.address").split(":"); // hack
+                if (txvr.length == 2) {
+                    returnAddresses = new InetSocketAddress[1];
+                    returnAddresses[0] = new InetSocketAddress(hostName, Integer.parseInt(txvr[1]));
```
msirek: If dfs.datanode.address is malformed, like hostname:abc, we'll throw an exception instead of allowing the compaction to complete with a warning.  Is this intended?

jyuanca:
@dgomezferro Passing a RegionServerService will cause coprocessors to be loaded. Coprocessors may want to access things that a dummy RegionServerService doe not have. There are scoping and API compatibility problem that make refactoring code hard.

dgomezferro:
@jyuanca Where does the coprocessor loading happen?

jyuanca:
@dgomezferro In the constructor of HRegion()
if (rsServices != null) {
      this.rsAccounting = this.rsServices.getRegionServerAccounting();
      // don't initialize coprocessors if not running within a regionserver
      // TODO: revisit if coprocessors should load in other cases
      **this.coprocessorHost = new RegionCoprocessorHost(this, rsServices, conf);**
      this.metricsRegionWrapper = new MetricsRegionWrapperImpl(this);
      this.metricsRegion = new MetricsRegion(this.metricsRegionWrapper);

      Map<String, Region> recoveringRegions = rsServices.getRecoveringRegions();
      String encodedName = getRegionInfo().getEncodedName();
      if (recoveringRegions != null && recoveringRegions.containsKey(encodedName)) {
        this.recovering = true;
        recoveringRegions.put(encodedName, this);
      }

dgomezferro:
@jyuanca I see, good point, thanks for the pointer!

PR #1934, opened 2018-06-06
DB-6891 make rpm include sqlshell.ch when packaging

liprais:
1.make sqlshell.cn included in rpm
2.modify sqlshell.cn work under /var/lib/splicemachine


assembly/common/src/main/resources/bin/sqlshell.sh
```diff
@@ -153,6 +153,8 @@ if [[ "$SPLICE_LIB_DIR" == *"##"* ]]; then
    # ends in bin, go up and over to lib
    if [[ "$CURDIR" = *"bin" ]]; then
       SPLICE_LIB_DIR="${CURDIR}/../lib"
+   elif [[ "$CURDIR" = "/var/lib/splicemachine" ]]; then #hdp2.6.3 rpm needed
```
MurrayBrown: The whole point of this section is to handle the *fallback* behavior  if the ##SPLICE_LIB_DIR## is *not* token replaced. the default action for any built artifact should be to replace that in a platform independent way.

I dont think this is the right answer. We should find a way to replace the token for hdp2.6.3 with the indicated directory at build or install time.

wtang16:
@liprais please close this PR and create a new one.

PR #1935, opened 2018-06-07
SPLICE-2190 MPartition calls size on ConcurrentSkipListSet prior to s…

jleach4:
…can (size is not constant time)


mem_storage/src/main/java/com/splicemachine/storage/MPartition.java
```diff
@@ -502,9 +502,6 @@ private void delete(MDelete mDelete,Lock rowLock) throws IOException{
 
     private NavigableSet<DataCell> getAscendingScanSet(DataScan scan){
         NavigableSet<DataCell> dataCells;
-        if(memstore.size()<=0)
```
dgomezferro: Would it make sense to substitute it for isEmpty() ? It might not be constant-time either but it would be much faster than .size()

PR #1939, opened 2018-06-09
DB-6419: add validate_backup system procedure

jyuanca:
Adding a system procedure syscs_util.validate_backup to detect backup corruption caused by file transfer or media corruption, etc. The system procedure makes sure all files for the backup are in place and verifies their checksums.

Please also review ee branch.

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/hbase/CheckTableIT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/hbase/CheckTableIT.java
@@ -266,6 +266,24 @@ public class CheckTableIT extends SpliceUnitTest {
     @Test
     public void negativeTests() throws Exception {
 
+
+        try {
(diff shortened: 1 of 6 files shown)


hbase_sql/src/test/java/com/splicemachine/hbase/CheckTableIT.java
```diff
@@ -266,6 +266,24 @@ public void testIndexExcludeDefaults() throws Exception {
     @Test
     public void negativeTests() throws Exception {
 
+
+        try {
+            spliceClassWatcher.execute(String.format("call syscs_util.check_table('%s', '%s', null, 1, null)", SCHEMA_NAME, A, getResourceDirectory(), A));
```
dgomezferro: Are theses tests related to this change?

jyuanca: Not to validate_backup. It is for check_table

PR #1945, opened 2018-06-12
DB-6389: improve bulk import locality

jyuanca:


hbase_sql/src/main/java/com/splicemachine/derby/stream/function/HFileGenerationFunction.java
```diff
@@ -17,13 +17,19 @@
...
+import com.splicemachine.access.hbase.HBaseConnectionFactory;
 import com.splicemachine.derby.stream.iapi.OperationContext;
 import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.HRegionLocation;
 import org.apache.hadoop.hbase.KeyValue;
+import org.apache.hadoop.hbase.*;
```
changli6: I think it is better to "import org.apache.hadoop.hbase.MetaTableAccessor;"

PR #1949, opened 2018-06-12
DB-7084 Link Ranger JAR files to ranger plugins directory and create needed hdfs directories (2.5)

wb14123:


assembly/hdp2.6.3/src/main/resources/common-services/SPLICEMACHINE/2.5.1/package/scripts/splice_install_master.py
```diff
@@ -0,0 +1,103 @@
...
+                        type="directory",
+                        action="create_on_execute",
+                        owner=hbase_user)
+
+    params.HdfsResource("/user/splice/history",
+                        type="directory",
+                        action="create_on_execute",
+                        owner=hbase_user)
+
+    params.HdfsResource("/user/splice/spark-warehouse",
```
wtang16: Good catch. Currently spark.sql.warehouse.dir is set to '/user/splice/spark-warehouse'. Do we need to update the installation guide for CDH as well (in which spark-warehouse directory is not created)?

wb14123: In my test it seems `'/user/splice/spark-warehouse` is not needed. Deleted these code.

PR #1954, opened 2018-06-13
SPLICE-2185 Fix KillOperationIT

dgomezferro:
Synchronize SpliceBaseOperation.closeables, fix contextManager usage and close TransactionResourceImpl when exception is raised during prepare

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
@@ -521,9 +521,7 @@ public final class TransactionResourceImpl
 			SanityManager.ASSERT(cm != null, "cannot cleanup on error with null context manager");
 
         //DERBY-4856 thread dump
-        boolean result = cm.cleanupOnError(e, diagActive);
-		csf.removeContext(cm);
-		return result;
+        return cm.cleanupOnError(e, diagActive);
 	}
 
 	boolean isIdle()
--- a/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
@@ -521,7 +521,9 @@ public final class TransactionResourceImpl
 			SanityManager.ASSERT(cm != null, "cannot cleanup on error with null context manager");
 
         //DERBY-4856 thread dump
-        return cm.cleanupOnError(e, diagActive);
+        boolean result = cm.cleanupOnError(e, diagActive);
+        csf.removeContext(cm);
+        return result;
 	}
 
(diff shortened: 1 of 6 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
```diff
@@ -521,9 +521,7 @@ boolean cleanupOnError(Throwable e, boolean diagActive)
 			SanityManager.ASSERT(cm != null, "cannot cleanup on error with null context manager");
 
         //DERBY-4856 thread dump
-        boolean result = cm.cleanupOnError(e, diagActive);
-		csf.removeContext(cm);
-		return result;
+        return cm.cleanupOnError(e, diagActive);
```
yxia92: @dgomezferro You mentioned that csf.removeContext(cm) is called outside the function cleanupOnError() thus can be removed here, but I couldn't find it at the caller to cleanupOnError(), could you point me to the code?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
```diff
@@ -125,8 +125,10 @@ public void close() {
                 LOG.error("Exception while closing operation", e);
             }
         }
-        for (Future<Pair<OperationContext, Iterator<ExecRow>>> future : futures) {
-            future.cancel(true);
+        if (futures != null) {
```
yxia92: @dgomezferro Could you take a look at my comment at https://splicemachine.atlassian.net/browse/DB-7126?focusedCommentId=54759&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-54759 and see if it is a valid concern with the fix to just check futures !=null.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
dgomezferro: There are 3 call sites:
- TransactionResourceImpl.handleException()
- EmbedConnection.close()
- EmbedConnection constructor

The latter 2 close it explicitly calling restoreContextStack(), but the first one doesn't. Several callers of TransactionResourceImpl.handleException() close it too, but not all. At first this 'double' cleanup was creating problems for me, but now it seems to be fine, so I'll revert this change to keep the current behaviour.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/NLJoinFunction.java
dgomezferro: @yxia92 I've answered on that JIRA with more details, but I don't think that's problematic for this case. There might be some issues with CachedOperations below NLJoins or other operations that clone() the operation tree though.

PR #1967, opened 2018-06-18
DB-7154 Mark CachedOperation as populated to prevent double call

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/CachedOperation.java
```diff
@@ -71,8 +71,8 @@ public void init(SpliceOperationContext context) throws StandardException {
             super.init(context);
             source.init(context);
             if (!populated) {
-                populateCache();
                 populated = true;
+                populateCache();
             }
```
jyuanca: Can you add a comment for ordering changes, so that no one may accidentally reorder it?

dgomezferro: Sure, will do, thanks!

PR #1973, opened 2018-06-20
DB-7115 save value of the flag materialized in Ser/De for Activation …

yxia92:
…class, also remove logic to bind expression twice for UpdateNode.

Please see comments in [DB-7115](https://splicemachine.atlassian.net/browse/DB-7115) for explanation of the fix.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/serialization/ActivationSerializer.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/serialization/ActivationSerializer.java
@@ -119,6 +119,7 @@ public class ActivationSerializer {
             try{
                 visit(baseActClass.getDeclaredField("row"),baseFields);
                 visit(baseActClass.getDeclaredField("resultSet"),baseFields);
+                visit(baseActClass.getDeclaredField("materialized"),baseFields);
             } catch (NoSuchFieldException e) {
                 SpliceLogUtils.warn(LOG, "Could not serialize current row list");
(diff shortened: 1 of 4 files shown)


splice_machine/src/main/java/com/splicemachine/derby/serialization/ActivationSerializer.java
```diff
@@ -119,6 +119,7 @@ private void visitBaseActivationFields() throws IOException, IllegalAccessExcept
             try{
                 visit(baseActClass.getDeclaredField("row"),baseFields);
                 visit(baseActClass.getDeclaredField("resultSet"),baseFields);
+                visit(baseActClass.getDeclaredField("materialized"),baseFields);
```
msirek: If this code is copying field values for a clone operation, should we perhaps handle all of the field values defined in the BaseActivation class?

yxia92: Thanks @msirek . I also had a question why we didn't put all the fields defined in the BaseActivation class in visitBaseActivationFields() in the first place, but couldn't get an answer that I'm confident. My guess is that most of the fields are no longer used in a copy or could be re-computed, and serialize the value may add CPU overhead and also increase the size of the serialized object? Since I don't fully understand the usage of all the variables. I chose to be conservative and only make changes to the one that is known to be problematic.

PR #1976, opened 2018-06-21
SPLICE-1923 Remove race condition from Vacuum

dgomezferro:
Make it not wait for concurrent transactions

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
@@ -68,8 +68,6 @@ public class Vacuum{
 
     public void vacuumDatabase() throws SQLException{
 
-        ensurePriorTransactionsComplete();
-
         //get all the conglomerates from sys.sysconglomerates
         PreparedStatement ps = null;
         ResultSet rs = null;
@@ -91,6 +89,7 @@ public class Vacuum{
 
         //get all the tables from HBaseAdmin
         try {
+            long oldestActive = getOldestActiveTransaction();
             Iterable<TableDescriptor> hTableDescriptors = partitionAdmin.listTables();
 
             if (LOG.isTraceEnabled()) {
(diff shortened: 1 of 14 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -68,8 +68,6 @@ public Vacuum(Connection connection) throws SQLException {
 
     public void vacuumDatabase() throws SQLException{
 
-        ensurePriorTransactionsComplete();
-
```
jyuanca: Vacuum still needs to wait for all active txns to complete. An long running txn may want to access a table which was created by another txn, and dropped while it is running. The table descriptor's txnId could be null or less than oldest txn id. In either case, it would be deleted by vacumme.

dgomezferro:
Could you elaborate a little bit more the example, Jun? I don't understand
completely the scenario you are describing. If possible with a timeline of
events. Thanks!

jyuanca:
@dgomezferro Here is an example. The last select fails with a TableNotFound error. 

session 1                                              session 2 
create table a(col1 int);
insert into a values 1,2,3;
autocommit off;
select * from t;
                                                               drop table t;          
                                                               vacuum();

select * from t;

jyuanca:
@dgomezferro Here is an example. Run the following command from two sqlshell session(s1 and s2). At time 7, the select will fail with TableNotFound error. 

1. s1 - create table a(col1 int);
2. s1 - insert into a values 1,2,3;
3. s1 - autocommit off;
4. s1 - select * from t;
5. s2 - drop table t;          
6. s2 - vacuum();
7.  s1 - select * from t;

dgomezferro:
Thanks @jyuanca , you are right, working on a fix.

splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -132,72 +145,35 @@ public void vacuumDatabase() throws SQLException{
...
-                        }
+            activeTxn = txn.getEffectiveBeginTimestamp();
+
+            TxnView next;
+            try (Stream<TxnView> activeTransactions = reader.getActiveTransactions()){
+                while((next = activeTransactions.next())!=null){
+                    long txnId = next.getEffectiveBeginTimestamp();
+                    if(txnId < activeTxn){
+                        activeTxn = txnId;
+                        break;
```
msirek: The 'if' clause breaks out of the while loop without searching all of the active transactions, so it might not find the oldest active transaction.

PR #1979, opened 2018-06-21
Revert "DB-7115 save value of the flag materialized in Ser/De for Activation …(2.7)"

jyuanca:
Reverts splicemachine/spliceengine#1975


jyuanca:
accidentally merged it. reverting...

PR #1982, opened 2018-06-21
SPLICE-2131 Allow Spark Adapter to run queries directly inside Splice

dgomezferro:


jyuanca:
Can you add some ITs?

wtang16:
@dgomezferro I have a couple of questions:
1. what caused the embedded splice not checking the view permission (i.e., the root cause of the issue)?
2. what kind of performance impact are we expecting with the export (which may be slow)? Is there a more efficient way to pass the result to the spark driver?

dgomezferro:
@wtang16 
1. The embedded splice checks view permissions, but we can't trust it because it runs on an user-controlled jvm, so we have to perform further checks during execution. During execution we only see the Spark application is trying to access the base table, if we allow access the Spark application would have (potentially) access to all data on the base table, not just the columns defined on the view.
2. The performance impact is somewhat mitigated by only generating 1 replica in HDFS, rather than the conventional 3 replicas, and it will be proportional to the size of the Result set. There are more efficient ways of passing data around, we could implement a service similar to the external shuffle service that doesn't require the data to be exported to hdfs, but it will require more development time. I'll open a JIRA to track the effort.

wtang16:
Thanks @dgomezferro for the explanation! 

Question: if something does wrong with the EXPORT or SpliceSpark.getSession.read.csv(), will exception/stack trace be logged in a log file? This will help problem diagnosis.

splice_spark/src/main/spark2.1/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -229,10 +243,69 @@ class SplicemachineContext(url: String) extends Serializable {
...
+        fs.delete(path, true)
+      }
+    })
+    path.toString
+  }
+
+  def getRandomName(): String = {
+    val name = new Array[Byte](20)
+    new SecureRandom().nextBytes(name)
+    Bytes.toHex(name)
```
yxia92: SecureRandom().nextBytes() does not seem to guarantee uniqueness. Though the chance of name conflict is slim, theoretically it is still possible. Should we append the random name with a timestamp to make it unique?

PR #1995, opened 2018-06-23
DB-7161 avoid system table lookup when doing schema permission check,…

yxia92:
… cache null for permission cache


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -9266,7 +9278,7 @@ else if( key instanceof PermDescriptor)
             }
         }
 
-        dataDictionaryCache.permissionCacheAdd(key, permissions);
+        dataDictionaryCache.permissionCacheAdd(key, permissions==null? Optional.<PermissionsDescriptor>absent():Optional.of(permissions));
```
dgomezferro: You could use Optional.fromNullable() here (minor nitpick)

splice_machine/src/test/java/com/splicemachine/derby/security/PermissionIT.java
```diff
@@ -0,0 +1,169 @@
...
+
+import static com.splicemachine.derby.test.framework.SpliceUnitTest.assertFailed;
+import static com.splicemachine.derby.test.framework.SpliceUnitTest.resultSetSize;
+import static java.lang.String.format;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by yxia on 6/22/18.
+ */
+@Category(value = {SerialTest.class,HBaseTest.class})
```
dgomezferro: Does it have to be Serial?

yxia92: Yes, since the two test cases testCachedSchemaPrivilege() and testCachedTablePrivilege() work on the same set of schema/tables and each of them have grant/revoke privileges operations. In case they run in parallel, the permission check result may vary.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
yxia92: Thanks @dgomezferro , I will incorporate the suggested change shortly.

PR #2002, opened 2018-06-27
include sqlshell.sh in hdp2.6.3 rpm

liprais:
include sqlshell.sh in hdp2.6.3 rpm
also fixed a typo in pom.xml

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -611,7 +611,7 @@
                                     <classifier>splicemachine</classifier>
                                     <version>${envClassifier}.${project.version}.${parcel.patch}</version>
                                     <group>Applications/Internet</group>
-                                    <licence>Copyright © 2018 Splice Machine Inc. Licensed under the Apache License, Version 2.0.</licence>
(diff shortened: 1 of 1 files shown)


assembly/pom.xml
```diff
@@ -624,7 +624,24 @@
...
+                                            <filemode>
+                                                755
+                                            </filemode>
+                                            <directoryIncluded>
+                                                false
+                                            </directoryIncluded>
+                                        </mapping>
                                     </mappings>
+                                    <postinstallScriptlet>
+                                      <script>sed -i 's/##SPLICELIBDIR##/\/var\/lib\/splicemachine/g' /var/lib/splicemachine/sqlshell.sh</script>
```
MurrayBrown: one technique to make this sed easier to read, is use a different statment separating character, instead of / which then has to be escaped, use e.g. @. 

's@##SPLICELIBDIR##@/var/lib/splicemachine@g' /var/lib/splicemachine/sqlshell.sh

PR #2003, opened 2018-06-27
DB-7157 Support HDP2.6.1 (master)

perryluopl:
Add HDP2.6.1 support for master branch.


perryluopl:
I built it on Jenkins. Here is the link.

http://206.225.8.98:8080/view/parallel/job/DB-7157/

PR #2011, opened 2018-06-28
DB-7178 Check kerberos principal when opening JDBC connection

dgomezferro:


dgomezferro:
Let me test this on the cluster before merging, thanks

PR #2029, opened 2018-07-04
DB-7205 Create RM client in authenticated context

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerMaster.java
```diff
@@ -205,10 +176,12 @@ public void run() {
...
 
         SpliceSpark.getContextUnsafe(); // kickstart Spark
-        
+
         while(!end.get()) {
             Thread.sleep(10000);
             ugi.checkTGTAndReloginFromKeytab();
+            if (yarnUgi != null)
+                yarnUgi.checkTGTAndReloginFromKeytab();
         }
```
jyuanca: Why need to login twice using the same principal/keytab and relogin from both UGI?

wtang16: Do we know why it didn't work?

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerMaster.java
```diff
@@ -89,58 +96,22 @@ private void run() throws Exception {
...
-                end.set(true);
-            }
-        };
-        AMRMClientAsync<AMRMClient.ContainerRequest> rmClient = AMRMClientAsync.createAMRMClientAsync(1000, allocListener);
-        rmClient.init(conf);
-        rmClient.start();
-
-        // Register with ResourceManager
-        rmClient.registerApplicationMaster(Utils.localHostName(), 0, "");
+        UserGroupInformation.isSecurityEnabled();
```
OlegMazurov: The result is not used. Do we somehow rely on a side-effect of this call?

dgomezferro: Yes, it's mentioned in YARN documentation as a way to trigger Kerberos login, from https://hadoop.apache.org/docs/r2.7.3/hadoop-yarn/hadoop-yarn-site/YarnApplicationSecurity.html :

> This means you have a relative similar workflow across secure and insecure clusters.
> 1. Suring AM startup, log in to Kerberos. A call to UserGroupInformation.isSecurityEnabled() will trigger this operation.

PR #2048, opened 2018-07-09
DB-7206 Display incompatible data type in error message

changli6:


jyuanca:
Can you add a negative test?

PR #2059, opened 2018-07-13
DB-7234:add c3p0 related jars to hdp rpm

liprais:
add c3p0 related jars to hdp rpm


wtang16:
You may want to prepare 2.5/2.7 PRs too.

PR #2062, opened 2018-07-15
DB-7141 Allow Splice usernames with dashes or a period

changli6:


changli6:
The current commit may break parseMultiPartSQLIdentifier(). Pending.

PR #2065, opened 2018-07-17
DB-6045 Support compression for avro external table

changli6:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -426,23 +429,23 @@ public void createEmptyExternalFile(ExecRow execRows, int[] baseColumnMap, int[]
...
-                        empty.write().option("compression",compression).option("timestampFormat", "yyyy-MM-dd'T'HH:mm:ss.SSSZZ").mode(SaveMode.Append).csv(location);
-                    }
+            if (storedAs!=null) {
+                if (storedAs.toLowerCase().equals("p")) {
+                    empty.write().option("compression",compression).partitionBy(partitionByCols.toArray(new String[partitionByCols.size()]))
+                            .mode(SaveMode.Append).parquet(location);
+                }
+                else if (storedAs.toLowerCase().equals("a")) {
+                    empty.write().partitionBy(partitionByCols.toArray(new String[partitionByCols.size()]))
+                            .mode(SaveMode.Append).format("com.databricks.spark.avro").save(location);
```
jyuanca: why option("compression", compression) does not work for avro?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -412,11 +412,14 @@ public StructType getExternalFileSchema(String storedAs, String location) throws
...
             StructType schema = AvroUtils.supportAvroDateType(execRows.schema(),storedAs);
-
-            Dataset<Row> empty = SpliceSpark.getSession()
-                        .createDataFrame(new ArrayList<Row>(), schema);
+            SparkSession spliceSpark = SpliceSpark.getSession();
+            if (storedAs != null && storedAs.toLowerCase().equals("a")){
+                if (compression.toLowerCase().equals("zlib"))
+                    compression = "deflate";
+                spliceSpark.conf().set("spark.sql.avro.compression.codec",compression);
+            }
```
jyuanca: spark session and configuration is shared by all spark jobs. If two "create external table" are being executed, and one wants to compress avro file, the other does not, or they try to compress in different format, setting a global configuration may not work

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
changli6: I tried but not work.
In the Spark-Avro doc (https://docs.databricks.com/spark/latest/data-sources/read-avro.html), the example showed the compression code was accepted by spark conf settting. 

You can also specify Avro compression options:
Copy to clipboardCopy
import com.databricks.spark.avro._
// configuration to use deflate compression
spark.conf.set("spark.sql.avro.compression.codec", "deflate") 

Others raised the same confusion of compression code setting, but was not responded.
(https://github.com/databricks/spark-avro/issues/259)

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
changli6: Thanks @jyuanca  Global session is improper, let me address it.

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -881,34 +881,59 @@ public void testWriteReadFromCompressedParquetExternalTable() throws Exception {
...
+    }
+    @Test
+    public void testWriteReadFromCompressedZlibAvroExternalTable() throws Exception {
+        methodWatcher.executeUpdate(String.format("create external table compressed_zlib_avro_test (col1 varchar(24))" +
+                " COMPRESSED WITH ZLIB STORED AS AVRO LOCATION '%s'", getExternalResourceDirectory()+"compressed_zlib_avro_test"));
+        int insertCount = methodWatcher.executeUpdate(String.format("insert into compressed_zlib_avro_test values ('XXXX')," +
+                "('YYYY')"));
+        Assert.assertEquals("insertCount is wrong",2,insertCount);
+        ResultSet rs = methodWatcher.executeQuery("select * from compressed_zlib_avro_test");
         Assert.assertEquals("COL1 |\n" +
```
jyuanca: The test cannot tell whether the table is compressed

changli6:
Global session level setting of Avro compression may impact other Avro-related spark jobs, and spark-avro doesn't support DataFrame level setting for compression so far, we have to defer this feature until Spark implemented the DataFrame level setting for Avro file.

PR #2076, opened 2018-07-20
DB-7255: fix external table partitioning

jyuanca:
The fix is for 2.5 only. Partitioning works for master and 2.7

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TableElementList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TableElementList.java
@@ -31,6 +31,7 @@
 
 package com.splicemachine.db.impl.sql.compile;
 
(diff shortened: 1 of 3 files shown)


yxia92:
I think though master has no issue, it has the same obsolete code in SparkDataSet.writeORCFile(), writeParquetFile() as well as writeArvoFile() to populate the list for cols:
            List<Column> cols = new ArrayList();
            for (int i = 0; i < baseColumnMap.length; i++) {
                cols.add(new Column(ValueRow.getNamedColumn(baseColumnMap[i])));
            }
it would be good to remove them too. but I'm fine if we want to piggyback it with some future PRs.

jyuanca:
@yxia92 I'm working on other Jira that will clean up this for master and 2.7 branch

yxia92:
@jyuanca got it, that is fine, thanks.

PR #2091, opened 2018-07-23
DB-7221: schema inference performance improvement

jyuanca:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -382,25 +388,57 @@ public Partitioner getPartitioner(DataSet<ExecRow> dataSet, ExecRow template, in
...
+                        dataset = SpliceSpark.getSession().read().format("com.databricks.spark.avro").load(location);
                     } else if (storedAs.toLowerCase().equals("o")) {
-                        schema = SpliceSpark.getSession().read().orc(location).schema();
+                        dataset = SpliceSpark.getSession().read().orc(location);
                     }
                     if (storedAs.toLowerCase().equals("t")) {
                         // spark-2.2.0: commons-lang3-3.3.2 does not support 'XXX' timezone, specify 'ZZ' instead
                         schema = SpliceSpark.getSession().read().option("timestampFormat", "yyyy-MM-dd'T'HH:mm:ss.SSSZZ").csv(location).schema();
                     }
+                    dataset.printSchema();
```
yxia92: Is this line for debugging purpose? Do we want this to be printed out unconditionally?

jyuanca: Yes, this should not be too verbose, and will be helpful trouble shooting schema mismatch problem.

jyuanca:
@wtang16 spark.sql.parquet.mergeSchema is false by default. I will add an example in Jira so that Gary can document.

wtang16:
Cool. Thanks @jyuanca

PR #2103, opened 2018-07-27
SPLICE-1809: fix partition column data type

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/stream/utils/ExternalTableUtils.java
```diff
@@ -0,0 +1,204 @@
...
+        for (int pos : partitionColumnMap) {
+            partitionColumns.add(pos);
+        }
+        int index = 0;
+        for (int i = 0; i < tableFields.length; ++i) {
+            if (!partitionColumns.contains(i)) {
+                partitionedTableFields[index++] = tableFields[i];
+            }
+        }
+        for (int pos:partitionColumnMap) {
```
yxia92: It seems that we don't need to populate the partition columns here, as we are not check the consistency of partitioning columns between data schema and table schema.

yxia92:
@jyuanca SPLICE-1809 said Spark infers partioncolumn data type poorly, my understanding of the fix is not to use the partition column type in the table definition directly to replace that inferred from the data. In ExternalTableUtils.checkSchema(), we also skip the type check for partition columns.  In case, we put the wrong type in the external table definition for the partition column, what would happen? We report error at the run time? Thanks.

splice_machine/src/main/java/com/splicemachine/derby/stream/utils/ExternalTableUtils.java
jyuanca: Yes

jyuanca:
@yxia92 Yes, a type conversion exception will be thrown at runtime. There is not reliable approach to infer partition column data types, the user needs to provide correct types when creating external table.

splice_machine/src/main/java/com/splicemachine/derby/stream/utils/ExternalTableUtils.java
```diff
@@ -0,0 +1,204 @@
+/*
+ * Copyright (c) 2012 - 2017 Splice Machine, Inc.
+ *
```
msirek: Copywrite should span up to 2018.

splice_machine/src/main/java/com/splicemachine/derby/stream/utils/ExternalTableUtils.java
```diff
@@ -0,0 +1,204 @@
...
+        // Infer schema from external files\
+        StructType dataSchema = null;
+        try {
+            dataSchema = getDataSchema(dsp, tableSchema, partitionColumnMap, location, storeAs, false);
+        }
+        catch (StandardException e) {
+            String sqlState = e.getSqlState();
+            if (sqlState.equals(SQLState.INCONSISTENT_NUMBER_OF_ATTRIBUTE) ||
+                    sqlState.equals(SQLState.INCONSISTENT_DATATYPE_ATTRIBUTES)) {
+                dataSchema = getDataSchema(dsp, tableSchema, partitionColumnMap, location, storeAs, true);
```
msirek: If we hit INCONSISTENT_DATATYPE_ATTRIBUTES error, how would calling getDataSchema with mergeSchema as true resolve the error?  I thought merging schemas would just pick up more columns that exist in other parquet files, but if two columns in 2 separate parquet files have the same name but different data types, wouldn't this still fail, or is it actually smart enough to do stuff like combine an int and a long into a long?  If not, maybe it's best to let this fail on the first call to getDataSchema so we could avoid a lengthy schema inference, when we know in advance it's going to fail anyway.

jyuanca: There could be schema evolution, for example dropping a column and add another column. When mergeSchema is false, it only reads schema from one file, which may have outdated metadata. If data types does not match, fall back with mergeSchema=true to get complete schema.

msirek: @jyuanca I see.  If the user accidentally enters the wrong data type for a column, but no schema evolution took place, it may take a long time for them to get back an error message.  It would be nice if there were a user option or database property to specify not to check for schema evolution.  It may be worth creating another JIRA.

jyuanca: @msirek The user should get the error when he was trying to create the table on existing data. We have an option for the user to merge schema or not. Once the table is created, we assume table and data schema matches.
 
When a user reads the table, if schema does not match when mergeSchema is false, we are pretty sure it should match if mergeSchema is true.

msirek: @jyuanca OK, so you're saying the above code is only invoked after the external table is already created, so therefore retrying the call to getDataSchema with mergeSchema as true should always succeed?

jyuanca: @msirek It should always succeed, unless the data is tampered. In that case, we will find a mismatch and throw an error.

PR #2110, opened 2018-07-30
DB-7160 Fix projection pruning for joins where column references from…

msirek:
… predicates are lost.

Please see comment in [DB-7160](https://splicemachine.atlassian.net/browse/DB-7160) for discussion of the fix.


dgomezferro:
I'm OK with the change but I'd rather have @yxia92 review it, since she implemented the projection pruning

yxia92:
I'm still reviewing the problem case. The problem I see only happens when the series of join nodes are in a subquery, not when the series of join nodes occur in the main query. for both cases, at the binding time , the join condition of each join node is correctly pointing to the immediate child JoinNode's result columns. This is the tree structure projection pruning assumes and is based on. However, for the subquery case, when it comes to the projection pruning phase, the column references in the join condition is changed to point to the result columns of the leaf nodes(FromBaseTable nodes). This violate the assumption of the projection pruning optimization. I haven't located where this change happens. I'll  discuss with Mark further before approving this PR. The fix may still be OK if the change of the column references is legitimate. Thanks.

yxia92:
Mark and I had a discussion, the direct cause of the problem is that a SubqueryNode for an exists subquery is bound twice, this could break the linkage between the result columns of parent resultset(e.g., JoinNode) and the result columns of the immediate child result sets(e.g., another JoinNode, FromBaseTable). The source of parent result set's result columns is supposed to point to the IMMEDIATE child result sets' result columns, but the double binding break this relationship, which also breaks the assumption projection pruning optimization is based on. This is similar to DB-7069. I'm looking into the possibility of avoiding double binding.

msirek:
We've decided to go with Yi's fix.  Closing this PR...

PR #2111, opened 2018-07-30
SPLICE-2210 EE classes are duplicated in hbase_sql.jar

OlegMazurov:
Replaced including sources and resources from EE modules with corresponding dependencies in the test scope.

### DIFF
--- a/hbase_sql/pom.xml
+++ b/hbase_sql/pom.xml
@@ -1391,6 +1391,38 @@
             <properties>
                 <ee_out>,ee</ee_out>
             </properties>
+            <dependencies>
+                <dependency>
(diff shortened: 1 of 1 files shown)


dgomezferro:
@OlegMazurov Have you tested this change builds a usable parcel with all EE classes?

OlegMazurov:
@dgomezferro I made sure ITs include and successfully run EE tests. Building a parcel I now run into maven errors. Please mark this PR as "changes requested" so that it's not merged before the errors are resolved.

OlegMazurov:
This PR introduces cyclic dependencies between splice_backup and hbase_sql (even if for the test scope only). A different approach is needed.

PR #2112, opened 2018-07-31
SPLICE-2029: Added support for Python stored procedure

Zeninma:
Added support for Python Procedure.
[For documentation](https://docs.google.com/document/d/1fxsoCP1H2qFpcswp8DzzqJ9WkarrBGWN1PFfDJ6WnII/edit)

File changes:
1. db-engine/src/main/java/com/splicemachine/db/catalog/types/RoutineAliasInfo.java -- Added language and compiledPyCode these two additional fields to the class.

2. db-engine/src/main/java/com/splicemachine/db/iapi/services/loader/ClassInspector.java -- Use method name to resolve the Java wrapper method DEMO.

3. Added an additional setValue method for each of the classes below. These setValue methods will be invoked by PyStoredProcedureResultSetFactory’s create method:
 db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
  * db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDouble.java
  * db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLInteger.java
  * db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLLongint.java
  * db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLReal.java
  * db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLSmallint.java

4. db-engine/src/main/java/com/splicemachine/db/impl/services/reflect/JarLoader.java -- Allow Python stored procedure related classes to be loaded into the installed Java wrapper method. If the Java wrapper method is added as a system function, then the changes to JarLoader are not needed.

5. db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java -- Reflect the changes made to RoutineAliasInfo’s constructor.

6. db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/Procedure.java  -- Reflect the changes made to RoutineAliasInfo’s constructor

7. db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateAliasNode.java -- Compile the Python script. Reflect the changes made to RoutineAliasInfo’s constructor.

8. db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/MethodCallNode.java -- If  MethodCallNode is for Python script, then upon code generation put all of its parameters into an array of Objects. Also added code to resolve the parameters’ type for the method call.

9. db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/StaticMethodCallNode.java -- For Python stored procedure, resolve the Java wrapper method for the method call. Modify the code-generation implementation so that the parameters for the wrapper method can be passed in properly.

10. db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj -- Add “PYTHON” as a token and allow arbitrary string to be the external name for the CreateAliasNode if "PYTHON" shows as the LANGUAGE, since the external name will contain the actual Python script.

11. db-engine/src/main/java/com/splicemachine/db/impl/sql/pyprocedure/PyCodeUtil.java -- This class contains two methods. One is used to compile a Python script into an array of byte. The other one is used to execute the compiled Python script.

12. db-engine/src/main/java/com/splicemachine/db/impl/sql/pyprocedure/PyInterpreterPool.java -- Implements a resource pool for Jython’s PythonInterpreter. This class is a singleton.

13. splice_machine/src/main/java/com/splicemachine/derby/impl/sql/pyprocedure/PyStoredProcedureResultSetFactory.java -- This is a Facotry class, which provides a static method to construct a ResultSet.

14. splice_machine/src/main/java/com/splicemachine/EngineDriver.java -- Add the initialization of the PyInterpreterPool singleton

15. splice_machine/src/main/java/com/splicemachine/derby/utils/BaseAdminProcedures.java -- Change the buildExecRow from protected into public, as it will be used by the PyStoredProcedureResultSetFactory.

16. txn-it-procs/src/main/java/org/splicetest/txn/TxnTestProcs.java -- This class was used to contain example Splice Machine Java stored procedure. Added the Java wrapper static method DEMO into this class.


jyuanca:
- Can Python method be defined in a source file instead of string literal?
- There are many hard-coded String literals.
-  Can you tell whether a method is in Java or python by means other than method name "DEMO"?

Zeninma:
@jyuanca 
1. No, currently it only supports using a string literal.
By using a source file, do you mean that when creating the Stored procedure the user can use a command like the one below?
```
CREATE PROCEDURE SPLICE.TEST_FUNC() PARAMETER STYLE JAVA READS SQL DATA LANGUAGE PYTHON DYNAMIC RESULT SETS 0 EXTERNAL NAME SOURCE_FILE_PATH;
```
where the SOURCE_FILE_PATH is the path to the Python source file.

2. For hard-coded String literals, I think there two kinds of them. The first ones are the scripts that get appended to the user's script. The second ones are the method name like "DEMO". I think by moving the wrapper function into a system function will help with the second ones. I am not sure whether how should I fix the first ones.

db-engine/src/main/java/com/splicemachine/db/iapi/types/TypeId.java
```diff
@@ -1162,6 +1162,39 @@ public String getCorrespondingJavaTypeName(){
...
+            case "char[]":
+                return "java.lang.Character[]";
+            case "boolean[]":
+                return "java.lang.Boolean[]";
+            case "double[]":
+                return "java.lang.Double[]";
+            case "float[]":
+                return "java.lang.Float[]";
+            case "int[]":
+                return "java.lang.Number[]";
```
msirek: What is the reason for using abstract class java.lang.Number for some of these instead of the specific type, like java.lang.Integer?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/MethodCallNode.java
```diff
@@ -1318,4 +1354,37 @@ public void setHashableJoinColumnReference(ColumnReference cr) {
...
+
+		QueryTreeNode qt = (QueryTreeNode) newParm;
+
+		/*
+		 ** Since we need the parameter to be in Java domain format, put a
+		 ** SQLToJavaValueNode on top of the parameter node if it is a
+		 ** SQLValueNode. But if the parameter is already in Java domain
+		 ** format, then we don't need to do anything.
+		 */
+		if ( ! (qt instanceof JavaValueNode))
```
msirek: Comment says we expect a JavaValueNode or SQLValue node here.  We could throw an exception here if qt is neither of those nodes.

hbase_sql/pom.xml
```diff
@@ -1197,6 +1197,7 @@
                                 <argument>-Xdebug</argument>
                                 <!-- <argument>-XX:+UnlockCommercialFeatures</argument>
                                 <argument>-XX:+FlightRecorder</argument> -->
+                                <!--<argument>-Dderby.debug.true=DumpClassFile</argument>-->
```
msirek: Did you have "DumpClassFile" turned on for debugging?  I think we may want to not have this enabled by default on all systems.

hbase_sql/src/test/java/com/splicemachine/derby/transactions/CallableTransactionIT.java
```diff
@@ -78,8 +78,8 @@
...
-	private static final String DROP_PROC_INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE = String.format("DROP PROCEDURE %s.INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE", SCHEMA_NAME);
+//	private static final String CREATE_PROC_INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE = String.format("CREATE PROCEDURE %s.INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE(IN tableName VARCHAR(40), IN id INT, IN fname VARCHAR(20), IN lname VARCHAR(30), IN fname2 VARCHAR(20), IN lname2 VARCHAR(30), IN id2 INT) PARAMETER STYLE JAVA MODIFIES SQL DATA LANGUAGE JAVA DYNAMIC RESULT SETS 1 EXTERNAL NAME 'org.splicetest.txn.TxnTestProcs.INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE'", SCHEMA_NAME);
+//	private static final String DROP_PROC_INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE = String.format("DROP PROCEDURE %s.INSERT_UPDATE_GETx2_DELETE_AND_GET_EMPLOYEE", SCHEMA_NAME);
```
msirek: Were these commented out for a reason?

Zeninma: They were commented out as these two Strings were not used in the tests.
This reminds me that I should deleted the commented out Strings in my own test PyProcedureIT.

hbase_sql/pom.xml
Zeninma: This argument has been commented out and I left it here thought this might be something that people want to turn on for debugging.

msirek: OK, I see.  Thanks.

db-engine/src/main/java/com/splicemachine/db/catalog/types/RoutineAliasInfo.java
```diff
@@ -356,12 +379,14 @@ public void writeExternal( ObjectOutput out )
...
 		out.writeBoolean(calledOnNullInput);
-		out.writeInt(0); // future expansion
+		out.writeInt(1); // future expansion
 		if (parameterCount != 0) {
 			ArrayUtil.writeArrayItems(out, parameterNames);
 			ArrayUtil.writeArrayItems(out, parameterTypes);
 			ArrayUtil.writeIntArray(out, parameterModes);
 		}
+		out.writeObject(language);
+		out.writeObject(compiledPyCode);
```
yxia92: Prefer to put these two lines inside the "if (expansionNum ==1)" statement.

db-engine/src/main/java/com/splicemachine/db/iapi/types/TypeId.java
Zeninma: Array type of Objects among parameters are used for OUT parameters. Upon code generation, they would be converted to NumberDataType via the setValue method. For int, byte, short, it is the setValue((int) theValue) will be invoked.

Since there is no setValue method for type Integer, Byte, Short but only a method for Number, I convert them to Number.

db-engine/src/main/java/com/splicemachine/db/catalog/types/RoutineAliasInfo.java
Zeninma: If the user is using the version with Python stored procedure, he/she always needs to write these two fields even though they are null. However, if the user is not using this version, then these two fields will not be written as the code will not be here.

Therefore, I think it is OK to leave the code as it is. May I know what is your concern?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/StaticMethodCallNode.java
```diff
@@ -704,6 +691,100 @@ private void resolveRoutine(FromList fromList,
...
+				if(origParmCnt > 0) {
+					System.arraycopy(this.routineInfo.getParameterModes(), 0, updatedParmModes,0,origParmCnt);
+					System.arraycopy(this.routineInfo.getParameterNames(), 0, updatedParmNames,0,origParmCnt);
+					System.arraycopy(this.routineInfo.getParameterTypes(), 0, updatedParmTypes, 0, origParmCnt);
+				}
+				updatedParmModes[origParmCnt] = 1; // States the input script is an IN parameter
+				// Here Need to use a unique name
+				updatedParmNames[origParmCnt] = "PYCODE";
+				// I am not sure whether the format Id is correct or not
+				updatedParmTypes[origParmCnt] = new TypeDescriptorImpl(new BaseTypeIdImpl(442),
```
yxia92: I assume 442 is from the BLOB type defined in StoredFormatIds.java, could you use BLOB_TYPE_ID_IMPL to make it more readable?

Zeninma: Yes, I was experiencing some issue with using BLOB_TYPE_ID_IMPL as it resulted in 440 (or some other int) and hence used a hardcoded int instead. Then I forgot to figure out what caused the problem.

I will replace it with BLOB_TYPE_ID_IMPL.

splice_machine/src/main/java/com/splicemachine/derby/utils/PyRoutineWrapper.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class PyRoutineWrapper {
+    /**
+     * A wrapper static method for Python Stored Procedure
+     *
+     * @param args if the registered stored procedure has n parameters, the args[0] to args[n-1] are these parameters.
+     * The args[n] must be a String which contains the Python script. The rest of elements in the args are ResultSet[].
+     * In the current test version only allows one ResultSet[]
+     */
+    public static void PyProcedureWrapper(Object... args)
```
dgomezferro: Nitpick, methods are usually lower case

wtang16:
@dgomezferro do you approve @Zeninma's latest change?

dgomezferro:
There are a few issues that can be handled in follow up JIRAs:

1. Provide better error reporting when dealing with problematic python (note the indentation as conn.commit()):
`CREATE PROCEDURE PYPROC_TYPE_UNIT_TEST() PARAMETER STYLE JAVA READS SQL DATA LANGUAGE PYTHON DYNAMIC RESULT SETS 1 AS 'def run(rs):
 c = conn.cursor()
 stmt = "select * from sys.systables"
 c.execute(stmt)
 d = c.description
 result = c.fetchall()
 rs[0] = factory.create([d,result])
conn.commit()
 c.close()
 conn.close()';
ERROR XJ001: Java exception: 'java.lang.IllegalArgumentException: Cannot create PyString from null!'.
ERROR XJ001: Java exception: 'Cannot create PyString from null!: java.lang.IllegalArgumentException'.`

2. Handle null values:
` call PYPROC_TYPE_UNIT_TEST();
ERROR 08006: A network protocol error was encountered and the connection has been terminated: the requested command encountered an unarchitected and implementation-specific condition for which there was no architected message (additional information may be available in the derby.log file on the server)
ERROR 08006: DERBY SQL error: SQLCODE: -1, SQLSTATE: 08006, SQLERRMC: A communications error has been detected: {0}.::SQLSTATE: XJ001Java exception: 'java.lang.NullPointerException: java.io.IOException'.::SQLSTATE: XJ001Java exception: ': java.lang.NullPointerException'.`

PyStoredProcedureResultSetFactory.create() uses the first row to construct a templateRow, if the first row has NULLs it will fail when calling "MethodType mt = MethodType.methodType(void.class, templateObj.getClass());"


Could you also add some examples doing transformations on the Python side? Something simple like uppercasing a column or joining two columns together

PR #2119, opened 2018-07-31
DB-6865: write external table using data schema

jyuanca:
Please check [DB-6865](https://splicemachine.atlassian.net/browse/DB-6865) for explanation


dgomezferro:
@jyuanca  can you write a short explanation on the ticket about the problem/solution to aid the review? Thanks

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -774,20 +780,24 @@ public static DataSet toSpliceLocatedRow(JavaRDD<Row> rdd, OperationContext cont
...
-                    schema);
+                    dataSchema);
 
-            List<Column> cols = new ArrayList();
-            for (int i = 0; i < baseColumnMap.length; i++) {
-                cols.add(new Column(ValueRow.getNamedColumn(baseColumnMap[i])));
-            }
             List<String> partitionByCols = new ArrayList();
             for (int i = 0; i < partitionBy.length; i++) {
                 partitionByCols.add(ValueRow.getNamedColumn(partitionBy[i]));
```
changli6: I think we need to handle partitioned columns in Orc and Avro.

jyuanca: @changli6 Yes. The removed lines are useless.

changli6: The name was not modified for the partitioned columns?  We may treat them same as in write ParquetFile : partitionByCols.add(dataSchema.fields()[partitionBy[i]].name());

yxia92:
@jyuanca why we don’t need to replace the partition column name with the name from data schema for Avro file? Thanks.

jyuanca:
@changli6 Good catch. I push a fix with tests.

changli6:
Thanks @jyuanca commit. In writeORCFile, I think we could declare "insertDF" same as Parquet, Avro, and  modified the partitioned columns' name also.

jyuanca:
@changli6 Orc implementation on 2.7 and master is very different from parquet and avro. It has no problem writing data.

PR #2120, opened 2018-08-01
DB-7296: remove the behavior to always quote usernames

MurrayBrown:
add an option '-U' to encapsulate a new feature to quote usernames


wtang16:
We will need the fix for 2.5/2.7 users.

MurrayBrown:
@wtang16 thanks for the reminder: all three PRs are in and should look about the same

PR #2142, opened 2018-08-02
DB-7298: Do not allow incremental backup if SpliceHFileCleaner is not configured

jyuanca:
Refer [DB-7298](https://splicemachine.atlassian.net/browse/DB-7298) for an explanation. Please also review code changes on ee branch

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
+++ b/hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
@@ -184,21 +184,24 @@ public class BackupUtils {
     public static boolean shouldCaptureIncrementalChanges(FileSystem fs,Path rootDir) throws StandardException{
         boolean shouldRegister = false;
         try {
-            RecoverableZooKeeper zooKeeper = ZkUtils.getRecoverableZooKeeper();
-            String spliceBackupPath = HConfiguration.getConfiguration().getBackupPath();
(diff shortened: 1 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
```diff
@@ -340,4 +343,11 @@ private static boolean backupTimedout() throws Exception {
         }
         return timedout;
     }
+
+    public static boolean incrementalBackupEnabled() {
+
+        Configuration conf = HConfiguration.unwrapDelegate();
+        String fileCleaners = conf.get("hbase.master.hfilecleaner.plugins");
+        return fileCleaners.contains("SpliceHFileCleaner");
```
wtang16: Incremental backup is enabled via the following:
hbase.master.hfilecleaner.plugins = com.splicemachine.hbase.SpliceHFileCleaner,
org.apache.hadoop.hbase.master.cleaner.TimeToLiveHFileClean

Should we check whether both are set instead of just "SpliceHFileCleaner"?

jyuanca: TimeToLiveHFileCleaner is the default HBase file cleaner, which is not related to incremental backup

wtang16: Got it. Thanks.

PR #2152, opened 2018-08-08
DB-7297: Remove password from JDBC url in order to not log it

wb14123:


splice_machine/src/main/java/com/splicemachine/client/SpliceClient.java
```diff
@@ -129,10 +121,13 @@ public void run() {
...
-    public static String maskJDBCPassword(String jdbcUrl) {
-        return jdbcUrl.replaceAll(
-                "password=([a-zA-Z0-9%])*(;|$|&|\\s)",
-                "password=[Omitted]$2");
+    private static String jdbcPasswordRegex = "password=([a-zA-Z0-9%]*)(;|$|&|\\s)";
+
+    static String parseJDBCPassword(String jdbcUrl) throws SqlException {
+
+        Properties properties = ClientDriver.tokenizeURLProperties(jdbcUrl, null);
+        System.out.println(properties);
```
dgomezferro: Remove

splice_machine/src/main/java/com/splicemachine/client/SpliceClient.java
```diff
@@ -154,7 +149,19 @@ public static DataSource getConnectionPool(boolean debugConnections, int maxConn
                     pool = new ComboPooledDataSource();
                     try {
                         pool.setDriverClass("com.splicemachine.db.jdbc.Driver40");
+
+                        // parse and mask password in JDBC url in order to make it not logged
+                        String password = parseJDBCPassword(connectionString);
+                        if (password != null) {
+                            connectionString = connectionString.replace(
+                                    "password="+password+";", "");
+                            connectionString = connectionString.replace(
```
dgomezferro: Duplicate line

wb14123: This will first remove if the URL is like `....password=XXXX;username=XXX`, which have semicolon after password, and then remove the case like "username=XXXX;password=", which don't have anything after password.

splice_machine/src/main/java/com/splicemachine/client/SpliceClient.java
wb14123: Removed

splice_machine/src/test/java/com/splicemachine/client/SpliceClientTest.java
```diff
@@ -1,23 +1,21 @@
...
 
 import static org.junit.Assert.*;
 
 public class SpliceClientTest {
 
     @Test
-    public void testMaskJDBCPassword() {
+    public void testParseJDBCPassword() throws SqlException {
+        String password = "Abd98*@80EFg";
         String raw = "No suitable driver found for jdbc:splice://dwbdtest1r1w3.wellpoint" +
```
wtang16: Please replace the host name. Use generic names instead.

splice_machine/src/main/java/com/splicemachine/client/SpliceClient.java
dgomezferro: I see, thanks!

splice_machine/src/test/java/com/splicemachine/client/SpliceClientTest.java
wb14123: @wtang16 Fixed

PR #2157, opened 2018-08-08
DB-7316: fix external table schema for numeric column

jyuanca:


yxia92:
@jyuanca Why do we have problem inferring numeric type from the variable template? I see that in createTable(), template is populated with Null value of the corresponding type:
            // Get a template value for each column
            if (col_info.defaultValue != null) {
            /* If there is a default value, use it, otherwise use null */
                template.setColumn(ix + 1, col_info.defaultValue);
            }
            else {
                template.setColumn(ix + 1, col_info.dataType.getNull());
            }
Thanks.

jyuanca:
@yxia92 That will lose precision and scale for numeric type

PR #2158, opened 2018-08-09
SPLICE-1814 Disable JMX interface

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -975,25 +1137,46 @@ public V getThird(){
...
+        }
+
+        List<ExecRow> rows = new ArrayList<>();
+        for (HostAndPort server : servers) {
+            try (Connection connection = RemoteUser.getConnection(server.toString())) {
+                try (PreparedStatement ps = connection.prepareStatement("values SYSCS_UTIL.SYSCS_GET_DATABASE_PROPERTY(?)")) {
+                    ps.setString(1, key);
+                    try (ResultSet rs = ps.executeQuery()) {
+                        while (rs.next()) {
+                            ExecRow row = new ValueRow(1);
```
yxia92: ValueRow should be initialized with 2 fields instead of 1.

dgomezferro: Good catch, thanks!

PR #2162, opened 2018-08-10
SPLICE-2221 Remove RuntimeExceptions from logs

dgomezferro:


changli6:
In hbase1.1.8 and hbase1.1.2.2.6, MemStoreFlushAwareScanner also have 
```
        if (LOG.isDebugEnabled()) {
            SpliceLogUtils.debug(LOG, "close", new RuntimeException());
        }
```
Do we need to remove them also?

dgomezferro:
@changli6 Yes, good catch, thanks!

PR #2169, opened 2018-08-15
SPLICE-2029: Added support for Python stored procedure(2.7)

yxia92:
port change from master to branch-2.7

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/pyprocedure/PyStoredProcedureResultSetFactoryIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/pyprocedure/PyStoredProcedureResultSetFactoryIT.java
@@ -1,6 +1,5 @@
 package com.splicemachine.derby.impl.sql.pyprocedure;
 
-import com.splicemachine.derby.impl.sql.execute.operations.DropTableIT;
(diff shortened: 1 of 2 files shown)


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/pyprocedure/PyStoredProcedureResultSetFactoryIT.java
```diff
@@ -0,0 +1,224 @@
...
+            pyRow[13] = pyResultSet.getString(14);      // VARCHAR
+
+            pyRow[14] = getStr(pyResultSet.getClob(15));// CLOB
+
+            pyRow[15] = getStr(pyResultSet.getClob(16));// TEXT
+            pyRows.add(pyRow);
+        }
+        pyResultSet.close();
+
+        ResultSet javaResultSet = methodWatcher.executeQuery(CALL_JPROC);
```
OlegMazurov: This call fails in master in the mem profile. Fix or disable, make sure mem ITs pass.

yxia92: Thanks Oleg, I pushed a change to fix men platform.

PR #2171, opened 2018-08-16
DB-6691 Fix sentry build for CDH (2.5)

wb14123:


assembly/pom.xml
```diff
@@ -108,6 +108,13 @@
                     <value>parcel</value>
                 </property>
             </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>com.splicemachine</groupId>
+                    <artifactId>splice_sentry-${envClassifier}</artifactId>
+                    <version>2.5.0.1833-SNAPSHOT</version>
```
OlegMazurov: Use ${project.version} here and fix other occurrences of 2.5.0.1833-SNAPSHOT in dependencies as well.

OlegMazurov: The current version of the module has been changed to 2.5.0.1834-SNAPSHOT. This PR would introduce a new dependency on 2.5.0.1833-SNAPSHOT, which would be wrong and which would not be updated by the update script when the version changes to 1835. This is the reason for my request.

wb14123:
@OlegMazurov Fixed the version. I also tried to use `${project.version}` in ee's sentry module, but it  cannot compile. I think maybe we need to manually fix the version when merge.

wtang16:
Please note that the main part of the Sentry support resides in the -ee repo. Please review that one first: https://github.com/splicemachine/spliceengine-ee/pull/272.

We'll need to merge both when shipping DB-6691. Then we'll port to other branches.

assembly/pom.xml
```diff
@@ -108,6 +108,13 @@
                     <value>parcel</value>
                 </property>
             </activation>
+            <dependencies>
```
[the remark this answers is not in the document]
MurrayBrown: i agree this breaks the open-source build. the question i have is: what is the method for an enterprise feature to be correctly 'assembled' into our project? is there an example to copy for e.g. how the backup jars and feature is included?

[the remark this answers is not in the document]
OlegMazurov: Currently, it's running maven with -f platforms/hdp2.6.4/pom.xml -Phdp2.6.4,ee,hdp_services.
EE module are included in that pom file.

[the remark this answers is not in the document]
wb14123: @OlegMazurov I've tried to remove sentry dependency from assembly, but the built parcel will not include sentry jar in that way. (I've already added the dependencies in platforms/cdh*/pom.xml). I saw there is an ee profile in assembly's pom.xml, but Sentry is only needed in cdh platforms so I don't think I can add sentry dependency in that profile.

So here may have a more common question, how do I add a dependency when both ee and cdh are in the specified profile: e.g only add the dependency with `mvn install -Pee,cdh5.8.3`

wb14123:
Add a commit to stop splice after Spark IT so that sentry IT can run.

splice_spark/pom.xml
```diff
@@ -317,22 +317,7 @@
                     </execution>
                     <execution>
                         <id>############### stop-splice-its ###############</id>
-                        <phase>prepare-package</phase>
```
MurrayBrown: i'm confused about this 'stop test' removal. where is it being done now, instead?

wb14123: @MurrayBrown The stop tests task is in two phases before: prepare-package and pre-clean. prepare-package is not needed before because other IT will stop the Spice database. pre-client is too late. So I change the phases into post-integration-test so that the Splice database will stop after this module's IT.

PR #2173, opened 2018-08-17
SPLICE-2029 fix IT issue for mem platform

yxia92:


OlegMazurov:
You need to uncomment the Java test.
Also please change DropTableIT -> PyStoredProcedureResultSetFactoryIT and maybe get rid of CLASS_NAME and set SCHEMA_NAME directly.

yxia92:
Thanks Oleg, I've uncomment the test, and correct class name, but I still leave the CLASS_NAME there.  mem platform test passes on my local system.

PR #2175, opened 2018-08-20
Fix ZooKeeper liveness check in scripts

chuckwondo:
Changed `until` condition to stop when `COUNT` equals `0` so that the
condition following the loop will evaluate to true, as expected, when
ZooKeeper is not found to be alive.

Also, changed condition to break out of the loop to succeed when `nc`
returns a non-empty string instead of an empty string.


chuckwondo:
@changli6 Would you like me to fix `platform_it/src/test/bin/start-splice-its` as well as part of this PR? If so, would you like me to create a new JIRA issue, or just update SPLICE-2228 with information about that file as well?

changli6:
@chuckwondo You could use this PR and update SPLICE-2228. Thanks.

assembly/standalone/template/bin/start-splice.sh
```diff
@@ -128,11 +128,11 @@ _startZoo "${BASE_DIR}" "${ZOO_LOG}" "${LOG4J_FILE}" "${ZOO_DIR}" "${CP}"
 #######################################################################################################
 COUNT=65
 ZOO_UP=""
-until [ $COUNT -lt 0 ]; do
+until [ $COUNT -eq 0 ]; do
     sleep 1
     ZOO_UP=$(echo 'ruok' | nc localhost 2181)
-    if [ -z "$ZOO_UP" ]; then
-                    break
+    if [ -n "$ZOO_UP" ]; then
```
OlegMazurov: If ZooKeeper is up it returns 'imok'. Should we check for that here?

start-splice-cluster
```diff
@@ -184,11 +184,11 @@ echo "Starting ZooKeeper. Log file is ${ZOO_LOG}"
 #######################################################################################################
 COUNT=65
 ZOO_UP=""
-until [ $COUNT -lt 0 ]; do
+until [ $COUNT -eq 0 ]; do
     sleep 1
     ZOO_UP=$(echo 'ruok' | nc localhost 2181)
-    if [ -z "$ZOO_UP" ]; then
-                    break
+    if [ -n "$ZOO_UP" ]; then
```
OlegMazurov: Ditto.

assembly/standalone/template/bin/start-splice.sh
chuckwondo: As long as we receive a non-empty string response, we should be fine. In fact, I suggest that we do not check for an explicit response value because doing so could break the script in the future.

Specifically, the [use of Four Letter Words is a security risk](https://blog.cloudera.com/blog/2017/06/apache-zookeeper-four-letter-words-and-security/), and starting with ZooKeeper 3.4.10 and 3.5.3, a 4lw whitelist has been implemented, and by default only `srvr` is on the whitelist.

This means that when we upgrade to one of those versions or beyond, the response to `ruok` will not be `imok` if `ruok` is not added to the whitelist, but will instead be something along the lines of "ruok is not on the whitelist" (I don't recall the exact wording). Therefore, checking specifically for `imok` will break the script.

At some point, we may want to consider an alternative to using a 4lw here.

start-splice-cluster
chuckwondo: Ditto. :-)

chuckwondo:
@changli6 I've updated SPLICE-2228 along with the 3 scripts in which the ZooKeeper liveness check appears, which now appear in this PR.

assembly/standalone/template/bin/start-splice.sh
OlegMazurov: If the script breaks because of a change in the protocol, that's a good thing  (and whoever makes that change will have to fix the script).
Not so good when the script continues when things are in fact broken (exactly what you are fixing here).

chuckwondo: Sure, that's a valid argument. I'd be happy to update this to check for `imok` instead of any non-empty response, if you'd prefer that approach. Just let me know.

OlegMazurov: Please do.

chuckwondo: Done.

PR #2177, opened 2018-08-21
DB-7365 Check external table directory emptiness

changli6:


dgomezferro:
@jyuanca did you mean to close the PR? I agree an IT would be needed.

jyuanca:
@dgomezferro my mis-operation :-)

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreateTableConstantOperation.java
```diff
@@ -469,10 +469,10 @@ protected void createTable(Activation activation) throws StandardException {
                 DistributedFileSystem fileSystem = SIDriver.driver().getFileSystem(location);
                 FileInfo fileInfo = fileSystem.getInfo(location);
                 String[] files = fileSystem.getExistingFiles(location, "*");
-                if(!fileInfo.exists() || (files.length == 1 && files[0].equals("_SUCCESS"))) {
+                if(!fileInfo.exists() || (files.length == 0) || (files.length == 1 && files[0].equals("_SUCCESS"))) {
```
OlegMazurov: Can you use ExternalTableUtils.isEmptyDirectory here as well (extend with an optional DistributedFileSystem parameter)?

changli6: The "fileInfo" was used several times in the code block, I'd like to keep it.  The other part of the IF condition used ExternalTableUtils.isEmptyDirectory.

PR #2184, opened 2018-08-22
DB-7361 Close timestamp client/server connection on request timeout

dgomezferro:


splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampClient.java
```diff
@@ -261,6 +261,9 @@ public void operationComplete(ChannelFuture future) throws Exception {
         try {
             boolean success = callback.await(timeoutMillis);
             if (!success) {
+                // We timed out, close the channel so that the next request recreates the connection
+                channel.close();
+
```
msirek: Would it be useful to do the following here? :
state.set(State.DISCONNECTED);

dgomezferro: When we close the channel the method channelDisconnected() is triggered and that handles the state update and reconnection.

PR #2189, opened 2018-08-22
DB-7370: check null pointer when adding warning to activation

jyuanca:


dgomezferro:
I'm OK with the NPE check, but something doesn't make sense. We are generating splits automatically, we generate duplicate splits, and since that fails, we try to add a warning to the activation (or log it). I think we should be removing the duplicates from the generated splits instead. If the user is the one providing the splits then the warning makes sense.

jyuanca:
@dgomezferro This is common code path for both user provided split points and generated split points. I'm fixing NPE to unblock customer first.

dgomezferro:
@jyuanca Understood, then I'm +1

PR #2190, opened 2018-08-22
DB-7213 clean up Maven dependencies

OlegMazurov:
1) spliceenine/utility shades its dependencies: hppc, log4j, snappy-java (relocated)
2) all platform provided dependencies are marked as such
3) many unnecessary dependencies have been removed
4) [almost] all guava classes are imported from org.spark-project.guava - prerequisite for DB-7369 
5) mem profile has been upgraded to use HBase 1.2, Spark 2.2.0

Tested: ITs, sanity + TPC-H in cluster


hbase_storage/pom.xml
```diff
@@ -25,148 +25,46 @@
         <dependency>
             <groupId>com.splicemachine</groupId>
             <artifactId>pipeline_api</artifactId>
-            <version>2.8.0.1834-SNAPSHOT</version>
+            <version>${project.version}</version>
```
MurrayBrown: does @perryluopl have a python script which needs to change as a result of this edit? I like it the change, but if a script exists he should be aware before he builds this.

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/TxnLifecycleEndpoint.java
```diff
@@ -16,7 +16,7 @@
 
 import java.io.IOException;
 
-import com.google.common.primitives.Longs;
+import org.spark_project.guava.primitives.Longs;
```
MurrayBrown: i am not saying no to this: but... why?

hbase_storage/pom.xml
OlegMazurov: Perry's script updates all occurrences of the current version, say 2.8.0.1834-SNAPSHOT, to next one in all pom.xml files. It should continue working correctly after the change.
One explicit version per pom.xml is sufficient, I believe.

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/TxnLifecycleEndpoint.java
OlegMazurov: Unification. This change ensures only one, currently prevailing, version of the guava library is used in the module.
See DB-7369 for details. There is some chance once DB-7369 is fixed all imports will be converted back to com.google.common, uniformly this time.

PR #2191, opened 2018-08-23
SPLICE-1923 Remove race condition from Vacuum

dgomezferro:
Make it not wait for concurrent transactions

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
@@ -110,6 +110,7 @@ public class Vacuum{
                         continue; //ignore system tables
                     }
                     boolean requiresDroppedId = false;
+                    boolean ignoreDroppedId = false;
                     if (table.getTransactionId() != null) {
                         TxnView txn = SIDriver.driver().getTxnSupplier().getTransaction(Long.parseLong(table.getTransactionId()));
                         if (txn.getEffectiveBeginTimestamp() >= oldestActiveTransaction) {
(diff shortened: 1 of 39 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -111,6 +109,42 @@ public void vacuumDatabase() throws SQLException{
...
+                                    LOG.info("Ignoring recently dropped table: " + table.getTableName() + " by transaction " + txn.getTxnId());
+                                }
+                                continue;
+                            }
+                        } else {
+                            // This table must have a dropped transaction id before we can process it for vacuum
+                            continue;
+                        }
+                    }
                     if(!activeConglomerates.contains(tableConglom)){
```
jyuanca: For old tables without a transaction id and drop transaction id, can vacuum physically drop it? If not, still needs to check sys.sysconglomerate table?

dgomezferro: Yes, it can. We always check sys.sysconglomerates anyway, the transactionId logic just makes sure that the tables that cannot be found in sys.sysconglomerates are safe to delete.

I have a test case for your scenario in VacuumIT.testVacuumDoesNotRemoveTablesWithoutTransactionId

While looking at this I realized there was a logic gap regarding tables without a transactionId (created before this change goes in) but with droppedTransactionId (deleted after this change). I wasn't checking whether they were still in use. I addressed that and added a test case: testVacuumDoesNotDeleteTableWithoutTransactionIdPossiblyInUse

PR #2197, opened 2018-08-23
DB-7365  Check external table location is not a directory

changli6:
When checking if a location is empty, checking it is a directory or not at first.


changli6:
The new PR is in https://github.com/splicemachine/spliceengine/pull/2198, close this one.

PR #2203, opened 2018-08-24
SPLICE-1855 allow add/drop primary key on empty table

yxia92:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/AlterTableConstantOperation.java
```diff
@@ -556,7 +633,8 @@ private void createPrimaryKeyConstraint(Activation activation,
         /*
          * modify the conglomerate descriptor with the new conglomId
          */
-        updateTableConglomerateDescriptor(tableDescriptor,
+        ConglomerateDescriptor droppedConglomerateDescriptor =
```
dgomezferro: This only recreates the base table conglomerate, right? Would it be possible to do that for all conglomerates (indexes...) instead? It would be safer from a concurrent writes perspective.

yxia92: @dgomezferro Yes, it only recreates the base table conglomerate. For empty table without concurrent write, the existing conglomerates for indexes are OK, but I agree that with concurrent writes or non-empty table, the index conglomerates have to be recreated. Let me see if I can make changes within this PR. Thanks.

dgomezferro: I'm fine with the PR as is, if it is trivial to recreate those conglomerates then go for it, if there's weird edge cases then we'll tackle it when we support data/concurrent writes.

PR #2205, opened 2018-08-27
DB-7377 Fix updateTest in TimestampIT.

msirek:
This fixes the failing IT test point from SPLICE-1212, and re-enables its execution.

The fix is to reset autoCommit to false after the rollback.  Rollback has a side effect of setting autoCommit to true.

I also changed the IT to execute serially, since SpliceWatcher is not thread safe.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TimestampIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TimestampIT.java
@@ -18,7 +18,9 @@ import com.splicemachine.derby.test.framework.SpliceUnitTest;
 import com.splicemachine.derby.test.framework.SpliceWatcher;
 import com.splicemachine.derby.test.framework.TestConnection;
 import com.splicemachine.homeless.TestUtils;
+import com.splicemachine.test.SerialTest;
 import org.junit.*;
+import org.junit.experimental.categories.Category;
 import org.junit.rules.RuleChain;
 import org.junit.rules.TestRule;
 import org.junit.runner.RunWith;
@@ -40,6 +42,7 @@ import static org.junit.Assert.*;
  */
 @FixMethodOrder(MethodSorters.NAME_ASCENDING)
 @RunWith(Parameterized.class)
+@Category(SerialTest.class)
 public class TimestampIT extends SpliceUnitTest {
     private static final String SCHEMA = TimestampIT.class.getSimpleName().toUpperCase();
     private Boolean useSpark;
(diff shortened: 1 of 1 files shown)


msirek:
Hi @dgomezferro, 
In SpliceWatcher.java is the comment:

/**
 * A TestWatcher that provides Connections, Statements, and ResultSets and then closes them when finished() is called.
 *
 * Not thread-safe, synchronize externally if using in a multi-threaded test case.
 */

I thought JUnit was single-threaded before, and am not sure exactly how it uses threading, but adding '@Category(SerialTest.class)' is needed to make the problem go away on 2.7. 
In the case of this JIRA, the properties of the connection, namely the autoCommit property seems to be:

1.  Modified by thread 1 to false.
2.  Updated by another thread to true, possibly when a rollback is done (is this perhaps a deferred asynchronous event).
3.  Accessed by thread 1 when it is executing a query, which dies when autoCommit is expected to be false.

We could synchronize on the connection in all places it's used, or create a new connection for each thread, or make the test serial.  Since it's an IT, I didn't think it was crucial to optimize it for performance and keep it multithreaded, so I just made the test serial.   If you have any insight about this JUnit option '@Category(SerialTest.class)',  or links you could share, I would appreciate it.

dgomezferro:
@msirek I find that surprising, SerialTest is just meant for tests that cannot run in parallel with other tests for some reason (checking the number of conglomerates, things like that), but currently we are running all ITs serially anyway, so adding that doesn't change the thread safety of the class at all. What it does is it runs those SerialTests at the beginning, and then the rest of the ITs, so it might be an interaction with some IT running before it.

I'd rather find the real root cause of the failure rather than masking it by making the IT Serial when it shouldn't need it.

msirek:
Hi @dgomezferro, 
    I found a way to attach to the running maven process that executes the ITs, and verified that on the mem platform there are two threads running TimestampIT.updateTest at the same time (4 threads are allocated, but only 2 being used).  One thread has useSpark set to true and and other has useSpark set to false.  I don't know if it's always two updateTests that are run simultaneously or if sometimes updateTest could be run at the same time as one of the other tests.  If it's the latter, then this test cannot use parallel execution because it involves a rollback, which would mess up the resultSet in the test running in the other thread.

    Since I verified we're doing multithreading, and that is the cause of the problem (we're not synchronizing updateTest between the two threads, and the rollback in one thread is updating the autoCommit property of the connection when the other thread is not protected via synchronization), I would like to go ahead with keeping this test in the SerialTest category and just avoid multithreading altogether.  What do you think?

BTW, here is the rule in mem_sql/pom.xml which I believe is determining the number of threads to use:

                    <execution>
                        <id>integration-tests-parallel</id>
                        <goals>
                            <goal>integration-test</goal>
                            <goal>verify</goal>
                        </goals>
                        <configuration>
                            <excludedGroups>com.splicemachine.test.SerialTest, com.splicemachine.test.HBaseTest, ${excluded.categories}</excludedGroups>
                            
                            <parallel>classes</parallel>
                            **<threadCount>4</threadCount>**

wb14123:
@msirek I find you've sent PR for SPLICE-1212 to branch-2.7 and it has been merged. So is this fix needed for branch 2.7?

msirek:
Hi @wb14123, I included this fix already in the splice-1212_2.7 branch, so it's already in 2.7.

PR #2213, opened 2018-08-29
DB-7375 Fix bad translation of imported and literal date data for years before 1900.

msirek:
Fixes the TO_DATE function to avoid building a java.sql.Date object specified as milliseconds sourced from a Joda DateTime object.  There are known timezone differences between these two types that makes conversion from one to the other problematic.  Instead, build a LocalDate object and use a standard API call to construct the Date object.


msirek:
Need to redo this fix...

PR #2217, opened 2018-08-30
DB-7404 CheckTableIT fails to run in IDE

OlegMazurov:
The `commons-dbutils` test scope dependency is not transitively propagated from the `splice_machine` project.
Has to be explicitly specified where used.


wb14123:
@OlegMazurov Do we need this for 2.5/2.7?

PR #2219, opened 2018-08-30
DB-7401 Remove splice engine/platforms/*/pom.xml

OlegMazurov:
Jenkins builds have been updated to stop using those files.


MurrayBrown:
I like this change, but i think we should wait at least one week: the next weekly build must pass all regression tests before we ‘burn the boats’.

wb14123:
@OlegMazurov @perryluopl I'll merge the PR. The build settings on Jenkins may need to change. Can any of you do this? Thanks!

wb14123:
@OlegMazurov Oh, I just found you have updated that. Just ignore my comment.

PR #2223, opened 2018-09-01
DB-7375 Fix bad translation of imported date data for years before 1884.

msirek:
Fixes incorrect dates getting imported into Splice due to improper conversion of a Joda DateTime object into a java.sql.Date object.  These two classes use different chronologies.

Please see comment in [DB-7375](https://splicemachine.atlassian.net/browse/DB-7375?focusedCommentId=55816&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-55816) for solution description.


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFunctions.java
```diff
@@ -113,7 +112,8 @@ public static Time TO_TIME(String source) throws SQLException {
      */
     public static Time TO_TIME(String source, String format) throws SQLException {
         if (source == null) return null;
-        return new Time(parseDateTime(source, format));
+        Date date = TO_DATE(source, format);
+        return new Time(date.getTime()) ;
```
yxia92: TO_DATE() will change the date to local date, as a result, we lose the hour/minute/second info. Won't the Time object created here all be 00:00:00?

PR #2225, opened 2018-09-04
DB-6775 Support Atlas

wb14123:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceSystemProcedures.java
```diff
@@ -33,10 +34,7 @@
 import com.splicemachine.procedures.external.ExternalTableSystemProcedures;
 
 import java.sql.Types;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
MurrayBrown: Do we have a splice style guide? this seems like a conflict between the school of thought that says 'explicitly list all imported classes' and 'it is okay to import an entire package'.  Where does our engineering team stand on that question in general? surely `import java.util.Collection` would also suffice?

wb14123:
@MurrayBrown @msirek Fixed the issues mentioned above.

PR #2227, opened 2018-09-04
DB-7098 allow statement with only comment difference to reuse stateme…

yxia92:
…nt cache

Please see comment in [DB-7098](
https://splicemachine.atlassian.net/browse/DB-7098?focusedCommentId=55929&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55929) for details.


splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/CommentStripIT.java
```diff
@@ -0,0 +1,157 @@
...
+
+import com.splicemachine.db.impl.sql.misc.CommentStripper;
+import com.splicemachine.db.impl.sql.misc.CommentStripperImpl;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Created by yxia on 8/30/18.
+ */
+public class CommentStripIT {
```
dgomezferro: This should be a Test, not an IT

msirek:
hi @yxia92, the JIRA shows query performance numbers.  Have you collected similar metrics on a build without the fix?  Just wondering what the regression amount is.

Also, just wondering why we use property names starting with 'derby.database' instead of 'splice'?  Not that this is bad, I'm just curious why new properties which are splice-specific are tagged with derby.  Or is this just to signify the property has effect in the derby parser instead of some other component in splice?

db-engine/src/main/java/com/splicemachine/db/impl/sql/misc/sqlcommentstrip.jj
```diff
@@ -0,0 +1,89 @@
...
+
+<IN_COMMENT> MORE :
+{
+    < ~[] >
+}
+
+TOKEN :
+{
+    <OPEN_QUOTE: "'"> : IN_QUOTATION
+}
```
msirek: As strange as it may seem, we could have a table name or column name of "/*", so we should probably parse out double quotes as well so such weird names aren't stripped out or cause the javaCC parser to choke.

yxia92:
Hi @msirek The performance number in JIRA is not the elapse time running/compiling the query. It is just the time for the function call of CommentStripper.stripStatement(). I  collected it through a separate java program which directly calls the function CommentStripper.stripStatement() passing in a sql string as input. This function parses an input sql (a string) and output a string with the comment removed. This is a new function and is the main overhead introduced by DB-7098.  I computed the delta before and after the function call to get the elapse time. 

For the name of the database property, I just thought it is related to parser, thus used a prefix similar to the existing properties. Thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/misc/sqlcommentstrip.jj
yxia92: Thanks @msirek for catching this. I think I missed that scenario, I'll work on a fix for it.

yxia92: Hi @msirek I've pushed a commit to take into consideration a statement like "table_name/*". Test is updated too. Thanks.

yxia92:
Based on the discussion in the morning, I pushed a commit to turn off the functionality to find cached plan match ignoring comments difference by default.

db-engine/src/main/java/com/splicemachine/db/impl/sql/misc/sqlcommentstrip.jj
msirek: Thanks @yxia92, looks good.

PR #2240, opened 2018-09-12
DBAAS-1583 Build Sqlshell artifact

bklo94:
Created the artifacts to be bundled together with sqlshell to allow it to run and connect. 
To test: 
1. Run ./start-splice-cluster 
2. After build is complete, start the sqlshell located in assembly/sqlshell/
3. Wait for it to connect to database
4. Run query 
5. Move sqlshell folder a different location
6. Repeat steps 2-4


OlegMazurov:
db-engine should not be a dependency for sqlshell.
db-client depends on db-shared and db-tools-i18n - these should be sufficient (if not we need to fix it, again).

assembly/pom.xml
```diff
@@ -128,6 +200,7 @@
                                         splice_ee-${envClassifier},
                                         splice_backup-${envClassifier},
                                         splice_sentry-${envClassifier},
+                                        splice_atlas-${envClassifier},
```
OlegMazurov: Why are you adding splice_atlas to CDH parcels?

assembly/pom.xml
```diff
@@ -573,7 +646,6 @@
                                         spark-avro_2.11,
                                         splice_ranger_admin-${envClassifier},
                                         splice_ranger_service-${envClassifier},
-                                        splice_atlas-${envClassifier},
```
OlegMazurov: ... and removing it from HDP rpms?

assembly/common/src/main/resources/bin/sqlshell.sh
```diff
@@ -149,22 +149,19 @@ if [[ "$jversion" < "1.8" ]]; then
    exit 1
 fi
 
-# set splice lib dir
-SPLICE_LIB_DIR="##SPLICELIBDIR##"
```
MurrayBrown: removing this will break the on-premise install for cloudera, hortonworks and mapr

assembly/common/src/main/resources/bin/sqlshell.sh
```diff
@@ -149,22 +149,19 @@ if [[ "$jversion" < "1.8" ]]; then
...
-   if [[ "$CURDIR" = *"bin" ]]; then
-      SPLICE_LIB_DIR="${CURDIR}/../lib"
-   else #subdir
-      SPLICE_LIB_DIR="${CURDIR}/lib"
-   fi
-fi
-
 # set up classpath to point to splice jars
-export CLASSPATH="${SPLICE_LIB_DIR}/*"
+IJ_SYS_ARGS="-Djdbc.drivers=com.splicemachine.db.jdbc.ClientDriver"
```
MurrayBrown: duplicates line 185?

assembly/common/src/main/resources/bin/sqlshell.sh
MurrayBrown: there is a part of the assembly pom which replaces this token ##SPLICELIBDIR## with the appropriate directory, such as /opt/cloudera/parcels/SPLICEMACHINE/lib or /opt/splicemachine/lib

assembly/common/src/main/resources/bin/sqlshell.sh
```diff
@@ -149,22 +149,19 @@ if [[ "$jversion" < "1.8" ]]; then
...
+CURDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DBCLIENT=$(find ${CURDIR}/lib/ -type f -name 'db-client*-SNAPSHOT.jar')
+DBTOOLS=$(find ${CURDIR}/lib/ -type f -name 'db-tools*-SNAPSHOT.jar')
+DBENGINE=$(find ${CURDIR}/lib/ -type f -name 'db-engine*-SNAPSHOT.jar')
+
+#check if all the tools needed to connect are found
+if [ -z "$DBCLIENT"] || [ -z "$DBTOOLS"] || [ -z "$DBENGINE"]; then
```
MurrayBrown: i like this added check! perhaps we use the old 'SPLICE_LIB_DIR' to search for these files?

assembly/sqlshell/sqlshell-2.8.0.1836-SNAPSHOT.sh
```diff
@@ -0,0 +1,244 @@
+#!/bin/bash
```
MurrayBrown: this file should probably live in e.g. assembly/target/sqlshell/

MurrayBrown:
> db-engine should not be a dependency for sqlshell.
> db-client depends on db-shared and db-tools-i18n - these should be sufficient (if not we need to fix it, again).

I just tried that 'locally' on another copy of this same artifact.

if i remove db-engine, I can connect and make a query, but when i 'quit;' i get this error:

splice> quit;
Exception in thread "main" java.lang.NoClassDefFoundError: com/splicemachine/db/jdbc/AutoloadedDriver
	at com.splicemachine.db.impl.tools.ij.utilMain.cleanupGo(utilMain.java:438)
	at com.splicemachine.db.impl.tools.ij.utilMain.go(utilMain.java:257)
	at com.splicemachine.db.impl.tools.ij.Main.go(Main.java:231)
	at com.splicemachine.db.impl.tools.ij.Main.mainCore(Main.java:186)
	at com.splicemachine.db.impl.tools.ij.Main.main(Main.java:77)
	at com.splicemachine.db.tools.ij.main(ij.java:67)
Caused by: java.lang.ClassNotFoundException: com.splicemachine.db.jdbc.AutoloadedDriver
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 6 more

MurrayBrown:
@OlegMazurov I agree that SPLICE-1348 should fix this.  Sadly that was only merged to the 'master' branch, and we've built this 2.7.0.1835 standalone from branch-2.7.

Perhaps we merge SPLICE-1348 to 2.7?

OlegMazurov:
I wonder, does it really work without db-shared dependency?

MurrayBrown:
I have not been able to get that stacktrace to go away, but it only occurs on 'quit'.  should i be looking at the master version instead? is it easy to port SPLICE-1348 to branch-2.7

OlegMazurov:
If a similar PR is also considered for 2.7 then SPLICE-1348 should also be backported (or included), which is trivial.

Answering my previous question: db-client shades (includes) all its dependencies so there is no need to add db-shared.

I don't like copying stuff from target and using artifact version in db-client-${project.version}.jar.
There should be a maven way to properly name artifacts by their id and version.

chuckwondo:
I agree with @OlegMazurov regarding copying things from a target directory. @bklo94 see docs on the Maven Assembly Plugin. This might help: https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/including-and-excluding-artifacts.html. You shouldn't need to use the `copy-rename-maven-plugin` for copying the jar files. Same thing for the `sqlshell.sh` file. All of the files being copied should not be specified in `pom.xml` but rather in an assembly descriptor. I'm not familiar with the repo yet, but is there an assembly descriptor somewhere? I don't see one offhand.

MurrayBrown:
perhaps this example:
https://github.com/splicemachine/splice-zeppelin/blob/master/pom.xml#L541-L579

PR #2242, opened 2018-09-12
SPLICE-2246 Missing io.airlift dependencies in HDP cluster deployment

OlegMazurov:


OlegMazurov:
The order is not important but we have two lists of artifacts in the pom file, one for parcels another one for rpms. I'd rather have them ordered similarly because the hidden message should be "changed one, change the other" and not "this one is for rpm, it's probably OK if it's different".

dgomezferro:
@OlegMazurov if the list of artifacts is the same (or very similar) for parcels and RPMs would it make sense to define it as a variable and reuse it to prevent changes to one but not the other?

OlegMazurov:
Introduced commonArtifacts property to reduce redundancy and have only one list of common artifacts to maintain (although my goal is to get rid of this list as **difficult** to maintain).

PR #2245, opened 2018-09-13
SPLICE-2247 Log exceptions during deserialization

dgomezferro:


yxia92:
@dgomezferro Which rule is used to determine what module's readExternal() should be wrapped with the logic to log deserialization exception? For example, I see that the fix adds logging for readExternal() in SpliceBaseOperation, but not in BroadcastJoinOperation. Is it based on how trivial it is?

dgomezferro:
@yxia92 ideally all of them would be wrapped, I tried to focus on some special cases that could create more problems (SparkOperationContext, SpliceBaseOperation...) and ignored some of the simpler ones. However, the current PR missed the classes from the offending issue that prompted the creation of this PR (BulkImportPartitioner and BulkImportPartition specifically) so it probably makes sense to wrap them all.

PR #2248, opened 2018-09-14
DB-7458: fix byte array deserialization

jyuanca:


yxia92:
@jyuanca @dgomezferro I saw the discussion in the slack channel about this fix: 
> read() doesn't guarantee reading all expected bytes, so it only populates X amount of bytes and returns
But I'm still not very clear about the difference between in.read() and in.readFully(). Does X above mean the value obtained from `int size = in.readInt();`?
If so, if after reading X amount of bytes,there are still unread bytes left in the stream, does that indicate  an inconsistency in serialization? Thanks.

dgomezferro:
@yxia92 When we serialize an array of bytes (check writeByteArray on the same class) we write an int with the length of the array and then all the bytes of the array. When we deserialize we read the number of bytes (an int) and then we know there should be that number of bytes following on the stream. With readFully we guarantee that we read all those bytes, it blocks until it fills the byte[] provided in the call.

There could be more bytes on the stream because we might have serialized a bigger object with several byteArray fields, so all those fields would be serialized on the same stream.

yxia92:
Thanks @dgomezferro I got it now that read() is non-blocking and it could return before it reads all the expected bytes, we need extra logic(like a loop) to guarantee we get back the expected number of bytes. readFully() is blocking and it returns only when it reads all the expected bytes.

MurrayBrown:
Thanks for putting in additional ITs @jyuanca much appreciated!

PR #2250, opened 2018-09-17
SPLICE-2239 Version 4.0 Date serializer. Store year, month and day only, not time.

msirek:
This changes the Date encoding from storing milliseconds to:

(year << 9) + (month << 5) + day

Which changes what's stored from a point in DateTime, to simply a pure date.
See [SPLICE-2239](https://splice.atlassian.net/browse/SPLICE-2239?focusedCommentId=27567&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27567)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDate.java
@@ -104,11 +104,26 @@ public final class SQLDate extends DataType
         return BASE_MEMORY_USAGE;
     } // end of estimateMemoryUsage
 
-    int getEncodedDate()
+    public int getEncodedDate()
     {
         return encodedDate;
     }
-    
+
+	// Try to make modern dates close to the year 2020 take up less space.
+	// Subtracting this value from a disk-encoded date allows us to use
+	// 1 or 2 bytes for dates around year 2020 plus or minus 15 years,
+	// vs. 3 or 4 bytes.
+	public static final int DATE_ENCODING_OFFSET = 0xFC800;
+
+	public int getDiskEncodedDate()
+	{
+		return (getYear(encodedDate)  << 9)  +
+		       (getMonth(encodedDate) << 5)  +
+		        getDay(encodedDate)          -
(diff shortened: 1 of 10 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDate.java
```diff
@@ -104,11 +104,26 @@ public int estimateMemoryUsage()
...
     {
         return encodedDate;
     }
-    
+
+	// Try to make modern dates close to the year 2020 take up less space.
+	// Subtracting this value from a disk-encoded date allows us to use
+	// 1 or 2 bytes for dates around year 2020 plus or minus 15 years,
+	// vs. 3 or 4 bytes.
+	public static final int DATE_ENCODING_OFFSET = 0xFC800;
```
yxia92: Doesn't 0xFC800 represent year 1988? I don't quite get the relationship of this number with year 2020. Could you elaborate a bit?

msirek: Hi @yxia92, the new disk encoding is:
getYear(encodedDate) << 9)  + (getMonth(encodedDate) << 5) + getDay(encodedDate)
... so, the year is shifted 9 bits to the left.  So, for an encoded date of 0xFC800, to extract the year, we shift 9 bits to the right, to get 0x7E4, which is 2020 in decimal.  I subtract this encoded year during encoding and add it in during decoding, so that the encoded value is closer to zero for dates around the year 2020.  By adding/subtracting an encoded year (instead of adding/subtracting 2020 to/from the year before encoding), the initial encoding step (the above expression) produces only positive numbers, so there is no weirdness or complexity dealing with 2's complement in adding a negative year component with positive month and day.

yxia92: Thanks @msirek for the explanation. I miscounted the bits for year and got the number 0x7C4(1988), thus got myself confused. The code looks good to me.

PR #2253, opened 2018-09-17
DB-7446 treat inlist with expression as regular OrList at code genera…(2.5)

yxia92:
…tion phase, and fix wrong result


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -2663,6 +2664,77 @@ private void generateSingleQualifierCode(MethodBuilder consMB,
...
+        consMB.push(false);
+
+        /* Negate comparison result? */
+        consMB.push(false);
+
+        /* variantType for qualifier's orderable */
+        consMB.push(valNode.getOrderableVariantType());
+
+        int numArgs=10;
+        /* Qualifier's string representation */
```
wtang16: Not sure if we want to have a fixed string representation for any qualifiers. But we can change it later.

PR #2257, opened 2018-09-18
DB-7367 Write correct column name in external table data file

changli6:


db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/ValueRow.java
```diff
@@ -68,6 +68,7 @@
 	private int ncols;
 	private int hash = 0; // Cached value of hashCode(). Invalidate on any change to the object.
 	private byte[] key;
+	private String[] names;
```
yxia92: I see that the column names stored in ValueRow is only used in SparkDataSet.writeORCFile(). If that is the case, to avoid the 4 bytes extra added to every ValueRow, why don't we directly pass the column names through the DWLWriteOperation?

mem_sql/src/main/java/com/splicemachine/derby/impl/sql/ControlOnlyDataSetProcessorFactory.java
```diff
@@ -127,7 +127,7 @@ public void setup(Activation activation,String description,String schedulerPool)
         }
 
         @Override
-        public void createEmptyExternalFile(ExecRow execRow, int[] baseColumnMap, int[] partitionBy, String storageAs, String location, String compression) throws StandardException {
+        public void createEmptyExternalFile(StructFields fields, int[] baseColumnMap, int[] partitionBy, String storageAs, String location, String compression) throws StandardException {
```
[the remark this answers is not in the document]
changli6: Good catch, Thanks @msirek.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSet.java
```diff
@@ -829,13 +830,14 @@ public static DataSet toSpliceLocatedRow(JavaRDD<Row> rdd, OperationContext cont
                     context.getOperation().getExecRowDefinition().schema());
 
             String[] partitionByCols = new String[partitionBy.length];
+            StructField[] fields = context.getOperation().getExecRowDefinition().schema().fields();
```
msirek: GetExecRowDefinition() is called on lines 830 and 833, so there is some unnecessary overhead.  Could we save the schema the first time and re-use it at line 833?

changli6:
Thanks @yxia92 @msirek . The latest implementation uses DMLWriteOperation to pass column names, removed the column names member in ValueRow.

msirek:
+1

PR #2261, opened 2018-09-19
DB-7468 Remove service when stop region coprocessor to avoid memory leak (2.5)

wb14123:


hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexEndpoint.java
```diff
@@ -178,6 +178,7 @@ public void stop(CoprocessorEnvironment env) throws IOException{
         if(service!=null)
             try{
                 service.shutdown();
+                DatabaseLifecycleManager.manager().deregisterGeneralService(service);
```
chuckwondo: If `service.shutdown()` throws an exception, deregistration will not occur. Perhaps the line above should go into a `finally` block instead, if you want deregistration to occur regardless.

hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexObserver.java
```diff
@@ -146,8 +146,10 @@ public void stop(CoprocessorEnvironment e) throws IOException {
             super.stop(e);
             if (region != null)
                 region.close();
-            if(service!=null)
+            if(service!=null) {
                 service.shutdown();
+                DatabaseLifecycleManager.manager().deregisterGeneralService(service);
```
chuckwondo: If `service.shutdown()` throws an exception, deregistration will not occur. Perhaps the line above should go into a `finally` block instead (along with a null check of `service`), if you want deregistration to occur regardless.

PR #2268, opened 2018-09-21
DB-7484: batch load HFile

jyuanca:
By default, move HFiles from 30 regions into a staging directory, and load all of them. The number of regions to be batch loaded is configured by "splice.region.toLoad.perTask"

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
@@ -15,15 +15,16 @@
 
 package com.splicemachine.derby.stream.function;
 
+import com.clearspring.analytics.util.Lists;
 import com.splicemachine.access.HConfiguration;
 import com.splicemachine.access.api.PartitionFactory;
 import com.splicemachine.derby.impl.SpliceSpark;
 import com.splicemachine.si.impl.driver.SIDriver;
-import com.splicemachine.storage.ClientPartition;
 import com.splicemachine.storage.Partition;
 import com.splicemachine.storage.SkeletonHBaseClientPartition;
 import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.client.HTable;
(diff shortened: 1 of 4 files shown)


jyuanca:
The existing ITs can make sure the PR does not introduce regression.

hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
```diff
@@ -53,23 +59,40 @@ public BulkImportFunction(String bulkImportDirectory) {
...
+            for (int i = 1; i < partitionList.size(); ++i) {
+                Path sourceDir = new Path(partitionList.get(i).getFilePath());
+                if (fs.exists(sourceDir)) {
+                    FileStatus[] statuses = fs.listStatus(sourceDir);
+                    for (FileStatus status : statuses) {
+                        Path filePath = status.getPath();
+                        Path destPath = new Path(path, filePath.getName());
+                        fs.rename(filePath, destPath);
+                        if (LOG.isDebugEnabled()) {
+                            SpliceLogUtils.debug(LOG, "Move file %s to ", filePath.toString(), destPath.toString());
```
yxia92: It seems an "%s" is missing in the debug message.

hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
```diff
@@ -53,23 +59,40 @@ public BulkImportFunction(String bulkImportDirectory) {
...
-        PartitionFactory tableFactory= SIDriver.driver().getTableFactory();
-        try(Partition partition=tableFactory.getTable(Long.toString(conglomerateId))){
-            Path path = new Path(importPartition.getFilePath()).getParent();
-            if (fs.exists(path)) {
-                loader.doBulkLoad(path,(HTable) ((SkeletonHBaseClientPartition)partition).unwrapDelegate());
-                fs.delete(path, true);
-            } else {
-                LOG.warn("Path doesn't exist, nothing to load into this partition? " + path);
+        for (Long conglomId : partitionMap.keySet()) {
+            PartitionFactory tableFactory= SIDriver.driver().getTableFactory();
```
yxia92: I think this line can be moved out from the loop.

hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
```diff
@@ -83,4 +106,18 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
+    private void init(Iterator<BulkImportPartition> importPartitions) throws Exception {
+        partitionMap = new HashMap<>();
+        while (importPartitions.hasNext()) {
+            BulkImportPartition partition = importPartitions.next();
+            Long conglom = partition.getConglomerateId();
+            List<BulkImportPartition> partitionList = partitionMap.get(conglom);
+            if (partitionList == null) {
+                partitionList = Lists.newArrayList();
+                partitionMap.put(conglom, partitionList);
+            }
```
msirek: The code makes the assumption that there is only one unique conglomerate possible in the importPartitions list, because partitionMap.put is only called once.  Could index conglomerates also be in the list?  Either we should add a sanity check to make sure each call to getConglomerateId() returns the same conglomerate (if we only expect there to ever be just one), or we should make a call to partitionMap.put for each unique conglomerate.

jyuanca: This code add one entry to partitionMap for each conglomerate, including base table and indexes.

msirek: I misread it, didn't see the call to partitionMap.get.  Looks good.

PR #2272, opened 2018-09-22
DB-7433 Run Vacuum in the background and tuning JVM params

wb14123:
This PR includes many modifications. Please review it carefully and better to run it on your own machine or laptop. After this fix the IT should be able to run within 10G memory. The successfully build on Jenkins: http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/.

The modifications:

1. Start a Vacuum progress that running in the background.
2. Because some test case is conflict with Vacuum running in background, so move these test cases into one category and disable them for now. This expected failed IT are `KillOperationIT`, `VacuumIT`, the unexpected failed IT are `TempTableHBaseIT` (cannot delete temp table when cleanup), `IndexIT`, `AddColumnWithDefaultIT`.
3. Init JVM with less memory using `-Xms`.
4. Make JVM less aggressive in memory allocation with `-XX:MinHeapFreeRatio=10` and `-XX:MaxHeapFreeRatio=50`.
5. Reduce JVM max memory for Splice, Zookeeper and Kafka.
6. Use G1 gc.

**This is not included in this PR and need to config on Jenkins or your own machine**: 

```
MAVEN_OPTS="-Xms64m -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=50 -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=YYYY-MM-dd|HH:mm:ss,SSS"
```


platform_it/pom.xml
```diff
@@ -1233,6 +1271,9 @@
                             <mainClass>com.splicemachine.test.SpliceTestYarnPlatform</mainClass>
                             <arguments>
                                 <!-- Root of the serve classpath. Yarn config file (yarn-site.xml) will be written here. -->
+                                <!--<argument>-agentpath:/opt/jprofiler/bin/linux-x64/libjprofilerti
```
MurrayBrown: i do not understand why we're checking in additional commented out parameters.

platform_it/pom.xml
```diff
@@ -1274,6 +1315,10 @@
                                 <argument>-XX:+FlightRecorder</argument> -->
                                 <argument>-Djava.security.krb5.conf=${project.build.directory}/krb5.conf</argument>
                                 <argument>-Dlog4j.configuration=${log4j.config.uri}</argument>
+                                <!--<argument>-agentpath:/opt/jprofiler/bin/linux-x64/libjprofilerti.so=port=8851</argument>-->
```
MurrayBrown: same here: if the intent is to say, e.g. : to run the profiler uncomment this line - then i would accept the comment and the line, but it just fills the file with extra content?

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -137,7 +137,7 @@ public static Configuration create(String hbaseRootDirUri,
         config.setInt("hbase.master.info.port", masterInfoPort);
         config.setInt("hbase.regionserver.port", regionServerPort);
         config.setInt("hbase.regionserver.info.port", regionServerInfoPort);
-        config.setInt("hbase.master.jmx.port", HConfiguration.DEFAULT_JMX_BIND_PORT); // this is set because the HBase master and regionserver are running on the same machine and in the same JVM
+        // config.setInt("hbase.master.jmx.port", HConfiguration.DEFAULT_JMX_BIND_PORT); // this is set because the HBase master and regionserver are running on the same machine and in the same JVM
```
MurrayBrown: why remove the jmx port? not used anymore?

wb14123:
@MurrayBrown Pushed a new commit to cleanup the comments.

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/actions/TempTableHBaseIT.java
```diff
@@ -36,6 +34,7 @@
  * @author Scott Fines
  *         Date: 1/19/16
  */
+@Category(NoVacuumTest.class)
```
wtang16: Do you mean TempTableBaseIT does not work with Vacuum?

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -257,7 +260,7 @@ public static Configuration create(String hbaseRootDirUri,
         //
 
         config.setLong("splice.ddl.drainingWait.maximum", SECONDS.toMillis(15)); // wait 15 seconds before bailing on bad ddl statements
-        config.setLong("splice.ddl.maxWaitSeconds",120000);
+        config.setLong("splice.ddl.maxWaitSeconds",240000);
```
wtang16: Is 4min really necessary for DDL operations?

wb14123:
Added a commit to re-enable NoVacuum tests.

wb14123:
@wtang16 Fixed the issues you mentioned.

wb14123:
BTW, the newest test result: http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/default/1235/console

wtang16:
We may need some further discussion though. There are some concerns regarding the introduction of the NoVacuumTest, asynchronous vs. synchronous execution of vacuum. The memory and GC settings look fine.

splice_machine/src/test/java/com/splicemachine/derby/impl/load/CaseSensitiveImportIT.java
```diff
@@ -43,11 +44,7 @@
     private TestConnection conn;
 
     @BeforeClass
-    public static void beforeClass() throws Exception {
-        cleanSchema(SCHEMA_NAME, methodWatcher);
```
OlegMazurov: Has schema cleaning before execution been moved somewhere else?
Should it be done for any test that creates a new schema?
cleanSchema is a private method not used anywhere else. Should be removed.

wb14123: Yes, every test case will clean the created schema within the ClassRule of SpliceSchemaWatcher. I'll remove the unused `cleanSchema`.

wb14123:
When run Vacuum synchronously, it takes more than 4 hours to run the tests (http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/1238/). Which is more than 1 hour longer than run Vacuum async. (http://206.225.8.98:8080/job/spliceengine-branch/job/spliceengine-branch/1236/)

dgomezferro:
I agree async Vacuum is the way to go, but I'd like it to be run in a separate thread rather than a separate process as Oleg said.

I'd remove the NoVacuum category and move any test that doesn't work with concurrent vacuum to the SerialTests and only start the concurrent Vacuum for non-serial ITs.

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -186,15 +186,18 @@ public static Configuration create(String hbaseRootDirUri,
         //
         // Threads, timeouts
         //
-        config.setLong("hbase.rpc.timeout", MINUTES.toMillis(2));
+        config.setLong("hbase.rpc.timeout", MINUTES.toMillis(5));
```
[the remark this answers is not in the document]
wb14123: Foreign key related IT will timeout if set the timeout to 2 minutes.

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -186,15 +186,18 @@ public static Configuration create(String hbaseRootDirUri,
...
         config.setInt("hbase.client.max.perserver.tasks",50);
         config.setInt("hbase.client.ipc.pool.size",10);
         config.setInt("hbase.rowlock.wait.duration",0);
 
         config.setLong("hbase.client.scanner.timeout.period", MINUTES.toMillis(2)); // hbase.regionserver.lease.period is deprecated
         config.setLong("hbase.client.operation.timeout", MINUTES.toMillis(2));
-        config.setLong("hbase.regionserver.handler.count", 100);
-        config.setLong("hbase.regionserver.metahandler.count", 100);
+        config.setLong("hbase.regionserver.handler.count", 10);
```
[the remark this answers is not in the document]
wb14123: This will reduce the threads, thus reduce the memory usage. Since by default, each thread will use 1M memory. After the modification, the threads reduced from more than 1000 to 500, which saves about 500M memory.

wb14123:
@dgomezferro I'm looking up a way to run the Vacuum in a single thread. However, JUnit don't have a trigger like BeforeSuite. So it seems tricky to run in a single thread. Still finding the way...

OlegMazurov:
I implemented async vacuuming in a separate thread.
System property splice.cleanupMode is used in platform_it/pom.xml to control how vacuuming is to be done. For SerialTests there is not vacuuming, and for parallel tests vacuuming is run asynchronously in a separate thread.
All tests previously marked as NoVacuumTest have been moved to SerialTests.

PR #2274, opened 2018-09-25
DB-7179: full backup/restore for tables

jyuanca:
Please refer to [DB-7179](https://splicemachine.atlassian.net/browse/DB-7179) for details.

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
+++ b/hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
@@ -16,11 +16,8 @@
 package com.splicemachine.si.impl.server;
 
 import com.splicemachine.si.api.txn.TxnView;
-import com.splicemachine.si.impl.server.CompactionContext;
-import com.splicemachine.si.impl.server.SICompactionState;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.regionserver.InternalScanner;
-import org.apache.hadoop.hbase.regionserver.ScannerContext;
 import org.apache.log4j.Logger;
 
 import java.io.IOException;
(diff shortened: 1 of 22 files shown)


splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
```diff
@@ -412,4 +463,73 @@ public static void SYSCS_CANCEL_BACKUP(ResultSet[] resultSets) throws StandardEx
...
+                                           boolean validate,
+                                           ResultSet[] resultSets) throws StandardException, SQLException {
+        try {
+            destSchema = EngineUtils.validateSchema(destSchema);
+            destTable = EngineUtils.validateTable(destTable);
+            sourceSchema = EngineUtils.validateSchema(sourceSchema);
+            sourceTable = EngineUtils.validateTable(sourceTable);
+            validateTable(destSchema, destTable);
+            BackupManager backupManager = EngineDriver.driver().manager().getBackupManager();
+            backupManager.restoreTable(destSchema, destTable, sourceSchema, sourceTable, directory, backupId, validate);
```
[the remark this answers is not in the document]
yxia92: I think this comment can be dismissed. Jun shows an example in the review for ee branch where if restore fails with validation, warning can be returned to the user console. And if restore is successful, we will just see "Statement executed". I think that is fine.

hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
```diff
@@ -84,7 +80,8 @@ public boolean next(List<Cell> list) throws IOException{
             compactionState.mutate(entry.cells, txns, list, purgeDeletedRows);
             if (!more) {
                 timer.cancel();
-                context.close();
+                if (context != null)
```
dgomezferro: Why are these guards needed?

jyuanca: @dgomezferro Because I did not create a compaction context for table backup.

dgomezferro: Can you create one? I think it's cleaner to have it

PR #2277, opened 2018-09-26
DB-7409 Create a new standalone installer

bklo94:
Also fixed a bug in platform_it by generating a test-jar, since some required java test classes in platform_it were needed for standalone to function.


OlegMazurov:
Could you please describe in the JIRA what is broken and how you are fixing it?
Seeing a dependency on spark-mllib in assemply/pom.xml is incomprehensible.
Also, adding test-jar to platform_it (which is not deployed, by the way) contradicts some work-in-progress (SPLICE-2251). We need to coordinate.

MurrayBrown:
I've written a better description into the DB-7409 ticket, in case that helps.

MurrayBrown:
Brandon - you may want to abandon this PR, do a rebase of your edits, and then submit a new PR. I appreciate that the branch is 'good', and has your work; but it is every hard to read given the many merges?

PR #2280, opened 2018-09-27
SPLICE-2249 Allow schema to be passed in JDBC connection string

wb14123:


db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -477,7 +481,9 @@ protected SchemaDescriptor initDefaultSchemaDescriptor() throws StandardExceptio
         */
         if(cachedInitialDefaultSchemaDescr==null){
             DataDictionary dd=getDataDictionary();
-            String authorizationId=getSessionUserId();
+            if (defaultSchema != null) {
+                return dd.getSchemaDescriptor(defaultSchema, getTransactionCompile(), true);
```
yxia92: I think we need to set cachedInitialDefaultSchemaDescr to the returned schema descriptor before returning from the function.

splice_machine/src/test/java/com/splicemachine/jdbc/JdbcApiIT.java
```diff
@@ -127,4 +112,31 @@ public void testTimeoutControl() throws Exception {
         }
     }
 
+    @Test
+    public void testUrlWithSchema() throws Exception {
+        String url = "jdbc:splice://localhost:1527/splicedb;create=true;user=%s;password=%s;schema="
+                + CLASS_NAME;
```
yxia92: I think the attribute "create=true" should not be put in the test case, it is used the very first time to create the database.

splice_machine/src/test/java/com/splicemachine/jdbc/JdbcApiIT.java
```diff
@@ -127,4 +112,31 @@ public void testTimeoutControl() throws Exception {
...
+            try (Statement statement = connection.createStatement()) {
+                statement.executeQuery("select * from a");
+            }
+        }
+    }
+
+    @Test
+    public void testUrlWithNonExistSchema() throws Exception {
+        String url = "jdbc:splice://localhost:1527/splicedb;create=true;user=%s;password=%s;schema="
+                + "nonexist";
```
yxia92: remove "create=true" here too.

wb14123:
@yxia92 Fixed the issues.

PR #2286, opened 2018-09-28
DB-7312: fx Hive integration(2.5)

jyuanca:


MurrayBrown:
Suggestion: lets not put this into 2.5.  The prospect here is using Splice 2.7, and this feature is one more thing that would be better targetted at the updated build infrastructure on Master/2.7.

wtang16:
We can close the port to 2.5.

PR #2288, opened 2018-09-28
DB-7312: fix Hive integration

jyuanca:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreateTableConstantOperation.java
```diff
@@ -320,7 +320,7 @@ protected void createTable(Activation activation) throws StandardException {
         String createAsVersion2String = null;
         if (lcc != null) {
             createAsVersion2String =
-            PropertyUtil.getCachedDatabaseProperty(lcc.getTransactionCompile(),
+            PropertyUtil.getCachedDatabaseProperty(lcc, lcc.getTransactionCompile(),
```
msirek: Since lcc is now passed to getCachedDatabaseProperty, perhaps parameter 2 could be removed because getTransactionCompile() could now be called inside the method body.

PR #2292, opened 2018-10-01
DBAAS-1817 Add Anonymous Login Creds for S3 

bklo94:
Uses the AnonymousAWSCredentials() to create credentials if none is given. Allows S3A to download from public buckets.


MurrayBrown:
I had asked Brandon to look into this, hoping it was 'just the config'. He has found the answer that this needs to be changed in our code, and that there is no config change that *can* make this work.

My thinking is: if we allow for non-credentialed s3 access, then we can have e.g. our standalone and our on-premise products both read from s3 *out-of-the-box*, so examples on our website and in our documents will 'just work'.

MurrayBrown:
perhaps with this change, we can re-enable this test:
spliceengine/hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/S3IT.java

PR #2296, opened 2018-10-03
DB-7472: fix a race condition in TableDescriptor

jyuanca:


jyuanca:
If "alter table" changes the heap conglomerate id, it should invalidate the cache, instead of changing it in TableDescriptor.

"alter table add primary key" for empty table changes conglomerate id for base table and invalidate cache. "add/drop primary key" for non-empty table is probably not a supported feature yet.

I don't know any other alter table command that changes base table conglomerate. If they do, they should invalidate the cache.

PR #2303, opened 2018-10-09
DB-7501 Register dropped conglomerates in HBase table

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -130,8 +158,8 @@ public void vacuumDatabase(long oldestActiveTransaction) throws SQLException{
                             requiresDroppedId = true;
                         }
                     }
-                    if (!ignoreDroppedId && table.getDroppedTransactionId() != null) {
-                        TxnView txn = SIDriver.driver().getTxnSupplier().getTransaction(Long.parseLong(table.getDroppedTransactionId()));
+                    if (!ignoreDroppedId && droppedConglomerates.containsKey(tableConglom)) {
```
yxia92: It seems that for tables that are dropped between 1838 and 1840, since they are not registered in  the table DROPPED_CONGLOMERATES, they won't get a chance to be vacuumed. 
maybe the code to check table.getDroppedTransactionId() should still be retained?

splice_si_api/src/main/java/com/splicemachine/si/constants/SIConstants.java
```diff
@@ -50,6 +50,9 @@
     public static final long SUBTRANSANCTION_ID_MASK = 0xFFl;
     public static final long TRANSANCTION_ID_MASK= 0xFFFFFFFFFFFFFF00l;
 
+    // Dropped conglomerates
+    public static final String DROPPED_CONGLOMERATES_TABLE_NAME = "DROPPED_CONGLOMERATES";
```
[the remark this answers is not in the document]
dgomezferro: Moved it to HBaseConfiguration

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -64,6 +64,7 @@ else if(v1==v2)
         scripts.put(new Splice_DD_Version(sdd,2,8,1), new UpgradeScriptForModifySchemaPermissionAndDefaultRole(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,2,8,0, 1812), new UpgradeScriptToCleanSysRoutinePerms(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,2,8,0, 1817), new UpgradeScriptForSysTokens(sdd,tc));
+        scripts.put(new Splice_DD_Version(sdd,2,8,0, 1840), new UpgradeScriptForDroppedConglomerates(sdd,tc));
```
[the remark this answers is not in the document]
dgomezferro: Updated it

splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
dgomezferro: Added code to handle that case, good catch! I also added an IT on VaccumIt specifically for this case.

Thanks for the review!

splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -130,8 +159,16 @@ public void vacuumDatabase(long oldestActiveTransaction) throws SQLException{
                             requiresDroppedId = true;
                         }
                     }
```
jyuanca: What is the differences between transactionId and droppedTransactionId? I cannot find usage of droppedTransactionId.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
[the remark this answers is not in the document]
yxia92: Thanks Daniel for addressing all my comments. I think the build number for the upgrade script should be bumped up to 1842 now that we are in the weekly build for 1842 and this PR is likely to merge in this week.

dgomezferro:
Thanks @yxia92, I bumped the script version

splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
dgomezferro: transactionId is the Id of the transaction that created the table, droppedTransactionId is the id of the txn that dropped it.

PR #2306, opened 2018-10-09
DB-7472: fix a race condition in TableDescriptor

jyuanca:
refer to DB-7472[https://splicemachine.atlassian.net/browse/DB-7472] for details


jyuanca:
Truncate table does not modify table descriptor. It creates a new one.

PR #2309, opened 2018-10-12
SPLICE-2257 Correctly parse timestamps in the 1-hour gap when switching over to daylight savings.

msirek:
See [SPLICE-2257](https://splice.atlassian.net/browse/SPLICE-2257) for details.


wtang16:
Need this in 2.7 as well.

PR #2310, opened 2018-10-12
DB-7545 SynchronousWriteControl creates contention

OlegMazurov:
Use java.util.concurrent.atomic.LongAdder to update individual counters. Don't use WriteStatus as simultaneous updates of two counters don't have to be "transactional": f.e. it's OK to read dependentWriteThreads == 0 and dependentWriteCount != 0.


OlegMazurov:
I'm proposing an alternative implementation of AtomicSpliceWriteControl, which is based on one AtomicLong for thread/count updates and compareAndSet(). It provides tight control and higher throughput when there is more contention (the max number of threads is much lower than the number of worker threads). See DB-7545 for benchmark results.

jyuanca:
Will you port to branch-2.5 and 2.7?

OlegMazurov:
@jyuanca 2.5/2.7 PRs are ready

PR #2317, opened 2018-10-16
DB-7376 Fix length of table type in metadata getTableTypes

changli6:


db-testing/src/test/java/com/splicemachine/dbTesting/functionTests/tests/jdbcapi/DatabaseMetaDataTest.java
```diff
@@ -2077,7 +2077,7 @@ public void testTableTypes() throws SQLException
         
         JDBC.assertFullResultSet(rs, new String[][]
           {
-            {"SYNONYM"},{"SYSTEM TABLE"},{"TABLE"},{"VIEW"},               
+            {"SYNONYM"},{"SYSTEM TABLE"},{"TABLE"},{"VIEW"},{"EXTERNAL TABLE"}
```
dgomezferro: Do we run this test as part of our IT run?

changli6: It's not. I questioned on PD channel how to run these test cases.

PR #2323, opened 2018-10-16
SPLICE-201 Use shared thread local Calendar in SQLTimestamp for performance.

msirek:
See [SPLICE-201](https://splice.atlassian.net/browse/SPLICE-201) for details.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
@@ -86,6 +86,11 @@ public final class SQLTimestamp extends DataType
 						implements DateTimeDataValue
 {
 
+    protected static final ThreadLocal<Calendar> localCal = new ThreadLocal<Calendar>() {
+        protected Calendar initialValue() {
+            return Calendar.getInstance();
+        }
+    };
     static final int MAX_FRACTION_DIGITS = 9; // Only nanosecond resolution on conversion to/from strings
     static final int FRACTION_TO_NANO = 1; // 10**(9 - MAX_FRACTION_DIGITS)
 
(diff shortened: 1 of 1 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
```diff
@@ -1093,8 +1098,8 @@ private void setNumericTimestamp(Timestamp value, Calendar cal) throws StandardE
 		if (value != null)
 		{
             if( cal == null) {
-				cal = Calendar.getInstance();
-			}
+                cal = localCal.get();
```
OlegMazurov: There is already a thread local created for that same purpose which is used in other locations in SQLTimestamp.
```suggestion
                cal = SQLDate.GREGORIAN_CALENDAR.get();
```

msirek: @OlegMazurov That is a good idea.  Though, Calendar.getInstance may end up instantiating a BuddhistCalendar or JapaneseImperialCalendar depending on the system locale, so always using GregorianCalendar would represent a change, which systems in Japan or Thailand would notice (a given year in those other calendars would be a different year in the Gregorian calendar, yielding different timestamps than those originally loaded in a table).  However, the fact that SQLDate uses a Gregorian calendar and SQLTimestamp sometimes doesn't probably means use of these other locales is not really supported in Splice.  I'm OK with always using GregorianCalendar, though maybe the Splice documentation should be updated saying locales that use the Buddhist calendar or Japanese imperial calendar are not supported.  Comments?

OlegMazurov: There are 11 uses of SQLDate.GREGORIAN_CALENDAR in SQLTimestamp.java introduced with SPLICE-937 on 9/15/16 and only one Calendar.getInstance() added earlier on 1/20/16.
It's likely the SPLICE-937 update just missed this usage.
I don't think setNumericTimestamp() is somehow special to use a different calendar.

PR #2325, opened 2018-10-17
DB-7543 restore fix of DB-7402 and fix NPE for expression with SQL fu…

yxia92:
…nction

This PR restores the code change of DB-7402. The newly added code change is just SQLToJavaValueNode.setHashableJoinColumnReference() and JoinWithFunctionIT.testJoinExpressionWithSQLFunctionWithOneColumnReference() which fixes and tests the NPE issue in DB-7543.
Details is in [DB-7543](https://splicemachine.atlassian.net/browse/DB-7543)


dgomezferro:
@yxia92 just a minor comment, since this was restoring a previous commit with a fix on top, I think it would have been better to have two distinct commits, the original plus the one with the fix.

PR #2330, opened 2018-10-17
DB-7541 Add export_binary operation and use it for internalDf

dgomezferro:


db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLTimestamp.java
```diff
@@ -1606,7 +1606,7 @@ public void write(UnsafeRowWriter unsafeRowWriter, int ordinal) {
 		        Platform.putInt(holder.buffer, holder.cursor + 4, encodedTime);
 		        Platform.putInt(holder.buffer, holder.cursor + 8, nanos);
 		        unsafeRowWriter.setOffsetAndSize(ordinal, 12);
-				holder.cursor = 12;
```
MurrayBrown: How does this change relate to this bug fix? I'm not hating it, but it seems somewhat out of place to me?

dgomezferro: It's not related now, at some point I was writing ExecRows into UnsafeRows before writing the Parquet file in ControlDataset, and hit this bug. Then I moved to using the RowEncoder mechanism and left this fix in.

PR #2338, opened 2018-10-17
DB-7563 Add Xenial support and fix parcel.json replace bug

bklo94:
A bug was fixed where the replace token was not detecting the new modified values and therefore did not replace the token with the new value.


chuckwondo:
Hey @MurrayBrown I suggested this fix to @bklo94 for the short term. He has inherited what appears to me to be a frankly awful build solution, so I suggested this fix for now, but that we will need to revisit this in the larger scope of constructing a better overall build approach.

PR #2340, opened 2018-10-17
DB-7563 Add Xenial support and fix parcel.json replace bug (2.5)

bklo94:
A bug was fixed where the replace token was not detecting the new modified values and therefore did not replace the token with the new value.


bklo94:
No need for 2.5 PR feature.

PR #2342, opened 2018-10-18
DB-7566 Make SplicemachineContext.internalConnection a singleton

dgomezferro:


splice_spark/src/main/spark2.1/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -568,5 +577,16 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
     new StructType(columns.map(name => fieldMap(name)))
   }
 
+  private[this] def withoutAutocommit(op: => Unit) = {
+    val autocommit = getConnection().getAutoCommit
+    if (autocommit) {
+      getConnection().setAutoCommit(false)
+      op
+      getConnection().setAutoCommit(autocommit)
+      getConnection().commit()
```
msirek: What if there is an error in the connection in between the call to get getConnection().setAutoCommit(false) and the call to getConnection().commit() ?  If we were to make getConnection() establish a new connection in this scenario, then we'd be using two separate connections here.  So, if getConnection() were modified with this behavior, perhaps these method calls should use the same connection object (hold the connection in a variable) so that if the connection goes bad, the commit errors out instead of succeeding and making it look like the transaction was committed, when in reality it did not.

dgomezferro:
Opened new PR: https://github.com/splicemachine/spliceengine/pull/2368

PR #2352, opened 2018-10-19
DB-7501 Register dropped conglomerates in HBase table 

dgomezferro:


dgomezferro:
The fix for the the previous merge failure is on the -ee branch: https://github.com/splicemachine/spliceengine-ee/pull/294

This PR is the same that got merged earlier.

The failure was due to the added system table being missing from the system tables list used by Backup, which lead to the Restore test deleting the DROPPED_CONGLOMERATES table.

PR #2353, opened 2018-10-19
DB-6726 Add export API to Spark adapter

dgomezferro:


jyuanca:
This adds export APIs for spark adapter. However, export is still executed by hbase user. Can you explain how it impersonate the user that executes export?

dgomezferro:
@jyuanca this export runs inside the Spark adapter process, in the mini-splice DB it starts, which is running as the user that submits the spark app, not hbase

PR #2357, opened 2018-10-19
DB-7541 Add export_binary operation and use it for internalDf

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
```diff
@@ -44,6 +52,18 @@
...
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.collections.IteratorUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder;
+import org.apache.spark.sql.catalyst.encoders.RowEncoder;
+import org.apache.spark.sql.execution.datasources.parquet.ParquetWriteSupport;
```
OlegMazurov: You can remove now unused import

PR #2382, opened 2018-11-01
SPLICE-2250 Convert string to date/time to match frequency items type  in selectivity estimate logic

changli6:


db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -228,10 +233,41 @@ public long selectivity(DataValueDescriptor element) {
...
+        // When look up predicate value in quantilesSketch, the element's data type
+        // must match with the data type stored the ItemsSketch.hashMap
+        // Otherwise the same value of element in ItemsSketch.hashMap can't be queried out
+        // This handle the situation of string predict compare with DATE/TIME/TIMESTAMP
+        DataValueDescriptor lookUpElement = element;
+        int typeFormatId = dvd.getTypeFormatId();
+        int eTypeFormatId = element.getTypeFormatId();
+        boolean isConverted = false;
+        if (typeFormatId != eTypeFormatId) {
+            if (eTypeFormatId == StoredFormatIds.SQL_CHAR_ID) {
```
yxia92: Does VarChar have the same problem?

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -228,10 +233,41 @@ public long selectivity(DataValueDescriptor element) {
...
+                        case StoredFormatIds.SQL_TIMESTAMP_ID:
+                            lookUpElement = new SQLTimestamp(element.getTimestamp(new GregorianCalendar()));
+                    }
+                    isConverted = true;
+                } catch (StandardException e) {
+                    LOG.warn("Data type conversion failure when looking up frequencies.ItemsSketch", e);
+                }
+            }
+        }
+        if (typeFormatId == eTypeFormatId || isConverted) {
```
yxia92: The check of "typeFormatId == eTypeFormatId" may be too strict. Would this affect the number data types? for example, between INT and BIGINT. 
Prefer to restrict the fix to just Char and VarChar type. Something like 
if (eTypeformatId !=StoredFormatIds.SQL_CHAR_ID && eTypeFormatId !=StoredFormatIds.SQL_VARCHAR_ID || isConverted)

msirek: I agree with Yi.  I did some tests, and estimates when comparing dates with timestamps is currently correct, so no need to bypass the current code.  Maybe just reverse the order of the if statements from:

if (typeFormatId != eTypeFormatId) {
            if (eTypeFormatId == StoredFormatIds.SQL_CHAR_ID) {

... to:

if (eTypeFormatId == StoredFormatIds.SQL_CHAR_ID) {
            if (typeFormatId != eTypeFormatId) {

For the Char test, you could also do: "if (element instanceof SQLChar)" which would handle VARCHAR to since SQLVarchar is a subclass of SQLChar.

changli6:
Thanks @yxia92 @msirek  comments.

The root cause is whether the predicate could be found in hashMap. In another word, we need to identify whether the implementations of data type's hashCode function are same between predicate and sketchItems. 

For hashCode functions:
SQLVarchar hashCode function inherited from SQLChar. They can be treated the same.
SQLInteger, SQLSmallint, and SQLTinyint are same, SQLLongint has the same hashcode result in SQLInteger / SQLTinyint / SQLSmalint range. But if SQLongint is large (over 4 bytes), it may have the same hashcode with a small value in SQLInteger / SQLSmallint / SQLTinyint. SQLLongint should be excluded from hashcode comparing with other 3 types. They are classified as integer family.
SQLReal, SQLDouble, and SQLDecimal have similar hashCode functions. They are classified as float family.
SQLDate, SQLTime, and SQLTimestamp are different from each other, we need to treat them separately.

We should strict the different data types frequenciesSketch looking up in the same family or string to time conversion. Otherwise, the predicate object may match with a coincident key in the hashmap and get a wrong value.

changli6:
The fix focuses on hash code function between predicate and column:
1. If they are in same type (same hash code function), then look up in hashMap
2. If they are not in same type but have same hash code function, address them as same family, then look up in hashMap 
3. If they are not in same type and not have same hash code function, and column type precedence is greater than predicate type precedence, then convert the predicate to column data type and look up in hashMap
4. Otherwise, compare predicate with each element in hashMap iteratively

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -228,9 +234,85 @@ public long selectivity(DataValueDescriptor element) {
...
+                    case StoredFormatIds.SQL_TIMESTAMP_ID:
+                        lookUpElement = new SQLTimestamp(element.getTimestamp(SQLDate.GREGORIAN_CALENDAR.get()));
+                        break;
+                    case StoredFormatIds.SQL_DATE_ID:
+                        lookUpElement = new SQLDate(element.getDate(SQLDate.GREGORIAN_CALENDAR.get()));
+                        break;
+                    case StoredFormatIds.SQL_DOUBLE_ID:
+                    case StoredFormatIds.SQL_REAL_ID:
+                    case StoredFormatIds.SQL_DECIMAL_ID:
+                        lookUpElement = new SQLDouble(element.getDouble()); // These types are in the same family
```
yxia92: Could you add default clause here to cover the other data types?

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
```diff
@@ -228,9 +234,85 @@ public long selectivity(DataValueDescriptor element) {
         // Use Null Data
         if (element == null || element.isNull())
             return nullCount;
+
         // Frequent Items Sketch?
-        long count = frequenciesSketch.getEstimate(element);
-        if (count>0)
+
+        // When look up predicate value in quantilesSketch, the hashCodes of two different data types having different values may be same by coincident
```
yxia92: This should be the frequent item sketch instead of the quantileSketch.

PR #2385, opened 2018-11-01
DB-7186 Add cdh5.14.2 support (2.7)

bklo94:


assembly/cdh5.14.2/src/main/resources/meta/splicemachine_env.sh
```diff
@@ -0,0 +1,42 @@
...
+    exit 1
+}
+
+PREPENDSTRING=`echo ${MYLIBDIR}/*.jar ${PARCELS_ROOT}/SPARK2/lib/spark2/jars/*.jar | sed 's/ /:/g'`
+echo "prepending $PREPENDSTRING to HBASE_CLASSPATH_PREFIX"
+if [ -z $HBASE_CLASSPATH_PREFIX ] ; then
+    export HBASE_CLASSPATH_PREFIX="${PREPENDSTRING}"
+else
+    export HBASE_CLASSPATH_PREFIX="${PREPENDSTRING}:${HBASE_CLASSPATH_PREFIX}"
+fi
```
chuckwondo: You can simplify the entire `if/else` block above to a single line (i.e., no need for the if/else at all):
```bash
export HBASE_CLASSPATH_PREFIX="${PREPENDSTRING}${HBASE_CLASSPATH_PREFIX:+:$HBASE_CLASSPATH_PREFIX}"
```
To explain, the expression `${HBASE_CLASSPATH_PREFIX:+:$HBASE_CLASSPATH_PREFIX}` will either resolve to an empty string if `HBASE_CLASSPATH_PREFIX` is either undefined or empty, or to the evaluation of everything following the `:+`, which is `:$HBASE_CLASSPATH_PREFIX`.

In other words, if `HBASE_CLASSPATH_PREFIX` is empty or undefined, the entire expression (the right side of the assignment) evaluates only to the value of `PREPENDSTRING`. Otherwise it evaluates to the value of `PREPENDSTRING`, followed by a colon (`:`), followed by the value of `HBASE_CLASSPATH_PREFIX`.

For reference, see https://www.tldp.org/LDP/abs/html/parameter-substitution.html

assembly/cdh5.14.2/src/main/resources/meta/splicemachine_env.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo "Could not find splice parcel lib dir, exiting" >&2
+    exit 1
+}
+
+[ -d ${PARCELS_ROOT}/SPARK2 ] || {
+    echo "Could not find SPARK2 parcel lib dir, exiting" >&2
+    exit 1
+}
+
+PREPENDSTRING=`echo ${MYLIBDIR}/*.jar ${PARCELS_ROOT}/SPARK2/lib/spark2/jars/*.jar | sed 's/ /:/g'`
```
chuckwondo: Please use `$(...)` instead of `` `...` ``.

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+  This function will scan the specified directory, identify any parcel files
+  in it, and then build a manifest from those files. Certain metadata will be
+  extracted from the parcel and copied into the manifest.
+  @param path: The path of the directory to scan for parcels
+  @param timestamp: Unix timestamp to place in manifest.json
+  @return: the manifest.json as a string
+  """
+  manifest = {}
+  manifest['lastUpdated'] = int(timestamp * 1000)
+  manifest['parcels'] = []
```
chuckwondo: You can initialize `manifest` more succinctly like so:
```python
manifest = {
    'lastUpdated': int(timestamp * 1000),
    'parcels': []
}
```

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+
+  files = os.listdir(path)
```
chuckwondo: You can initialize `files` like so:
```python
files = (f for f in os.listdir(path) if f.endswith('.parcel'))
```
Doing so then allows you to eliminate the initial `if` block in the loop below.

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+  for f in files:
+    if not f.endswith('.parcel'):
+      continue
```
chuckwondo: Eliminate the `if` block above if you followed my advise on initializing `files` above.

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+
+    print("Found parcel %s" % (f,))
+    entry = {}
+    entry['parcelName'] = f
```
chuckwondo: You can simply initialize `entry` like so: `entry = {'parcelName': f}`

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+
+    with open(fullpath, 'rb') as fp:
+      entry['hash'] = hashlib.sha1(fp.read()).hexdigest()
+
+    with tarfile.open(fullpath, 'r') as tar:
+      try:
+        json_member = tar.getmember(os.path.join(_get_parcel_dirname(f),
+                                    'meta', 'parcel.json'))
+      except KeyError:
+        print("Parcel does not contain parcel.json")
```
chuckwondo: I suggest making the error message above more informative by including the name of the parcel file:
```python
print("Parcel does not contain parcel.json: %s" % fullpath)
```

assembly/cdh5.14.2/src/main/resources/python/make_manifest.py
```diff
@@ -0,0 +1,113 @@
...
+        continue
+      try:
+        parcel = json.loads(tar.extractfile(json_member).read().decode(encoding='UTF-8'))
+      except:
+        print("Failed to parse parcel.json")
```
chuckwondo:
```python
print("Failed to parse parcel.json in parcel file %s" % fullpath)
```

assembly/cdh5.14.2/src/main/resources/scripts/copy-yarn-site.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
chuckwondo: Please use `#!/usr/bin/env bash`

assembly/cdh5.14.2/src/main/resources/scripts/flatten.sh
```diff
@@ -0,0 +1,259 @@
...
+	;;
+	u)
+		ssh_user=$OPTARG
+	;;
+	\?)
+	show_help
+	exit 1
+	;;
+	esac
+done
```
chuckwondo: Please make your indentation in the block above consistent.

pom.xml
```diff
@@ -1253,6 +1253,38 @@
...
+                <envClassifier>cdh5.14.2</envClassifier>
+                <envHbase>hbase1.2.0</envHbase>
+                <hadoop.version>2.6.0-cdh5.14.2</hadoop.version>
+                <hbase.version>1.2.0-cdh5.14.2</hbase.version>
+                <hive.version>1.1.0-cdh5.14.2</hive.version>
+                <spark.major.version>spark2.2</spark.major.version>
+                <zookeeper.version>3.4.5-cdh5.14.2</zookeeper.version>
+                <spark.version>2.2.0.cloudera2</spark.version>
+                <spark-assembly-id>spark-assembly-hadoop2.6.0-cdh5.14.2</spark-assembly-id>
+                <kafka.version>0.10.2-kafka-2.2.0</kafka.version>
```
OlegMazurov: 0.10.0-kafka-2.1.0

bklo94: For the branch-2.7, i'm not sure for the reason, but the kafka version is different from master (0.10.2-kafka-2.2.0 is a more recent version than 0.10.0-kafka-2.1.0). 
Above this shows 
```
<kafka.version>0.10.2-kafka-2.2.0</kafka.version>
```
For the other 9 cloudera platforms. Should master match/be updated?

OlegMazurov: I believe kafka.version is currently irrelevant in 2.7. I had to downgrade it when I added Spark 2.3 support. If we backport it to 2.7 we'll need to use the correct version. I still think it's better to set it right for 5.14.2 now as it won't be part of the Spark2.3 backport.

PR #2386, opened 2018-11-05
DB-7553: adding spec files for mapr 6.0

MurrayBrown:
this initial set of files built rpms that worked on mapr 6.0.1, based on the same binary files compiled for mapr 6.0.0 and tested there. I think this structure therefore works on both versions of mapr 6


MurrayBrown:
wrong destination branch

PR #2388, opened 2018-11-06
DB-7412: Build Splice Machine Zeppelin Interpreter

edriggers:


zeppelin-splicemachine/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-pool2</artifactId>
+            <version>${commons.pool2.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
+            <version>${commons.logging.version}</version>
+        </dependency>
+        <dependency>
```
bklo94: This appears to be a duplicate of lines 61-68. It causes a warning
```
[WARNING] Some problems were encountered while building the effective model for com.splicemachine:zeppelin-splicemachine-0.8.0:jar:2.8.0.1845-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.slf4j:slf4j-log4j12:jar -> version (?) vs ${slf.logging.version} @ com.splicemachine:zeppelin-splicemachine-0.8.0:[unknown-version], /Users/user/Github/spliceengine/zeppelin-splicemachine/pom.xml, line 83, column 21
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.slf4j:slf4j-api:jar -> version (?) vs ${slf.logging.version} @ com.splicemachine:zeppelin-splicemachine-0.8.0:[unknown-version], /Users/user/Github/spliceengine/zeppelin-splicemachine/pom.xml, line 88, column 21
```

Do you need both set of dependencies to function? Looking at the build it looks like it only shades lines 61-68 and ignores the duplicate.

zeppelin-splicemachine/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-enforcer-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
```
bklo94: The plugin is missing a version. And causes a warning
```
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-resources-plugin is missing. @ com.splicemachine:zeppelin-splicemachine-0.8.0:[unknown-version], /Users/user/Github/spliceengine/zeppelin-splicemachine/pom.xml, line 107, column 21
```
This is  version that was used in assembly.xml
```
<version>3.1.0</version>
```

zeppelin-splicemachine/src/test/com/splicemachine/zeppelin/SpliceMachineInterpreterTest.java
```diff
@@ -0,0 +1,635 @@
...
+import org.apache.zeppelin.scheduler.ParallelScheduler;
+import org.apache.zeppelin.scheduler.Scheduler;
+import org.apache.zeppelin.user.AuthenticationInfo;
+import org.apache.zeppelin.user.UserCredentials;
+import org.apache.zeppelin.user.UsernamePassword;
+
+/**
+ * JDBC interpreter unit tests.
+ */
+public class JDBCInterpreterTest extends BasicJDBCTestCaseAdapter {
```
chuckwondo: Rename the class to match the filename and update the javadoc comment as well.

Did you run these tests?

zeppelin-splicemachine/pom.xml
MurrayBrown: I think we presently have our poms as children of 'our parent', aka the spliceengine-parent; that is the structure all of our poms currently have. It would be antithetical to have a pom in this project taht was a child of a third-party pom, i think?

I agree that the transitive dependencies now have to be resolved, though. Perhaps this pom just declares them, or if appropriate we edit teh spliceengine-parent pom.xml to provide the needed version variable(s)?

zeppelin-splicemachine/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-pool2</artifactId>
+            <version>${commons.pool2.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
+            <version>${commons.logging.version}</version>
+        </dependency>
```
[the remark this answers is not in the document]
MurrayBrown: I think that we only get a splice-derby.log via the slf4j logging functions that our codebase uses. since the zeppelin-splice-interpreter is really just a mini version of our database engine, 'our' logging does (i think) require that logging framework.

liprais:
I think this Interpreter should have its own GitHub repo and build process.
it is meant for zeppelin and there is no need to put it into spliceengine repo,which may bring in conflicts and confusions

zeppelin-splicemachine/pom.xml
chuckwondo: @MurrayBrown making `interpreter-parent` the parent of this might be inconsistent with the rest of the modules in this project, but it should not be problematic. It is not generally antithetical to specify a parent that is different from the project in which it is listed as a module. There is no strict need for the relationship to be 2-way. In this case, making `interpreter-parent` the parent will likely resolve all of the build problems for this module.

chuckwondo:
@liprais We were discussing whether to make this a separate repo or not, and we were leaning towards doing so. However, @OlegMazurov had suggested making it a module of `spliceenggine`, so we went with his recommendation. Personally, I was leaning towards making it a separate repo, but I'm still learning the build structure/process here, so we didn't go that route. I'm still open to making it a separate repo if there is a consensus to doing that.

@MurrayBrown, @OlegMazurov, @edriggers, what's the consensus?

edriggers:
@chuckwondo - I don't have a preference.

MurrayBrown:
If the conversation is about storing and building this in its own repository, then I would guess that 'splice-zeppelin' is the right place to put this?

PR #2392, opened 2018-11-08
DB-7636 Ignore recent transactions on read only transactions

dgomezferro:


yxia92:
@dgomezferro In the JIRA's problem description, you mentioned that for write that happens in the same transaction, we need to "wrap the Spark scan on a new transaction (1100)", is this part of the current PR? I don't see code related to that. Or this has already been implemented? Thanks.

dgomezferro:
@yxia92 in the end I only implemented it for read only transactions, because I was having some IT failures for triggers etc. I created DB-7638 to track the generalization of the optimization to writable transactions.

yxia92:
Got it, thanks @dgomezferro !

PR #2407, opened 2018-11-13
DB-7499 Add spark adapter to release package

bklo94:
After discussing how to add the spark adapter, I added the spark adapter to the binary distributions since we currently already have the jars on nexus and having them on S3 would cause more issues with documentation/use.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -31,7 +31,6 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <!-- Get all projects and required dependencies transitively -->
             <groupId>${project.groupId}</groupId>
             <artifactId>splicemachine-${envClassifier}-${spark.version}_${scala.binary.version}</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


OlegMazurov:
I thought we would just add an artifactId to commonArtifacts (and installer). Did it not work for some reason?

bklo94:
I attempted to add it to under the copy-dependency goal (where commonArtifacts is located), but the goal actually pulled and packaged all of spark adapter's dependencies instead of just getting the single spark adapter artifact.

OlegMazurov:
As a matter of fact, splicemachine-${envClassifier}-${spark.version}_${scala.binary.version} is already included in commonArtifacts. It's not added to the parcel because the artifact is not found among module dependencies. Adding it does the trick:

`    <dependencies>
        ...
        <dependency>
            <artifactId>splicemachine-${envClassifier}-${spark.version}_${scala.binary.version}</artifactId>
        </dependency>
    </dependencies>
 `
For _installer_ the artifact also should be added to _includeArtifactIds_.

PR #2409, opened 2018-11-14
DB-6913 Query fails when output of CONVERT_TO_DATE is used in GROUP BY clause

OlegMazurov:
Ensure ContextManager is always setup after ActivationHolder deserialization (affects Control, OlapServer, Spark executors).
Tested: ITs (mem, cdh), TPC-H 100g, HTAP-25 in cluster.
Monitored RS, OlapServer, Spark executors for memory leaks (number of alive ContextManagers).

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
@@ -128,8 +128,9 @@ public class ActivationHolder implements Externalizable {
         }
     }
 
-    public Activation getActivation() {
-        // Reinitialize if deserialized
+    public synchronized Activation getActivation() {
+        // Only directly instantiated ActivationHolders have initialized == true
+        // Those deserialized will check if impl.get() and activation are both set
         if (!initialized) {
             init(txn, true);
         }
@@ -150,7 +151,7 @@ public class ActivationHolder implements Externalizable {
         SIDriver.driver().getOperationFactory().writeTxn(txn,out);
     }
 
-    private synchronized void init(TxnView txn, boolean reinit){
+    private void init(TxnView txn, boolean reinit){
         try {
             SpliceTransactionResourceImpl txnResource = impl.get();
             if (txnResource != null && activation != null) return;
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
@@ -15,6 +15,7 @@
 package com.splicemachine.derby.stream;
 
 import com.splicemachine.EngineDriver;
+import com.splicemachine.db.iapi.services.context.ContextService;
 import com.splicemachine.db.impl.jdbc.EmbedConnection;
 import com.splicemachine.db.impl.jdbc.EmbedConnectionContext;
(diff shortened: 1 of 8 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
```diff
@@ -130,7 +129,10 @@ private void addSubOperations(Map<Integer, SpliceOperation> operationsMap, Splic
     }
 
     public Activation getActivation() {
-        init();
+        // Reinitialize if deserialized
+        if (!initialized) {
+            init(txn, true);
+        }
         return activation;
```
msirek: If we're returning the activation outside of synchronization, is there a chance some other thread is updating it (e.g. a call to reinitialize), and we could get a partially updated copy?  General question:  Will each thread deserialize its own ActivationHolder objects, or could these be shared?

splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
```diff
@@ -148,30 +150,28 @@ public synchronized void writeExternal(ObjectOutput out) throws IOException {
...
+            new EmbedConnectionContext(ContextService.getService().getCurrentContextManager(), internalConnection);
 
-            SpliceOperationContext context = SpliceOperationContext.newContext(activation);
-            for(SpliceOperation so: operationsList){
-                so.init(context);
+            if (reinit) {
+                SpliceOperationContext context = SpliceOperationContext.newContext(activation);
+                for (SpliceOperation so : operationsList) {
+                    so.init(context);
+                }
```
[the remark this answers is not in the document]
OlegMazurov: That's deliberate but I realize it's confusing. The "initialized" is used only for instantiated ActivationHolders. Those that are created via deserialization are initialized when their ThreadLocal and activation are both initialized (too complicated).

splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
OlegMazurov: I agree, activation should not be read outside the synchronized block.

splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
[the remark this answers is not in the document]
OlegMazurov: The overhead of a subsequent call to getActivation() is synchronization and checking whether txnResource and activation have already been initialized. Previously we would close ActivationHolder immediately after initialization and rely on reinitialize(), which was already synchronized, before actual use. From that perspective, I'm just closing some holes without changing the dynamics much.

splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
msirek: I did some reading and apparently assignment and reading of references is atomic in Java:[https://stackoverflow.com/questions/44283378/why-is-reference-assignment-atomic-in-java/44284401](https://stackoverflow.com/questions/44283378/why-is-reference-assignment-atomic-in-java/44284401).  So, it sounds like we won't ever see a reference which is partially updated.  But I'm wondering if in the future activation should be made thread local, or use some copy-on-write solution.  I've seen places in the code where elements of the activation are updated at runtime.

PR #2412, opened 2018-11-15
DB-7583 Adjust right start key of merge join based on first read left row.

msirek:
This is a performance enhancement for merge join to adjust the right table start key using the values from the first read row of the left table when there is a range term like join_col > 5.  

Without the range term present, we currently use the left table key value as the right table start key.
With the range term, this doesn't happen even though the left table value could be much greater than the constant value in the range term.  This fix will use the greater of the two start keys.

See [DB-7583](https://splicemachine.atlassian.net/browse/DB-7583?focusedCommentId=58106&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-58106) for more info.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/function/merge/AbstractMergeJoinFlatMapFunction.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/function/merge/AbstractMergeJoinFlatMapFunction.java
@@ -17,6 +17,7 @@ package com.splicemachine.derby.stream.function.merge;
 import com.splicemachine.EngineDriver;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
+import com.splicemachine.db.iapi.types.DataValueDescriptor;
 import com.splicemachine.db.impl.sql.execute.BaseActivation;
 import com.splicemachine.db.impl.sql.execute.ValueRow;
 import com.splicemachine.derby.iapi.sql.execute.SpliceOperation;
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/function/merge/AbstractMergeJoinFlatMapFunction.java
```diff
@@ -119,49 +119,65 @@ protected void initRightScan(PeekingIterator<ExecRow> leftPeekingIterator) throw
...
                             newStartKey.setColumn(i + 1, firstHashRow.getColumn(j + 1));
-                            len++;
-                            break;
                         }
-                    }
-                    //no match found for the given key column position
-                    if (j >= rightHashKeys.length)
+                        else {
+                            // If a descending index, reverse the result of the comparison.
+                            boolean replaceWithLeftRowVal = (startPosition == null) ||
```
dgomezferro: This logic is a bit hard to follow, could you break it down into one or more utility methods or at least use a few more booleans if possible?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/merge/AbstractMergeJoinFlatMapFunction.java
```diff
@@ -112,56 +113,103 @@ private boolean isKeyColumn(int[] columnOrdering, int col) {
...
+
+                            // Replace the right key value to seek to with left row data value if:
+                            // - The right key is null, in which case left is always >= right , or
+                            // - If neither left key nor right key is null, compare
+                            //   left and right key values, and do the replacement if:
+                            //   - If the key is ascending and the current right key start
+                            //     position is less than the left data value, or
+                            //   - If the key is descending and the current right key start
+                            //     position is greater than the left data value.
+                            boolean replaceWithLeftRowVal = false;
```
yxia92: I'm thinking if replaceWithLeftRowVal should be a decision based on every column, or we should do it once for the leading index/pk columns, and let the subsequent columns honor the decision. 
For example, given firstHashRow (1,3), startPosition (2,1), and all columns in asc order, for the first column, we will pick 2 from startPosition, for the subsequent columns, we should continue to pick 1 from startPosition instead of 3 from firstHashRow. I think the existing logic also has an issue.
Here is an example:
create table t1 (a1 int, b1 int, c1 int, primary key (a1, b1));
create table t2 (a2 int, b2 int, c2 int, primary key (a2, b2));
insert into t1 values (1,3,3), (2,1,1);
insert into t2 values (1,3,3), (2,1,1);

select * from t1, t2 --splice-properties joinStrategy=merge
where a1=a2 and b1=b2 and a2>=4/2;

yxia92: In this case, since a2>=4/2 contains an expression, TC does not kick in, so the left right may not start from a1=2 but a row with a smaller value. I guess we cannot rely on TC to always be kicking in for the logic initRightScan().

msirek: Good catch @yxia92.  I updated the code to keep comparing key columns as long as they are equal, and once a partial key is found to be greater than the other (left vs. right), we pick that one as the start key and stick with that decision.

PR #2420, opened 2018-11-16
DBAAS-1929 Increase retry deployment count

bklo94:
Adding this configuration hopefully will increase the number of passing builds since rate limiting to the artifactory does not seem to be enough


MurrayBrown:
looks good - this may help avoid more failures during networking issues pulling/pushing from our nexus artifact repositories (repository.splicemachine.com and nexus.splicemachine.com)

PR #2426, opened 2018-11-16
SPLICE-2277 Prevent Disruptor from blocking region server shutdown (2.5)

dgomezferro:


MurrayBrown:
Approved for 2.5 because shutdown is a thing everyone needs to do.

PR #2431, opened 2018-11-19
DB-6812 Fix artifactId and invalid transitive pom warnings

bklo94:
These changes fix the artifactId warning along with the invalid transitive POM warning. Tested on DB-6812 job in Jenkins and changes can be seen in the nexus repo under version 0.0.0.1. 

Removed 
`jar` goal from maven-jar-plugin since it is the default goal.

Removed 
Prerequisite since we currently have the same configuration in the maven enforcer plugin and it also removes another WARNING.

Changes to jar locations in Nexus
Old
hbase_sql-envClassifier/
hbase-storage-envClassifier/
Splice machine-envClassifier-platform/
hbase_pipeline-envClassifier/

New 
hbase_sql/
hbase-storage/
Splicemachine/
hbase_pipeline/


assembly/pom.xml
```diff
@@ -14,7 +14,7 @@
   -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <artifactId>splice_machine-assembly-${envClassifier}</artifactId>
+    <artifactId>splice_machine-assembly</artifactId>
```
MurrayBrown: do we not have to add a classifier at the top level here?

splice_spark/pom.xml
```diff
@@ -19,7 +19,7 @@
         <groupId>com.splicemachine</groupId>
         <version>2.8.0.1845-SNAPSHOT</version>
     </parent>
-    <artifactId>splicemachine-${envClassifier}-${spark.version}_${scala.binary.version}</artifactId>
+    <artifactId>splicemachine</artifactId>
```
MurrayBrown: this artifact name probably now needs changing.  I think calling it 'splicemachine' is flawed. since this will be a change that impacts all users, perhaps we use this opportunity to name the artifact 'splice_spark' ?

assembly/pom.xml
MurrayBrown: I see that the pattern here is completely the same, so i wont re-post this question elsewhere. I guess what i am missing is: how does an artifact get its classifier?

bklo94: Some artifacts do not have classifiers that differentiate them on nexus. If there was a need to have a differentiation for the artifact, I either added a configuration to add the envClassifier or if they didn't have the maven-jar-plugin, it was added. 

For, splice_machine assembly is a little odd since it's included into the shaded jar. We stopped deploying it to Nexus after 1825.

bklo94:
The actual {envClassifier} is set under profiles in the parent/top level pom.xml. So when you use the flag -P it looks for it and sets all the needed properties.
i.e
`<profile>
            <id>cdh5.8.5</id>
            <activation>
                <property>
                    <name>env</name>
                    <value>cdh5.8.5</value>
                </property>
            </activation>
            <properties>
                <cdh.parcel.version>5.8.5</cdh.parcel.version>
                <envClassifier>cdh5.8.5</envClassifier>`

splice_spark/pom.xml
liprais: <dependency>
                    <artifactId>splicemachine-cdh5.12.2-2.2.0.cloudera2_2.11</artifactId>
                    <version>${spliceVersion}</version>
                </dependency>
                <dependency>
                    <artifactId>hbase_sql-cdh5.12.2</artifactId>
                    <version>${spliceVersion}</version>
                </dependency>
                <dependency>
                    <artifactId>hbase_storage-cdh5.12.2</artifactId>
                    <version>${spliceVersion}</version>
                </dependency>
these 3 artifacts is used to make spark adapter work,it is customer-faced,we should make a note so that customers also make changes to their pom when upgrading.
I am good with the changes,spark adapter tests can be easily adapted to this change

PR #2432, opened 2018-11-20
SPLICE-1958 Fix reference column matching when update target table has alias

changli6:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UpdateNode.java
```diff
@@ -1393,16 +1397,29 @@ private void checkTableNameAndScrubResultColumns(ResultColumnList rcl)
...
+                column.clearTableName();
+            }
+            else {
+                /* Make the column's table name to table's exposed name
+                 which is consistent with getMatchingColumn in FromBaseTable
+                */
+                ValueNode expression = column.getExpression();
+                if (expression instanceof ColumnReference && expression.getColumnName().equals(column.getName()))
+                {
+                    ((ColumnReference) expression).setTableNameNode(tableNameNode);
```
msirek: If the table name is set here when foundMatchingTable is false, we might be setting the wrong table name, right?  Maybe need to only do this when foundMatchingTable is true?

msirek: If columnTableName is null, then the exception will not be thrown because it is within the 'if (columnTableName != null)' block, and we can get to line 1420 when no match is found.

changli6: Thanks @msirek  that I found the column name comparing was wrong here. I pushed a new commit to compare between result column's column reference table name and targetTable base name. The "if (columnTableName != null) " scope only used for checking table match now. While I'm not sure there was a chance columnTableName is null?

msirek: OK, @changli6.   The current code looks good.  But, the comments in ColumnReference say tableName could be null.  So instead of doing this:
`if (expression != null && 
    expression instanceof ColumnReference && 
    expression.getTableName().equals(baseTableName))`

... maybe we should reverse the order of evaluation to avoid potential NPE? : 

`if (expression != null && 
    expression instanceof ColumnReference && 
    baseTableName.equals(expression.getTableName()))`

changli6: Good catch.

PR #2433, opened 2018-11-20
SPLICE-1958 Fix reference column matching when update target table has alias (2.7)

changli6:


MurrayBrown:
Approved for 2.7. Thanks @changli6 !

PR #2434, opened 2018-11-20
SPLICE-1958 Fix reference column matching when update target table has alias (2.5)

changli6:


MurrayBrown:
Approved for 2.5 (although I'm not sure that anyone needs it on that branch?)

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/UpdateOperationIT.java
```diff
@@ -684,4 +685,21 @@ private int doTestUpdateMultiColumnOneSubSyntax(String outerWhere) throws Except
...
+
+        new TableCreator(methodWatcher.getOrCreateConnection())
+                .withCreate("create table alias2 (a2 int, b2 int, c2 int)")
+                .create();
+
+        Assert.assertTrue("expected result having Subquery",
+                SpliceUnitTest.getExplainMessage(3,"explain update alias1 X set a1 = (select y.a2 from alias2 y where b1=y.b2)",methodWatcher).contains("Subquery"));
+        Assert.assertTrue("expected result having BroadcastJoin",
+            SpliceUnitTest.getExplainMessage(3,"explain update alias1 X set (a1) = (select y.a2 from alias2 y where b1=y.b2)",methodWatcher).contains("BroadcastJoin"));
+
```
jyuanca: The difference between the two update statements is a pair of parenthesis around column a1. They are semantically equivalent, but why they have very different plans?

changli6: The columns braced by parenthesis are used for updating multiple columns, which is optimized as join update. 
The single column without braced parenthesis update from a subquery.

PR #2436, opened 2018-11-20
DB-7672 start-splice-cluster -f is broken (2.7)

OlegMazurov:
Removing redundant core-site.xml which enforces local FS in standalone/ITs.
Also removing duplicated SpliceTestDfsPlatform.java


MurrayBrown:
Approved for 2.7

PR #2439, opened 2018-11-20
DB-7508: display engine name system for system procedures(2.5)

jyuanca:


MurrayBrown:
Since the customer who requested this will be upgrading to Splice 2.7 with their next update, we do not need this on the Splice 2.5 branch.  Thanks Jun!

PR #2440, opened 2018-11-21
DB-7651 Shuffle partitions when writing data from Spark Adapter

dgomezferro:


scala_util/src/main/scala/com/splicemachine/spark/splicemachine/ShuffledDependency.scala
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+
+package com.splicemachine.spark.splicemachine
+
+import org.apache.spark.NarrowDependency
+import org.apache.spark.rdd.RDD
+
+class ShuffledDependency[T](rdd: RDD[T]) extends NarrowDependency[T](rdd) {
+  val partitions = util.Random.shuffle[Int, IndexedSeq]((0 to (rdd.getNumPartitions - 1)).toArray)
```
msirek: This is a very useful enhancement.  If the shuffle is random though, it might make benchmarking difficult as performance results might differ from run to run.  Is there something else we can do besides a random shuffle to make it deterministic?  Maybe spread the partitions out evenly?  For example, if range is 0 to 99, create the list as 0,9,19,29,39,49,59,69,79,89,99,1,10,20,30 ...  Could maybe use the square root of the number of partitions as the increment value.  That's just one idea, there are probably other better ways to spread out the partitions deterministically.

dgomezferro: I think that's a fair requirement, however currently we are already shuffling partitions in a random order when running IMPORT. I'll create a JIRA to come up with some standardized way to get more determinism. Maybe by introducing a configuration property that always sets the seed to a constant and is only set during benchmarking.

scala_util/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+// limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>spliceengine-parent</artifactId>
+        <groupId>com.splicemachine</groupId>
+        <version>2.8.0.1847-SNAPSHOT</version>
+    </parent>
+    <artifactId>scala_util-${envClassifier}-${spark.version}_${scala.binary.version}</artifactId>
```
OlegMazurov: DB-6812 fixed all artifactIds in spliceengine/spliceengine-ee to be constant and stop using expressions like ${envClassifier}. Please use splice-spark for reference.

dgomezferro:
Good catch @OlegMazurov , added it on all branches

PR #2446, opened 2018-11-23
DB-7673 Fix NPE in nested loop join with IndexLookup.

msirek:
Fixes a regression of [DB-6345](https://splicemachine.atlassian.net/browse/DB-6345).

See [DB-7673](https://splicemachine.atlassian.net/browse/DB-7673).


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/NestedLoopJoinOperationIT.java
```diff
@@ -79,6 +135,22 @@ public static void createDataSet() throws Exception {
...
+    /* Regression test for DB-7673 */
+    @Test
+    public void testIndexLookupWithNestedJoin() throws Exception {
+        ResultSet rs = null;
+        try {
+            rs = methodWatcher.executeQuery("SELECT * FROM v1 --splice-properties useSpark=true\n" +
+                "where M_ID >=1 and M_ID <=100");
+        } catch (SQLException e) {
+            fail("Query not expected to fail.");
+        }
```
chuckwondo: Catching the exception and invoking `fail` is redundant. You can eliminate the try/catch.

PR #2449, opened 2018-11-26
DB-7679 (De)serialize user types when using Spark

dgomezferro:


chuckwondo:
+1

db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
```diff
@@ -715,7 +730,7 @@ public void updateThetaSketch(UpdateSketch updateSketch) {
 
 	@Override
 	public void setSparkObject(Object sparkObject) throws StandardException {
-
+		value = sparkObject;
```
msirek: It looks like sparkObject could now be a byte[], so should setSparkObject call deserialize here, the same as the code you added in getLength() ?

dgomezferro: I think that's right, even though that's only currently being used in SQLArray for which we don't have much support. Thanks

PR #2451, opened 2018-11-26
DB-7679 (De)serialize user types when using Spark (2.7)

dgomezferro:


db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
```diff
@@ -249,7 +254,27 @@ public Object getObject()
...
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+		} finally {
+			try {
+				bos.close();
+			} catch (IOException ex) {
+				// ignore close exception
+			}
+		}
+	}
```
chuckwondo: How about making use of a try-with-resources block to simplify things a bit?
```java
	@Override
	public Object getSparkObject() throws StandardException {
		try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
			ObjectOutput out = new ObjectOutputStream(bos);
			out.writeObject(value);
			out.flush();
			return bos.toByteArray();
		} catch (IOException e) {
			throw new RuntimeException(e);
		}
	}
```
Also, why are you throwing `RuntimeException` instead of `StandardException`?

db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
```diff
@@ -698,7 +723,23 @@ public void read(Row row, int ordinal) throws StandardException {
...
+				throw new RuntimeException(e);
+			} finally {
+				try {
+					if (in != null) {
+						in.close();
+					}
+				} catch (IOException ex) {
+					// ignore close exception
+				}
+			}
```
chuckwondo: Here's another spot where a try-with-resources block can simplify things a bit.
```java
		else {
			isNull = false;
			byte[] bytes = (byte[]) row.get(ordinal);
			ByteArrayInputStream bis = new ByteArrayInputStream(bytes);

			try (ObjectInput in = new ObjectInputStream(bis)) {
				value = in.readObject();
			} catch (IOException | ClassNotFoundException e) {
				throw new RuntimeException(e);
			}
		}
```
Again, why are you throwing `RuntimeException` instead of `StandardException`?

db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
dgomezferro: @chuckwondo good points, I simplified it by using SerializationUtils.serialize()/deserialize() as used elsewhere on the class.

db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
```diff
@@ -723,22 +708,11 @@ public void read(Row row, int ordinal) throws StandardException {
...
-				try {
-					if (in != null) {
-						in.close();
-					}
-				} catch (IOException ex) {
-					// ignore close exception
-				}
+			Object object = row.get(ordinal);
+			if (object instanceof byte[]) {
+				value = SerializationUtils.deserialize((byte[]) row.get(ordinal));
```
chuckwondo: Don't you want to pass `object` to `deserialize` instead of calling `row.get(ordinal)` a second time?

db-engine/src/main/java/com/splicemachine/db/iapi/types/UserType.java
```diff
@@ -257,21 +253,10 @@ public Object getObject()
...
-		} finally {
-			try {
-				bos.close();
-			} catch (IOException ex) {
-				// ignore close exception
-			}
+		if (value instanceof Serializable) {
+			return SerializationUtils.serialize((Serializable) value);
+		} else {
+			return super.getSparkObject();
```
chuckwondo: How about this?
```java
return (value instanceof Serializable)
	? SerializationUtils.serialize((Serializable) value)
	: super.getSparkObject();
```

dgomezferro: I think that looks good, but it's a matter of personal taste.

dgomezferro:
@chuckwondo thanks for the review! If you want to +1 or request more changes please use https://github.com/splicemachine/spliceengine/pull/2449 (the PR for the master branch)

PR #2455, opened 2018-11-26
DB-7409 Create standalone tarball

bklo94:
New PR for the standalone profile since the last PR with rebase included all the commits. Added a test-jar to the platform-it in order for YARN to correctly start, since one of the java tests was needed. When you use the standalone profile it will create a tarball with the name of SPLICEMACHINE-{project.version}-{envClassifier}.tar.gz. Once unzipped it will create a folder with SPLICEMACHINE-{project.version}.

Tried to get rid of the duplicate versioning, but there's still some jars with duplicate versioning that are still necessary since they pull from their own repository (i.e. jersey-core from org.glassfish vs com.sun.jersey)


chuckwondo:
+1

PR #2456, opened 2018-11-26
DB-7409 Create standalone tarball (2.7)

bklo94:


assembly/assembly.xml
```diff
@@ -0,0 +1,27 @@
+<assembly xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
chuckwondo: Assembly files are conventionally placed under a `src/assembly` directory rather than at the top level next to the `pom.xml` file, so please create a `src/assembly` directory here and move this file there. Also, the name of the assembly file should indicate what it is assembling, so I suggest renaming the file to `standalone.xml`.

assembly/pom.xml
```diff
@@ -733,6 +734,424 @@
...
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-assembly-plugin</artifactId>
+                        <configuration>
+                            <descriptor>assembly.xml</descriptor>
+                            <finalName>SPLICEMACHINE-${project.version}</finalName>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <phase>package</phase>
```
chuckwondo: I'm glad to see your use of the assembly plugin with an assembly descriptor file! Let's just add an `id` to this execution, perhaps `<id>make-standalone-assembly</id>`.

PR #2460, opened 2018-11-27
DB-7656 Modify installation defaults for better load throughput

dgomezferro:


assembly/hdp2.6.1/build/var/lib/ambari-server/resources/common-services/SPLICEMACHINE/2.5.1/service_advisor.py
```diff
@@ -195,22 +195,22 @@ def getHBaseSiteDesiredValues(self):
...
         "hbase.hstore.defaultengine.compactor" : "com.splicemachine.compactions.SpliceDefaultCompactor",
         "hbase.coprocessor.region.classes" : "org.apache.hadoop.hbase.security.access.SecureBulkLoadEndpoint,org.apache.ranger.authorization.hbase.RangerAuthorizationCoprocessor,com.splicemachine.hbase.MemstoreAwareObserver,com.splicemachine.derby.hbase.SpliceIndexObserver,com.splicemachine.derby.hbase.SpliceIndexEndpoint,com.splicemachine.hbase.RegionSizeEndpoint,com.splicemachine.si.data.hbase.coprocessor.TxnLifecycleEndpoint,com.splicemachine.si.data.hbase.coprocessor.SIObserver,com.splicemachine.hbase.BackupEndpointObserver",
         "hbase.htable.threads.max" : "96",
         "hbase.ipc.warn.response.size" : "-1",
         "hbase.ipc.warn.response.time" : "-1",
         "hbase.master.loadbalance.bytable" : "true",
+        "hbase.master.balancer.stochastic.regionCountCost" : "1500",
         "hbase.htable.threads.max" : "96",
-        "hbase.regions.slop" : "0.01",
+        "hbase.regions.slop" : "0",
```
msirek: Would having zero slop cause rebalancing to occur too often (use resources unnecessarily)?  The HBase default value for this is 0.001.  Should we use this instead?

MurrayBrown: Does the algorithm that we are employing use the slop setting?  It seems like it might be 'okay' to have a *slightly* different count of regions-per-regionserver ?

OlegMazurov: The balancer will tolerate node loads (number of regions) in the inclusive range
_floor_(regions/nodes*(1 - slop)) .. _ceil_(regions/nodes*(1 + slop))
Say, regions = nodes. If slop > 0, the balancer will tolerate any distribution with 0..2 regions per node.
On the other hand, if slop = 0, the balancer will try to fix any deviation from 1 region per node.
Once we get regions = nodes + 1, slop = 0.01 makes a difference only when the number of nodes is > 100.
I'd say, Daniel tries to be more aggressive when the number of regions is low. The effect is minimal in my opinion but the price is also negligible.

msirek: OK, so slop of 0 means we only tolerate a region count at most 1 away from the average.  I think I misinterpreted the details of slop.  Also, I am not sure if the LoadBalancer we use looks at the slop number.  I don't think there's an issue here.

PR #2463, opened 2018-11-27
DB-7682 Encode full txn stack for Spark queries

dgomezferro:


splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnOperationFactory.java
```diff
@@ -170,6 +188,29 @@ public TxnView readTxn(ObjectInput oi) throws IOException{
...
+        ois.readFully(bytes);
+
+        MultiFieldDecoder decoder=MultiFieldDecoder.wrap(bytes, 0, bytes.length);
+        long txnId=decoder.decodeNextLong();
+        long beginTs=decoder.decodeNextLong();
+        boolean additive=decoder.decodeNextBoolean();
+        Txn.IsolationLevel level=Txn.IsolationLevel.fromByte(decoder.decodeNextByte());
+        boolean allowsWrites=decoder.decodeNextBoolean();
+
+        TxnView parent=Txn.ROOT_TRANSACTION;
```
yxia92: this line can be removed.

PR #2487, opened 2018-12-01
DB-7715 Don't override logs in IT runs

dgomezferro:


platform_it/src/test/bin/start-splice-its
```diff
@@ -93,18 +93,18 @@ fi
 
 # Start SpliceTestYarnPlatform in background.
 echo "Starting SpliceTestYarnPlatform, log file is ${YARNLOG}"
-mvn -B exec:java -Denv=${HBASE_PROFILE} -P${PROFILE},spliceYarn > ${YARNLOG} 2>&1 &
+mvn -B exec:java -Denv=${HBASE_PROFILE} -P${PROFILE},spliceYarn >> ${YARNLOG} 2>&1 &
```
yxia92: To be consistent, should we do the same for the zoo log too at line 71:

> mvn -B exec:exec -Denv=${HBASE_PROFILE} -P${PROFILE},spliceZoo > ${ZOOLOG} 2>&1 &

dgomezferro: Good catch, done!

PR #2500, opened 2018-12-04
DB-6913 Query fails when output of CONVERT_TO_DATE is used in GROUP BY clause

OlegMazurov:
Redux: taking care of a ContextManager memory leak in Spark adapter application master.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
@@ -100,9 +100,6 @@ public final class SpliceTransactionResourceImpl implements AutoCloseable{
 
 
     public void close(){
-        while(!cm.isEmpty()){
-            cm.popContext();
-        }
         csf.resetCurrentContextManager(cm);
         csf.removeContext(cm);
     }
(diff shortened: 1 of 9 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/ActivationHolder.java
```diff
@@ -148,30 +151,31 @@ public synchronized void writeExternal(ObjectOutput out) throws IOException {
...
-        initialized = true;
+    private void init(TxnView txn, boolean reinit){
         try {
-            impl = new SpliceTransactionResourceImpl();
-            prepared =  impl.marshallTransaction(txn);
-            activation = soi.getActivation(this, impl.getLcc());
+            SpliceTransactionResourceImpl txnResource = impl.get();
+            if (txnResource != null) {
+                if (activation != null) return;
+                txnResource.close();
```
OlegMazurov: New change: closing a leftover SpliceTransactionResourceImpl

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
```diff
@@ -100,9 +100,6 @@ public boolean marshallTransaction(TxnView txn) throws StandardException, SQLExc
 
 
     public void close(){
-        while(!cm.isEmpty()){
```
OlegMazurov: New change: don't destroy ContextManager, it may still be used via an explicit  reference

PR #2503, opened 2018-12-04
DB-7717: reset isFlushing flag if a region is stuck in split state

jyuanca:


dgomezferro:
@jyuanca can you elaborate on why this is needed and how the system gets on that state?

hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
```diff
@@ -138,6 +138,18 @@ public void prepareBackup(
         if (isSplitting.get() || isFlushing.get() || isCompacting.get()) {
             SpliceLogUtils.info(LOG, "table %s region %s is not ready for backup: isSplitting=%s, isCompacting=%s, isFlushing=%s",
                     tableName , regionName, isSplitting.get(), isCompacting.get(), isFlushing.get());
+
+            // If the region is flush, wait until it is done. Check the flag. If it is still true and the region
+            // gets stuck in split state, flush it to reset the flag
+            if (isFlushing.get()) {
+                region.waitForFlushesAndCompactions();
```
dgomezferro: Why the special case for flush?

hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
```diff
@@ -271,11 +282,12 @@ public void postFlush(ObserverContext<RegionCoprocessorEnvironment> e, Store sto
         try {
             // Register HFiles for incremental backup
             String filePath =  resultFile != null?resultFile.getFileInfo().getFileStatus().getPath().toString():null;
-            SpliceLogUtils.info(LOG, "Flushing store file %s", filePath);
+            SpliceLogUtils.info(LOG, "Flushed store file %s", filePath);
             if (!BackupUtils.isSpliceTable(namespace, tableName))
                 return;
             BackupUtils.captureIncrementalChanges(conf, region, fs, rootDir, backupDir,
                     tableName, resultFile.getPath().getName(), preparing.get());
+            isFlushing.set(false);
```
dgomezferro: We already set isFlushing to false in 

    public void postFlush(ObserverContext<RegionCoprocessorEnvironment> e) throws IOException {

why is it needed to do it here too?

jyuanca:
@dgomezferro Exactly how a region is stuck in flushing state is still unknown. This PR adds some tracing to help identify the problem, set the flag to false in more postFlush hook just in case some hook is not invoked for some reasons. It also checks when there is no flushes, whether the region stucks in flushing state, if so, flush the region to reset the flag.

hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/hbase/BackupEndpointObserver.java
```diff
@@ -138,6 +138,18 @@ public void prepareBackup(
...
+
+            // If the region is flush, wait until it is done. Check the flag. If it is still true and the region
+            // gets stuck in split state, flush it to reset the flag
+            if (isFlushing.get()) {
+                region.waitForFlushesAndCompactions();
+                if (isFlushing.get()) {
+                    SpliceLogUtils.warn(LOG, "The region %s may be stuck in split state, flush it to reset flag",
+                            regionName);
+                    HBasePlatformUtils.flush(region);
+                }
```
msirek: If we're waiting for the flushes and compactions to complete, why do we still fall through and call responseBuilder.setReadyForBackup(false) ?  Could we just re-apply the splitting/flushing/compacting test here?, e.g.

if (!(isSplitting.get() || isFlushing.get() || isCompacting.get()))
    return super.preCompact(e, store, scanner, scanType);

jyuanca: Yes, we could, but I prefer to make simple code changes here. This is to handle a rare case. Returning false and letting backup(for this region) retry does not have a significant impact on performance.

msirek: OK, I didn't realize if we return false, the backup will get retried.  It seems fine.  Thanks.

PR #2506, opened 2018-12-04
DB-7692: stop vacuum during database restore

jyuanca:
Please review code changes on ee branch as well.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
@@ -53,6 +53,7 @@ import com.splicemachine.si.api.txn.TxnView;
 import com.splicemachine.si.impl.driver.SIDriver;
 import com.splicemachine.stream.Stream;
 import com.splicemachine.stream.StreamException;
+import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.log4j.Logger;
 
 /**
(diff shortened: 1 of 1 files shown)


splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -199,6 +199,9 @@ public void vacuumDatabase(long oldestActiveTransaction) throws SQLException{
 
             List<DataDelete> deletes = new ArrayList<>();
             for (long removed : toDelete) {
+                boolean restoreMode = SIDriver.driver().lifecycleManager().isRestoreMode();
+                if (restoreMode)
+                    break;
```
yxia92: Should we break out here or return directly? It is possible that when restoreMode becomes true, the _deletes_ list is non-empty, in this case, do we want "dropped.delete(deletes)" to be executed or not?

jyuanca: Good point. We may want to terminate vacuum immediately during a database restore.

splice_machine/src/main/java/com/splicemachine/derby/utils/Vacuum.java
```diff
@@ -199,6 +199,9 @@ public void vacuumDatabase(long oldestActiveTransaction) throws SQLException{
 
             List<DataDelete> deletes = new ArrayList<>();
             for (long removed : toDelete) {
+                boolean restoreMode = SIDriver.driver().lifecycleManager().isRestoreMode();
+                if (restoreMode)
+                    return;
```
yxia92: On a second thought, I think this code may have an issue. When we get here, all the eligible dropped tables have been deleted, and this FOR loop is just to prepare for the deletion of the corresponding entries in the DROPPED_CONGLOMERATES_TABLE. If we abort here, the tables deleted will never get a chance to be removed from the DROPPED_CONGLOMERATES_TABLE. Correct me if I missed something.

dgomezferro: That's right @yxia92 , we want to break out of the "for(TableDescriptor table:hTableDescriptors){" loop, not this one.

jyuanca:
@yxia92 @dgomezferro I think that's my original intention, but I later got confused ...

PR #2509, opened 2018-12-05
DB-7603 do not add dependency of FK constraint on role or permission …

yxia92:
…descriptors

Please see solution description in [DB-7603](https://splicemachine.atlassian.net/browse/DB-7603?focusedCommentId=58276&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-58276)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLConstantOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLConstantOperation.java
@@ -159,8 +159,11 @@ public abstract class DDLConstantOperation implements ConstantAction, ScopeNamed
 		DataDictionary dd = lcc.getDataDictionary();
 		DependencyManager dm = dd.getDependencyManager();
         String currentUser = lcc.getCurrentUserId(activation);
-		SettableBoolean roleDepAdded = new SettableBoolean();
-        if (! currentUser.equals( dd.getAuthorizationDatabaseOwner()) ) {
+        List<String> dependedRoleList = new ArrayList<>();
+		List<String> groupuserList = lcc.getCurrentGroupUser(activation);
+		String dbo = dd.getAuthorizationDatabaseOwner();
+        if (! currentUser.equals(dbo)
+				&& !(groupuserList != null && groupuserList.contains(dbo))) {
 			PermissionsDescriptor permDesc;
 			// Now, it is time to add into dependency system the FOREIGN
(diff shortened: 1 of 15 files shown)


dgomezferro:
I haven't had a chance to look over it yet, does it work correctly with dependencies that have been removed erroneously by the existing code?

yxia92:
@dgomezferro The existing logic adds erroneous dependency of FK constraint on role descriptors and permission descriptors, so the fix is to stop adding the erroneous dependencies and also have an upgrade script to remove the existing erroneous dependencies. I tested a scenario to upgrade from version 1845 to version 1849 where we created 1000 table with FK constraints, and check on 1849 that the erroneous dependencies in the sys.sysdepends table have been removed, the FK constraints are retained.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLConstantOperation.java
```diff
@@ -159,8 +159,11 @@ protected void storeConstraintDependenciesOnPrivileges(Activation activation,
 		DataDictionary dd = lcc.getDataDictionary();
 		DependencyManager dm = dd.getDependencyManager();
         String currentUser = lcc.getCurrentUserId(activation);
-		SettableBoolean roleDepAdded = new SettableBoolean();
-        if (! currentUser.equals( dd.getAuthorizationDatabaseOwner()) ) {
+        List<String> dependedRoleList = new ArrayList<>();
+		List<String> groupuserList = lcc.getCurrentGroupUser(activation);
+		String dbo = dd.getAuthorizationDatabaseOwner();
+        if (! currentUser.equals(dbo)
+				&& !(groupuserList != null && groupuserList.contains(dbo))) {
```
msirek: @yxia92 What is this groupuserList check trying to do?  Do all users in the group user list automatically have permissions to eachother's objects?

yxia92: Hi @msirek , groupUserList is added as part of DB-6889. Basically, an (LDAP) user could belong to multiple groups, and the user will inherit the privileges granted to those groups. Some groups have admin privileges, like a database owner. If one of the groups is a dbo, we treat this user also as a dbo. DB-6889 adds logic to check groupusers(an example is in GenericAuthorizer.authorize()) in addition to the check of current user, but it misses the case in this function. I spot it, so just added it. It is not really related to DB-7603. thanks.

PR #2519, opened 2018-12-11
DB-7456 Fix SQLDecimal write and read external of precision and scale

changli6:


changli6:
Thanks @dgomezferro. Push a new commit to write precision and scale unconditionally.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -393,12 +393,14 @@ private boolean evaluateNull()
 	 */
 	public void writeExternal(ObjectOutput out) throws IOException {
         out.writeBoolean(isNull);
-        if (isNull)
-            return;
+		out.writeByte(precision);
```
dgomezferro: Sorry, I meant writing them unconditionally when the value isNull, basically move these 2 lines inside the isNull if.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -445,17 +447,20 @@ public void writeExternal(ObjectOutput out) throws IOException {
...
-	 * Note the use of rawData: we reuse the array if the
+	 * Note the use of rawData: we reuse the array if the\
 	 * incoming array is the same length or smaller than
 	 * the array length.
 	 *
 	 * @see java.io.Externalizable#readExternal
 	 */
 	public void readExternal(ObjectInput in) throws IOException {
+		boolean nullValue = in.readBoolean();
+		precision = in.readByte();
```
dgomezferro: Similarly, move the precision/scale reads inside the if(nullValue) check

PR #2522, opened 2018-12-11
DB-7570 Fix broadcast join on different datatype

changli6:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -150,28 +150,38 @@ public ValueNode bindExpression(FromList fromList,
...
+				targetTypeId = rightTypeId;
+
+			rightOperand = (ValueNode)
+					getNodeFactory().getNode(
+							C_NodeTypes.CAST_NODE,
+							rightOperand,
+							new DataTypeDescriptor(
+									targetTypeId,
+									true,
+									rightTypeServices.getMaximumWidth()),
```
yxia92: For the case where we need to cast to the left type, should we use the left type's typeServices.getMaximumWidth()?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -150,28 +150,38 @@ public ValueNode bindExpression(FromList fromList,
...
+				targetTypeId = leftTypeId;
 			leftOperand =  (ValueNode)
 				getNodeFactory().getNode(
 					C_NodeTypes.CAST_NODE,
 					leftOperand, 
 					new DataTypeDescriptor(
-							leftTypeId,
+							targetTypeId,
 							true, 
 							leftTypeServices.getMaximumWidth()),
```
yxia92: Same here, the type service should be consistent with the targetTypeId.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
```diff
@@ -414,4 +441,104 @@ public void testInequalityBroadCastJoin() throws Exception {
...
+        Assert.assertTrue("rs.next() failed", rs.next());
+        c = rs.getInt(1);
+        Assert.assertTrue("count(s1.a1) returned incorrect number of rows:", (c == 1));
+
+        rs.close();
+    }
+
+    @Test
+    public void testCharDateColumnsBroadCastJoin() throws Exception {
+        String sqlText = format("select count(s1.a1) from " + s1 +
```
yxia92: Can you add a negative test case where the char value  is not a legitimate date type? Would it be able to error out gracefully with the cast operation?

changli6: Sure. It's an error show out.

changli6:
Good catch @yxia92. Push a new commit fix and having a negative test case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -149,31 +149,49 @@ public ValueNode bindExpression(FromList fromList,
...
+			TypeId targetTypeId;
+			DataTypeDescriptor targetTypeServices;
+			if (leftTypeId.isBooleanTypeId() || leftTypeId.isDateTimeTimeStampTypeId()) {
+				targetTypeId = leftTypeId;
+				targetTypeServices = leftOperand.getTypeServices();
+			}
+			else {
+				targetTypeId = rightTypeId;
+				targetTypeServices = rightOperand.getTypeServices();
+			}
```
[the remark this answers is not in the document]
changli6: String type is only comparable with a boolean or DateTime type, but the error out for other cases is from bindComparisonOperator() which is after here. We can't cast the string to non-boolean/DateTime because it will break the checking in bindComparisonOperator.

I think we should keep the if/else logic. We just need to cast the string type to boolean or DateTime. For other cases (else block), we could raise an error, since it will inevitably raise error out from bindComparisonOperator later.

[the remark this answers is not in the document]
msirek: OK, sounds good.

changli6:
Please hold on merge, there are failed test cases on mem platforms.

PR #2542, opened 2018-12-18
DB-6830:Enable ldap server in distributed standalone cluster

liprais:
DB- 6830:Enable ldap server in distributed standalone cluster
how to start full distributed standalone cluster:
0. change EE_KEY to the real ee key in ee repo.
1.spliceengine and spliceengine-ee should be under the same path and on the same version
2./start-splice-cluster -pcdh5.12.2,ee -f to start the cluster


platform_it/src/test/java/com/splicemachine/test/SpliceTestDfsPlatform.java
```diff
@@ -85,8 +87,11 @@ public void start(int nodeCount, String directory) throws Exception {
             conf.set(DFSConfigKeys.IGNORE_SECURE_PORTS_FOR_TESTING_KEY, "true");
 
             dfsCluster = new MiniDFSCluster.Builder(conf).clusterId("localDfs").format(true).numDataNodes(nodeCount).nameNodePort(58878).build();
+
             dfsCluster.waitActive();
 
+            System.out.println("liprais" + dfsCluster.getHttpUri(0));
+
```
jyuanca: remove debugging info?

liprais: removed

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -125,6 +125,14 @@ public static Configuration create(String hbaseRootDirUri,
             config.set("hbase.master.keytab.file", keytab);
             config.set("yarn.nodemanager.principal", "yarn/example.com@EXAMPLE.COM");
             config.set("yarn.resourcemanager.principal", "yarn/example.com@EXAMPLE.COM");
+            config.set("splicemachine.enterprise.key","EE_KEY");
```
dgomezferro: Can you make it so that the enterprise key is not required? Either by picking it automatically from the -ee repo, not enabling LDAP by default, or somehow avoiding the EE check on standalone.

dgomezferro:
This is great @liprais , nice job! It would be good if we didn't have to provide the EE key on standalone though, specially by changing code. There should be a way to provide it from the CLI, or just not require it to start standalone with the default options.

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
liprais: I think the easiest method is that we use sed tool to substitute "EE_KEY" string with our EE-key provided from cli.I will try to achieve that.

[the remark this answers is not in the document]
liprais: @MurrayBrown The purpose of starting a full distributed standalone cluster is to execute IT tests under secure env.I think we can start a distributed standalone from ee repo ,I will explore that too.

liprais:
I have made the following changes:
when starting a full distributed standalone cluster, the ee key will be read from /platform_it/src/test/resources/ee.txt ,which has been added to git ignore file.
if the file doesn't exists ,the cluster will start without the ldap server and use native auth.
if the ee key has been setup correctly ,a full kerberos + ldap full standalone cluster will start

dgomezferro:
@liprais I'm not sure what happened but the PR is a bit of a mess, right now it includes changes that are not relevant to the JIRA this is fixing. If it's not possible to clean up the PR so that it only includes the relevant changes, please open a new PR with only those changes. 

Your modification to read the EE key from a file looks good to me, but I'd rather have a clean PR before merging. Thanks!

liprais:
> @liprais I'm not sure what happened but the PR is a bit of a mess, right now it includes changes that are not relevant to the JIRA this is fixing. If it's not possible to clean up the PR so that it only includes the relevant changes, please open a new PR with only those changes.
> 
> Your modification to read the EE key from a file looks good to me, but I'd rather have a clean PR before merging. Thanks!

The branch is out of dated and I performed a rebase on the branch ,I will close this pull request and make a new one,Thanks for the review !

PR #2557, opened 2019-01-03
DB-7621 Multiprobe scan for multiple IN lists.

msirek:
This is a performance feature to enhance MultiProbeScan to use more index columns in the presence of multiple IN lists to reduce the number of rows probed.

For example, for 
```
create table t1 (pk_col1 int, pk_col2 int, col3 int, primary key(pk_col1, pk_col2));
select * from t1 where pk_col1 in (1,2,3) and pk_col2 in (1,2);
```
The in lists are combined to produce 6 sets of probe values:
`(pk_col1,pk_col2) IN ((1,1), (1,2), (2,1), (2,2), (3,1), (3,2))`

Each probe specifies two primary key columns, so may scan many less rows, given the more specific start/stop key, than just using `pk_col1 in (1,2,3)` to probe the first primary key column.

For more details, please see [DB-7621](https://splicemachine.atlassian.net/browse/DB-7621?focusedCommentId=58723&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-58723)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
@@ -1238,7 +1238,7 @@ public abstract class DataValueFactoryImpl implements DataValueFactory, ModuleCo
             BIT(StoredFormatIds.SQL_BIT_ID),
             ROW_LOCATION(StoredFormatIds.ACCESS_HEAP_ROW_LOCATION_V1_ID),
             ARRAY(StoredFormatIds.SQL_ARRAY_ID),
-            LIST(StoredFormatIds.LIST_ID),;
+            LIST(StoredFormatIds.LIST_ID);
 
             private final int storedFormatId;
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
@@ -31,27 +31,24 @@
 
 package com.splicemachine.db.iapi.types;
 
+import com.splicemachine.db.iapi.db.DatabaseContext;
 import com.splicemachine.db.iapi.error.StandardException;
-import com.splicemachine.db.iapi.services.sanity.SanityManager;
+import com.splicemachine.db.iapi.reference.Attribute;
+import com.splicemachine.db.iapi.reference.SQLState;
+import com.splicemachine.db.iapi.services.context.ContextService;
 import com.splicemachine.db.iapi.services.i18n.LocaleFinder;
 import com.splicemachine.db.iapi.services.io.RegisteredFormatIds;
 import com.splicemachine.db.iapi.services.io.StoredFormatIds;
 import com.splicemachine.db.iapi.services.monitor.ModuleControl;
 import com.splicemachine.db.iapi.services.monitor.ModuleFactory;
 import com.splicemachine.db.iapi.services.monitor.Monitor;
-import com.splicemachine.db.iapi.reference.Attribute;
-import com.splicemachine.db.iapi.reference.SQLState;
-import java.sql.Blob;
-import java.sql.Clob;
-import java.sql.Date;
-import java.sql.Time;
-import java.sql.Timestamp;
+import com.splicemachine.db.iapi.services.sanity.SanityManager;
+
+import java.sql.*;
 import java.text.Collator;
 import java.text.RuleBasedCollator;
-import java.util.Properties;
 import java.util.Locale;
-import com.splicemachine.db.iapi.db.DatabaseContext;
-import com.splicemachine.db.iapi.services.context.ContextService;
+import java.util.Properties;
 
 /**
  * Core implementation of DataValueFactory. Does not implement
@@ -1240,7 +1237,8 @@ public abstract class DataValueFactoryImpl implements DataValueFactory, ModuleCo
             BLOB(StoredFormatIds.SQL_BLOB_ID),
             BIT(StoredFormatIds.SQL_BIT_ID),
             ROW_LOCATION(StoredFormatIds.ACCESS_HEAP_ROW_LOCATION_V1_ID),
-            ARRAY(StoredFormatIds.SQL_ARRAY_ID);
+            ARRAY(StoredFormatIds.SQL_ARRAY_ID),
+            LIST(StoredFormatIds.LIST_ID),;
 
             private final int storedFormatId;
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
@@ -74,7 +74,9 @@ import java.sql.ResultSet;
  * <p>
  * If any of the elements of the ListDataType holds a null, the
  * ListDataType item itself can be considered as a null value since
- * (null OP X) AND (A OP B) and (C OP D) ... is always UNKNOWN.
+ * (null OP X) AND (A OP B) and (C OP D) ... is never TRUE,
+ * and a multicolumn IN list is never negated (we never apply
+ * the NOT operator on it).
  * <p>
  */
 public final class ListDataType extends DataType {
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
@@ -1,0 +1,563 @@
+/*
+ * This file is part of Splice Machine.
+ * Splice Machine is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Affero General Public License as published by the Free Software Foundation, either
+ * version 3, or (at your option) any later version.
+ * Splice Machine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Some parts of this source code are based on Apache Derby, and the following notices apply to
+ * Apache Derby:
+ *
+ * Apache Derby is a subproject of the Apache DB project, and is licensed under
+ * the Apache License, Version 2.0 (the "License"); you may not use these files
+ * except in compliance with the License. You may obtain a copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ * Splice Machine, Inc. has modified the Apache Derby code in this file.
+ *
+ * All such Splice Machine modifications are Copyright 2012 - 2018 Splice Machine, Inc.,
+ * and are licensed to you under the GNU Affero General Public License.
+ */
+
+package com.splicemachine.db.iapi.types;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+import com.splicemachine.db.iapi.services.cache.ClassSize;
+import com.splicemachine.db.iapi.services.io.ArrayInputStream;
+import com.splicemachine.db.iapi.services.io.StoredFormatIds;
+import com.splicemachine.db.iapi.services.sanity.SanityManager;
+import com.splicemachine.db.iapi.sql.execute.ExecRow;
+import com.splicemachine.db.iapi.types.DataValueFactoryImpl.Format;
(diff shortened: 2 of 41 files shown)


msirek:

spliceci:
TEST FAILED

msirek:

dgomezferro:
Does this somehow conflict with DB-7778 ?

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
```diff
@@ -1240,7 +1237,8 @@ protected LocaleFinder getLocaleFinder()
             BLOB(StoredFormatIds.SQL_BLOB_ID),
             BIT(StoredFormatIds.SQL_BIT_ID),
             ROW_LOCATION(StoredFormatIds.ACCESS_HEAP_ROW_LOCATION_V1_ID),
-            ARRAY(StoredFormatIds.SQL_ARRAY_ID);
+            ARRAY(StoredFormatIds.SQL_ARRAY_ID),
+            LIST(StoredFormatIds.LIST_ID),;
```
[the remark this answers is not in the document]
msirek: Done.

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
```diff
@@ -0,0 +1,563 @@
...
+ * elements and the comparison operation is true for each pairwise
+ * comparison of same-positioned elements from the two lists,
+ * for example, the comparison:
+ * (1, '2001-01-01', 'Widget') = (1, '2001-01-01', 'Sprocket')
+ * is evaluated as the following boolean expression:
+ * (1 = 1) AND ('2001-01-01' = '2001-01-01') and 'Widget' = 'Sprocket'
+ * <p>
+ * If any of the elements of the ListDataType holds a null, the
+ * ListDataType item itself can be considered as a null value since
+ * (null OP X) AND (A OP B) and (C OP D) ... is always UNKNOWN.
```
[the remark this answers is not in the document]
msirek: Updated the comment to clarify.

msirek:
jenkins please test @cdh5.12.2

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
```diff
@@ -0,0 +1,565 @@
...
+    public String getTraceString() throws StandardException
+    {
+        return getString();
+    }
+
+    @Override
+    public boolean isNull() {
+        boolean localIsNull = true;
+        
+        if (numElements > 1) {
```
yxia92: Why the check of numElements > 1 here? if numElements=1, we may not set the nullability correctly.

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
```diff
@@ -0,0 +1,565 @@
...
+    /*
+     * Hash code
+     */
+    @Override
+    public int hashCode() {
+        final int prime = 37;
+        int result = 17;
+    
+        for (int i = 0; i < numElements; i++) {
+            result = result * prime + dvd[i].hashCode();
```
yxia92: We need to check null before reference dvd[i] here.

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
```diff
@@ -0,0 +1,565 @@
...
+		return Format.LIST;
+		
+	}
+    
+    private static final int BASE_MEMORY_USAGE = ClassSize.estimateBaseFromCatalog(ListDataType.class);
+    
+    public int estimateMemoryUsage() {
+        int memEstimate = BASE_MEMORY_USAGE;
+        for (int i = 0; i < numElements; i++) {
+            memEstimate += dvd[i].estimateMemoryUsage();
```
yxia92: check null before reference dvd[i].

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
msirek: Removed the if condition.

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
msirek: Added the check.

db-engine/src/main/java/com/splicemachine/db/iapi/types/ListDataType.java
msirek: Added the check, plus a couple others in this module.

msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/InListMultiprobeIT.java
```diff
@@ -32,8 +39,19 @@
 /**
  * Test the IN list predicates with the multiprobe index scan
  */
+@RunWith(Parameterized.class)
+@Category(SerialTest.class)
```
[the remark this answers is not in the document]
msirek: Because the test involves calling syscs_set_global_database_property.  If the tests are run in parallel, the call might happen at the wrong time for one of the tests and change the results, which then might not match expected results.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNodeList.java
```diff
@@ -427,10 +431,15 @@ public boolean containsAllConstantNodes()
...
 		for (int index = 0; index < size; index++)
 		{
-			if (! ((ValueNode) elementAt(index) instanceof ConstantNode))
+			ValueNode literalVal = (ValueNode) elementAt(index);
+			if (! (literalVal instanceof ConstantNode))
 			{
 				return false;
 			}
+			else if (literalVal instanceof ListValueNode) {
+				if (! ((ListValueNode)literalVal).containsAllConstantNodes())
```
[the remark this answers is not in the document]
msirek: OK, I am moving the else if block inside the if block.  The way this function is used, it should only check for ConstantNodes, not ParameterNodes, because it is used to see when we can eliminate duplicates in the parser.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/InListSelectivity.java
```diff
@@ -45,26 +45,74 @@
...
         this.useExtrapolation = isExtrapolationEnabled();
     }
+    
+    private double multiplySelectivity(ValueNode vn, int columnNumber, double localSelectivity, boolean useExtrapolation) {
+        if (vn instanceof ConstantNode) {
+            ConstantNode cn = (ConstantNode)vn;
+            if (localSelectivity == -1.0d)
+                localSelectivity = storeCost.getSelectivity(columnNumber, cn.getValue(), true, cn.getValue(), true, useExtrapolation);
+            else
+                localSelectivity *= storeCost.getSelectivity(columnNumber, cn.getValue(), true, cn.getValue(), true, useExtrapolation);
```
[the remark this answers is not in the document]
msirek: Discussed this briefly with @yxia92.  This modification will be postponed until SPLICE-2287.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/AndNode.java
```diff
@@ -307,4 +307,40 @@ public boolean verifyChangeToCNF(){
...
+            return inListOp1.isEquivalent(inListOp2);
+        }
+        return false;
+    }
+    
+    /**
+     * Are the current node and next node in the AND chain both IN list probe
+     * predicates deriving from the same original IN list?
+     */
+    boolean isNextAndedPredInSameMultiprobeSet() throws StandardException {
```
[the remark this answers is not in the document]
msirek: Removed isGroupedMultiprobePreds as it is never used.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -773,41 +847,224 @@ private void orderUsefulPredicates(Optimizable optTable,
...
+            ValueNodeList vnl = (ValueNodeList)getNodeFactory().getNode(
+                                 C_NodeTypes.VALUE_NODE_LIST,
+                                  getContextManager());
+            ValueNodeList groupedConstants =
+                   (ValueNodeList) getNodeFactory().getNode(
+                                   C_NodeTypes.VALUE_NODE_LIST,
+                                   getContextManager());
+            boolean multiColumnInListBuilt = false;
+            if (optTable instanceof FromBaseTable &&
+                (!((FromBaseTable) optTable).isSpark(((FromBaseTable) optTable).getDataSetProcessorType()) ||
```
yxia92: The decision of whether to go for Spark path or not does not depend on one step. It is possible that for this FromBaseTable node, we don't choose spark path, but another FromBaseTable node would choose spark path. The final decision is made after all the steps in the plan have been checked. The check here may not block the multi-probe scan completely for multiple inlists for Spark path.

yxia92: The dataset processor type finally used is the one tracked in the current CompilerContext.dataSetProcessorType. During the code generation phase, as we go through each operation, this variable will be updated. Please see the call in FromBaseTable.getScanArguements()->acb.setDataSetProcessorType(dataSetProcessorType);

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -1349,19 +1608,34 @@ void pullExpressions(int numTables,ValueNode searchClause) throws StandardExcept
...
-                addPredicate(newPred);
+                topAnd = (AndNode) topAnd.getRightOperand();
+                if (!thisAnd.isNextAndedPredInSameMultiprobeSet()) {
+                    /* Set the rightOperand to true */
+                    thisAnd.setRightOperand(trueNode);
+
+                    /* Add the top AndNode to the PredicateList */
+                    newJBitSet=new JBitSet(numTables);
+                    newPred=(Predicate)getNodeFactory().getNode(C_NodeTypes.PREDICATE,thisAnd,newJBitSet, contextManager);
+                    addPredicate(newPred);
```
yxia92: This logic does not look right, thisAnd node here could be part of the AndedPredInSameMultiprobeSet, but we are separating it out. This may explain why in the following explain we are seeing two duplicate inlists:
```
create table t1 (a1 int, b1 int, c1 int, d1 int, primary key(a1,b1));
splice> explain select * from t1 where a1=1 and b1=10 or a1=2 and b1=20;
Plan                                                                                                                                                                                                                                                            
Cursor(n=4,rows=1,updateMode=READ_ONLY (1),engine=control)                                                                                                                                                                                                      
  ->  ScrollInsensitive(n=3,totalCost=8.046,outputRows=1,outputHeapSize=2 B,partitions=1)                                                                                                                                                                       
    ->  ProjectRestrict(n=2,totalCost=4.04,outputRows=1,outputHeapSize=2 B,partitions=1,preds=[((A1[0:1],B1[0:2]) IN ((1,10),(2,20))),((A1[0:1],B1[0:2]) IN ((1,10),(2,20)))])                                                                                  
      ->  TableScan[T1(1648)](n=1,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=2 B,partitions=1)                                                                                                                                                  

4 rows selected
```

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/InListMultiprobeIT.java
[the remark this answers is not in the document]
yxia92: I see, I overlooked that one test case has the database property change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: Yes, I've corrected the logic, thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: Opened [SPLICE-2289](https://splice.atlassian.net/browse/SPLICE-2289) for this issue.

msirek:

spliceci:
TEST SUCCEEDED +1

dgomezferro:
Regarding the Spark bit, even if it's slower I feel like the example used to showcase it performs poorly is not a real scenario since it's a very small table, so the overhead of creating multiple RDDs really shows. In a real query that gets selected for Spark execution the overhead wouldn't be as noticeable, and having the MultiProbe could prove much better. So my suggestion would be to remove the check for Spark execution and always use the MultiProbe if the requirements are met.

msirek:
@dgomezferro Thanks, I had previously tested some cases with millions of rows and also observed regression, though the number of qualified rows was very low.  I will do more testing and figure out the best resolution through [SPLICE-2289](https://splice.atlassian.net/browse/SPLICE-2289).  Maybe there is even an option to reduce the number of RDDs used for MultiProbe, and reduce the overhead.

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3436/

PR #2563, opened 2019-01-04
DB-7684 fix costing issue when JoinNode is served as the inner table of another join

yxia92:
The PR fixes a costing issue and some explain display issue(details please see [DB-7684](https://splicemachine.atlassian.net/browse/DB-7684). I separate the code into two commits. The first one is the fix for the costing issue. The second one contains the fix for the explain display issue,most files changes are just interface change. I also incorporated review comment for inlist selectivity from #2548 .


spliceci:
TEST SUCCEEDED +1

PR #2565, opened 2019-01-08
DB-7646: fix properties issues in custom ambari service in hdp2.6.4 of branch-2.7

liprais:
fix hdp version and add props for backup functions:
1.fixed hdp version for hdp2.6.4
2.added props for backup functions to work
3.remove ranger coprocessor from "hbase.coprocessor.regionserver.classes"


liprais:
@wb14123 I haven't review the configs for a while, but I guess so.

PR #2570, opened 2019-01-09
DB-6830:Enable ldap server in distributed standalone 

liprais:
usage: ./start-splice-cluster -pcdh5.12.2,ee -f to start the cluster.
when starting a full distributed standalone cluster, the ee key will be read from /platform_it/src/test/resources/ee.txt ,which has been added to git ignore file.
if the ee key doesn't exists ,the cluster will start without the ldap server and use native auth.
if the ee key has been setup correctly ,a full kerberos + ldap full standalone cluster will start

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,3 +221,4 @@ tramp
 # cask packages
 .cask/
 
+/platform_it/src/test/resources/ee.txt
--- a/platform_it/src/test/java/com/splicemachine/test/SpliceTestKDCPlatform.java
+++ b/platform_it/src/test/java/com/splicemachine/test/SpliceTestKDCPlatform.java
@@ -15,11 +15,15 @@
 
 package com.splicemachine.test;
 
+import org.apache.directory.server.core.api.DirectoryService;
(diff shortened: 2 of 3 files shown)


OlegMazurov:
Why does LDAP authentication depend on the EE key? Open source users who don't have access to the EE repository should be able to use it. If we want LDAP to be the default for the distributed standalone, that should be done in 'secure/insecure' profiles.
 
What is the need to provide the EE key in a separate file? If the EE repository is present, one can call SYSCS_UTIL.SYSCS_ENABLE_ENTERPRISE('EE-KEY'). That can be [semi-]automated.
Am I missing anything?

genedavis:
Currently we only support LDAP with the Enterprise Edition

genedavis:
(For business reasons not technical)

liprais:
@OlegMazurov 
1.the dependencies  of ee repo and ee key is as @genedavis says ,a business reason.
2.in order to setup ldap auth, the ee key has to passed in in the startup process, which is before splice machine can process queries, so we cannot enable ee function by calling a system procedure.

PR #2576, opened 2019-01-15
DB-7812 Update Copyright Dates

bklo94:
Command I ran 
`find . -path *test-data* -prune -o \( -name '*.txt' -o -name '*.java' -o -name '*.xml' -o -name '*.scala' -o -name '*.md' \) -print0 | xargs -0 sed -i "" "s/- 2018 Splice Machine, Inc./- 2019 Splice Machine, Inc./g"`

While I was at it I also updated some old 2017 and 2016 copyrights that were left hanging around.


changli6:
Please also update the files in following directories. Most of them are *.properties.

hbase_sql/src/main/resources
hbase_pipeline/*/src/main/resources
hbase_storage/*/src/main/resources
db-drda/src/main/properties
db-drda/src/main/resources

bklo94:
Closing and reopening with another branch, since Murray noticed that if I use grep instead of find. It doesn't rip off the newlines.

bklo94:
Closing and reopening with another branch, since Murray noticed that if I use grep instead of find. It doesn't rip off the newlines.

PR #2581, opened 2019-01-15
DB-7612 Remove master installation for Splice Ambari serivce

wb14123:


assembly/hdp2.6.1/src/main/resources/common-services/SPLICEMACHINE/2.5.1/package/scripts/splice_install.py
```diff
@@ -21,7 +21,21 @@ def install(self, env):
   def configure(self, env):
     import params
     print 'Configure the client'
+
+    hbase_user = params.config['configurations']['hbase-env']['hbase_user']
+
+    params.HdfsResource("/user/splice",
```
MurrayBrown: what guarantees that this only runs once?  

perhaps what i am asking is: if this is run for each and every region server, what guarantees this doest error the second time?

wb14123: Sorry for the late response, I've tested this on the cluster and this command can run multiple times without error. The implementation of `HdfsResource` should have handled the situation when there is already a directory there.

MurrayBrown: sounds good

PR #2585, opened 2019-01-15
DB-7476: Make bulk import sampling rate a database property

jyuanca:


jyuanca:
jenkins please test branch @cdh5.13.0

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkInsertDataSetWriter.java
```diff
@@ -194,8 +197,17 @@ public TxnView getTxn(){
...
             if (!skipSampling) {
                 if (sampleFraction == 0) {
-                    SConfiguration sConfiguration = HConfiguration.getConfiguration();
-                    sampleFraction = sConfiguration.getBulkImportSampleFraction();
+                    LanguageConnectionContext lcc = activation.getLanguageConnectionContext();
+
+                    String sampleFractionString = PropertyUtil.getCachedDatabaseProperty(lcc,
+                            Property.BULK_IMPORT_SAMPLE_FRACTION);
+                    if (sampleFractionString != null) {
+                        sampleFraction = Double.parseDouble(sampleFractionString);
```
msirek: If the user sets the splice.bulkImport.sample.fraction to a value which is not a valid Double, parseDouble will throw an error and the import will fail.  Maybe add a try block to ignore bad settings by the user:
try {
    sampleFraction = Double.parseDouble(sampleFractionString);
}
catch (Exception e) {
                        SConfiguration sConfiguration = HConfiguration.getConfiguration();
                        sampleFraction = sConfiguration.getBulkImportSampleFraction();
}

Also, this code is repeated in a few places.  It may be good to create a new method for it.

jyuanca:
@msirek Good suggestions, thanks!

PR #2587, opened 2019-01-15
DB-7476: Make bulk import sampling rate a database property(2.5)

jyuanca:


jyuanca:
It can help existing 2.5 customers to change the property w/o restarting splice.

PR #2592, opened 2019-01-17
DB-7847: flush should not block backup(2.5)

jyuanca:


MurrayBrown:
I am aware of no customer issues related to this ticket, so I do not think we need this on the 2.5 branch.

PR #2598, opened 2019-01-22
DB-7229: remove unused backup system tables and procedures

jyuanca:


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
```diff
@@ -230,17 +230,15 @@
     int SYSPERMS_CATALOG_NUM=20;
     int SYSUSERS_CATALOG_NUM=21;
     int SYSBACKUP_CATALOG_NUM=22;
-    int SYSBACKUPFILESET_CATALOG_NUM=23;
```
MurrayBrown: is it better to remove these, or leave the integers in the schema as 'unused placeholders', to reduce reverse compatibility problems after an upgrade?

jyuanca: Good question. We never dropped system tables before. I'm not aware of any consequences if we modify existing table number. I've run upgrade tests, and it looks good to me.

MurrayBrown: My suggestion is to leave the unchanged numbers alone, i.e. remove '23 and 24', but leave the numbers 'the same' for all the others 25 through 33, so we'd end the list on:
  SYSSNAPSHOT_NUM=32; and SYSTOKENS_NUM=33;

In this way, we wont have a problem accessing any records written to a pre-existing database that has already stored a data value based on the integers 25 through 33.  

Either that, or I am not understanding the data dictionary correctly!

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
MurrayBrown: Agreed, I think i'm happier with that sort of approach (just to be safe?).

jyuanca:

jyuanca:
Please also review code changes on ee branch

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1163,4 +1115,31 @@ public void upgradeSysColumnsWithUseExtrapolationColumn(TransactionController tc
             SpliceLogUtils.info(LOG, "SYS.SYSCOLUMNS upgraded: added column: USEEXTRAPOLATION.");
         }
     }
+
+    public void removeUnusedBackupTables(TransactionController tc) throws StandardException {
+        SchemaDescriptor sd = getSystemSchemaDescriptor();
+        TableDescriptor td = getTableDescriptor("SYSBACKUPFILESET", sd, tc);
+        if (td != null) {
+            dropTableDescriptor(td, sd, tc);
```
yxia92: Is it enough to just drop the table descriptor if we want to remove the unused backup tables? what about the columns of these tables, and the conglomerate as well as index conglomerates on these tables? Looking at DropTableConstantOperation.executeConstantAction(), dropping a user table seems to have more things to do. Is it similar for dictionary table? Thanks.

jyuanca: Good catch!

spliceci:
TEST SUCCEEDED +1

jyuanca:

spliceci:
TEST SUCCEEDED +1

PR #2618, opened 2019-02-08
DB-7817 Autocast char to decimal.

msirek:
Allow numeric expressions to be compared with CHAR and VARCHAR expressions, for example:
select * from t1 where numeric_column = '12345';

More details in [DB-7817](https://splicemachine.atlassian.net/browse/DB-7817)

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
@@ -625,15 +625,15 @@ public class BroadcastJoinIT extends SpliceUnitTest {
     }
 
     @Test
-    public void testUncomparableCharColumnsBroadCastJoin() throws Exception {
+    public void testComparableCharColumnsBroadCastJoin() throws Exception {
         String sqlText = format("select count(s1.a1) from " + s1 +
                 " inner join " + s2 + " --SPLICE-PROPERTIES joinStrategy=BROADCAST,useSpark=%s \n" +
                 " on s1.b1 = s2.a2" , useSpark
         ) ;
+
+        // The following query with CHAR/INT comparison should not throw an error.
         try (ResultSet rs = classWatcher.executeQuery(sqlText)) {
 
-        } catch (SQLSyntaxErrorException e) {
-            Assert.fail("Exception should not be thrown.");
         }
     }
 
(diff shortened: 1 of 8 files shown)


msirek:

spliceci:
TEST FAILED

msirek:

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
```diff
@@ -631,9 +631,9 @@ public void testUncomparableCharColumnsBroadCastJoin() throws Exception {
                 " on s1.b1 = s2.a2" , useSpark
```
dgomezferro: Please change the name of this test since it's no longer 'uncomparable'

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
```diff
@@ -631,9 +631,9 @@ public void testUncomparableCharColumnsBroadCastJoin() throws Exception {
...
         ) ;
         try (ResultSet rs = classWatcher.executeQuery(sqlText)) {
-            Assert.fail("Exception not thrown");
+
         } catch (SQLSyntaxErrorException e) {
```
dgomezferro: You can remove the catch if a raised exception is cause for failure

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
msirek: Changed Uncomparable to Comparable.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
msirek: Removed the catch.

spliceci:
TEST SUCCEEDED +1

yxia92:
This implicit casting may impact the selectivity logic similar to SPLICE-2250, as the selectivity logic is not aware of the cast?

msirek:
@yxia92 Regarding selectivity estimation, it looks like we are skipping over the CAST in BinaryRelationOperatorNode.getCompareValue.  This doesn't seem correct because what if we cast something like 12.3 to Dec(3,0)... it should get converted to 12.0, but if we ignore the cast, selectivity logic would use the 12.3 for comparison.  Also, do we even have constant folding logic?  Maybe constant folding needs to be added to turn arbitrarily complex constant expressions into simple constants.  

Also, just to mention, the casting added by this JIRA could also be done explicitly in SQL: CAST ('12345' AS DOUBLE) = 12345.  So, any selectivity issues present would apply to this case as well.

yxia92:
Yes, @msirek We are on the same page as I've mentioned in my updated comment. We should open a separate ticket to address the casting issue. Thanks.

yxia92:
@msirek I opened SPLICE-2305 to track the selectivity estimation issue for predicate with casting.

PR #2620, opened 2019-02-12
DB-7066 Implement show create table command and procedure

changli6:


spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
```diff
@@ -0,0 +1,273 @@
...
+            Assert.fail("Expected to fail");
+        }
+        catch (SQLException e)
+        {
+            String sqlState = e.getSQLState();
+            Assert.assertEquals("42X05",sqlState);
+        }
+    }
+
+}
```
msirek: This looks like it will be a very useful feature.  Here are some general comments:

1. Could we perhaps print a semi-colon ';' at the end of the CREATE TABLE output, so the output of SHOW CREATE TABLE commands could be run from another sqlshell script without having to manually add them?

2.  The output of the primary key columns does not always follow the same order as the original CREATE TABLE, which means the output doesn't show the correct primary key.  Example:

create table t2 (a int, b int, primary key(b,a));
splice> show create table t2;
DDL
---
CREATE TABLE "SPLICE"."T2" (
"A" INTEGER NOT NULL
,"B" INTEGER NOT NULL
, CONSTRAINT SQL190213163554700 PRIMARY KEY(A,B)) <--- should be (B,A)

3.  The output does not show any of the index DDL for indexes on the table, so we would still need to manually reconstruct the CREATE INDEX statements, so limits it's usefulness in setting up tests based on customers' tables.  Is there a possibility of adding the CREATE INDEX statements to the output?  If this JIRA is going on too long, maybe open another JIRA for it.  SHOW CREATE VIEW would be another good one.

changli6: Thanks Mark reviewing!
1. Add a semi-colon at the end of DDL
2. It is a bug to be addressed. Good test.
3. We could file new Jiras for INDEX and VIEW.

changli6:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2145,4 +2146,349 @@ public static void INVALIDATE_GLOBAL_DICTIONARY_CACHE() throws Exception {
...
+            Connection connection = getDefaultConn();
+            Statement stmt = connection.createStatement();
+
+            EngineUtils.verifyTableExists(connection, schemaName, tableName);
+
+            ResultSet tableIdRs = stmt.executeQuery("SELECT T.TABLEID, T.TABLETYPE, T.COMPRESSION, T.DELIMITED, " +
+                    "T.ESCAPED, T.LINES, T.STORED, T.LOCATION" +
+                    " FROM SYS.SYSTABLES T, SYS.SYSSCHEMAS S " +
+                    "WHERE T.TABLETYPE IN ('T','E') AND T.SCHEMAID = S.SCHEMAID " +
+                    "AND T.TABLENAME = '" + tableName + "' AND S.SCHEMANAME = '" + schemaName + "'");
```
yxia92: the code change in ij.jj suggests that we could have commands like "show create table schema_name.*;", but this query does not seem to handle that.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2145,4 +2146,349 @@ public static void INVALIDATE_GLOBAL_DICTIONARY_CACHE() throws Exception {
...
+                    "FROM SYS.SYSCOLUMNS C, SYS.SYSTABLES T WHERE T.TABLEID = ? " +
+                    "AND T.TABLEID = C.REFERENCEID AND C.PARTITIONPOSITION > -1 ORDER BY C.PARTITIONPOSITION");
+
+            String tableId = "" ;
+            boolean firstCol = true;
+
+            //External Table
+            String isExternal = "";
+            StringBuilder extTblString = new StringBuilder("");
+            while (tableIdRs.next()){
```
yxia92: The while loop suggests that we could be handling more than one table here. If we have more than one external tables, the specifics for the multiple external tables will be lumped together. Or maybe what you want is an IF instead of WHILE here?

db-drda/src/main/properties/metadata.properties
```diff
@@ -716,8 +719,9 @@ getImportedKeys=\
           SELECT PK.CONGLOMERATEID from SYS.SYSPRIMARYKEYS PK where PK.CONSTRAINTID = C.CONSTRAINTID \
     ) = CONGLOMS.CONGLOMERATEID \
     AND F.KEYCONSTRAINTID = C.CONSTRAINTID \
-    and FK_INFO.FK_PID = F.KEYCONSTRAINTID \
-    AND (CASE WHEN CONGLOMS.DESCRIPTOR IS NOT NULL THEN CONGLOMS.DESCRIPTOR.getKeyColumnPosition(COLS.COLUMNNUMBER) ELSE 0 END) <> 0 \
+    AND FK_INFO.FK_PID = F.KEYCONSTRAINTID \
+    AND FK_INFO.FK_ID = F.F.CONSTRAINTID \
```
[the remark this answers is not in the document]
changli6: Oops, a typo.  Here is a parser's bug allowing this wrong syntax. Filed a ticket.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
changli6: show create table schema_name.* is intending support. 
Clean these semi-finished codes and implement in a separate ticket.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
changli6: Also intending to support multiple tables.
Change the WHILE to IF  and reorganize scope.

PR #2623, opened 2019-02-13
DB-7854 Deadlock in EmbedPreparedStatement.getMetaData()

OlegMazurov:
Refactor to avoid an endless loop in a critical section.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
@@ -1120,20 +1120,21 @@ public abstract class EmbedPreparedStatement extends EmbedStatement implements E
                 //Second check - if the statement was revalidated since last getMetaData call,
                 //then gcDuringGetMetaData wouldn't match with current generated class name
 
-                GeneratedClass currAc;
-                ResultDescription resd;
+                GeneratedClass currAc = null;
+                ResultDescription resd = null;
 
-                synchronized(execp) {
-                    // DERBY-3823 Some other thread may be repreparing
-                    do {
-                        while (!execp.upToDate()) {
+                // DERBY-3823 Some other thread may be repreparing
+                do {
+                    synchronized (execp) {
+                        if (!execp.upToDate()) {
                             execp.rePrepare(lcc);
                         }
-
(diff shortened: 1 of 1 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
```diff
@@ -1120,20 +1120,21 @@ public void clearParameters() throws SQLException {
...
-
-                        currAc = execp.getActivationClass();
-                        resd = execp.getResultDescription();
-                    } while (currAc == null);
-                }
+                        else {
+                            currAc = execp.getActivationClass();
+                            resd = execp.getResultDescription();
+                        }
+                    }
```
msirek: Should we put a short Thread.sleep() time in this loop so it's not consuming CPU if the blocking thread takes a long time:  
if (currAc == null)
    Thread.sleep(x);

OlegMazurov:
@dgomezferro 
> Are you going to address the lookupStatement()/putIfAbsent() issue too?
Not in this PR. That change is too intrusive: all caches in DataDictionaryCache should be evaluated, ManagedCache should expose guave.cache.Cache.get(K, Callable<V>), which provides similar to putIfAbsent functionality but doesn't combine well with dd.canWriteCache/canReadCache checks.
Although the current behavior of lookupStatement() helps to create conditions for the deadlock I don't think it's the root cause. From that perspective, cache rework deserves a separate JIRA.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
OlegMazurov: All operations are still executed in a synchronized block so it's not busy waiting.
There still is a chance it continuously fails to rePrepare execp due to broken logic somewhere else (not because other threads updating execp are blocked -- that's fixed here). In that case, I'd prefer it to burn CPU so that it's easily visible (I used this symptom to quickly check whether my benchmark runs were affected in AWS) rather than be hidden behind a sleep, which wouldn't help with that broken logic anyway.

msirek: OK

PR #2625, opened 2019-02-13
SPLICE-2291 Avoid slow IndexLookups when row count estimates are bad.

msirek:
This fix decouples the selected max concurrency of index row lookups from optimizer row count estimates.  For large number of rows, we should still allow concurrency, even if the optimizer predicted we have less rows.

More info in: [SPLICE-2291](https://splice.atlassian.net/browse/SPLICE-2291?oldIssueView=true)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/IndexRowReader.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/IndexRowReader.java
@@ -91,16 +91,20 @@ public class IndexRowReader implements Iterator<ExecRow>, Iterable<ExecRow>{
         this.outputTemplate=outputTemplate;
         this.txn=txn;
         batchSize=lookupBatchSize;
-        this.numBlocks=numConcurrentLookups;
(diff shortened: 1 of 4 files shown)


msirek:

msirek:
@MurrayBrown OK, I can add a test that verifies runtime of a large test case is less than a certain amount when ANALYZE TABLE is not run.

MurrayBrown:
if the test case has to have a 'large' quantity of data, then that might be more of an 'IT' vs a Unit test, but I wasnt necessarily hoping for a test that is large.  Testing based on timing sounds... tenuous at best.

Is there some small test case which proves that two threads are used? what is the smallest possible test to cover this change?

msirek:
For using an index, I guess it would be very difficult to write a true unit test (as opposed to an IT), because there is no test driver to generate valid splice tables outside of running splice SQL, but I could create a simple one where the IndexRowReader class is instantiated with a request for 1 thread, and then check that the class has a max concurrency setting of 2.

spliceci:
TEST SUCCEEDED +1

PR #2627, opened 2019-02-14
DB-7931: change export default compression format to bz2

jyuanca:


jyuanca:

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportOperationIT.java
```diff
@@ -14,6 +14,7 @@
 
 package com.splicemachine.derby.impl.sql.execute.operations.export;
 
+import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
```
MurrayBrown: surely, if we are adding this, we can remove the GZipInputStream on line 38?

spliceci:
TEST SUCCEEDED +1

genedavis:
I second the motion of making it an option.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportOperation.java
```diff
@@ -208,7 +208,15 @@ public void writeExternal(ObjectOutput out) throws IOException {
             case "parquet": {
                 OperationContext<?> writeContext = dsp.createOperationContext(this);
                 long start = System.currentTimeMillis();
-                dataset.writeParquetFile(dsp, new int[]{}, exportParams.getDirectory(), exportParams.isCompression() ? "snappy" : "none", writeContext);
+                String compression = null;
+                if (exportParams.getCompression() == ExportFile.COMPRESSION.SNAPPY) {
```
dgomezferro: We should handle gz and bzip2 here too

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportParams.java
```diff
@@ -117,8 +117,27 @@ private void setFormat(String format) throws StandardException {
         this.format = format;
     }
 
-    private void setCompression(Boolean compression) throws StandardException {
-        this.compression = compression;
+    private void setCompression(String compression) throws StandardException {
```
dgomezferro: This breaks the current uses of Export(), we should handle true/false too

db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ResultSetFactory.java
```diff
@@ -1780,7 +1780,7 @@ NoPutResultSet getExportResultSet(NoPutResultSet source,
 									  Activation activation,
 									  int resultSetNumber,
 									  String exportPath,
-									  boolean compression,
```
dgomezferro: Please make this change for Binary_export too

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportParams.java
jyuanca: We cannot keep the current export() API and introduce different compression options while still keeping true/false.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportOperation.java
jyuanca: Parquet does not support bzip2 compression

jyuanca:

spliceci:
TEST SUCCEEDED +1

dgomezferro:
Looks good to me! Thanks for the updates +1

PR #2629, opened 2019-02-14
DB-7693 Batched inserts using JDBC batch API

dgomezferro:
Please see https://splicemachine.atlassian.net/browse/DB-7693 for more details


dgomezferro:

spliceci:
TEST FAILED

dgomezferro:

spliceci:
TEST SUCCEEDED +1

dgomezferro:

spliceci:
TEST SUCCEEDED +1

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -4549,7 +4551,13 @@ private boolean parseEXCSQLSTTobjects(DRDAStatement stmt) throws DRDAProtocolExc
 		
 		if (! gotEXTDTA) {
 			stmt.ps.clearWarnings();
-			result = stmt.execute();
+			if (updateCounts.length > 1) {
+				int[] results = stmt.executeBatch();
```
yxia92: For batch execution, the returned value result is not set, so it goes with the default value of false, is this desired behavior?

yxia92: Also at line 4534 below:
					case CodePoint.EXTDTA:	
						readAndSetAllExtParams(stmt, true);
						stmt.ps.clearWarnings();
						result = stmt.execute();   <--- line 4534
						gotEXTDTA = true;
						break;
It handles execution of single statement, do we need to worry about executeBatch() here?

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
```diff
@@ -1061,6 +1062,71 @@ public void addBatch() throws SQLException {
         }
     }
 
+    @Override
+    public int[] executeBatch() throws SQLException {
+        checkExecStatus();
+        synchronized (getConnectionSynchronization()) {
+            setupContextStack();
+            int i = 0;
```
yxia92: What is "I" here for? It doesn't seem to be very useful.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
```diff
@@ -1061,6 +1062,71 @@ public void addBatch() throws SQLException {
...
+
+                sqle = handleException(se);
+            } catch (SQLException sqle2) {
+                sqle = sqle2;
+            } finally {
+                activation.setBatch(null, null); // reset the batch state
+                restoreContextStack();
+            }
+
+            int successfulUpdateCount[] = new int[i];
```
yxia92: I think "I" here should be changed to "r".

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
dgomezferro: Yep, bad copy paste.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedPreparedStatement.java
dgomezferro: Currently for batched execution we fail all statements, so I've set this explicitly to an empty array.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
dgomezferro: result should be false, yes, since batched execution is only for DML statements that don't have a result set.

Regarding line 4534, I'm not sure what role that plays or when that gets executed. We could add logic to handle batched execution here but I'm not sure how to test it.

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSetWriter.java
```diff
@@ -101,6 +104,18 @@ public ControlDataSetWriter(ControlDataSet<ExecRow> dataSet,AbstractPipelineWrit
                     }
                 }
             }
+            if (updateCounts != null) {
+                int total = 0;
```
msirek: recordsWritten is a long, so total should probably also be a long.

PR #2631, opened 2019-02-14
SPLICE-2306 NPE in query with multiple AND'ed and OR'ed IN lists.

msirek:
There are some problems with the part of DB-7621 which converts predicates like (a = 1 and b = 2) or (a = 3 and b = 4) into (a,b) in ((1,2), (3,4)):

1. There was no checking in canConvertToInList that we found at least one column in the OR'ed predicate.

2. There was no checking that we have IN list probe predicates when more than one column is in the predicates. Predicates such as "a in (1,2) or a in (3,4)" can be combined, but the code wasn't meant to handle terms such as "a1 in (1,2) and b1 in (10,20) or a1 in (1,3) and b1 in (30,40)"

3. There was no checking that a column had already been added to the columnNumbers list and columns HashMap, and so we get duplicates sometimes, when each unique column is only supposed to be added once.

I have added the above checks plus some more test cases to the IT.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OrNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OrNode.java
@@ -34,6 +34,7 @@ package com.splicemachine.db.impl.sql.compile;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.sql.compile.C_NodeTypes;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.apache.commons.lang3.mutable.MutableInt;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OrNode.java
```diff
@@ -145,12 +157,14 @@ else if (vn instanceof BooleanConstantNode)
...
-				return false;
+			if (!bron.getRightOperand().isConstantOrParameterTreeNode() ||
+				(bron.isInListProbeNode() && columnNumbers.size() > 1))
+			return false;
 		}
         else if (bron.getRightOperand() instanceof ColumnReference) {
 			cr = (ColumnReference) bron.getRightOperand();
-			if (!bron.getLeftOperand().isConstantOrParameterTreeNode())
+			if (!bron.getLeftOperand().isConstantOrParameterTreeNode() ||
+				(bron.isInListProbeNode() && columnNumbers.size() > 1))
```
yxia92: Mark, at the check, we may not have seen all the columns yet. Can the fix handle the following scenario:
a1 in (1,2,3) and b1 =1 or a1=4 and b1=2?

msirek: Good find Yi.  I think I was contemplating a similar scenario last night, but it was quite late and I ended up missing it.  I pushed a commit to fix this case by making canConvertToInList process the first term twice.  Once with FirstTime == true, and then with FirstTime == false.  I also pushed some logic outside of the function into the function, because that's really where it belongs.

msirek: Refined the fix to avoid calling canConvertToInList twice for the first OR term.

PR #2634, opened 2019-02-15
DB-7819 Resubmit to Spark when resub.U is raised

dgomezferro:


dgomezferro:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
```diff
@@ -431,7 +431,11 @@ public void openCore(DataSetProcessor dsp) throws StandardException{
             resubmitDistributed(e);
         }catch(Exception e){ // This catches all the iterator errors for things that are not lazy
             checkInterruptedException(e);
-            throw Exceptions.parseException(e);
+            StandardException se = Exceptions.parseException(e);
```
msirek: The fix looks good.  I'm wondering though if there is a need to improve parseException.  It only looks at the direct rootCause exception, but sometimes I've seen stack traces where the original Exception is nested a few layers deep.  Should parseException continue calling getRootCause all the way to the original exception?

PR #2638, opened 2019-02-18
DB-7874 Add locale support for UPPER, LOWER, UCASE and LCASE

wb14123:


wb14123:

spliceci:
TEST SUCCEEDED +1

changli6:
@wb14123 I commented on the ticket [DB-7874](https://splicemachine.atlassian.net/browse/DB-7874)
about the behavior of ß upper and lower case.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
```diff
@@ -2787,6 +2788,65 @@ public StringDataValue upper(StringDataValue result)
...
+
+        if (rightOperand == null || rightOperand.isNull() || rightOperand.getString() == null) {
+            throw StandardException.newException(
+                    SQLState.LANG_INVALID_FUNCTION_ARGUMENT, "NULL", isUpper?"UPPER":"LOWER");
+        }
+        String localeStr = rightOperand.getString();
+        Locale locale = Locale.forLanguageTag(localeStr);
+        if (!validLocale(locale)) {
+            throw StandardException.newException(
+                    SQLState.LANG_INVALID_FUNCTION_ARGUMENT, localeStr, isUpper?"UPPER":"LOWER");
```
msirek: splice> values UPPER('abc', 'a');
ERROR 22008: 'a' is an invalid argument to the UPPER function.

.... would it be useful to give more information here, like "'a' is not a valid locale." ?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SimpleLocaleStringOperatorNode.java
```diff
@@ -0,0 +1,164 @@
...
+		/*
+		** The result type of upper()/lower() is the type of the operand.
+		*/
+
+		setType(new DataTypeDescriptor(operandType,
+				leftOperand.getTypeServices().isNullable(),
+				leftOperand.getTypeCompiler().
+					getCastToCharWidth(leftOperand.getTypeServices())
+						)
+				);
```
msirek: This code is the reason the following SQL doesn't display properly:
`             values upper('ß', 'de-DE');`
Since the String result isn't created as unicode, it converts to two uppercase S's.  So, the result string could be up to twice as long as the source.  So we need to do something like the following instead:

          TypeId newTypeId = null;
		if (leftOperand.getTypeCompiler().
			getCastToCharWidth(leftOperand.getTypeServices()) * 2 > Limits.DB2_LOB_MAXWIDTH) {
			// throw new StandardException();
		}
		int maxWidth = leftOperand.getTypeCompiler().
			           getCastToCharWidth(leftOperand.getTypeServices()) * 2;

		if (maxWidth > Limits.DB2_VARCHAR_MAXWIDTH) {
			newTypeId = TypeId.getBuiltInTypeId(TypeId.CLOB_NAME);
		}
		else {
			newTypeId = TypeId.getBuiltInTypeId(TypeId.VARCHAR_NAME);
		}

		setType(new DataTypeDescriptor(newTypeId, true, maxWidth));

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
wb14123: Exception with SQLState.LANG_INVALID_FUNCTION_ARGUMENT give error information like this. I think it is better to not add a new SQLState.

wb14123:
@msirek Thanks for reviewing the code, pushed new commits to address that.

wb14123:

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ProjectRestrictOperationIT.java
```diff
@@ -312,6 +314,20 @@ public void testUpperLowerWithLocale() throws Exception{
         rs = methodWatcher.executeQuery("values lcase('I', 'tr-TR')");
         rs.next();
         Assert.assertEquals("lower incorrect with tr-TR", "ı", rs.getString(1));
+        Assert.assertEquals("lower return type incorrect", Types.CHAR,
+                rs.getMetaData().getColumnType(1));
+
+        String from = "";
+        String to = "";
+        for (int i = 0; i < Limits.DB2_CHAR_MAXWIDTH - 1; i++) {
```
msirek: Use Limits.DB2_VARCHAR_MAXWIDTH here.

spliceci:
TEST SUCCEEDED +1

wb14123:
@msirek I find it is a little tricky to extend the string length. I've re-implement the result type based on these rules:

1. If `width*2` still in the limit of current type, then remain the current type.
2. If `width*2` exceeds current type's limit, cast the type into a longer length type.

wb14123:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -7405,22 +7405,47 @@ characterValueFunction() throws StandardException :
         return getReplaceNode( value, str1, str2 );
 	}
 |
-	( upperTok = <UPPER> | lowerTok = <LOWER> ) <LEFT_PAREN> value = additiveExpression(null,0) <RIGHT_PAREN>
+	( upperTok = <UPPER> | lowerTok = <LOWER> ) <LEFT_PAREN> str1 = additiveExpression(null,0) [ <COMMA> str2 = additiveExpression(null,0) ] <RIGHT_PAREN>
```
dgomezferro: Please use more descriptive names, I think str1 is value and str2 locale, is that right?

wb14123: @dgomezferro Pushed new commit

spliceci:
TEST SUCCEEDED +1

msirek:
@wb14123 OK, I'm not sure I understand the issues with casting.  Limits.DB2_CHAR_MAXWIDTH is only 254, so with the code how it currently is, we will cast to CLOB if width*2 is more than 254.  But we don't need to use CLOB unless width*2 goes above Limits.DB2_VARCHAR_MAXWIDTH, right?

wb14123:
@msirek We will cast it to VARCHAR if it exceeds DB2_CHAR_MAXWIDTH, cast it to LONGVARCHAR if it exceeds DB2_VARCHAR_MAXWIDTH, cast it to CLOB if it exeeds DB2_LONGVARCHAR_MAXWIDTH.

wb14123:
@msirek To make it clear, I've pushed new commits after your second review.

msirek:
@wb14123 OK, sorry, I did not see the new changes.   Code looks good, thanks.

PR #2641, opened 2019-02-19
DB-7876 Support STRIP() function

changli6:


changli6:

spliceci:
TEST SUCCEEDED +1

changli6:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2589,12 +2589,14 @@ TOKEN [IGNORE_CASE] :
...
 |	<ANY: "any">
 |	<ARE: "are">
+|	<ARRAY: "array">
 |	<AS: "as">
 |	<ASC: "asc">
 |	<ASSERTION: "assertion">
 |	<AT: "at">
 |	<AUTHORIZATION: "authorization">
 |	<AVG: "avg">
+|	<B: "b">
```
yxia92: We should put "b", "L", "T" in the non-reserved keyword section.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
```diff
@@ -84,8 +84,9 @@
 	public static final int TIMESTAMPADD = 4;
 	public static final int TIMESTAMPDIFF = 5;
 	public static final int REPLACE = 6;
-	static final String[] TernaryOperators = {"trim", "LOCATE", "substring", "like", "TIMESTAMPADD", "TIMESTAMPDIFF", "replace"};
-	static final String[] TernaryMethodNames = {"ansiTrim", "locate", "substring", "like", "timestampAdd", "timestampDiff", "replace"};
+	public static final int STRIP = 7;
```
yxia92: This change in this module does not seem necessary. I don't see a case where the operator type will be set to STRIP, as in sqlgrammer.jj, STRIP is eventually generated as a TrimOperatorNode. Please correct me if I missed something.

changli6:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
changli6: Moved them to non-reserved keywords section.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
changli6: Good catch. The STRIP function returns TrimOperatorNode node type with TernaryOperatorNode.TRIM operator, which is the same as TRIM. STRIP related stuff is unnecessary here.

spliceci:
TEST SUCCEEDED +1

PR #2643, opened 2019-02-20
DB-7848: asynchronous replication

jyuanca:


jyuanca:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -68,7 +68,7 @@ else if(v1==v2)
         scripts.put(new Splice_DD_Version(sdd,2,8,0, 1849), new UpgradeScriptToRemoveFKDependencyOnPrivileges(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,2,8,0, 1851), new UpgradeScriptToAddUseExtrapolationInSysColumns(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,2,8,0, 1901), new UpgradeScriptToRemoveUnusedBackupTables(sdd,tc));
-
+        scripts.put(new Splice_DD_Version(sdd,2,8,0, 1905), new UpgradeScriptForReplication(sdd, tc));
```
MurrayBrown: does this need to be modified to match the latest version number as of the time of the merge?

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampServer.java
```diff
@@ -33,7 +33,7 @@
     /**
      * Fixed number of bytes in the message we expect to receive from the client.
      */
-    static final int FIXED_MSG_RECEIVED_LENGTH = 2; // 2 byte client id
+    static final int FIXED_MSG_RECEIVED_LENGTH = 3; // 2 byte client id
```
MurrayBrown: should this comment be adjusted to reflect:  '2 byte client id plus refresh boolean'?

hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
```diff
@@ -246,6 +246,19 @@ public boolean createSpliceHBaseTables(){
                         com.splicemachine.si.constants.SIConstants.SEQUENCE_TABLE_NAME+" created");
             }
 
+            if(!admin.tableExists(TableName.valueOf(namespace, HConfiguration.MASTER_SNAPSHOTS_TABLE_NAME))){
```
dgomezferro: I don't think we need a persisted table to keep this information

hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
```diff
@@ -246,6 +246,19 @@ public boolean createSpliceHBaseTables(){
...
+                HTableDescriptor td=generateNonSITable(HConfiguration.MASTER_SNAPSHOTS_TABLE_NAME);
+                admin.createTable(td);
+                SpliceLogUtils.info(LOG,
+                        HConfiguration.MASTER_SNAPSHOTS_TABLE_NAME + " created");
+            }
+
+            if(!admin.tableExists(TableName.valueOf(namespace, HConfiguration.SLAVE_REPLICATION_PROGRESS_TABLE_NAME))){
```
dgomezferro: Is it possible to keep this info in-memory too? It's just a mapping from regionId to LSN, right?

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampSource.java
```diff
@@ -132,4 +132,14 @@ public synchronized void shutdown() {
             _tc = null;
         }
     }
+
+    @Override
+    public void refresh() {
```
dgomezferro: I don't like the refresh() approach, we should take care of that when we fail over

hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
jyuanca: @dgomezferro I agree we should rethink how to keep these information and process them more efficiently. How about file another Jira as an improvement?

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampSource.java
jyuanca: We need to advance timestamp whenever replica has replicated all WAL edits of a snapshot.

hbase_storage/src/main/java/com/splicemachine/access/hbase/H10PartitionAdmin.java
```diff
@@ -422,6 +422,50 @@ public void markDropped(long conglomId, long txn) throws IOException {
...
+
+            for (HColumnDescriptor hcd : htd.getFamilies()) {
+                hcd.setScope(HConstants.REPLICATION_SCOPE_GLOBAL);
+            }
+            admin.modifyTable(tn, htd);
+            SpliceLogUtils.info(LOG, "enabled replication for table %s", tn);
+        }
+        finally {
+            if (tn != null && !admin.isTableEnabled(tn)) {
+                admin.enableTable(tn);
```
msirek: What if the table was already disabled to begin with?  The act of enabling replication should not enable the table in that case, right?  Also, should we only allow enabling of replication on tables that are currently enabled in HBase, and error out other cases, or is it OK to enable these as well?

hbase_storage/src/main/java/com/splicemachine/access/hbase/H10PartitionAdmin.java
```diff
@@ -422,6 +422,50 @@ public void markDropped(long conglomId, long txn) throws IOException {
...
+
+            for (HColumnDescriptor hcd : htd.getFamilies()) {
+                hcd.setScope(HConstants.REPLICATION_SCOPE_LOCAL);
+            }
+            admin.modifyTable(tn, htd);
+            SpliceLogUtils.info(LOG, "disabled replication for table %s", tn);
+        }
+        finally {
+            if (tn != null && !admin.isTableEnabled(tn)) {
+                admin.enableTable(tn);
```
msirek: Similar to the enable replication case, we should probably not enable tables in HBase that started out as disabled before this method was called.

hbase_storage/hbase1.1.1/src/main/java/com/splicemachine/access/hbase/HBaseConnectionFactory.java
dgomezferro: That'd be fine, yes.

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampSource.java
dgomezferro: I thought we are advancing the timestamp directly in ZK and then calling this method to refresh the Timestamp server, however we don't really need to update the Timestamp server since we don't create new transactions, do we?

dgomezferro:
Does it make sense to only replicate certain tables and not others? What happens if you stop replication for some table and then reenable it? Does it catch up or only replays the new writes since you reenabled it?

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampSource.java
jyuanca: We should allow creating read-only transactions on replica database, because it may be used to serve read requests. At least the DBA may want to run some queries on replica to trouble-shooting replication, check its progress, etc.

jyuanca:
@dgomezferro Unfortunately, it does not catch up. Once a table is disabled from replication, its WAL edits will not be sent to replica until it's re-enabled. By default, we should enable all tables for replication. However, we still should expose the API, in case the user wants to determine which table should be enabled or disabled for replication.

PR #2649, opened 2019-02-22
DB-7992 Improve error handling in TimestampClient

dgomezferro:


dgomezferro:

spliceci:
TEST SUCCEEDED +1

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampClient.java
```diff
@@ -263,10 +282,12 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
                 // We timed out, close the channel so that the next request recreates the connection
                 channel.close();
+                clientCallbacks.remove(clientCallId);
                 
                 doClientErrorThrow(LOG, "Client timed out after %s ms waiting for new timestamp: %s", null, timeoutMillis, callback);
             }
         } catch (InterruptedException e) {
+            clientCallbacks.remove(clientCallId);
             doClientErrorThrow(LOG, "Interrupted waiting for timestamp client: %s", e, callback);
         }
```
yxia92: For the below chunk of code starting at line 298:
        if (timestamp < 0) {
            doClientErrorThrow(LOG, "Invalid timestamp found for client: %s", null, callback);
        }
Do we need to call "clientCallbacks.remove(clientCallId);" also before erroring out?

dgomezferro: I think the only way that could happen is if callback.error() is called (which calls countdown() and would unblock the previous callback.await()) and that only happen at line 253/272, after removing the clientCallId in line 252/271.

However I think it's pretty safe to add a remove call there just in case something changes.

PR #2652, opened 2019-02-22
DB-7952 ContextManager leak in Spark Adapter application driver

OlegMazurov:


OlegMazurov:
With this change, the path that avoided releasing ContextManager is taken under stricter conditions, thus taking care of the leak.
Clearing the entire list of ContextManagers was a bug in the original Derby code but I believe its condition was too strict for that code to ever be executed. Fixing it anyway so that it doesn't raise brows anymore.

PR #2656, opened 2019-03-05
DB-7960 Fix order sensitivity and overflows of aggregations on numerics.

msirek:
Fixes precision errors in SUM and AVG aggregations on floating point data that appears when evaluating rows in different orders.
Fixes unnecessary overflow errors in arithmetic expressions, e.g.
```
splice> values 2000000000 + 2000000000;
ERROR 22003: The resulting value is outside the range for the data type INTEGER.
```
Fixes incorrect results in SUM and AVG aggregations which overflow Long and switch to a Decimal aggregator.

For more details, see [DB-7960](https://splicemachine.atlassian.net/browse/DB-7960?focusedCommentId=59838&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-59838)


msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
```diff
@@ -81,7 +83,7 @@ public void merge(ExecAggregator addend) throws StandardException {
 				if(addend==null) return; //treat null entries as zero
 				//In Splice, we should never see a different type of an ExecAggregator
 				DecimalBufferedSumAggregator other = (DecimalBufferedSumAggregator)addend;
-            if (other.isNull){
+            if (other.isNull && other.position == 0){
```
yxia92: Why do we need to add the check of other.position == 0? So other.isNull is not enough to represent that other is  empty?

msirek:
> @msirek By keeping the aggregation of float or double in a treemap, do you notice any performance impact for a large dataset?

@yxia92 I tried a sum of about 8 million Doubles, with and without my changes, and the runtime was virtually identical, about 6040 ms, give or take a few ms.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
msirek: I was trying to handle the case where we overflow during incrementPosition(), and so isNull got set to false, but the new value never got added, so the total sum should still be null.  But now I see this check is bad because position could be 0 when sum is non-zero.  I think it can be fixed by delaying the setting of isNull in incrementPosition().  I will make the change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryOperatorNode.java
```diff
@@ -679,9 +694,11 @@ public void generateExpression(ExpressionClassBuilder acb,
...
 			*/
 
-			rightOperand.generateExpression(acb, mb);			
-			mb.cast(receiverType); // cast the method instance
-			// stack: right
+            if (dupReceiver)
+                rightOperand.generateExpression(acb, mb);
+			else
+                acb.generateNull(mb, getTypeCompiler(), getTypeServices());
+			// stack: receiver
```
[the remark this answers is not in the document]
msirek: > Is mb.cast(receiverType) missing here?

Yes, thanks.  I've added it in.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/AggregateNode.java
```diff
@@ -434,6 +434,27 @@ public ValueNode bindExpression(FromList fromList,
...
+            // For now, the receiver data type picked by
+            // getAggregator() must match the data type of the
+            // operand when processing window functions
+            // to avoid ClassCastExceptions for NativeSparkDataSets.
+            // For non-windowed aggregations, we process the aggregation using
+            // the original data type of the operand for performance.  If it overflows,
+            // a new aggregator that uses the final result type is allocated.
+            // When there is no overflow, a final CAST is applied in
+            // methoc SpliceGeneratorAggregator.finish.
+            operand =
```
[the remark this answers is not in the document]
msirek: Good catch.  We likely only want to do the CAST to avoid overflow when we're doing a SUM or AVG function, which will have a compatible data type with the operand.  I will add a condition to only do the cast for SUM or AVG:
`uad instanceof SumAvgAggregateDefinition`

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/FloatBufferedSumAggregator.java
```diff
@@ -35,55 +35,69 @@
...
 import java.util.Arrays;
+import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * @author Scott Fines
  *         Date: 5/16/14
  */
 public class FloatBufferedSumAggregator extends SumAggregator {
-		private final float[] buffer;
```
dgomezferro: Isn't this a copy of DoubleBufferedSumAggregator now? What's the difference?

dgomezferro:
Other than the (seemingly)  duplicated Double/FloatBufferedSumAggregator I'm +1, nice one!

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/FloatBufferedSumAggregator.java
msirek: @dgomezferro I didn't mention it in the JIRA, but I was considering removing FloatBufferedSumAggregator entirely from the source base, since it is no longer being used (it is never instantiated).  The only reason I left it in is in case we changed our minds a sprint or two from now and wanted to re-introduce it.  Do you think I should just go ahead and expunge it from the source entirely now?

dgomezferro: @msirek if it's not used then yes, definitely. Thanks!

msirek: @dgomezferro I have gone ahead and removed all traces of FloatBufferedSumAggregator.

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3435/

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3435/

PR #2658, opened 2019-03-06
DB-7976 flatten non-aggregate correlated exists/not exists with singl…

yxia92:
…e table as inner (anti) join,
and fix a wrong result for merge join in the presence of null.

Please see comment in [DB-7976](https://splicemachine.atlassian.net/browse/DB-7976?focusedCommentId=59891&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-59891)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryPredicate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryPredicate.java
@@ -110,6 +110,18 @@ class ExistsSubqueryPredicate implements org.spark_project.guava.base.Predicate<
         // The subquery is correlated
         // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
+        // For correlated exists/not exists subquery, if it is just a non-aggregate subquery with one table,
+        // inner join(for exists) and anti-join(for not exists) with the base table directly(through the flattening logic
+        // in SubqueryNode) are heuristically more efficient than rewriting the exists subquery to a derived table(in this module).
+        // So for such cases, return false here and let the flattening logic in SubqueryNode take care of them.
+        FromList fromList = subqueryResultSet.getFromList();
+        if (subqueryNode.singleFromBaseTable(fromList) != null) {
+            boolean hasAggregation = subquerySelectNode.hasAggregatesInSelectList() || subquerySelectNode.hasHavingClause();
+            if (!hasAggregation &&
+                    !subquerySelectNode.getOriginalWhereClauseHadSubqueries())
+                return false;
+        }
+
         /* correlated subquery cannot contain a union */
         if (subqueryResultSet.getFromList().containsNode(UnionNode.class)) {
             return isUnionSubqueryOk(subqueryNode, subquerySelectNode.getNestingLevel(), notExistsSubquery);
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryPredicate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryPredicate.java
@@ -116,7 +116,8 @@ class ExistsSubqueryPredicate implements org.spark_project.guava.base.Predicate<
         // So for such cases, return false here and let the flattening logic in SubqueryNode take care of them.
         FromList fromList = subqueryResultSet.getFromList();
         if (subqueryNode.singleFromBaseTable(fromList) != null) {
-            boolean hasAggregation = subquerySelectNode.hasAggregatesInSelectList() || subquerySelectNode.hasHavingClause();
+            boolean hasAggregation = subquerySelectNode.hasAggregatesInSelectList() || subquerySelectNode.hasHavingClause()
+                    || (subquerySelectNode.getGroupByList()!= null && subquerySelectNode.getGroupByList().isRollup());
             if (!hasAggregation &&
                     !subquerySelectNode.getOriginalWhereClauseHadSubqueries())
                 return false;
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
@@ -2707,4 +2707,12 @@ public class SelectNode extends ResultSetNode{
 
         return true;
     }
+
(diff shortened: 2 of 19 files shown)


yxia92:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryPredicate.java
```diff
@@ -110,6 +110,18 @@ private boolean doWeHandle(SubqueryNode subqueryNode) throws StandardException {
         // The subquery is correlated
         // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
+        // For correlated exists/not exists subquery, if it is just a non-aggregate subquery with one table,
+        // inner join(for exists) and anti-join(for not exists) with the base table directly(through the flattening logic
+        // in SubqueryNode) are heuristically more efficient than rewriting the exists subquery to a derived table(in this module).
+        // So for such cases, return false here and let the flattening logic in SubqueryNode take care of them.
+        FromList fromList = subqueryResultSet.getFromList();
+        if (subqueryNode.singleFromBaseTable(fromList) != null) {
+            boolean hasAggregation = subquerySelectNode.hasAggregatesInSelectList() || subquerySelectNode.hasHavingClause();
```
msirek: Does the rewrite work properly when subquerySelectNode.groupByList.isRollup() is true or subquerySelectNode.whereAggregates is not empty?  Just wondering what the flattening looks like for these cases and if they should be restricted, or, if the presence of whereAggregates should trigger this rewrite.

yxia92: Thanks @msirek. You are right, I should include the check in the fix as these scenarios should take the original path and not be affected by this JIRA. I'll push a commit shortly.

Also, for the question about whether the existing rewrite work properly when subquerySelectNode.groupByList.isRollup() is true, I think it is currently not working properly, NOT because of the rewrite (the rewrite itself is fine, it should be similar to subquery with regular group by), BUT because we don't support syntax like "group by a1, rollup(b1)". As a result, any extra column added to the group by list is treated as part of the rollup column list, which is wrong. For example,
`select * from t1 where not exists (select 1 from t2 where a1=a2 group by rollup(b2) having b2>1);`
Here is the plan:
Plan                                                                                                                                                                                                                                                            
Cursor(n=12,rows=20,updateMode=READ_ONLY (1),engine=control)                                                                                                                                                                                                    
  ->  ScrollInsensitive(n=11,totalCost=20.7,outputRows=20,outputHeapSize=80 B,partitions=1)                                                                                                                                                                     
    ->  ProjectRestrict(n=10,totalCost=12.3,outputRows=20,outputHeapSize=80 B,partitions=1)                                                                                                                                                                     
      ->  ProjectRestrict(n=9,totalCost=12.3,outputRows=20,outputHeapSize=80 B,partitions=1,preds=[is null(ExistsFlatSubquery-0-1.A2[8:2])])                                                                                                                    
        ->  BroadcastLeftOuterJoin(n=8,totalCost=12.3,outputRows=20,outputHeapSize=80 B,partitions=1,preds=[(A1[8:1] = ExistsFlatSubquery-0-1.A2[8:2])])                                                                                                        
          ->  ProjectRestrict(n=7,totalCost=12.3,outputRows=20,outputHeapSize=80 B,partitions=1)                                                                                                                                                                
            ->  ProjectRestrict(n=6,totalCost=24.6,outputRows=20,outputHeapSize=80 B,partitions=1)                                                                                                                                                              
              ->  GroupBy(n=5,totalCost=24.6,outputRows=20,outputHeapSize=80 B,partitions=1)                                                                                                                                                                    
                ->  ProjectRestrict(n=4,totalCost=4.04,outputRows=20,outputHeapSize=40 B,partitions=1)                                                                                                                                                          
                  ->  TableScan[T2(1600)](n=3,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=40 B,partitions=1)                                                                                                                                     
          ->  ProjectRestrict(n=2,totalCost=4.04,outputRows=20,outputHeapSize=40 B,partitions=1)                                                                                                                                                                
            ->  TableScan[T1(1584)](n=1,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=40 B,partitions=1)    

The GroupBy should be processed as "group by rollup(b2), a2", but currently we are treating it as "group by rollup (b2, a2)", thus is wrong.

For the other question about subquerySelectNode.whereAggregates, as I understand, we don't support aggregation in the where clause, not even for subquery in the having clause.

msirek: OK, thanks @yxia92.  Do we need any special handling of window functions, since they are similar to aggregation?  For example, a query such as the following (not sure if this one is legitimate, but in general if we use a window function, can we still do the rewrite and get the same results?) :

select a1 from t1 where not exists (select a2, max(a1) over (Partition by a2 ORDER BY a1 ROWS BETWEEN CURRENT ROW AND 2 FOLLOWING) from t2 where a1=a2 and 1=0);

yxia92: Thanks @msirek . In theory, we don't need to worry about what is put in the select clause of the exists/not exists subquery(except for aggregation), as what we care is whether there is a matching row or not from the subquery. In fact, there is logic in SubqueryNode.bindExpression() which wipes out the original select clause with just a ParTrue node. There is however a problem with this logic, as it does not clean the window functions from the SELECT node completely, and still leave entries in the widnowDefinitionList, which could cause failure in the later logic. On my local system, I got the following assertion failure:
splice> explain select * from t1 where not exists (select max(a2) over (partition by b2 order by c2) from t2 where a1=a2);
ERROR XJ001: DERBY SQL error: SQLCODE: -1, SQLSTATE: XJ001, SQLERRMC: java.lang.IllegalStateExceptionASSERT FAILED generatedRef is null.  replaceAggregateWithColumnReference() has not been called on this AggergateNode.  Make sure the node is under a ResultColumn as expected.XJ001.U

This, however, is not related to this PR. I'll open a public JIRA for this issue.

msirek: OK, thanks @yxia92

PR #2663, opened 2019-03-07
 DB-7934 Add mapr6.1.0 support

bklo94:
Added new maprpm profile to assembly to create rpms to install 6.1.0 onto a centos 7 cluster. 

I had to add <maven.skip.tests> property to the mapr6.1.0 profile in platform_it in since there were major changes to the kafka cluster api, and therefore test. Thus it did not test-compile and I had to call the surfire maven skip test property to skip test-compile.

Cleaned up the mapr6.1.0 directory of most files, since most of the files were for the manual process.


OlegMazurov:
What caused parquet-spark2.3 to be introduced? How is it different from orc-spark-2.3?

dgomezferro:
Why do we need assembly/mapr6.1.0/src/main/resources/{bin|examples|hbase-webapps}

MurrayBrown:
> Why do we need assembly/mapr6.1.0/src/main/resources/{bin|examples|hbase-webapps} ?

This is because these new 'splicemachine mapr rpms' are installing a full copy of hbase, not just 'our software'.

assembly/mapr6.1.0/src/main/resources/examples/conf/hbase-env.secure.sh
```diff
@@ -0,0 +1,236 @@
+#
```
MurrayBrown: do we want these files to be stored as 'examples' ? I think of them more as 'templates', since we copy in one or the other of these files into the rpm at build time. I wonder if the directory name will be confusing to future maintainers.

assembly/pom.xml
```diff
@@ -666,6 +666,602 @@
...
+                                <goals>
+                                    <goal>attached-rpm</goal>
+                                </goals>
+                                <configuration>
+                                    <name>splicemachine-${envClassifier}-hbase-secure-conf</name>
+                                    <targetOS>linux</targetOS>
+                                    <classifier>splicemachine</classifier>
+                                    <version>${project.version}.${parcel.patch}</version>
+                                    <group>Applications/Internet</group>
+                                    <license>Copyright © 2018 Splice Machine Inc. Licensed under the Apache License, Version 2.0.</license>
```
msirek: Should this be copyright 2019?  There are a few other occurrences in this PR.
Most of this PR seems to be just duplication of files for the new platform, so I'm not reviewing all of the duplicated content.  Wondering if we can unify some of these files in the future so they can be shared between like platforms.

assembly/mapr6.1.0/src/main/resources/examples/conf/hbase-env.secure.sh
bklo94: Moved them all into a scripts folder and separated them out based on service. I don't think change between platforms (other than past installation methods). Maybe mapr should be the first unified assembly platform?

PR #2665, opened 2019-03-08
DB-8003: report duplicate rows in decoded format

jyuanca:


jyuanca:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/function/HFileGenerationFunction.java
```diff
@@ -268,4 +372,81 @@ private static InetSocketAddress getFavoredNode(BulkImportPartition partition) t
...
+            execRow = decoder.decode(key, value);
+        }
+        if (operationContext != null && operationContext.isPermissive()) {
+
+            operationContext.recordBadRecord("Unique constraint violation, row: "
+                    + execRow.toString(), null);
+        }
+        else {
+            SpliceLogUtils.error(LOG, "Unique constraint violation, row: %s", execRow.toString());
+            DataDictionary dd = ah.getActivation().getLanguageConnectionContext().getDataDictionary();
```
dgomezferro: Isn't it possible to get the activation from the operationContext rather than passing around an ActivationHolder?

jyuanca: Not always possible. If it is not a SpliceOperation (in case create index), Activation will not be serialized

dgomezferro: Is it possible to create an operationContext with a dummy operation and make "create index" use it so that we always have an operationContext?

jyuanca: It is possible, but I think it's a bit complicated to fake a SpliceOperation than serde activation.

jyuanca:

spliceci:
TEST SUCCEEDED +1

PR #2672, opened 2019-03-19
SPLICE-2301 NativeSparkDataSet for ProjectRestrict with expressions.

msirek:
Support NativeSparkDataSet for ProjectRestrict operations with a filter (WHERE clause) and projected expressions, for example:
`explain select coalesce(a,b,c,e,d) from t3 --splice-properties useSpark=true
where d+e > 0;

Cursor(n=5,rows=7,updateMode=READ_ONLY (1),engine=Spark)                                                                                                                                                                                                        
  ->  ScrollInsensitive(n=4,totalCost=8.106,outputRows=7,outputHeapSize=33 B,partitions=1)                                                                                                                                                                      
    ->  **ProjectRestrict**(n=3,totalCost=4.04,outputRows=7,outputHeapSize=33 B,partitions=1)                                                                                                                                                                       
      ->  **ProjectRestrict**(n=2,totalCost=4.04,outputRows=7,outputHeapSize=33 B,partitions=1,**preds=[((D[0:4] + E[0:5]) > 0)]**)                                                                                                                                     
        ->  TableScan[T3(1616)](n=1,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=33 B,partitions=1)            
`
This feature helps reduce transformations back and forth between RDDs and DataFrames, which could provide some speedup to queries with many operations.

For more details, see [SPLICE-2301](https://splice.atlassian.net/browse/SPLICE-2301).


msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/catalog/types/TypeDescriptorImpl.java
```diff
@@ -456,6 +456,12 @@ public String	toString()
 		return s;
 	}
 
+    public String toSparkString()
+	{
+		String s = getSQLstring();
+		return s;
```
jaceklaskowski: Why not `return getSQLstring();`?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/Row.java
```diff
@@ -69,4 +69,14 @@
...
 	void	setColumn (int position, DataValueDescriptor value);
 
+    /**
+     * Set a DataValueDescriptor in a Row by ordinal position (1-based),
+     * preserving the data type of the target column.
+     * For example, if we are setting (SQLDouble)null = (SQLInteger)1,
+     * The column will end up holding (SQLDouble)1.0, preserving
+     * the data type of the DVD that is being overwritten.
+     *
+	 * @param position	The ordinal position of the column.
```
jaceklaskowski: Something's wrong with the tabs here and in other places.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
```diff
@@ -1602,6 +1602,10 @@ public String toString(){
         return typeDescriptor.toString();
     }
 
+    public String toSparkString(){
```
jaceklaskowski: In TypeDescriptorImpl.java you had the opening `{` on a separate line. Be consistent.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -537,7 +537,12 @@ protected int typeCompare(DataValueDescriptor arg) throws StandardException
     public DataValueDescriptor cloneValue(boolean forceMaterialization)
 	{
 		try {
-			return new SQLDecimal(getBigDecimal(), precision, scale);
+		    SQLDecimal decimal = new SQLDecimal(getBigDecimal(), precision, scale);
+		    if (decimal.getPrecision() != precision)
+		        decimal.setPrecision(precision);
+		    if (decimal.getScale() != scale)
+		        decimal.setScale(scale);
+			return decimal;
```
jaceklaskowski: Think the ident is spurious. Please use `{` and `}` to denote where `if` starts and ends.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryOperatorNode.java
```diff
@@ -1058,5 +1058,7 @@ public boolean isConstantOrParameterTreeNode() {
 		return leftOperand.isConstantOrParameterTreeNode() && rightOperand.isConstantOrParameterTreeNode();
 	}
 
+	// Is this a REPEAT operator?
```
jaceklaskowski: Remove the comment  since it adds nothing but a noise.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
```diff
@@ -1168,6 +1170,55 @@ public void generateResultSet(ExpressionClassBuilder acb, MethodBuilder mb) thro
...
+                if (expressions[i] == null) {
+                    genExpressions = false;
+                    numberOfValues = 0;
+                    break;
+                }
+                i++;
+            }
+        }
+
+        String stringClassName = "java.lang.String";
```
jaceklaskowski: That could be generated from `String.class...`

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -838,6 +860,27 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
         }
     }
 
+    /**
+     * Take a Splice DataSet in the consumer's context, with a single source,
```
jaceklaskowski: s/DataSet/Dataset

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -154,6 +154,17 @@ public NativeSparkDataSet(JavaRDD<V> rdd, String ignored, OperationContext conte
         }
     }
 
+    // Build a Native Spark Dataset from the context of the consumer operation.
+    // Only works for an op with a single child op... not joins.
```
jaceklaskowski: "or unions"?

msirek:

splice_machine/src/main/java/com/splicemachine/derby/stream/function/ProjectRestrictMapFunction.java
```diff
@@ -76,7 +76,7 @@ public boolean hasNativeSparkImplementation() {
         ProjectRestrictOperation op = (ProjectRestrictOperation) operationContext.getOperation();
         if (op.projection != null) {
```
jaceklaskowski: That combination of `if` with `return false` and then `return true` could be rewritten (simpler?), e.g.

```
return op.projection == null || op.hasExpressions()
```

(or similar = that begs for some tests)

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -82,23 +135,166 @@ public static String toString(PredicateList predicateList) {
...
+     */
+    public static String opToSparkString(ValueNode operand) {
+        String retval = null;
+
+        // Do not throw any errors encountered.  An error condition
+        // just means we can't generate a valid spark representation
+        // of the SQL expression to apply to a native spark Dataset,
+        // so should not be considered a fatal error.
+        try {
+            SPARK_EXPRESSION.set(Boolean.TRUE);
```
dgomezferro: We should pass down a parameter into opToString2() and related functions rather than using a ThreadLocal. If we are not sharing much code in opToString2() please consider using a completely separate function for generating the Spark representation, in most cases it seems we are not sharing anything at all.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -82,23 +135,166 @@ public static String toString(PredicateList predicateList) {
...
-        }else if(operand instanceof UnaryOperatorNode){
+        } else if (operand instanceof UntypedNullConstantNode)
+            return " null ";
+        else if(operand instanceof UnaryOperatorNode){
             UnaryOperatorNode uop=(UnaryOperatorNode)operand;
-            return format("%s(%s)",uop.getOperatorString(),opToString(uop.getOperand()));
+            String operatorString = uop.getOperatorString();
+            if (SPARK_EXPRESSION.get()) {
+                if (operand instanceof IsNullNode) {
+                    RELATIONAL_OP_DEPTH.get().increment();
```
dgomezferro: I think we should use a normal function parameter rather than a ThreadLocal here too.

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -537,7 +537,12 @@ protected int typeCompare(DataValueDescriptor arg) throws StandardException
     public DataValueDescriptor cloneValue(boolean forceMaterialization)
 	{
 		try {
-			return new SQLDecimal(getBigDecimal(), precision, scale);
+		    SQLDecimal decimal = new SQLDecimal(getBigDecimal(), precision, scale);
+		    if (decimal.getPrecision() != precision)
```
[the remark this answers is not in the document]
msirek: The default behavior when constructing a SQLDecimal is to set the precision and scale to the that of the BigDecimal literal we're building:
`	private void setCoreValue(BigDecimal theValue)
	{
		value = theValue;
		if (value !=null) {
			precision = value.precision();
			scale = value.scale();
		}`

However, there is a special case where precision and scale may not match the underlying BigDecimal.  In IndexRowToBaseRowOperation, SPLICE-1700 added code to write column default values into the index template row:

`            //update compactRow with the default value info
            if (defaultRow != null && defaultValueMap != null) {
                for (int i=defaultValueMap.anySetBit(); i>=0; i=defaultValueMap.anySetBit(i)) {
                    compactRow.setColumnValue(i+1, defaultRow.getColumn(i+1));
                }
            }`

If the default value has scale and precision less than that of the column definition, it causes overflow when inserting values that should fit.  For example, if column is defined as Dec(5,2), but the default value is 100.1, the template row gets a DVD in it with data type DEC(3,1).  If we then try to populate the template row with a larger value, say 1000.1, this causes overflow, even though it should fit in DEC(5,2).  So, because of SPLICE-1700, we have a need to set the precision and scale of the DVD to something other than the precision and scale of the literal value held in the DVD (done by introducing new method setColumnValue instead of using setColumn).  Therefore, cloneValue needs to preserve the different scale and precision if it has been modified from the original value.  Since this is special-purpose code, I don't think we want to modify how all SQLDecimals are constructed just to accomodate this case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -82,23 +135,166 @@ public static String toString(PredicateList predicateList) {
...
+                typeFormatId == DOUBLE_TYPE_ID   ||
+                typeFormatId == REAL_TYPE_ID);
+    }
+
+    private static boolean isOverflowSensitive(ValueNode operand) throws StandardException {
+        return (operand.getTypeId().getTypeFormatId() == LONGINT_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DECIMAL_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DOUBLE_TYPE_ID);
+    }
+    private static void checkOverflowHidingCases(BinaryArithmeticOperatorNode bao) throws StandardException {
```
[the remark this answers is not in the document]
msirek: > This is very limiting, there's a Spark PR targeting 3.0 that could fix this https://issues.apache.org/jira/browse/SPARK-26218
> 
> We could include that fix on the service and cover more cases

Agree

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
```diff
@@ -158,6 +158,7 @@
 	boolean     DEFAULT_MULTICOLUMN_INLIST_PROBE_ON_SPARK_ENABLED = false;
 	boolean     DEFAULT_CONVERT_MULTICOLUMN_DNF_PREDICATES_TO_INLIST = true;
 	boolean     DEFAULT_DISABLE_PREDICATE_SIMPLIFICATION = false;
+	double      DEFAULT_SPLICE_SPARK_MAJOR_VERSION = 2.2;
```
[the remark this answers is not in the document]
msirek: @dgomezferro Are you saying we should have an entirely new class devoted to holding the Spark version we're running?  I currently have interface getSparkMajorVersion() in the DatabaseVersion object that you mention, as spark version seems part of the database version info.  getSparkMajorVersion() retrieves a string value, picked up from a POM file setting, and its value is something like "spark2.3", which is then converted into a java property when splice is booted, with the prefix "spark" stripped off.  For ease of comparison of version numbers, this property value is read in and converted to a double value in GenericStatement.java, stored in the CompilerContext.  This way we can use '>' or '<' comparisons to check that our Spark version is at a given level or higher, e.g.

sparkVersion() >= 2.3

We can chat about this more if you like.

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/NativeTransformationsTest.scala
```diff
@@ -89,7 +89,7 @@ class NativeTransformationsTest extends FunSuite with TestContext with BeforeAnd
     assert("SortMergeJoin".r.findAllMatchIn(plan).length == 2)
   }
 
-  ignore("complex project on join over join") {
+  test("complex project on join over join") {
     val df = splicemachineContext.df(
       """select s1.colsequence * 10 from sys.systables s1
```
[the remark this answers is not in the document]
msirek: > Based on the overflow limitations I thought this shouldn't work, and it turns out we are promoting the Int to BigInteger. Where does that happen? That might be a good strategy to deal with integer overflow if we then convert it back to Integer after making sure it didn't overflow, but we can't return a BigInteger were we are expecting an integer, right?

@dgomezferro The result data type of 's1.colsequence * 10' is promoted to BigInt in NumericTypeCompiler.resolveArithmeticOperation:

		if (java.sql.Types.TINYINT == highTypeId ||
	            java.sql.Types.SMALLINT == highTypeId ||
		    java.sql.Types.INTEGER == highTypeId)
		{
			return new DataTypeDescriptor(
					TypeId.getBuiltInTypeId(java.sql.Types.BIGINT),
					nullable
				);
		}

This type of operation should never overflow because max integer * max integer is still less than max bigint.  For other operations that could overflow, I am not sure how to accomplish  "making sure it didn't overflow" programatically.  We can revisit this later, but I agree we need some solution to enable more cases.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
```diff
@@ -3668,6 +3676,8 @@ public void computeDistinctCardinality(CostEstimate costEstimate) throws Standar
 
         MethodBuilder cb = acb.getConstructor();
 
+        boolean retval = true;
```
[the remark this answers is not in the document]
msirek: > I'd like a more descriptive name here, supportedSparkExpressions or something similar?

Renamed retval to isExpressableInSparkSQL.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/ProjectRestrictOperation.java
```diff
@@ -310,9 +360,14 @@ public boolean apply(ExecRow row) throws StandardException {
...
         try {
             operationContext.pushScope();
-			if (restrictionMethodName != null)
-				sourceSet = sourceSet.filter(new ProjectRestrictPredicateFunction<>(operationContext));
-			return sourceSet.map(new ProjectRestrictMapFunction<>(operationContext));
+            // Temporary code to upgrade the source DataSet to a SparkNativeDataSet
+            // to flush out bugs.
+            // TODO:  Remove this after thorough regression testing.
+            if (hasFilterPred() || hasExpressions())
+                sourceSet = sourceSet.upgradeToSparkNativeDataSet(operationContext);
```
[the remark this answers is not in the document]
msirek: > I think it's better to remove this conversion, as we keep adding more supported operations we'll get better testing anyway, and this introduces a forced conversion that we are trying to avoid.

Removed the call to upgradeToSparkNativeDataSet.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -838,6 +860,27 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
...
 
+    /**
+     * Take a Splice SparkDataSet (RDD) in the consumer's context, with a single source,
+     * and convert it to a NativeSparkDataSet (Dataset<Row>) doing a map.
+     * @param context
+     * @return
+     * @throws Exception
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    static <V> Dataset<Row> sourceRDDToSparkRow(JavaRDD<V> rdd, OperationContext context) {
```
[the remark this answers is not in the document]
msirek: Removed the new NativeSparkDataSet constructor and all of its callers.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -82,23 +135,166 @@ public static String toString(PredicateList predicateList) {
...
+        // just means we can't generate a valid spark representation
+        // of the SQL expression to apply to a native spark Dataset,
+        // so should not be considered a fatal error.
+        try {
+            SPARK_EXPRESSION.set(Boolean.TRUE);
+            initRelationalOpDepth();
+            retval = opToString2(operand);
+
+        }
+        catch (Exception e) {
```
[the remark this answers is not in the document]
msirek: > Catch only the LANG_DOES_NOT_IMPLEMENT exception

Made the change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
msirek: > We should pass down a parameter into opToString2() and related functions rather than using a ThreadLocal. If we are not sharing much code in opToString2() please consider using a completely separate function for generating the Spark representation, in most cases it seems we are not sharing anything at all.

Moved the thread local variables into a "vars" parameter, collectively held together in an instantiated OperatorToString object, to make the code more legible (some lines of code call OperatorToString multiple times).  This also makes it easier to add new parameters in the future.

I considered using a separate function originally, but thought that some of the changes I'm adding might be good to also incorporate into explain output at some point.  Currently some SQL expressions aren't fully printed in the explain output, so we are not getting full information.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
msirek: Removed all ThreadLocals in OperatorToString.

msirek:

spliceci:
TEST FAILED

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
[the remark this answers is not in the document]
dgomezferro: I see, that makes sense, thanks!

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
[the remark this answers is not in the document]
dgomezferro: Sure, we can discuss this. Yes, I was suggesting using a (simple) class for this. My reasoning is that some features might be available in Spark 2.3.2 but not int 2.3.0, so a double comparison wouldn't work. Also if we have our own patched Spark in the service a proper class could encapsulate that logic.

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/NativeTransformationsTest.scala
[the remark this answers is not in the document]
dgomezferro: Oh I see, in that case the overflow logic is a quite a bit less restrictive than I thought.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -37,6 +37,7 @@
 import com.splicemachine.db.iapi.sql.compile.OptimizablePredicate;
 import com.splicemachine.db.iapi.types.*;
 import org.apache.commons.lang3.mutable.MutableInt;
+import org.apache.zookeeper.Op;
```
dgomezferro: Was this added by mistake?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
```diff
@@ -1440,7 +1447,13 @@ private void generateMinion(ExpressionClassBuilder acb,
         mb.push(costEstimate.getEstimatedCost());
         mb.push(printExplainInformationForActivation());
 
-        String filterPred = OperatorToString.opToSparkString(restriction);
+        String filterPred = null;
+        try {
+            OperatorToString.opToSparkString(restriction);
```
dgomezferro: You are missing the assignment to filterPred here.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
```diff
@@ -1189,8 +1189,15 @@ public void generateResultSet(ExpressionClassBuilder acb, MethodBuilder mb) thro
...
+                try {
+                    expressions[i] = OperatorToString.opToSparkString(rc.getExpression());
+                    if (expressions[i] == null) {
+                        genExpressions = false;
+                        numberOfValues = 0;
+                        break;
+                    }
+                }
+                catch (StandardException e) {
                     genExpressions = false;
```
dgomezferro: This would be an unexpected exception, we should log it at least, but I would prefer to make it bubble up. If we don't trigger it on the regression tests we can assume nothing breaks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
```diff
@@ -1440,7 +1447,13 @@ private void generateMinion(ExpressionClassBuilder acb,
         mb.push(costEstimate.getEstimatedCost());
         mb.push(printExplainInformationForActivation());
 
-        String filterPred = OperatorToString.opToSparkString(restriction);
+        String filterPred = null;
+        try {
+            OperatorToString.opToSparkString(restriction);
+        }
+        catch (StandardException e) {
```
dgomezferro: At least log it, but I'd rather not catch the exception here.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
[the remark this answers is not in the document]
msirek: @dgomezferro Added SimpleSparkVersion class.  We now set splice.spark.compile.version automatically, from POM file properties, to something like 2.2.0.cloudera2, which is used to initialize the SimpleSparkVersion class.  The cloudera2 part is ignored when comparing SimpleSparkVersion objects.  If the user sets system property splice.spark.version, and it is a valid version string, then it will override the setting in splice.spark.compile.version.  This is for the case where the cluster is running a different version of spark than the splice parcel compile-time setting for spark version.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
msirek: > Was this added by mistake?

cleaned it up, thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
msirek: > You are missing the assignment to filterPred here.

Fixed, thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
msirek: > This would be an unexpected exception, we should log it at least, but I would prefer to make it bubble up. If we don't trigger it on the regression tests we can assume nothing breaks.

OK, removed the try/catches, so unexpected exceptions are caught.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
msirek: > At least log it, but I'd rather not catch the exception here.

Removed the try/catch.

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/catalog/types/TypeDescriptorImpl.java
msirek: Removed toSparkString.  Changed to use getSQLstring.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
msirek: Splice calls it DataSet, not Dataset.

`/**
 * Stream of data acting on an iterable set of values.
 */
public interface DataSet<V> extends //Iterable<V>,
        Serializable {`

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -861,8 +861,8 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
     }
 
     /**
-     * Take a Splice DataSet in the consumer's context, with a single source,
-     * and convert it to a Spark Dataset doing a map.
+     * Take a Splice SparkDataSet (RDD) in the consumer's context, with a single source,
+     * and convert it to a NativeSparkDataSet (Dataset<Row>) doing a map.
```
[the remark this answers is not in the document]
msirek: This method has been removed.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/ProjectRestrictMapFunction.java
msirek: One form is shorter.  The other form is easier to read.  No need to change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/JavaToSQLValueNode.java
```diff
@@ -214,12 +214,12 @@ public JavaValueNode getJavaValueNode()
 	// Is this a system function (defined in the SYSFUN schema)?
 	public boolean isSystemFunction() {
 	    if (javaNode != null &&
```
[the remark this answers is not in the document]
msirek: Yes, they could, but sometimes the longer form is preferable as it's quicker to clearly see what the code is doing.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -82,23 +122,190 @@ public static String toString(PredicateList predicateList) {
...
+                typeFormatId == DOUBLE_TYPE_ID   ||
+                typeFormatId == REAL_TYPE_ID);
+    }
+
+    private static boolean isOverflowSensitive(ValueNode operand) throws StandardException {
+        return (operand.getTypeId().getTypeFormatId() == LONGINT_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DECIMAL_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DOUBLE_TYPE_ID);
+    }
+    private static void checkOverflowHidingCases(BinaryArithmeticOperatorNode bao,
```
yxia92: Do we need to worry about hidden overflow for operator other than BinaryOperator? for example, Coalesce, ConditionalNode?

msirek:
Hi @yxia92, for some reason I can't find your comments in Github, which I see in e-mail:
"+    private static boolean isOverflowSensitive(ValueNode operand) throws StandardException {
+        return (operand.getTypeId().getTypeFormatId() == LONGINT_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DECIMAL_TYPE_ID ||
+                operand.getTypeId().getTypeFormatId() == DOUBLE_TYPE_ID);
+    }
+    private static void checkOverflowHidingCases(BinaryArithmeticOperatorNode bao,
Do we need to worry about hidden overflow for operator other than BinaryOperator? for example, Coalesce, ConditionalNode?"

Response:
For coalesce, we do increment the relationalOpDepth, so will check for overflow when opToString2 is called on the argumentsList : 
       else if (operand instanceof CoalesceFunctionNode) {
            vars.relationalOpDepth.increment();

For ConditionalNode, it seems like we want to increment there too, so I'm adding it.

I am also adding handling for TernaryOperatorNode, BinaryListOperatorNode, ArrayConstantNode and ListValueNode.  Some of the functions that take only CHAR data don't necessarily need to be handled, since we can't cast Double to CHAR:

values cast(double(3) as char(4));
ERROR 42846: Cannot convert types 'DOUBLE' to 'CHAR'.

... as long as we increment on any function that could convert a double or similar numeric to char.

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
This PR is ready to merge.

pom.xml
```diff
@@ -631,6 +631,8 @@
                                 <Release>${project.version}</Release>
                                 <Build-Time>${maven.build.timestamp}</Build-Time>
                                 <URL>http://www.splicemachine.com</URL>
+                                <Spark-Major-Version>${spark.major.version}</Spark-Major-Version>
```
OlegMazurov: This change affected all jars including those built in the _core_ profile for which _spark.major.version_ and _spark.version_ are not defined.

PR #2688, opened 2019-04-06
DB-8138: extract columns correctly for update trigger

jyuanca:


jyuanca:

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
```diff
@@ -495,4 +518,49 @@ private static ExecRow extractColumns(ExecRow resultSet, boolean firstHalf) thro
...
+                }
+            }
+            int nCols = resultSet.nColumns()/2;
+            int sourceIndex = (firstHalf ? 1 : nCols + 1);
+            int pos = 1;
+
+            for (int i = allTriggerColumns.anySetBit(); i >= 0; i = allTriggerColumns.anySetBit(i)) {
+                int index = 0;
+                for (int j = heapList.anySetBit(); j >= 0; j = heapList.anySetBit(j)) {
+                    if (j == i) {
```
msirek: Minor comment.. this could use heapList.isSet(i) instead of looping through heapList.  Most likely no one will have a trigger with hundreds or columns or be updating hundreds of columns, so it's probably OK as-is.

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
```diff
@@ -495,4 +518,49 @@ private static ExecRow extractColumns(ExecRow resultSet, boolean firstHalf) thro
...
+                        result.setColumn(pos++, resultSet.getColumn(sourceIndex + index).cloneValue(false));
```
yxia92: I think we can break out after finding the match.

dgomezferro:
@wb14123 please don't merge if there unadressed comments

PR #2689, opened 2019-04-09
DB-8147 Fix some stored procedures under SSL connection

wb14123:


carolp-503:
These changes look necessary, but are the complete?  Can we do some additional validation in an SSL environment to make sure that nothing has slipped through?

wb14123:
@carolp-503 We can start an SSL instance for IT. But that may double testing time. I think we can keep this ticket for this specific bug, and create a new one for SSL testing since the scope would be big.

carolp-503:
@WB14123, I agree that it's a big effort, but it's one that our users are counting on. We want to put out a product we have confidence in, and that means more testing here.  There's some urgency here because we know people want this fix.  But it's important that we take the time to test it on a cluster that uses SSL and Ambari, not just our usual testing.  At a minimum, I suggest we install and test on a cluster with SSL and do sanity checking, even if that takes longer.  Security is something to take very seriously.  Whether we do the basic and extensive cluster testing as part of this ticket or another; it needs to be high priority  and this fix should be sanity checked at least on a cluster with SSL before going to customers.

wb14123:

wb14123:
@carolp-503 Sure. I'll test this build on our colo cluster and maybe also run some regression test under Xiao's help.

spliceci:
TEST FAILED

wb14123:

spliceci:
TEST FAILED

spliceci:
TEST SUCCEEDED +1

MurrayBrown:
@wb14123  can you also fix the same sqlshell error in this file:
assembly/common/src/main/resources/bin/sqlshell.sh

PR #2694, opened 2019-04-10
DB-8163: make schema inference temp directory name unique

jyuanca:


jyuanca:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -425,7 +425,7 @@ public StructType getExternalFileSchema(String storedAs, String location, boolea
                 fs = FileSystem.get(URI.create(location), conf);
                 String fileName = getFile(fs, location);
                 if (fileName != null) {
-                    temp = new Path(location, "temp");
```
dgomezferro: I think this could conflict with other systems, would it be better to create it under '_temporary' which seems more in line with what other systems do?

jyuanca:
For partitioned table, spark does not always return correct data type for partition column. For example, if the partition column is a string, however the partition values looks like an integer, spark will return an integer type. SPLICE-1809 infers schema for one data file and then fill in correct partition column types

dgomezferro:
@jyuanca my question is: why can't we use just our metadata to create the data schema? Why do we need Spark to infer the schema?

jyuanca:
@dgomezferro I'm a little puzzled why we never tried to read external data using table schema. Let me file a jira for that.

dgomezferro:
I'm Ok with merging this as long as we prioritize the new Jira created by @jyuanca to remove the inference if possible.

PR #2698, opened 2019-04-13
DB-8070 support WITH RECURSIVE

yxia92:
Please see the solution description [DB-8070](https://splicemachine.atlassian.net/browse/DB-8070) for details.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
@@ -27,6 +27,10 @@ import com.splicemachine.utils.Pair;
 import org.apache.log4j.Logger;
 import org.spark_project.guava.base.Strings;
 
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+
 /**
  * Created by yxia on 3/22/19.
  */
@@ -34,6 +38,7 @@ public class RecursiveUnionOperation extends UnionOperation {
     private static Logger LOG = Logger.getLogger(RecursiveUnionOperation.class);
     private DataSet rightDS;
     protected static final String NAME = RecursiveUnionOperation.class.getSimpleName().replaceAll("Operation","");
+    private int iterationLimit;
 
     @Override
     public String getName() {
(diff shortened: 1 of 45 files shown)


yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
```diff
@@ -117,6 +117,22 @@ public int partitions() {
...
+        return materialize();
+    }
+
+    @Override
+    public void unpersistIt() {
+        return;
+    }
+
+    @Override
+    public Pair<Pair<DataSet, DataSet>, Integer> persistIt2() {
```
dgomezferro: I don't like the "persistIt2" method too much. Would it be possible to add a getClone() method to the API instead? for SparkDataSet it would return `this', for Control you could return a special subclass in materialize() ( for instance MaterializedControlDataSet extends ControlDataSet ) that contains a List of values and calls super(list.iterator()) on the constructor, the getClone() in this case would return new MaterializedControlDataSet(list);

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
```diff
@@ -0,0 +1,120 @@
...
+        // so we have to return two datasets
+        Pair<Pair<DataSet, DataSet>, Integer> right = leftResultSet.getDataSet(dsp).persistIt2();
+        DataSet resultDS = right.getFirst().getFirst();
+        rightDS = right.getFirst().getSecond();
+
+        // compute the recursive body
+        right = rightResultSet.getDataSet(dsp).persistIt2();
+
+        int loop = 0;
+        while (right.getSecond() > 0 && loop < MAX_LOOP) {
```
dgomezferro: Other DBs raise an exception if the limit is reached, I think we should do the same. As someone else suggested it would be useful for MAX_LOOP to be a session level property.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
```diff
@@ -0,0 +1,120 @@
...
+    public DataSet<ExecRow> getDataSet(DataSetProcessor dsp) throws StandardException {
+        if (!isOpen)
+            throw new IllegalStateException("Operation is not open");
+
+        OperationContext operationContext = dsp.createOperationContext(this);
+        operationContext.pushScope();
+
+        // get the loop limit from configuration
+        SConfiguration configuration= EngineDriver.driver().getConfiguration();
+        int MAX_LOOP = configuration.getRecursiveQueryIterationLimit();
```
dgomezferro: Make MAX_LOOP lowercase.

yxia92:

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSet.java
yxia92: Thanks @dgomezferro, I've made the change accordingly.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
yxia92: recursion limit can now be controlled by a session level property recursivequeryiterationlimit.
Also runtime exception will be raised when limit is reached.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/RecursiveUnionOperation.java
yxia92: Fix is updated accordingly.

spliceci:
TEST SUCCEEDED +1

PR #2701, opened 2019-04-16
DB-8156 Fix TRIM() trim type same as alias

changli6:


changli6:

spliceci:
TEST FAILED

changli6:

spliceci:
TEST SUCCEEDED +1

carolp-503:
OK for R2.7 - Customer waiting

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TernaryOperationIT.java
```diff
@@ -213,9 +213,51 @@ public void testTRIM() throws Exception {
...
+        rs = methodWatcher.executeQuery(sql);
+        rs.next();
+        iCell = rs.getInt(1);
+        Assert.assertEquals("Wrong trim result", 4, iCell);
+
+        sql = "select count(*) from ( select trim (t.c) from a t ) tab";
+        rs = methodWatcher.executeQuery(sql);
+        rs.next();
+        iCell = rs.getInt(1);
+        Assert.assertEquals("Wrong trim result", 4, iCell);
```
yxia92: Can you add the following two test cases too:
1) negative test case:
select trim (leading.c) from a as leading;
2) positive test case:
select trim ("leading".c) from a as "leading";

changli6:
The STRIP function supports the following formats:

STRIP(trimSource, trimType, trimChar)
STRIP(trimSource, trimType)

The LOOKAHEAD(<COMMA>) in line 7652 is used to identify if the user specified a trimChar. It consumes <COMMA> after the trimType.
The ansiTrimSpecFollows() consumed <COMMA> is the one before the trimType.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TernaryOperationIT.java
changli6: sure, added these tests.

yxia92: Thanks @changli6

yxia92:
For the two scenarios of STRIP(without and without trimChar) that you listed, there is no ambiguity, and no need of LOOKAHEAD to determine which branch to go. That is why I asked to remove. If we look at the other path that does not contains TrimSpec, it also need to handle two scenarios:

STRIP(trimSource, trimChar)
STRIP(trimSource)

and no LOOKAHEAD is used.

```
        LOOKAHEAD ({!ansiTrimSpecFollows()})
        (
            // <source> COMMA <trimChar>
            <COMMA> trimChar = additiveExpression(null,0) <RIGHT_PAREN>
            {
                return getTrimOperatorNode(trimSpec, trimChar, trimSource, null);
            }
        |
            // <source>
            <RIGHT_PAREN>
            {
                return getTrimOperatorNode(trimSpec, null, trimSource, null);
            }

        )
```

changli6:
I got your point @yxia92 It is a useless line. Thank you!

PR #2705, opened 2019-04-22
DB-8200 fix NPE and wrong result for nested loop join with UNION oper…(2.7)

yxia92:
…ation as right table


carolp-503:
Approved for R2.7

PR #2707, opened 2019-04-24
DBAAS-1665 Add dbaas profile (2.7)

bklo94:
Adds a dbaas profile to create the needed shaded jars and tar.gz needed for the cloud builds.

Build with a 
`mvn clean install -Pcore,dbaas,ee`


jaceklaskowski:
I've got no knowledge about dbaas (which I think begs for some changes to README.md as part of the PR), but I'm wondering how this profile is different from existing profiles (e.g. cdhX)? My concern as an outsider is that I'm currently having a lot of troubles selecting the proper one to extend Splice Machine and yet another profile does not make my "integrator" life any easier.

Moreover, I can see that whatever build script(s) are used to publish SM to the public repo, only a subset of profiles are published (see http://repository.splicemachine.com/nexus/content/groups/public/com/splicemachine/hbase_storage/2.8.0.1914-SNAPSHOT/) and that's another reason why having a separate profile makes little to no sense (unless we delete the others to make room for new ones).

I also noticed that the profile relies on Spark-specific jars with some already deprecated, e.g. spark streaming is of no use these days and should be avoided in most cases.

bklo94:
@jaceklaskowski dbaas is for the cloud service, since their deployment is different, but is only locked into one profile. 

The latest snapshot didn't pass testing for cdh, so that wasn't deployed. If you look at last week's
(http://repository.splicemachine.com/nexus/content/groups/public/com/splicemachine/hbase_storage/2.8.0.1913-SNAPSHOT/), our current active profiles are present (mainly the missing cdh profiles that you see that are missing). Though I do agree that some of the profiles should be deleted from the pom.xml and assembly.

pom.xml
```diff
@@ -1530,6 +1550,7 @@
                 <spark.major.version>spark2.2</spark.major.version>
                 <zookeeper.version>3.4.5-cdh5.14.0</zookeeper.version>
                 <spark.version>2.2.2</spark.version>
+                <scala.version>2.11.8</scala.version>
```
jaceklaskowski: Why is this 2.11.8 while the others are 2.11.6? Why so many duplicates? Why not to default to 2.11.6 globally and overwrite with a custom value per profile?

PR #2711, opened 2019-04-26
SPLICE-2241 Cleanup Spark Adapter docs (2.7)

dgomezferro:


splice_spark/pom.xml
```diff
@@ -224,8 +224,32 @@
                 <version>1.5.0.Final</version>
             </extension>
         </extensions>
-        <plugins>
-            <plugin>
+	<plugins>
+		<plugin>
+			<groupId>org.apache.maven.plugins</groupId>
+			<artifactId>maven-site-plugin</artifactId>
+			<version>3.0</version>
```
jaceklaskowski: I think this and the other plugin versions should all be in one single place under `pluginManagement` in the top-level pom.xml. That way we cut the time needed to download all deps for the build (not to mention that there'll be no dups).

splice_spark/pom.xml
```diff
@@ -224,8 +224,32 @@
...
+					<plugin>
+						<artifactId>maven-project-info-reports-plugin</artifactId>
+						<version>2.2</version>
+					</plugin>
+					<plugin>
+						<groupId>net.alchim31.maven</groupId>
+						<artifactId>scala-maven-plugin</artifactId>
+						<version>3.4.2</version>
+						<configuration>
+							<jvmArgs>
```
jaceklaskowski: I'm curious about the number of spaces for indentation used? 8? Why not 4? Even for such a simple change we're almost at 144-char limit (!)

splice_spark/src/main/scala/com/splicemachine/spark/splicemachine/SpliceJDBCOptions.scala
```diff
@@ -24,6 +24,19 @@ object SpliceJDBCOptions {
     name
   }
 
+  /**
+    * JDBC Url including authentication mechanism (user/password, principal or principal/keytab)
```
jaceklaskowski: What do you think about adding an example to make the doc even better.

splice_spark/src/main/scala/com/splicemachine/spark/splicemachine/SpliceJDBCOptions.scala
```diff
@@ -24,6 +24,19 @@ object SpliceJDBCOptions {
...
+    */
+  val JDBC_URL = newOption("url")
+
+  /**
+    * Whether to create relations using SplicemachineContext.internalDf() by default. Defaults to "false"
+    */
   val JDBC_INTERNAL_QUERIES = newOption("internal")
+
+
```
jaceklaskowski: nit: 2 newlines?

splice_spark/src/main/scala/com/splicemachine/spark/splicemachine/SpliceJDBCOptions.scala
```diff
@@ -24,6 +24,19 @@ object SpliceJDBCOptions {
...
+  /**
+    * Temporary directory used by SplicemachineContext.internalDf() to hold temporary data. It has to be accessible by the client's user and SpliceMachine user
```
jaceklaskowski: nit: "temporary" used twice. My understanding is that a temporary directory holds temporary data by definition, doesn't it?

splice_spark/src/main/spark2.2/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -45,20 +45,25 @@ import org.apache.hadoop.hbase.util.Bytes
...
-  * @param options
+  * @param options Supported options are SpliceJDBCOptions.JDBC_URL, SpliceJDBCOptions.JDBC_INTERNAL_QUERIES, SpliceJDBCOptions.JDBC_TEMP_DIRECTORY
   */
 class SplicemachineContext(options: Map[String, String]) extends Serializable {
+  private[this] val url = options.get(JDBCOptions.JDBC_URL).get
 
-  val url = options.get(JDBCOptions.JDBC_URL).get
-  
+  /**
+    *
```
jaceklaskowski: nit: a new line

splice_spark/src/main/spark2.2/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -45,20 +45,25 @@ import org.apache.hadoop.hbase.util.Bytes
...
+    * Context for Splice Machine, specifying only the JDBC url.
+    *
+    * @param url JDBC Url with authentication parameters
```
jaceklaskowski: Why not to reference `JDBCOptions.JDBC_URL` so the doc is automatically copied / accessible?

splice_spark/src/main/spark2.2/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -407,6 +415,15 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
     }
   }
 
+  /**
+    *
```
jaceklaskowski: nit: a new line

splice_spark/src/main/spark2.3/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -340,7 +349,7 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
       execute(s"ANALYZE TABLE $tableName ESTIMATE STATISTICS SAMPLE $samplePercent PERCENT")
   }
 
-  lazy val tempDirectory = {
+  lazy private[this]val tempDirectory = {
```
jaceklaskowski: nit: `lazy` keyword after `private[this]` for consistency?

splice_spark/src/main/spark2.3/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -356,28 +365,28 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
...
   }
 
   /**
-    * SQL to Dataframe translation.  (Lazy)
+    * SQL to Dataset translation.  (Lazy)
     *
-    * @param sql
-    * @return
+    * @param sql SQL query
+    * @return Dataset[Row] with the result of the query
```
jaceklaskowski: `Dataset[Row]` is included in the signature so there's no need to include it in the `@return`. Moreover `Dataset[Row]` does not include anything but describes a distributed computation that once executed (e.g. `show`) gives a result (aka _query result_). I'd simply remove the line as it adds nothing at all.

splice_spark/src/main/spark2.3/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -356,28 +365,28 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
...
     */
   def df(sql: String): Dataset[Row] = {
     SparkUtils.resultSetToDF(internalConnection.createStatement().executeQuery(sql));
   }
 
   /**
-    * SQL to Dataframe translation.  (Lazy)
+    * SQL to Dataset translation.  (Lazy)
```
jaceklaskowski: The comments get worse every change :( `Dataset` is lazy by design so there's no point mentioning it (it's implied). Rather than describing what's included in the signature, I'd rather focus on why a developer would want to use it.

PR #2714, opened 2019-04-29
DBAAS-1665 Add dbaas profile

bklo94:
There is also a related EE PR.


pom.xml
```diff
@@ -1497,6 +1497,40 @@
...
+        <profile>
+            <id>dbaas</id>
+            <activation>
+                <property>
+                    <name>env</name>
+                    <value>dbaas</value>
+                </property>
+            </activation>
+            <properties>
+                <cdh.parcel.version>5.14.0</cdh.parcel.version>
```
jaceklaskowski: Just noticed all the uses of `5.14.0` and wonder if it could be somehow reused through the profile.

platform_it/pom.xml
```diff
@@ -1176,7 +1176,14 @@
                 <extraDependency3>splice_ee</extraDependency3>
             </properties>
         </profile>
-
+        <profile>
+            <id>dbaas</id>
+            <properties>
+                <extraDependency>splice_ee</extraDependency>
+                <extraDependency2>splice_ee</extraDependency2>
+                <extraDependency3>splice_ee</extraDependency3>
```
jaceklaskowski: nit: you may only guess how I (=a newcomer to the code) feel about `extraDependencyN` where this `N` says nothing about the purpose of the property. I'd consider it a naming bug.

platforms/dbaas/docs/installation.md
```diff
@@ -0,0 +1,4 @@
+
+# Installing and Configuring Splice Machine for dbaas profile
+
+The profile is for the cloud build service. Please refer to the cloud build confluence documentation in order to build and deploy this profile.
```
jaceklaskowski: And where is the _cloud build confluence documentation"_?! Why is there no link here? What's _"cloud build service"_?

splice_spark/pom.xml
```diff
@@ -404,6 +404,82 @@
         </plugins>
     </build>
     <profiles>
+        <profile>
+            <id>dbaas</id>
+            <properties>
+                <scala.version>2.11.8</scala.version>
```
jaceklaskowski: `dbaas` profile is configured in multiple places (begs for refactoring) and `scala.version` is dragged along. Why don't we have a "top-level" profile with all the shared properties. I'm concerned that it could lead to version mismatch eventually.

splice_spark/pom.xml
```diff
@@ -404,6 +404,82 @@
...
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                        <argLine>-Djava.library.path=${project.build.directory}/../../assembly/${envClassifier}/native/${os.detected.classifier}:/usr/local/lib/ -Dhbase.rootdir=${project.build.directory}/../../platform_it/target/hbase</argLine>
+                        </configuration>
+                    </plugin>
+                     <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-shade-plugin</artifactId>
+                        <version>3.2.1</version>
```
jaceklaskowski: nit: Remove `version` and define the plugin in `pluginManagement` in the parent/top-level pom.xml. Makes plugin management easier.

splice_spark/pom.xml
bklo94: I declared it in this profile since there actually is a version mismatch between the dbaas service versus the on-prem service due to differing dependencies. <scala.version> declared in line 26 is overwritten when activating the profile.

hbase_sql/pom.xml
```diff
@@ -299,4 +299,121 @@
...
+                            <goal>jar</goal>
+                          </goals>
+                        </execution>
+                      </executions>
+                    </plugin>
+
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-shade-plugin</artifactId>
+                        <version>3.2.1</version>
```
jaceklaskowski: Remove the version and define the plugin in `pluginManagement` in the parent/top-level pom.xml.

splice_spark/pom.xml
jaceklaskowski: Why is this then in two places?! Here and in `scala_util/pom.xml`? Why is this not part of the top-level `dbaas` profile in `pom.xml` (that this PR just defined)? I must be missing something...

bklo94:

spliceci:
Job Failure
 
TEST FAILED

spliceci:
TEST FAILED

PR #2716, opened 2019-05-01
DB-8195 Calculate average split size for the splits=n hint (2.7).

msirek:


carolp-503:
This merge should hold off until next week's 2.7 build to minimize the changes since we have customers waiting on this week's build.  But it is approved for R2.7 after May 3rd.

PR #2720, opened 2019-05-06
DB-8168 Implemented capability to pass arguments from a particular error scenario from server to client shell.

tjo7:


tjo7:

spliceci:
TEST SUCCEEDED +1

msirek:
@tjo7 The fix looks well thought out.  I believe the error message infrastructure may already have a way to pass arguments error message arguments around without having to introduce a new Exception class that includes arguments (SQLSyntaxErrorExceptionArg).  

In Exceptions.java, method parseException is used to intercept a SqlException and convert it into a StandardException.  I believe the method you use of decoding the arguments from the Sqlca class could also be applied in parseException to generate the proper StandardException that includes arguments.  I tested it, and it seems to work.  

Example code can be found here:
https://github.com/splicemachine/spliceengine/pull/2722/commits/4a84ce8360dae3ebac46eb3fa673fa6520c6fae7

Method generateCsSQLException ends up calling getSQLException to build an EmbedSQLException which propagates the arguments from this StandardException, so we could leverage this pre-existing infrastructure.

I am not sure which approach is best.. just trying to avoid extra infrastructure or classes.  Maybe @dgomezferro or @jyuanca could comment.  They are more familiar with the error message system.

splice_machine/src/main/java/com/splicemachine/pipeline/Exceptions.java
```diff
@@ -53,6 +54,17 @@ public static StandardException parseException(Throwable e,PipelineExceptionFact
             Throwable rootCause = Throwables.getRootCause(e);
             if (rootCause instanceof StandardException) {
                 return (StandardException) rootCause;
+            } else if (rootCause instanceof SqlExceptionArgs) {
+                SqlExceptionArgs sqlException = (SqlExceptionArgs) rootCause;
+                Throwable next = sqlException.getNextException();
+                Object[] args = sqlException.getArgs();
+                String messageID = null;
+                if (args.length >= 6) {
+                    messageID = (String) args[5];
```
jyuanca: Can you please document the magic number 6?

db-client/src/main/java/com/splicemachine/db/client/am/SQLExceptionFactory.java
```diff
@@ -48,6 +49,10 @@ public static SQLException notImplemented (String feature) {
     public SQLException getSQLException (String message, String sqlState, 
             int errCode) {
         return new SQLException (message, sqlState, errCode);           
-    }    
+    }
+
+    public SQLException getSQLException(String message, String sqlState, int errorCode, Object[] args) {
+        return new SQLException (message, sqlState, errorCode);
```
jaceklaskowski: nit: remove the space between `SQLException` and `(`

db-client/src/main/java/com/splicemachine/db/client/am/SQLExceptionFactory40.java
```diff
@@ -64,7 +57,7 @@
      * @param errCode db error code
      */
     public SQLException getSQLException (String message, String sqlState, 
-                                                            int errCode) { 
+                                                            int errCode) {
```
jaceklaskowski: While you're at it, could you make it consistent with the newly-added method below and indent `int errCode`?

db-client/src/main/java/com/splicemachine/db/client/am/SQLExceptionFactory40.java
```diff
@@ -104,4 +97,13 @@ public SQLException getSQLException (String message, String sqlState,
         }
         return ex;
     }
+
+    public SQLException getSQLException (String message, String sqlState,
+                                         int errCode, Object[] args) {
+        SQLSyntaxErrorExceptionArgs ex = null;
+        ex = new SQLSyntaxErrorExceptionArgs(message, sqlState, errCode);
```
jaceklaskowski: nit: Merge the lines 103 and 104

db-client/src/main/java/com/splicemachine/db/client/am/Sqlca.java
```diff
@@ -409,13 +414,18 @@ private void processSqlErrmcTokens(byte[] tokenBytes) {
...
+            int numMessages = tokens.length;
+            if (numMessages > 0) {
+                sqlErrorArgs = tokens[0].split(SQLERRMC_TOKEN_DELIMITER);
+            }
+
+            for (int i = 1; i < numMessages; i++) {
                 // All but the first message are preceded by the SQL state
-                // (five characters) and a colon. Extract the SQL state and
-                // clean up the token. See
+                // (five characters) and a colon. Extract the SQL state and// clean up the token. See
```
jaceklaskowski: nit: missing a new line before `// clean up`?

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/SQLExceptionFactory40.java
```diff
@@ -99,7 +93,8 @@ else if (sqlState.startsWith(SQLState.TRANSACTION_PREFIX)) {
             ex = new SQLTransactionRollbackException(message, sqlState,
                     severity, t);
         } else if (sqlState.startsWith(SQLState.LSE_COMPILATION_PREFIX)) {
-            ex = new SQLSyntaxErrorException(message, sqlState, severity, t);
+            //ex = new SQLSyntaxErrorException(message, sqlState, severity, t);
```
jaceklaskowski: remove the line (it's in the git repo anyway if anyone would revert)

db-client/pom.xml
```diff
@@ -54,6 +54,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.splicemachine</groupId>
+            <artifactId>db-engine</artifactId>
```
OlegMazurov: That will bloat client with lots of unnecessary dependencies.

PR #2722, opened 2019-05-07
DB-8168 Decode args from Sqlca.

msirek:
Do NOT merge this PR.  For discussion purposes only.

### DIFF
--- a/db-client/src/main/java/com/splicemachine/db/client/am/SqlException.java
+++ b/db-client/src/main/java/com/splicemachine/db/client/am/SqlException.java
@@ -102,7 +102,7 @@ public class SqlException extends Exception implements Diagnosable {
     /** 
      *  The message utility instance we use to find messages
      *  It's primed with the name of the client message bundle so that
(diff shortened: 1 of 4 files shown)


db-client/src/main/java/com/splicemachine/db/client/am/SqlException.java
```diff
@@ -102,7 +102,7 @@
     /** 
      *  The message utility instance we use to find messages
      *  It's primed with the name of the client message bundle so that
-     *  it knows to look there if the message isn't found in the
+     *  ∫it knows to look there if the message isn't found in the
```
jaceklaskowski: What's the change for?

msirek: @jaceklaskowski It's a mistaken key stroke, but no need to correct it.  This PR is not going to merge.  I just opened it to easily share some code for discussion purposes in a different PR.

PR #2726, opened 2019-05-10
DB-7752 Multiple OlapServers with different Yarn queues (2.7)

dgomezferro:


carolp-503:
We need to hold off on this PR while we work out our release strategy.  Let's merge this into Master only for now, and pull it into a 2.x customer release once we've got the overall release plan together.

Don't merge into 2.7

arnaud-splice:
Do we still want to bring that in 2.7?
If not, I'd suggest closing the PR.

PR #2735, opened 2019-05-23
DB-7752 Multiple OlapServers with different Yarn queues

dgomezferro:


dgomezferro:
Same PR with a fix for the mem profile

PR #2738, opened 2019-05-28
Splice-2343 Support builtin scalar function LEFT used by MicroStrategy

ngpham:
SPLICE-2343: Support builtin scalar function LEFT used by MicroStrategy. Essentially LEFT() with padding option.
@changli6 @dgomezferro @msirek @jyuanca

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
@@ -106,7 +106,7 @@ public class TernaryOperatorNode extends OperatorNode
 		{ClassName.DateTimeDataValue, "java.lang.Integer", ClassName.NumberDataValue}, // time.timestampadd( interval, count)
 		{ClassName.DateTimeDataValue, "java.lang.Integer", ClassName.DateTimeDataValue},// time2.timestampDiff( interval, time1)
 		{ClassName.ConcatableDataValue, ClassName.StringDataValue, ClassName.StringDataValue}, // replace{}
-		{ClassName.StringDataValue, ClassName.NumberDataValue, ClassName.StringDataValue}
+		{ClassName.StringDataValue, ClassName.NumberDataValue, ClassName.NumberDataValue} // left
 	};
 
 	/**
@@ -362,14 +362,8 @@ public class TernaryOperatorNode extends OperatorNode
 		{
 			leftOperand.generateExpression(acb, mb);
 			mb.upCast(leftInterfaceType);
-			if (rightOperand != null) {
-				rightOperand.generateExpression(acb, mb);
-				mb.upCast(rightInterfaceType);
-			} else {
-				mb.pushNull(rightInterfaceType);
-			}
 			mb.getField(field);
-			nargs = 3;
+			nargs = 2;
 			receiverType = receiverInterfaceType;
 		}
 		else if (operatorType == TIMESTAMPADD || operatorType == TIMESTAMPDIFF)
@@ -920,15 +914,11 @@ public class TernaryOperatorNode extends OperatorNode
 		{
 			leftOperand.setType(new DataTypeDescriptor(TypeId.INTEGER_ID, false));
 		}
-		if ((rightOperand != null) && rightOperand.requiresTypeFromContext())
-		{
-			rightOperand.setType(new DataTypeDescriptor(TypeId.getBuiltInTypeId(Types.VARCHAR), true));
-		}
 
 		bindToBuiltIn();
 
 		if (!leftOperand.getTypeId().isIntegerNumericTypeId())
-			throw StandardException.newException(SQLState.LANG_DB2_FUNCTION_INCOMPATIBLE, "LEFT", "FUNCTION");
+			throwBadType("LEFT", leftOperand.getTypeId().getSQLTypeName());
 
 		receiverType = receiver.getTypeId();
 		switch (receiverType.getJDBCTypeId())
@@ -944,21 +934,6 @@ public class TernaryOperatorNode extends OperatorNode
 			}
 		}
 
-		if (rightOperand != null)
-		{
-			switch (rightOperand.getTypeId().getJDBCTypeId())
-			{
-				case Types.CHAR:
-				case Types.VARCHAR:
-				case Types.LONGVARCHAR:
-				case Types.CLOB:
-					break;
-				default:
-				{
-					throwBadType("LEFT", rightOperand.getTypeId().getSQLTypeName());
-				}
-			}
-		}
 		if (receiverType.getTypeFormatId() == StoredFormatIds.CLOB_TYPE_ID) {
 			resultType = receiverType;
 		}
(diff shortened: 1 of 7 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2035,6 +2035,25 @@ public class SQLParser
 						    getContextManager());
     }
 
+    /**
+     * Get a left() node from
```
jaceklaskowski: s/Get/Gets

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TernaryOperationIT.java
```diff
@@ -284,8 +287,27 @@ public void testSubString() throws Exception {
...
+            ") tb(name, cut);";
+
+        List<String> expected = new ArrayList<String>(2);
+        expected.add("hello");
+        expected.add("hey dudeshortshortsh");
+
+        rs = methodWatcher.executeQuery(sql);
+        int i = 0;
+        while (rs.next()) {
+            Assert.assertEquals(expected.get(i), rs.getInt(1));
```
jaceklaskowski: Haven't executed it myself, and am curious about `rs.getInt(1)`. How can `int` be equal to `String` in this (test) case?

jaceklaskowski:
Please change the title to match the goal of the PR. Thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
ngpham: I fixed the typos, thanks! Also, I configured the IDE to remove trailing spaces on save, thus many chances in the file.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/TernaryOperationIT.java
ngpham: Thanks, good catch! Fixed. I was testing with length(), then changing to string compare. My concern is why this passed the test. For now, the fix should be correct. Please review.

jaceklaskowski: If you're at it, I'd rather use the array of strings `expected` as the source of truth, i.e. I'd iterate over the expected elements not `rs.next()` in `while`.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
```diff
@@ -885,6 +906,80 @@ public ValueNode substrBind()
...
+			switch (rightOperand.getTypeId().getJDBCTypeId())
+			{
+				case Types.CHAR:
+				case Types.VARCHAR:
+				case Types.LONGVARCHAR:
+				case Types.CLOB:
+					break;
+				default:
+				{
+					StandardException.newException(SQLState.LANG_DB2_FUNCTION_INCOMPATIBLE, "LEFT", "FUNCTION");
```
jaceklaskowski: Why are you using `StandardException.newException` while `throwBadType("LEFT", receiverType.getSQLTypeName());` above?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
```diff
@@ -885,6 +906,80 @@ public ValueNode substrBind()
...
+				}
+			}
+		}
+		if (receiverType.getTypeFormatId() == StoredFormatIds.CLOB_TYPE_ID) {
+			resultType = receiverType;
+		}
+
+		int resultLen = receiver.getTypeServices().getMaximumWidth();
+
+		if (leftOperand != null && leftOperand instanceof ConstantNode)
```
jaceklaskowski: Here you accept that `leftOperand == null` while it's not possible given the lines 919 and 930, is it?

ngpham: Yes, that null check is redundant.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
ngpham: Changed to throwBadType().

ngpham:
I already pushed commits addressing the comments.

ngpham:
Reimplemented without padding, as discussed at [https://splice.atlassian.net/browse/SPLICE-2343](https://splice.atlassian.net/browse/SPLICE-2343).

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2042,13 +2042,13 @@ public class SQLParser
...
-    ValueNode getLeftOperatorNode(ValueNode stringValue, ValueNode length, ValueNode padding) throws StandardException
+    ValueNode getLeftOperatorNode(ValueNode stringValue, ValueNode length) throws StandardException
     {
         return (ValueNode) nodeFactory.getNode(
                 C_NodeTypes.LEFT_OPERATOR_NODE,
                 stringValue,
                 length,
-                padding,
+                null,
                 ReuseFactory.getInteger(TernaryOperatorNode.LEFT),
```
msirek: LEFT now takes 2 arguments.  Should we make it a BinaryOperatorNode?

ngpham: Since LEFT() and RIGHT() are similar, please see my comment at: [https://github.com/splicemachine/spliceengine/pull/2746#discussion_r290761412](https://github.com/splicemachine/spliceengine/pull/2746#discussion_r290761412)

yxia92:
@ngpham I may have missed some discussions, since the decision is not to support padding, why don't we translate left and right to substr at sqlgrammer.jj? We could have saved two tree node types. Thanks!

ngpham:
@yxia92 Yes, I considered that, too, where we rewrite at parsing phase. This is quick and easy, however, the drawback is in error handling. i.e. when we have bad types, the error will be reported as a SUBSTR() function error, which could be confusing. I am not sure if there is a way to overcome that. Any idea?

yxia92:
A quick dirty way may be to generalize the error message in messages.xml. for example,            
```
 <msg>
                <name>22011</name>
                <text>The second or third argument of the SUBSTR function is out of range.</text>
</msg>
```
Instead of saying SUBSTR function, it could say the length parameter of the SUBSTR/LEFT/RIGHT is out of range.

To be more specific we need to pass down the function name/kind to the substr(), and further pass down as a parameter for the exception.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
msirek: OK

carolp-503:
We need this change in 2.7 as well as Master.  Please submit a separate PR for 2.7

PR #2742, opened 2019-06-03
DB-8036 allow more predicates to be pushed down into unions

yxia92:
Please refer to [DB-8036](https://splicemachine.atlassian.net/browse/DB-8036) for details.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
@@ -320,8 +320,8 @@ public class SelectNode extends ResultSetNode{
     public FromList getFromList(){ return fromList; }
 
     /**
-     * Remap the column reference node to underlying column reference in the select, if possible.
-     * The select node here corresponds to a union branch
+     * Find the corresponding column reference (as used in the select node) for the given column reference.
+     * The select node here corresponds to a branch underneath a UnionNode
      * @param    cr        ColumnReference to find a match from the SELECT node
      * @return ColumnReference    ColumnReference pointing to the source of cr, if found
      */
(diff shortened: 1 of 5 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -320,40 +320,22 @@ public void printSubNodes(int depth){
...
     /**
-     * Find colName in the result columns and return underlying columnReference.
-     * Note that this function returns null if there are more than one FromTable
-     * for this SelectNode and the columnReference needs to be directly under
-     * the resultColumn. So having an expression under the resultSet would cause
-     * returning null.
-     *
-     * @param    colName        Name of the column
-     * @return ColumnReference    ColumnReference to the column, if found
+     * Remap the column reference node to underlying column reference in the select, if possible.
```
jaceklaskowski: Does the method really "remap" or simply gives a source column reference? The name suggests the latter.

yxia92: Thanks @jaceklaskowski for the review. "Remap" here I meant to "find the corresponding source column reference". I updated the comment to make the comment and function name consistent.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -320,8 +320,8 @@ public void printSubNodes(int depth){
     public FromList getFromList(){ return fromList; }
 
     /**
-     * Remap the column reference node to underlying column reference in the select, if possible.
-     * The select node here corresponds to a union branch
+     * Given a column reference, find the corresponding column reference in the select, if possible.
```
jaceklaskowski: I'd word it as follows:

```
Finds the column reference (as used in select) for a given column reference
```

The rest is in `@param` and `@return`

yxia92: Thanks. I've updated the comment accordingly.

PR #2744, opened 2019-06-03
DB-8394 Remove extra wait for Spark queries

dgomezferro:


splice_machine/src/main/java/com/splicemachine/derby/stream/function/MergeAllAggregatesFunction.java
```diff
@@ -43,7 +43,6 @@ public ExecRow call(ExecRow locatedRow1, ExecRow locatedRow2) throws Exception {
             aggregates = op.aggregates;
             initialized = true;
         }
-        operationContext.recordRead();
```
yxia92: Why is this line removed?

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
```diff
@@ -174,16 +174,16 @@ private void updateLimitOffset() throws StandardException {
...
-            olapFuture.get(time, unit);
+            olapFuture.cancel(true);
     }
 
     @Override
     public void close() throws Exception {
         streamListener.stopAllStreams();
         if (olapFuture != null)
-            olapFuture.cancel(true);
+            olapFuture.cancel(false);
```
yxia92: Interrupt() and close() seem to be the same, as the parameter true or false for the cancel() function is not used. Is it intended to keep two separate functions so that in the future different operations will be applied?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/MergeAllAggregatesFunction.java
dgomezferro: It's an unrelated change, we were overcounting rows produced on GroupedAggregattes. I'd rather leave it in since it's a very simple change.

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
dgomezferro: You are right, I thought we were handling it in OlapFuture.cancel() and my limited testing didn't reproduce the exception from SPLICE-1895  

I addressed it now by ignoring the Cancel message when we cancel without interrupt.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/MergeAllAggregatesFunction.java
yxia92: sure, thanks!

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
yxia92: Thanks. Now that remoteQueryClient.close() is used for successful execution, interrupt() is used for timeout or killed scenario, should we change the call of remoteQueryClient.close() in SpliceBaseOperation.kill() and timeout() to interrupt() instead?

dgomezferro: Yep, good catch. Thanks for the review!

PR #2746, opened 2019-06-04
SPLICE-2344 Support builtin scalar function RIGHT used by MicroStrategy.

ngpham:
Implementation as per discussion at [https://splice.atlassian.net/browse/SPLICE-2344](url).


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TernaryOperatorNode.java
```diff
@@ -85,23 +85,26 @@
 	public static final int TIMESTAMPADD = 4;
 	public static final int TIMESTAMPDIFF = 5;
 	public static final int REPLACE = 6;
-	static final String[] TernaryOperators = {"trim", "LOCATE", "substring", "like", "TIMESTAMPADD", "TIMESTAMPDIFF", "replace"};
-	static final String[] TernaryMethodNames = {"ansiTrim", "locate", "substring", "like", "timestampAdd", "timestampDiff", "replace"};
-	static final String[] TernaryResultType = {ClassName.StringDataValue, 
+	public static final int RIGHT = 7;
+	static final String[] TernaryOperators = {"trim", "LOCATE", "substring", "like", "TIMESTAMPADD", "TIMESTAMPDIFF", "replace", "right"};
+	static final String[] TernaryMethodNames = {"ansiTrim", "locate", "substring", "like", "timestampAdd", "timestampDiff", "replace", "right"};
```
msirek: The RIGHT function takes 2 parameters, so should be added to BinaryOperatorNode instead of TernaryOperatorNode.

ngpham: Indeed, I was considering that option, then decided against that. I am still making acquaintance with the code base, so, correct me if my observation is wrong.... The logic is closer to TernaryOperatorNode, which is a misleading name, it is more like built-in Function, I think. Whereas, for BinaryOperatorNode, it really deals with A <OP> B, the operands have a wider scope, and the code generation path is longer. Thus, it is more efficient and logical to deal with RIGHT(), a function, in the built-in-function-like node TernaryOperatorNode.

msirek: I took a look at the code with Nguyen.  Indeed bindexpression calls genSQLJavaSQLTree to build SQLToJavaValueNodes on top of the inputs, which may have some overhead.  Also, we currently don't strictly follow the convention that TernaryOperatorNode only handles functions with 3 parameters.

carolp-503:
Please submit a PR for this for 2.7 as well as for Master.  Thanks!

PR #2758, opened 2019-06-18
DB-8250: support bulk load from spark adapter as non-hbase user

jyuanca:


jyuanca:

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkInsertDataSetWriter.java
```diff
@@ -248,8 +249,18 @@ public TxnView getTxn(){
                 assert rowAndIndexes instanceof SparkDataSet;
 
                 partitionUsingRDDSortUsingDataFrame(bulkImportPartitions, rowAndIndexes, hfileGenerationFunction);
+                Map<Long, Boolean> granted = new HashMap<>();
+                try {
+                    if (token != null && token.length > 0) {
+                        granted = grantCreatePrivilege(allCongloms);
```
dgomezferro: I think we should grant the Create privileges unconditionally in H10PartitionAdmin.grantPrivilegesIfNeeded() and then add checks in all places where the CREATE privilege is used: preAddColumn, preBulkLoad, preCreateTable, preDropTable...

The bulkLoad* methods we would allow if the user has the right token, all the rest we would disallow unless it's the admin user.

hbase_sql/src/main/java/com/splicemachine/derby/stream/function/BulkImportFunction.java
```diff
@@ -92,19 +96,37 @@ public void call(Iterator<BulkImportPartition> importPartitions) throws Exceptio
                     fs.delete(sourceDir.getParent(), true);
                 }
             }
+            writeToken(fs, path);
```
dgomezferro: This location is only known to the user and HBase?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkInsertDataSetWriter.java
jyuanca: When namespace privilege was grated at H10PartitionAdmin.grantPrivilegesIfNeeded() during startup, we don't know which tables need CREATE privilege. We don't want to grant CREATE privileges for ALL tables to the user that run bulk load.

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BulkInsertDataSetWriter.java
dgomezferro: That makes sense, thanks!

PR #2762, opened 2019-06-18
DB-8097 use numOptimizables instead of numTablesInQuery to determine …

yxia92:
…if we want to do heuristic or exhaustive join order search for a query block


yxia92:

spliceci:
TEST SUCCEEDED +1

PR #2765, opened 2019-06-20
SPLICE-2302 Native spark aggregates.

msirek:
This feature allows grouped and scalar aggregations (e.g. SUM, COUNT, MAX, MIN, AVG) to be evaluated as native Spark operations on a DataFrame (Dataset<Row>).  This will cause a performance benefit to queries that do one or more joins followed by an aggregation.  The larger the size of the intermediate result sets compared to the input tables, the greater the performance boost.  For example, cross join followed by aggregation will show the most speedup.  But a join with unique join keys followed by an aggregation may not show any speedup at all.

This JIRA also relaxes the rules regarding Native Spark projections with Decimal expressions so that such expressions are now evaluated natively in Spark by default.  A database property is added to control this behavior. 

For more details, please see [SPLICE-2302](https://splice.atlassian.net/browse/SPLICE-2302?oldIssueView=true)


msirek:

spliceci:
TEST FAILED

msirek:

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -1080,4 +1105,220 @@ public TableSamplerBuilder sample(OperationContext operationContext) throws Stan
...
+                        if (aggregates[i].isDistinct())
+                            return null;
+                        inputColumns.add(sourceColName);
+                        break;
+                    case "SpliceStddevSamp":
+                        // Disabling this path for now.
+                        // stddev_samp may return NaN unexpectedly,
+                        // which on Splice causes the query to error out.
+                        // See SPARK-13860.
+                        if (stddev_samp_Disabled)
```
dgomezferro: Could we compute stddev_samp from stdev_pop here? I think samp = sqrt(((pop^2)*count)/(count-1)) 
That's assuming stddev_pop handles NULLs gracefully.
I'm not sure how much precision we could lose, but since tpcds uses the function it would be good to have some native support.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
```diff
@@ -270,7 +294,10 @@ else if (notExistsRightSide)
             }
         }
 
-        result = result.map(new CountProducedFunction(operationContext), /*isLast=*/true);
+        // Counting rows in this manner disables native spark execution,
+        // so only do it for non-native spark.
+        if (!usesNativeSparkDataSet)
+            result = result.map(new CountProducedFunction(operationContext), /*isLast=*/true);
```
dgomezferro: You can add 
`
    @Override
    public boolean hasNativeSparkImplementation() {
        return true;
    }`

To CountProducedFunction so that it gets ignored in a NativeSparkDataset. 

In the JoinRestrictionPredicateFunction you could return true if restriction is != null for instance.

dgomezferro:
@msirek did you find cases where forcing the Native conversion for aggregates didn't pay off? I'm guessing for grouped aggregates it will always pay off, maybe not for very simple scalar aggregates. Could we be a bit more aggressive on the conversion?

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
msirek: @dgomezferro It looks to be a divide by zero problem, so I added a 'when' function to return null when the divisor is zero:
`column = when(countColumn.gt(lit(1)), column);`

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
msirek: OK, I'll add hasNativeSparkImplementation to CountProducedFunction.  For JoinRestrictionPredicateFunction, returning true without providing an implementation of a native spark join filter would cause queries to return incorrect results.  I have opened [SPLICE-2362](https://splice.atlassian.net/browse/SPLICE-2362) to handle the native spark join filter.

msirek:
> @msirek did you find cases where forcing the Native conversion for aggregates didn't pay off? I'm guessing for grouped aggregates it will always pay off, maybe not for very simple scalar aggregates. Could we be a bit more aggressive on the conversion?

I found that native spark aggregation did not help at all when scanning a single table, and was actually a little slower than traditional splice operation.  Also, joins with join keys having 1 row per value did not see any benefit.  I'd rather not force conversion of the source dataset to native spark unless a test case is found where this is beneficial.  Conversion of a join with no native spark implementation won't trigger whole-stage code gen and all rows need to be built and streamed to the next operation anyway, defeating then purpose of whole-stage code gen.  Maybe one thing we can consider is to avoid joins that don't have a native spark implementation when there is an aggregation.  Though perhaps this should be a cost-based decision depending on the uniqueness of join key values.

msirek:

spliceci:
TEST SUCCEEDED +1

dgomezferro:
For grouped aggregates I think we should be forcing conversion, for instance with the following setup:
`create table a ( i int, j int);
insert into a values (1,1);
insert into a select i+(select count(*) from a), j from a; -- (x20)
insert into a select * from a;
insert into a select * from a;`

This query takes ~40s without conversion and ~20 with conversion:
`select count(*) from (select i, count(j) from a group by i) a;`

On a different note, I tried to use `call syscs_util.syscs_set_global_database_property('splice.execution.nativeSparkAggregationMode', 'forced');` to force conversion but that wouldn't work, on the OlapServer GenericAggregateOperation.nativeSparkForced() returned false.

If we group on the lower cardinality column instead, doing the conversion is a little bit slower, but I think in general it pays off.

msirek:
> For grouped aggregates I think we should be forcing conversion, for instance with the following setup:
> 
> ```
> insert into a values (1,1);
> insert into a select i+(select count(*) from a), j from a; -- (x20)
> insert into a select * from a;
> insert into a select * from a;
> ```
> 
> This query takes ~40s without conversion and ~20 with conversion:
> `select count(*) from (select i, count(j) from a group by i) a;`
> 
> On a different note, I tried to use `call syscs_util.syscs_set_global_database_property('splice.execution.nativeSparkAggregationMode', 'forced');` to force conversion but that wouldn't work, on the OlapServer GenericAggregateOperation.nativeSparkForced() returned false.
> 
> If we group on the lower cardinality column instead, doing the conversion is a little bit slower, but I think in general it pays off.

@dgomezferro  OK, I changed the default to forced, so we always pick native spark aggregation if possible.  Does your script include 'CALL SYSCS_UTIL.SYSCS_EMPTY_STATEMENT_CACHE();' ?  If not, setting nativeSparkAggregationMode might have no effect since we don't pick up the new setting for cached query plans ( when the query is not recompiled we don't write the new setting into the operation tree).

msirek:

spliceci:
TEST FAILED

msirek:
I am fixing up some IT issues found by making forced native spark aggregation the default.  The most significant is the number of included columns being off in the case of semijoin.  Splice defines the merged row as including all columns from the left and right relations, but when executing the join on native spark, spark will prune out the right relation columns.  We end up getting ArrayIndexOutOfBoundsException when trying to access a named column sometimes, or access the wrong column.

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
jenkins please test branch @hdp2.6.5

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -1808,18 +1808,14 @@ public ResultSetNode modifyAccessPaths() throws StandardException{
...
-
-            /* Now we're finally ready to generate the JoinNode and have it
-             * replace the 1st 2 entries in the FromList.
-             */
-            JoinNode joinNode;
-            if (!rightResultSet.getFromSSQ()) {
+			/* Concatenate the 2 ResultColumnLists */
+            leftRCList.nondestructiveAppend(rightRCList);
+
+			/* Now we're finally ready to generate the JoinNode and have it
```
jaceklaskowski: nit: formatting

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -821,6 +821,58 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
         return (plan != null && !plan.isEmpty() ? plan : f.getSparkName());
     }
 
+    public static boolean joinProjectsOnlyLeftTableColumns(JoinType joinType) {
+        return (joinType == DataSet.JoinType.LEFTANTI ||
+                joinType == DataSet.JoinType.LEFTSEMI);
+    }
+    private Dataset<Row> fixupColumnNames(JoinOperation op, JoinType joinType,
```
jaceklaskowski: nit: a new line between methods

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -821,6 +821,58 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
...
+                                          Dataset<Row> leftDF, Dataset<Row> rightDF,
+                                          Dataset<Row> joinedDF) {
+        if (!joinProjectsOnlyLeftTableColumns(joinType) &&
```
jaceklaskowski: Can you create a new predicate method that would describe what the complex `if` uses to execute the `true` body? That would make comprehension slightly easier.

Since the entire body is pretty much the `true` block of the `if`, I'd recommend negate the `if` condition and `return joinedDF`. With such `if` you could execute the `if` body outside `if`.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -850,10 +902,17 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
             }
             DataSet joinedSet;
 
-            if (op.wasRightOuterJoin)
-                joinedSet =  new NativeSparkDataSet(rightDF.join(leftDF,expr,joinType.RIGHTOUTER.strategy()), context);
-            else
-                joinedSet =  new NativeSparkDataSet(leftDF.join(rightDF,expr,joinType.strategy()), context);
+            if (op.wasRightOuterJoin) {
+                joinedSet = new NativeSparkDataSet(rightDF.join(leftDF, expr, joinType.RIGHTOUTER.strategy()), context);
+                NativeSparkDataSet nds = (NativeSparkDataSet)joinedSet;
```
jaceklaskowski: I'd write the two lines as follows:

```
NativeSparkDataSet nds = new NativeSparkDataSet(...)
joinedSet = nds
```

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -850,10 +902,17 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
...
+                nds.dataset = fixupColumnNames(op, joinType, rightDF, leftDF, nds.dataset);
+            }
+            else {
+                joinedSet = new NativeSparkDataSet(leftDF.join(rightDF, expr, joinType.strategy()), context);
+                NativeSparkDataSet nds = (NativeSparkDataSet)joinedSet;
```
jaceklaskowski: Same as above

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -885,11 +944,16 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
...
             }
-            Dataset<Row> joinedSet = leftDF.crossJoin(rightDF);
+            Dataset<Row> joinedDF = leftDF.crossJoin(rightDF);
             if (expr != null) {
-                joinedSet = joinedSet.filter(expr);
+                joinedDF = joinedDF.filter(expr);
             }
-            return new NativeSparkDataSet(joinedSet, context);
+            DataSet joinedSet = new NativeSparkDataSet(joinedDF, context);
+            NativeSparkDataSet nds = (NativeSparkDataSet)joinedSet;
```
jaceklaskowski: Same as above

msirek:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
msirek: @jaceklaskowski Thanks for your comments.  I will incorporate the changes in my next push.

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -778,6 +820,74 @@ private String planIfLast(AbstractSpliceFunction f, boolean isLast) {
...
+    }
+
+    // If the left DataSet defines more columns than the left DataFrame,
+    // the column names will be off.  Let's fill the gaps to match the
+    // ExecRowDefinition.
+    private Dataset<Row> fixupColumnNames(JoinOperation op, JoinType joinType,
+                                          Dataset<Row> leftDF, Dataset<Row> rightDF,
+                                          Dataset<Row> joinedDF) {
+        boolean needsFixup =
+                leftDFOfJoinProjectsFewerColumnsThanDefined(op, joinType, leftDF);
```
yxia92: Why would we have a situation where leftDF project fewer columns than defined?

yxia92: I talked to @msirek , one scenario would be a regular join whose left source is a semi join. For derby, if it is a semi join, the immediate join row includes columns from the right even if it is not needed(these columns are pruned later), for spark, the columns from the right table of the semi-join won't be projected to the join result row. This difference could cause leftDF of the regular join to project few columns (as it is semi-join result computed from spark) than defined (which is determined in Derby).

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByNode.java
```diff
@@ -960,6 +1039,7 @@ private void addGroupingIdColumnAndUpdateGroupingFunction() throws StandardExcep
     private void addNewColumnsForAggregation() throws StandardException{
         aggInfo=new AggregatorInfoList();
         List<SubstituteExpressionVisitor> havingRefsToSubstitute=null;
+        VirtualColumnNode vcForGroupingFunction = null;
```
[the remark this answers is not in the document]
msirek: removing it...

PR #2770, opened 2019-06-27
DB-8214 Improve OperationContext cloning and perform it in parallel

dgomezferro:


dgomezferro:

spliceci:
TEST FAILED

dgomezferro:

spliceci:
TEST FAILED

dgomezferro:

spliceci:
TEST FAILED

dgomezferro:

spliceci:
TEST SUCCEEDED +1

start-splice-cluster
```diff
@@ -168,7 +168,7 @@ pushd "${RUN_DIR}" > /dev/null
 if [[ ${HBASE_PROFILE} == ${IN_MEM_PROFILE} ]]; then
   MEM_LOG="${RUN_DIR}/spliceMem.log"
   echo "Starting MEM. Log file is ${MEM_LOG}"
-  (${MVN} -Pmem exec:java > ${MEM_LOG} 2>&1) &    ## IN MEMORY
+  (MAVEN_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=4000" ${MVN} -Pmem exec:java > ${MEM_LOG} 2>&1) &    ## IN MEMORY
```
OlegMazurov: Is it intended for StartMemDatabase in mem_sql/pom.xml? Does -agentlib:jdwp really belong here?

msirek: Also, this is overwriting MAVEN_OPTS instead of appending an additional option to it.  Should we preserve whatever MAVEN_OPTS which are already set as an environment variable and append new options?

dgomezferro: @OlegMazurov this is intended for debugging the server when you manually start it with `./start-splice-cluster -p mem`

@msirek Good point, appending it now

OlegMazurov: For platform profiles, we specify "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=4000" in platform_it/pom.xml. I think we should be consistent and do the same in mem_sql/pom.xml rather than in the start-splice-cluster script.

dgomezferro: @OlegMazurov I see what you mean now. That's not possible because for platform profiles we use Maven's `exec:exec` goal, which forks a new process and passes the jdwp settings to it. However on the mem profile we use the `exec:java` goal, which runs the `main()` method directly on the same Maven process, so we need to pass the jdwp settings to the Maven process.

If you think we should be consistent I can make the mem profile use also the `exec:exec` goal, but that requires a larger change.

PR #2777, opened 2019-07-03
DB-8297 Schema restriction for Ranger

wb14123:


wb14123:
Please also review ee project

wb14123:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
```diff
@@ -271,4 +271,13 @@ public TupleDescriptor buildDescriptor(
 	public static final String SYSSCHEMASVIEW_VIEW_SQL1 = "create view sysschemasView as \n" +
 			SUPER_USER_SCHEMA;
 
+
+	public static final String RANGER_USER_SCHEMA =
+            "select S.* from SYS.SYSSCHEMAS as S where S.SCHEMANAME not in " +
```
yxia92: Why do we choose double-negation here to get the accessible schemas? That is, NOT IN + SchemaFilterVTI returning the list of not-accessible schemas. Why don't we use the combination of IN + SchemaFilterVTI returning the list of accessible schemas? The latter usually is more performant.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
```diff
@@ -235,7 +235,10 @@
     public static final int DEFAULT_RECURSIVE_QUERY_ITERATION_LIMIT = 20;
 
     public static final String METADATA_RESTRICTION_ENABLED = "splice.metadataRestrictionEnabled";
-    public static final boolean DEFAULT_METADATA_RESTRICTION_ENABLED = true;
+    public static final String METADATA_RESTRICTION_DISABLE = "DISABLE";
```
yxia92: Could you change the name from METADATA_RESTRICTION_DISABLE to METADATA_RESTRICTION_DISABLED, and the value from "DISABLE" to "DISABLED"?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1343,7 +1355,11 @@ public boolean isMetadataAccessRestrictionEnabled() {
 
     @Override
     public void setMetadataAccessRestrictionEnabled() {
-        metadataAccessRestrictionEnabled = SIDriver.driver().getConfiguration().getMetadataRestrictionEnabled();
+        String metadataRestriction =
+                SIDriver.driver().getConfiguration().getMetadataRestrictionEnabled();
+        metadataAccessRestrictionEnabled =
+                metadataRestriction.equals(SQLConfiguration.METADATA_RESTRICTION_NATIVE) ||
+                        metadataRestriction.equals(SQLConfiguration.METADATA_RESTRICTION_RANGER);
         SpliceLogUtils.trace(LOG,"metadataAccessRestritionEnabled=%b",metadataAccessRestrictionEnabled);
```
yxia92: Could you help to change SpliceLogUtils.trace() to SpliceLogUtils.info(), this way, we can see the value during the boot time with the default logging option. 
Could you also print the value metadataRestriction directly? Thanks!

splice_machine/src/main/java/com/splicemachine/derby/vti/SchemaFilterVTI.java
```diff
@@ -0,0 +1,95 @@
...
+        LanguageConnectionContext lcc = activation.getLanguageConnectionContext();
+
+        TransactionController tc = lcc.getTransactionExecute();
+        DataDictionaryImpl dd = (DataDictionaryImpl)lcc.getDataDictionary();
+        List<SchemaDescriptor> schemas= dd.getAllSchemas(tc);
+
+        for(SchemaDescriptor schema: schemas) {
+            String schemaName = schema.getSchemaName();
+            StatementPermission statementPermission = new StatementSchemaPermission(schemaName, null, Authorizer.ACCESS_PRIV);
+            if (!lcc.getAuthorizer().canSeeSchema(lcc.getLastActivation(), statementPermission)) {
```
yxia92: Is there a particular reason to use lcc.getLastActivation() instead of the activation variable above?
Also a repeat of the same question from SYSSCHEMASRowFactory: why returning the non-accessible schemas instead of the accessible schemas directly?

yxia92:
@wb14123 Could you test the accessible schemas of a non-splice user under the spark path, for example,
Connect to splicemachine with useSpark=true, something like 
   connect 'jdbc:splice://localhost:1527/splicedb;username=...;password=...;useSpark=true' as spark_con;
select * from sysvw.sysschemasView;

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/Authorizer.java
```diff
@@ -122,8 +124,10 @@
 	  @exception StandardException Thrown if the operation is not allowed
 	 */
 	void authorize(int operation) throws StandardException;
-    
-	/**
+
```
jaceklaskowski: nit: indent

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -669,6 +669,33 @@ private void getBuiltinSystemSchemas(){
 
     }
 
+    public List<SchemaDescriptor> getAllSchemas(TransactionController tc) throws StandardException{
+        List<SchemaDescriptor> lists =new ArrayList<>();
+        TabInfoImpl ti=coreInfo[SYSSCHEMAS_CORE_NUM];
+        SYSSCHEMASRowFactory rf=(SYSSCHEMASRowFactory)ti.getCatalogRowFactory();
+        ScanController sc=tc.openScan(
+                ti.getHeapConglomerate(),
+                false,   // don't hold open across commit
```
jaceklaskowski: Can you replace this and the following comments with their corresponding constants (or local values) and use them instead, e.g.

```
boolean dontHoldOpenAcrossCommit = false
```

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -669,6 +669,33 @@ private void getBuiltinSystemSchemas(){
...
+                null,                //
+                null, // stop position -through last row
+                0);                           // stopSearchOperation - none
+        ExecRow outRow=rf.makeEmptyRow();
+        SchemaDescriptor schemaDescr;
+        while(sc.fetchNext(outRow.getRowArray())){
+            schemaDescr=(SchemaDescriptor)rf.buildDescriptor(outRow,null,this);
+            lists.add(schemaDescr);
+        }
+        sc.close();
```
jaceklaskowski: What happens when an exception is thrown before `sc.close()`? Could this lead to a resource (descriptor) leak?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -669,6 +669,33 @@ private void getBuiltinSystemSchemas(){
 
     }
 
+    public List<SchemaDescriptor> getAllSchemas(TransactionController tc) throws StandardException{
+        List<SchemaDescriptor> lists =new ArrayList<>();
```
jaceklaskowski: Could you change `lists` to `allSchemas` or `schemas` to make the name a bit more self-explanatory?

splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
jaceklaskowski: I'm also curious why not to follow the pattern of `METADATA_RESTRICTION_ENABLED` to use `splice.metadataRestrictionDisabled`? Wait...why can't we use `METADATA_RESTRICTION_ENABLED`?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
wb14123: The reason is when using `in`, it will optimized to `join`. And in the join operation, `RangerAuthorizer.canSeeSchema` will through null pointer error from `lcc`. I don't know why join operation will have error but using `not in` to avoid join can avoid this situation.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SQLConfiguration.java
wb14123: @yxia92 Fixed this.

@jaceklaskowski  `METADATA_RESTRICTION_ENABLED` is a configuration name and `METADATA_RESTRICTION_DISABLED` is the value of this configuration.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
wb14123: @yxia92  Fixed this

splice_machine/src/main/java/com/splicemachine/derby/vti/SchemaFilterVTI.java
wb14123: @yxia92  I missed the activation above. I don't think there is any difference so changed the code.

wb14123:
@yxia92 I tested it with Spark by query like `select count(*) from sys.sysviews --splice-properties useSpark=true`. Is that OK or do I need to run your suggested query?

yxia92:
Thanks @wb14123 For the test of spark path, the test "select count(*) from sys.sysviews --splice-properties useSpark=true" may not be good as it is not going to execute the view SYSSCHEMASVIEW_VIEW_RANGER. Also hinting "useSpark=true" on a view may not force the spark execution as useSpark is a hint associated with a base table. A safer way is to put "useSpark=true" in the connection string.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
yxia92: I see @wb14123. I'm a little concerned about the null pointer exception for join. Maybe we should understand better this exception and see if the same problem could surface when the whole ranger view(SYSSCHEMASVIEW_VIEW_RANGER) is used as the right side of a join. Thanks.

yxia92:

spliceci:
TEST SUCCEEDED +1

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -294,7 +294,9 @@ public static Configuration create(String hbaseRootDirUri,
         config.set("splice.authentication.impersonation.users", "dgf=splice;splice=*");
         config.setBoolean("splice.authentication.impersonation.enabled", true);
         config.set("splice.authentication.ldap.mapGroupAttr", "jy=splice,dgf=splice");
-   //     config.setBoolean("splice.metadataRestrictionEnabled", true);
+        // config.set("splice.authorization.scheme", "RANGER");
+        // config.set("splice.metadataRestrictionEnabled", "RANGER");
+
```
jyuanca: Minor comments: remove unused code

yxia92: @jyuanca The reason to keep the commented out lines is to make it easier for future ranger testing on standalone system. I'll put a comment there indicating that there are the two parameters to set when testing ranger authorization on standalone system.

PR #2780, opened 2019-07-08
SPLICE-2364 Properly close internal region scanners

OlegMazurov:
Use try-with-resources to ensure underlying scanners are closed and associated memory is released (when MemStoreLAB is enabled).
Reduce splice.txn.completedTxns.cacheSize for ITs to ease memory pressure.

### DIFF
--- a/platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
+++ b/platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
@@ -294,6 +294,7 @@ class SpliceTestPlatformConfig {
         config.set("splice.authentication.impersonation.users", "dgf=splice;splice=*");
         config.setBoolean("splice.authentication.impersonation.enabled", true);
         config.set("splice.authentication.ldap.mapGroupAttr", "jy=splice,dgf=splice");
+        config.setInt("splice.txn.completedTxns.cacheSize", 4096);
    //     config.setBoolean("splice.metadataRestrictionEnabled", true);
 
         if (derbyPort > SQLConfiguration.DEFAULT_NETWORK_BIND_PORT) {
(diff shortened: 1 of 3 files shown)


platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -294,6 +294,7 @@ public static Configuration create(String hbaseRootDirUri,
         config.set("splice.authentication.impersonation.users", "dgf=splice;splice=*");
         config.setBoolean("splice.authentication.impersonation.enabled", true);
         config.set("splice.authentication.ldap.mapGroupAttr", "jy=splice,dgf=splice");
+        config.setInt("splice.txn.completedTxns.cacheSize", 4096);
```
msirek: How did you pick 4K?  The current default is 128K.  Would it be beneficial to cache more transactions?

OlegMazurov: In real environment, the default remains 128K. This setting, 4096, is used only in standalone environment and, most importantly, integration tests. The goal was to fit ITs into 2GB for Splice 3.0 (HBase 2.0) and that cache was a large contributor to accumulating objects, which were not even needed in the context of rather short tests.

msirek: Sure, I get that.  I was just asking why 4K was picked instead of 8K or 16K, etc.  The fix looks good.

OlegMazurov: It's somewhat arbitrary: severely reduced to prevent object accumulation during ITs but still adequate for standalone workloads (including ITs). I wouldn't have a strong opinion against 8K or 2K.

PR #2781, opened 2019-07-09
SPLICE-2338 Spurious column name in error message when querying parquet external tables

ngpham:
Please see comments at [https://splice.atlassian.net/browse/SPLICE-2338](https://splice.atlassian.net/browse/SPLICE-2338). Tested locally with TPC-DS1. @dgomezferro @msirek @jyuanca @yxia92 @OlegMazurov


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -763,7 +784,14 @@ private static Column createFilterCondition(Dataset dataset,String[] allColIdInS
                 if (q.getVariantType() == Qualifier.VARIANT)
                     continue; // Cannot Push Down Qualifier
                 q.clearOrderableCache();
-                Column orCol = dataset.col(allColIdInSpark[(baseColumnMap != null ? baseColumnMap[q.getStoragePosition()] : q.getStoragePosition())]);
+//                Column orCol = dataset.col(allColIdInSpark[(baseColumnMap != null ? baseColumnMap[q.getStoragePosition()] : q.getStoragePosition())]);
+                Column orCol = dataset.col(allColIdInSpark[q.getStoragePosition()]);
```
msirek: It's a big change to remove usage of baseColumnMap, though it looks like this is correct.  We might want to go through all types of external table paths and verify this is correct for all paths.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -729,7 +739,18 @@ private static Column createFilterCondition(Dataset dataset,String[] allColIdInS
...
             q.clearOrderableCache();
-            Object value = (probeValue==null || i!=0?q.getOrderable():probeValue).getObject();
+
+            DataValueDescriptor dvd = probeValue;
+            if (dvd == null || i != 0) {
+                dvd = q.getOrderable();
+            }
+
+            if (dvd instanceof  SQLChar) {
+                col = adaptSQLCharColumn(col, (SQLChar) dvd);
```
msirek: I set up parquet TPC-DS tables via instructions at https://splicemachine.atlassian.net/browse/DB-7852 but could not reproduce the problem reported with the following query:

select top 4 * from customer_demographics
where cd_education_status = 'Primary';

If there is an issue, then the fix looks fine.  But I could not reproduce it.  Are you doing something special in your test, like using the spark adapter?

ngpham: We can also test this with a smaller table, data must be populated from spark-sql:
```create table  ptba (col1 char(32))
  using parquet
  location '<some_path>/ptba';
insert into ptba values 
    ('abc'), 
    ('short string1');```
Then from splice sql, create the external table and query data:
```create external table  ptba (col1 char(32))
  stored as parquet
  location '<some_path>/ptba';
 select * from ptba where col1 = 'abc';```

Comment (msirek) [reply hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java]: OK, this answers my question.  If the external table data is populated using Spark SQL as opposed to splice SQL, the CHAR column data is not padded with spaces, otherwise it is.  I wonder if the padding should be applied in a more common location when the dataset is first created.  We could also have a case with a join predicate between an external table and an internal table.  What happens in that case?  Since the external table column is not padded, and the internal table column is, would the join return no rows?

File: hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java (hunk 1 above)
Comment (dgomezferro) [reply hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java]: Why is it correct? Is baseColumMap not needed here?

File: hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java [hunk 3]
```diff
@@ -729,7 +739,18 @@ private static Column createFilterCondition(Dataset dataset,String[] allColIdInS
...
+            }
+
+//            Object value = (probeValue==null || i!=0?q.getOrderable():probeValue).getObject();
```
dgomezferro: Please remove the commented out lines

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
ngpham: I believe that currently we do not have good support mixing of external, internal tables, or external tables DDL after creation. But we do want to improve that. I am thinking of the same direction, i.e. to validate and adapt the tables at creation time. It will be at a larger scope. Let's discuss more on Jira then?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
ngpham: As per my understanding, the qualifier q was already computed correctly, using index in the base dataset. If we use baseColumnMap, then we should do another reverse lookup to make it correct, which is redundant. In the existing code for 'and' qualifier, it is also the case that baseColumnMap is not used.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
ngpham: I made the change, and answered the question on baseColumMap in Mark's thread. Thanks!

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
msirek: OK, we can defer this to another Jira.  I was hoping you could do a quick test to see if it's an issue.  Then we could open another Jira for it.

dgomezferro: I agree with your @msirek, I think we should apply this transformation in SparkDataSetProcessor.processExternalDataset(), before calling createFilterCondition(), that way we don't have to do anything special with the filter and we'd handle joins etc correctly. There's too much stuff that can go wrong right now. 

Can you do that instead @ngpham ? After we get the dataset in processExternalDataset() we'd fix any SQLChar columns .

ngpham:
@dgomezferro , yes, let me implement that logic after processExternalDataset().

ngpham:
@dgomezferro @msirek @yxia92 Implemented a more robust solution, where all SQLChar columns are padded after Spark scan.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
```diff
@@ -799,13 +799,29 @@ public StringDataValue         getNullChar(StringDataValue previous,
...
-         }
+        }
+
+        public StringDataValue getNullChar(StringDataValue previous, int collationType, int maxSize)
+        throws StandardException
+        {
+            SQLChar ret = null;
+
+            if (collationType == StringDataValue.COLLATION_TYPE_UCS_BASIC) {
+                ret = (SQLChar) getCharDataValue((String) null);
```
yxia92: Why this path does not take into consideration the passed in argument _previous_ if it is not null? (I'm comparing it with the existing function getNullChar(StringDataValue, int).

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
```diff
@@ -927,6 +931,8 @@ private boolean evaluateNull()
     */
     public void writeExternal(ObjectOutput out) throws IOException {
         out.writeBoolean(isNull);
+        out.writeInt(sqlCharSize);
```
yxia92: My main concern of this fix is the Serde change for a basic data type. This may break the backward compatibility. For example, column stats info in system table may not be read correctly. Not sure if other data dictionary tables are affected.
If it is just column stats table that is affected, and the fix is going to a major release like 3.0, maybe it is OK, assuming a re-collection of stats is acceptable for a major release upgrade.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
ngpham: I made the logic a bit cleaner, the behavior is the same. We return a NULL value, always. If `previous` is defined, then the `previous` is also set to NULL.

yxia92: My understanding is that if _previous_ is not null, we want to reuse it. But with this function, we always create a new instance. Am I wrong?

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
ngpham: Yes, it might be the case... As per my understanding: We can remove the Ser/De part, to retain backward compatibility. For this fix, as long as OLAP reinitializes the `Activation`, which is the current design, the SQLChar size information will be repopulated.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataValueFactoryImpl.java
ngpham: Yes, you are right. I fixed the logic of using previous value.

ngpham:
> The code looks good. Why set sqlCharSize for varchar if we don't use it?
Since `char` and `varchar` are defined with size, we may need it in some other context, preallocation, for example. It is true that we do not need `varchar` size for now, but I would prefer to keep it.

msirek:
> > The code looks good. Why set sqlCharSize for varchar if we don't use it?
> > Since `char` and `varchar` are defined with size, we may need it in some other context, preallocation, for example. It is true that we do not need `varchar` size for now, but I would prefer to keep it.

OK

PR #2782, opened 2019-07-12
DB-7062 Add log4j layout to mask with regex

wb14123:


splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -0,0 +1,67 @@
...
+        if (matcher.find()) {
+            String maskedMessage = "";
+            for(int i = 0; i <= matcher.groupCount(); i++) {
+                int start;
+                int end;
+                if (i == 0) {
+                    start = 0;
+                } else {
+                    start = matcher.end(i);
+                    maskedMessage += MASK;
```
changli6: Please use StringBuilder for better performance.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -0,0 +1,67 @@
...
+                if (i == matcher.groupCount()) {
+                    end = message.length()-1;
+                } else {
+                    end = matcher.start(i+1) - 1;
+                }
+                if (start > end || end >= message.length()) {
+                    continue;
+                }
+                for (int j = start; j <= end; j++) {
+                    maskedMessage += message.charAt(j);
```
changli6: Please use StringBuilder for better performance.

jaceklaskowski:
I'm curious why there's no sample usage of the layout (even as a test)?

carolp-503:
From our discussion:  Did you test this for a badly formed regular expression?

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
msirek: Also, it's probably quicker to append a substring instead of a character at a time.
String has the replaceAll() method.  Maybe an interface such as this could be used.  It's likely already optimized.

wb14123: @changli6 Fixed with `StringBuilder` since this will not have race condition. @msirek replaceAll may replace the wrong string, for example, a regex `abc ([0-9]+)` only matches the second numbers in `123 abc 123`, `replaceAll` may replace the first one.

wb14123:
@jaceklaskowski Added some tests.

@carolp-503 Added the logic to handle regex compile exception.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -0,0 +1,85 @@
...
+                    start = matcher.end(i);
+                    maskedMessage.append(maskString);
+                }
+                if (i == matcher.groupCount()) {
+                    end = message.length() - 1;
+                } else {
+                    end = matcher.start(i + 1) - 1;
+                }
+                if (start > end || end >= message.length()) {
+                    continue;
```
changli6: break?

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public String getMaskString() {
+        return this.maskString;
+    }
+
+    public static String maskMessage(String message, Pattern maskPattern, String maskString) {
+        Matcher matcher = maskPattern.matcher(message);
+
+        if (matcher.find()) {
```
changli6: To handle the message having multiple substrings matched, we need to change to "while" here.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -0,0 +1,85 @@
...
+                if (i == matcher.groupCount()) {
+                    end = message.length() - 1;
+                } else {
+                    end = matcher.start(i + 1) - 1;
+                }
+                if (start > end || end >= message.length()) {
+                    continue;
+                }
+                for (int j = start; j <= end; j++) {
+                    maskedMessage.append(message.charAt(j));
```
msirek: Can you use substring instead of appending a character at a time?  Looping like this might be expensive.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
wb14123: This is actually not needed. Removed it

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
wb14123: @changli6 Changed the code to make it have the ability to match multiple times

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
wb14123: @msirek Changed to substring

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -1,23 +1,71 @@
...
     public String getMaskPattern() {
         return maskPattern.toString();
     }
 
     public void setMaskPattern(String pattern) {
-        maskPattern = Pattern.compile(pattern);
+        try {
+            maskPattern = Pattern.compile(pattern);
+        } catch (Exception e) {
+            logger.error("Error to compile regex for mask pattern, will not mask anything", e);
```
jaceklaskowski: Why not to include the incorrect `pattern`?

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
```diff
@@ -1,23 +1,71 @@
...
+
+    public void setMaskString(String maskString) {
+        this.maskString = maskString;
+    }
+
+    public String getMaskString() {
+        return this.maskString;
+    }
+
+    public static String maskMessage(String message, Pattern maskPattern, String maskString) {
```
jaceklaskowski: Why is this method `static`?

splice_encoding/src/test/java/com/splicemachine/utils/MaskPatternLayoutTest.java
```diff
@@ -0,0 +1,63 @@
+package com.splicemachine.utils;
+
+import com.splicemachine.utils.logging.MaskPatternLayout;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.regex.Pattern;
+
+public class MaskPatternLayoutTest {
```
jaceklaskowski: Please include tests for incorrect patterns

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
wb14123: This method has no need to be non-static and this also makes it easy to test.

splice_encoding/src/test/java/com/splicemachine/utils/MaskPatternLayoutTest.java
wb14123: Incorrect pattern is handled outside the method `maskMessage`, so the only way to test it is to pass through log4j configuration, which is incompatible with other tests thus makes this ticket very complex. I've already manually tested it.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
wb14123: Updated it.

splice_encoding/src/main/java/com/splicemachine/utils/logging/MaskPatternLayout.java
jaceklaskowski: Two out of three arguments are instance variables so I think that's what makes the method an instance method.

PR #2785, opened 2019-07-14
DB-8104 Fix date formats which have no separator characters (2.7).

msirek:
SQL such as 'VALUES TO_DATE('19000101','yyyyMMdd');' errors out when it should not.
Also, use of date format strings without separators in import statements similarly errors out when it should not.

See [DB-8104](https://splicemachine.atlassian.net/browse/DB-8104?oldIssueView=true)


msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
```diff
@@ -14,29 +14,88 @@
...
-        this.format = format != null ? format : "yyyy-M-d";
-        
-        if (format != null) {
+
+        // The default external format of "yyyy-MM-dd" may be used
+        // in error messages, but the corresponding internally used format
+        // is "yyyy-M-d".
+        if (format == null) {
+            format = "yyyy-M-d";
+            this.format = "yyyy-MM-dd";
```
jaceklaskowski: These two formats should really be constants. That might also get rid of the comment (that can easily get outdated).

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
```diff
@@ -14,29 +14,88 @@
...
+        else {
+            this.format = format;
+
+            // The valid format characters of DateTimeFormatter documented at the following link
+            // can be used to identify separator characters like '-' or '/' :
+            // See https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
+
+            // Any non-format character can be a separator character.
+            // This pattern matches one or more non-format characters.
+            String
```
jaceklaskowski: Why is the above comment not part of javadoc of the class?

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
```diff
@@ -14,29 +14,88 @@
...
 
+import com.splicemachine.db.iapi.reference.SQLState;
+import java.sql.SQLException;
+import static java.lang.String.format;
+
 public class SpliceDateFormatter {
 
     private String format = null;
-    private java.time.format.DateTimeFormatter formatter;
+    private java.time.format.DateTimeFormatter formatter = null;
```
jaceklaskowski: Could `DateTimeFormatter` be imported to avoid using fully-qualified name here and down the class?

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
```diff
@@ -14,29 +14,88 @@
...
+                                                    excludeFormatCharsPattern);
+            format = format.replaceFirst(matchPattern, "$1$3$4");
+        }
+        try {
+            this.formatter =
+                java.time.format.DateTimeFormatter.ofPattern(format);
+        }
+        catch (IllegalArgumentException e) {
+            // If the format is bad, this.formatter will be null, and the next time
+            // getFormatter() is called we'll throw a SQLException.
```
jaceklaskowski: We'd rather print out a WARN message to the logs.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
```diff
@@ -14,29 +14,88 @@
...
         }
-        this.formatter =
-              java.time.format.DateTimeFormatter.ofPattern(this.format);
     }
+
+    // Return the original date format used in the SQL.
+    // This may differ from the internal format used by DateTimeFormatter.
```
jaceklaskowski: Why is this not a javadoc?

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
msirek: yes, thanks.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
msirek: The class field can be final, but the other one needs to be variable because it's modified.  The comment is useful to explain there is a difference between the splice date format used in SQL and the date format we need to pass to DateTimeFormatter to get the expected behavior.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
msirek: This is an implementation detail, which should be an internal comment.  I'll add some javadoc comments referencing splice date formats.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
msirek: Thanks.. it seems the splice way of relaying syntax errors like this in the user's SQL is to throw a StandardException or a SQLException.  The constructor can't throw the exception, so the first time getformatter is called the exception will be thrown, and a warning such as the following will appear in the logs:

`2019-07-14 20:54:13,369 (DRDAConnThread_567) WARN  [c.s.p.Exceptions] - Parsing exception, original stacktrace:
java.lang.RuntimeException: ERROR 38000: The exception 'java.sql.SQLException: Error parsing datetime with pattern: yyyy-MM-dddd. Try using an ISO8601 pattern such as, yyyy-MM-dd'T'HH:mm:ss.SSSZZ, yyyy-MM-dd'T'HH:mm:ssZ or yyyy-MM-dd' was thrown while evaluating an expression.
`

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceDateFormatter.java
msirek: thanks, changed to javadoc

msirek:

spliceci:
TEST SUCCEEDED +1

PR #2792, opened 2019-07-23
DB-8392 Add new property splice.minimumSplitsPerTable

changli6:


changli6:

spliceci:
TEST SUCCEEDED +1

splice_access_api/src/main/java/com/splicemachine/access/configuration/StorageConfiguration.java
```diff
@@ -36,10 +36,13 @@
     public static final String SPLIT_BLOCK_SIZE = "splice.splitBlockSize";
     public static final int DEFAULT_SPLIT_BLOCK_SIZE=32*1024*1024;
 
+    public static final String SPLITS_PER_TABLE_MINIMUM = "splice.splitsPerTableMinimum";
```
jaceklaskowski: What do you think about a change to have the names like `MINIMUM_SPLITS_PER_TABLE` and `splice.minimumSplitsPerTable` (minimum is a prefix not a suffix).

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkScanSetBuilder.java
```diff
@@ -114,7 +114,9 @@ else if (storedAs.equals("O"))
         }
         if (op != null) {
             ScanOperation sop = (ScanOperation) op;
-            conf.setInt(MRConstants.SPLICE_SPLITS_PER_TABLE, sop.getSplits());
+            int splitsPerTableMin = HConfiguration.getConfiguration().getSplitsPerTableMin();
+            int requestedSplits = sop.getSplits();
```
jaceklaskowski: Could `requestedSplits` be negative?

splice_access_api/src/main/java/com/splicemachine/access/configuration/SConfigurationImpl.java
```diff
@@ -823,6 +824,10 @@ public long getRegionMaxFileSize() {
     public long getTableSplitSleepInterval() {
         return tableSplitSleepInterval;
     }
+    @Override
```
jaceklaskowski: nit: a new line between methods

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkScanSetBuilder.java
```diff
@@ -114,7 +114,9 @@ else if (storedAs.equals("O"))
...
+            conf.setInt(MRConstants.SPLICE_SPLITS_PER_TABLE, requestedSplits != 0 ? requestedSplits : splitsPerTableMin);
```
msirek: Please check that requestedSplits and splitsPerTableMin are positive integers before using them.

changli6: Thanks @msirek .
Addressed checking the "splice.minimumSplitsPerTable" value be >= 0. It could be 0 if we don't want to set the lower bound.
The "requestedSplit" is checked in FromBaseTable.verifyProperties to make sure it is  > 0. We won't check it again here.

msirek: @changli6 I didn't mean throw an exception if splitsPerTableMin is negative.  Then the user could cause queries not to run by setting a bad value.  I'm not sure we want to allow them this type of control to cause all spark queries to error out.   I just meant if splitsPerTableMin is negative it should be ignored, for example:

`conf.setInt(MRConstants.SPLICE_SPLITS_PER_TABLE, (requestedSplits != 0) ? requestedSplits : (splitsPerTableMin > 0) ? splitsPerTableMin : 0);`

changli6: Addressed. Thanks!

PR #2793, opened 2019-07-26
DB-8166 Quit join execution immediately when one side is empty.

msirek:
Performance improvement for joins (Broadcast, Merge, SortMerge, NestedLoop and Cross) to quit evaluation of the join if one of the "join tables" has no rows.  The left and right side of a join could really be a sequence of joins, ProjectRestricts, etc.  Such a sequence of operations that yields no rows would likewise allow us to skip the join.

The rules are:
- If the join is a semijoin or inner join, and either side has no rows, return an empty DataSet or row iterator instead of performing the join.
- If the join is an outer join and the outer table has no rows , return an empty DataSet or row iterator, instead of performing the join.

Also added the following:
- A fix to inner merge join to quit scanning the left when the right has no more rows.
- A fix to avoid using cross join for EXISTS subqueries.
- The addition of session property "explainMode", which when set to true in the JDBC connection string or in a "set session_property" SQL statement, causes all queries to be EXPLAINed instead of executed.
Example of setting explainMode in a connection string:
     `connect 'jdbc:splice://localhost:1527/splicedb;user=splice;password=admin;explainMode=true';`
Example of setting explainMode via SQL:
     `set session_property explainMode=true;`

See [DB-8166](https://splicemachine.atlassian.net/browse/DB-8166)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/db/Database.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/db/Database.java
@@ -92,6 +92,7 @@ public interface Database extends com.splicemachine.db.database.Database, Locale
 													 CompilerContext.DataSetProcessorType dataSetProcessorType,
 													 boolean skipStats,
 													 double defaultSelectivityFactor,
+													 boolean explainMode,
 													 String ipAddress,
 													 String defaultSchema,
 													 Properties sessionProperties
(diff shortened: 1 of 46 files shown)


msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
I found out some performance implications of using RDD.isEmpty() in this fix, so I'm going to rework the code and open a new PR later.

PR #2796, opened 2019-07-29
SPLICE-2335 Import timestamp data can't handle daylight savings time gaps.

msirek:
Importing timestamp data into Splice tables fails if the timestamp falls in the gap when we spring forward an hour due to daylight savings time.  
It is fixed by replacing use of Joda DateTimeFormat internally with the standard Java DateTimeFormatter when parsing row data in .csv files.  Previous fixes added use of DateTimeFormatter for timestamps without time zone offsets.  But this mixture of using two formatters may be problematic going forward.
Joda DateTimeFormat was also problematic for Splice anyway because Splice supports microseconds precision in timestamps, but Joda DateTime only supports milliseconds.  

The system functions TO_DATE, TO_TIMESTAMP and TO_TIME will all now use the new DateTimeFormatter.

See [SPLICE-2335](https://splice.atlassian.net/browse/SPLICE-2335)


msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
[DB-8204](https://splicemachine.atlassian.net/browse/DB-8204)

splice_machine/src/test/java/com/splicemachine/derby/impl/load/HdfsImportIT.java
```diff
@@ -905,6 +905,44 @@ public void testNoSeparatorDateTime() throws Exception {
...
+
+        ResultSet rs = methodWatcher.executeQuery(format("select * from %s.%s", spliceSchemaWatcher.schemaName,
+                TABLE_9));
+        List<String> results = Lists.newArrayList();
+        while (rs.next()) {
+            Timestamp order_date = rs.getTimestamp(1);
+            assertNotNull("order_date incorrect", order_date);
+            // The system may or may not be in a time zone that observes daylight savings.
+            Assert.assertTrue(String.format("Unexpected timestamp value: %s", order_date.toString()),
+                                order_date.toString().equals("2012-03-11 03:05:12.123456") ||
```
carolp-503: While we're here, should we also verify "fall back" dates?  Or are we doing that somewhere else?

msirek: Those were handled by SPLICE-2264.  There's a test for it here:  https://github.com/splicemachine/spliceengine/pull/2404/files#diff-83845c0e11fdf371b3bcfd3456fa9d33

PR #2797, opened 2019-07-29
DB-8204 Avoid native spark SortMergeJoin involving the Real data type.

msirek:
Wrong results can occur when executing joining on columns of the Real data type and the join is done natively on spark.  Execute the join via Splice functions in this case.

[DB-8204](https://splicemachine.atlassian.net/browse/DB-8204)

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
@@ -287,7 +287,7 @@ public class BroadcastJoinIT extends SpliceUnitTest {
     }
     
     @Test
-    public void testNumericColumnsBroadCastJoin() throws Exception {
+    public void testNumericJoinColumns() throws Exception {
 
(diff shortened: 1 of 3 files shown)


msirek:
jenkins please test branch @cdh5.16.1

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/joins/BroadcastJoinIT.java
```diff
@@ -287,94 +289,97 @@ public static void createDataSet() throws Exception {
     @Test
     public void testNumericColumnsBroadCastJoin() throws Exception {
 
-        String expected = "1 |\n" +
+        String [] joinTypes = {"NestedLoop", "SortMerge", "Broadcast", "Cross" };
```
yxia92: It may be better to remove the wording "Broadcast" from the test case name as now we test all 4 join strategies.

msirek: Updated the method name to testNumericJoinColumns.

PR #2798, opened 2019-07-29
DB-8193 Comments in CTAS mask the rest of the SELECT statement.

msirek:
A comment in CREATE TABLE ... AS, such as the following:
```
create table mws.t2 as
select * from mws.t1 --comment
where a = 1;
```
... masks out the rest of the SELECT.  In this case, it would mask out the WHERE clause, leading to an incorrect table definition.  The fix is in sqlgrammar.jj, to use a newline separator instead of single space.

[DB-8193](https://splicemachine.atlassian.net/browse/DB-8193)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -2353,9 +2353,25 @@ public class SQLParser
      * CREATE TABLE X AS [SELECT A, B, C FROM Y] WITH DATA
      */
    public String parseQueryString(String src) {
-       boolean inSection = false;
-       StringBuilder buf = new StringBuilder();
+        boolean inSection = false;
+        StringBuilder buf = new StringBuilder();
+
+        // Use a non-printable character as an end-of-hint marker.
+        char endOfHintMarker = 0x01;
+
+        // Strip out comments starting with --, except --splice-properties hints.
+        src = src.replaceFirst("(--)(?i)(?!splice-properties)(.*?)(\\n|\\r|\\r\\n)", " ");
+
+        // Add an end-of-hint marker to a --splice-properties hint, so we know where to
+        // place the newline later on.
+        src = src.replaceFirst("(--)(?i)(?=splice-properties)(.*?)(\\n|\\r|\\r\\n)",
+                               "$1$2$3"+Character.toString(endOfHintMarker));
+
+        // Strip out C-style comments:   /* This type of comment */
+        src = src.replaceFirst("(/\\*.*?\\*/)(\\n|\\r|\\r\\n)", " ");
         for (String part : src.split("\\s+")) {
+            if (part.charAt(0) == endOfHintMarker)
+                part = "\n" + part.substring(1);
             if (inSection && part.equalsIgnoreCase("with")) {
                 inSection = false;
             } else if (inSection) {
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -2359,16 +2359,17 @@ public class SQLParser
         // Use a non-printable character as an end-of-hint marker.
         char endOfHintMarker = 0x01;
 
+        // Strip out C-style comments:   /* This type of comment */
+        src = src.replaceAll("(/\\*.*?\\*/)", " ");
+
         // Strip out comments starting with --, except --splice-properties hints.
-        src = src.replaceFirst("(--)(?i)(?!splice-properties)(.*?)(\\n|\\r|\\r\\n)", " ");
+        src = src.replaceAll("(--)(?i)(?!splice-properties)(.*?)(\\n|\\r|\\r\\n)", " ");
 
         // Add an end-of-hint marker to a --splice-properties hint, so we know where to
         // place the newline later on.
-        src = src.replaceFirst("(--)(?i)(?=splice-properties)(.*?)(\\n|\\r|\\r\\n)",
+        src = src.replaceAll("(--)(?i)(?=splice-properties)(.*?)(\\n|\\r|\\r\\n)",
                                "$1$2$3"+Character.toString(endOfHintMarker));
 
-        // Strip out C-style comments:   /* This type of comment */
-        src = src.replaceFirst("(/\\*.*?\\*/)(\\n|\\r|\\r\\n)", " ");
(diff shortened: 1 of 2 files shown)


msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2359,7 +2359,7 @@ public class SQLParser
             if (inSection && part.equalsIgnoreCase("with")) {
                 inSection = false;
             } else if (inSection) {
-                buf.append(part).append(" ");
+                buf.append(part).append("\n");
```
yxia92: If we have multiple words in the comment, this fix may not work. For example:
```
create table t2 select a1 from t1 --comment where b1>3
where b1 < 3 with data;
```
We could be putting "where b1>3" in the comment in a separate line, and this would cause syntax error.

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
msirek: Thanks Yi... I think the pattern matching for the WITH clause needs to be fixed here too because it could be matching the word "with" inside a comment string.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/CreateTableWithDataIT.java
```diff
@@ -279,7 +279,7 @@ public void createTableWithViewJoins() throws Exception {
         String depsalTable = "depsal";
         String depsalTableRef = spliceSchemaWatcher.schemaName + "." + depsalTable;
         String depsalTableDef = format("create table %s as " +
-                "select dept, salary, ssn from %s with data", depsalTableRef, viewRef);
+                "select dept, salary, ssn from %s -- comment \n where dept < 3 with data", depsalTableRef, viewRef);
```
carolp-503: Can you expand this case or add a second case where the embedded comment includes the word "with"?

msirek: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
msirek: @yxia92 I reworked the fixed to explicitly match C-style comments and  '--' comment strings, except for --splice-properties hints, using regular expressions.  Please take a look.

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2353,9 +2353,25 @@ public class SQLParser
...
-       boolean inSection = false;
-       StringBuilder buf = new StringBuilder();
+        boolean inSection = false;
+        StringBuilder buf = new StringBuilder();
+
+        // Use a non-printable character as an end-of-hint marker.
+        char endOfHintMarker = 0x01;
+
+        // Strip out comments starting with --, except --splice-properties hints.
+        src = src.replaceFirst("(--)(?i)(?!splice-properties)(.*?)(\\n|\\r|\\r\\n)", " ");
```
yxia92: The fix may not work for some complex scenarios:
1) If "--" is inside the C-style comment, for example
create table v1 as 
select a1, b1 from t1 /* c-style comment starts with an embedded -- comment */
where a1 > 3 with data;
We could mistakenly remove the closing C-style comment. Maybe we should switch the order of processing for C-style comment and -- comment.
2) there could be more than one splice-properties hint, and replaceFirst is not sufficient.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -2353,9 +2353,25 @@ public class SQLParser
...
+
+        // Add an end-of-hint marker to a --splice-properties hint, so we know where to
+        // place the newline later on.
+        src = src.replaceFirst("(--)(?i)(?=splice-properties)(.*?)(\\n|\\r|\\r\\n)",
+                               "$1$2$3"+Character.toString(endOfHintMarker));
+
+        // Strip out C-style comments:   /* This type of comment */
+        src = src.replaceFirst("(/\\*.*?\\*/)(\\n|\\r|\\r\\n)", " ");
```
yxia92: I think C-style comment does not need to end with a new line, so we don't need to put (\\n|\\r|\\r\\n) as part of the pattern.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
msirek: Thanks @yxia92!  I'm updating the fix.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
msirek: I think you're right... I found an issue with nested C-style comments after removing the check for newlines and carriage returns, so I may need a new regex to handle that.

msirek: @yxia92 Java doesn't support recursive regular expressions, so I didn't find an easy way to handle nested C-style comments.  This is a low priority item, so I've opened [DB-8575](https://splicemachine.atlassian.net/browse/DB-8575) to handle this case.

msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
yxia92: Switching the order of processing the C-style comment and dash dash comment does not work either. For example:
```
create table v1 as
select a1, b1 from t1 /* c-style comment 
, t2 -- dash dash comment with an embedded */
*/
where a1 > 3 with data;
```
However, I don't see an easy way to cover this case except for writing a parser using maybe javacc. Since it is a rare case, I'm OK with proceeding the fix and add this test case to DB-8575 also.

spliceci:
TEST SUCCEEDED +1

PR #2803, opened 2019-08-01
DB-8559 do not include PK column in the setcolumnlist of update state…(2.7)

yxia92:
…ment for merge_data_from_file

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/load/UpportIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/load/UpportIT.java
@@ -506,8 +506,8 @@ public class UpportIT extends SpliceUnitTest {
 
         ResultSet resultSet = statement.executeQuery();
         //make sure that the bad records list is good
-        //all 5 rows are updated, no extra row inserted
-        validateMergeResults(resultSet,5,0, 0);
+        //no row should be updated, as we only use the pk column for the merge_data_from_file call
+        validateMergeResults(resultSet,0,0, 0);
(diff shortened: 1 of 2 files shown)


yxia92:

splice_machine/src/test/java/com/splicemachine/derby/impl/load/UpportIT.java
```diff
@@ -506,8 +506,8 @@ public void testMergeIntoOccupiedPartiallyTableWillUpdate() throws Exception {
 
         ResultSet resultSet = statement.executeQuery();
         //make sure that the bad records list is good
-        //all 5 rows are updated, no extra row inserted
-        validateMergeResults(resultSet,5,0, 0);
+        //no row should be updated, as we only use the pk column for the merge_data_from_file call
+        validateMergeResults(resultSet,0,0, 0);
```
jyuanca: do you mean non-PK columns?

yxia92: I meant PK columns. For this test case, all the columns(essentially 'a') passed to merge_data_from_file are PK columns, so after removing the pk columns from the set column list, nothing is left, and no update statement needs to be triggered, and we should see no row updated. That is what I meant. Thanks.

jyuanca: got it thanks!

spliceci:
TEST SUCCEEDED +1

PR #2806, opened 2019-08-06
DB-DB-8536: No need to close internal connections during shutdown

jyuanca:
Closing an internal connection needs to get a timestamp to start a txn. However, timestamp server may have shutdown. This will cause retry and prevent region server to shutdown within graceful shutdown timeout


jyuanca:

spliceci:
TEST SUCCEEDED +1

PR #2831, opened 2019-08-22
DB-8524 MERGE_DATA_FROM_FILE handle directories as well as files (2.7)

changli6:


changli6:
All comments were addressed.

PR #2834, opened 2019-08-22
DB-8535: copy yarn-site.xml from /etc/hadoop/conf

jyuanca:


jyuanca:
If one cloudera manager manages more than 1 cluster, the service name are appended by a number, and yarn-site.xml is stored at /etc/hadoop/conf.cloudera.yarn2, /etc/hadoop/conf.cloudera.yarn3, etc. The script copies yarn-site.xml from /etc/hadoop/conf.cloudera.yarn which may not exist. /etc/hadoop/conf is a link to /etc/hadoop/conf.cloudera.yarnXX which should always work.

PR #2837, opened 2019-08-24
DB-8532 Register nested class java.util.Arrays$ArrayList in Kryo.

msirek:
Nested loop join executed in a connection using impersonation throws an exception because the group users list is not registered in Kryo.


msirek:

spliceci:
TEST SUCCEEDED +1

PR #2842, opened 2019-08-27
DB-8570: invalidate dictionary cache after updating system procedures

jyuanca:


yxia92:
@jyuanca, on master, I see that upgradeSystemProcedures() is called only when a upgrade script is triggered (it is part of UpgradeScriptBase.run()). For version upgrade that does not trigger any upgrade script, createOrUpdateAllSystemProcedures() won't be executed.

PR #2875, opened 2019-09-04
DB-8677 Comment out ZK maven debugging option to avoid IT hangs.

msirek:
Zookeeper fails to start during some IT runs, causing failures.
The following was seen in the logs: 
> Listening for transport dt_socket at address: 4002

Sometimes this situation causes the process to suspend until the debugger is attached, even though "suspend=n" is specified.


OlegMazurov:
It's a good idea to exclude the debug flag during IT from all components. I'd even say that the debug flag should be enabled explicitly in standalone environment (maybe with -Pdebug) and be disabled by default.
However, I doubt that flag is responsible for IT timeouts.

PR #2882, opened 2019-09-09
DB-8696 trigger sysschemasview update only on master

yxia92:


yxia92:

spliceci:
TEST FAILED

yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1367,6 +1367,14 @@ public void setMetadataAccessRestrictionEnabled() {
 
     @Override
     public void updateSystemSchemasView(TransactionController tc) throws StandardException {
+        boolean toUpgrade = Boolean.TRUE.equals(EngineLifecycleService.toUpgrade.get());
+        // Only master can upgrade
```
arnaud-splice: I feel like the comment is a bit misleading, there's nothing from here which indicates that only master can upgrade. Only the ones with `toUpgrade` set can upgrade, and it just "so" happens that we set `toUpgrade` with `isMaster`. Now keep in mind that I don't have the whole picture just yet. :)

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1398,6 +1406,7 @@ public void updateSystemSchemasView(TransactionController tc) throws StandardExc
         // we need to re-generate the metadataSPS due to the definition change of sysschemasview
         updateMetadataSPSes(tc);
 
+        tc.commit();
         SpliceLogUtils.info(LOG, "SYSVW.SYSSCHEMAVIEW updated to " + metadataRestrictionEnabled);
```
jyuanca: Is this commit absolutely necessary?

yxia92: Not absolutely necessary. I put commit at the begin and end of this function so that the update of the schema view starts from a clean transaction and commit, all self-contained in this function. I'm not aware of side effect of redundant commit, if there is, we can remove.

PR #2886, opened 2019-09-17
DB-8714 Factor out common SpliceDefaultCompactor functions (2.8)

arnaud-splice:
With this commit, we only factor out the functions of
SpliceDefaultCompactor which are absolutely identical accross all
hbase_sql/hbase<...> versions.

A later commit may be necessary to factor out the meaty part as well.

### DIFF
--- a/hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
+++ b/hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
@@ -16,83 +16,45 @@ package com.splicemachine.compactions;
 
 import com.splicemachine.EngineDriver;
 import com.splicemachine.access.HConfiguration;
-import com.splicemachine.access.api.PartitionFactory;
 import com.splicemachine.access.api.SConfiguration;
-import com.splicemachine.constants.EnvUtils;
-import com.splicemachine.derby.stream.compaction.SparkCompactionFunction;
 import com.splicemachine.hbase.SICompactionScanner;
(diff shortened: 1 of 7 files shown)


arnaud-splice:

hbase_sql/hbase1.1.1/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -16,83 +16,45 @@
...
-
-        if (LOG.isDebugEnabled()) {
-            SpliceLogUtils.debug(LOG, "Initializing compactor: region=%s", ((HStore)this.store).getHRegion());
-        }
     }
 
     public SpliceDefaultCompactor(final Configuration conf, final Store store, long smallestReadPoint) {
-        this(conf, store);
-        this.smallestReadPoint = smallestReadPoint;
+        super(conf, store);
```
carolp-503: Resolve @yxia92 's comment on Master here before merging.

arnaud-splice: Done.

arnaud-splice:
> * All comments are resolved.

Done
> * @dgomezferro has reviewed master, and any corresponding changes are made both in Master and here.

I will wait for his review 
> * Pending changes for 3.0 branch can be merged separately, but we should make the same changes before moving onto the next phase of workload isolation/swimlane for compaction

No changes seem to be required in 3.0 to mirror that PR as the folder structure is different there and no version specific code exists.

PR #2891, opened 2019-09-17
DB-8718  BigDecimalEncoding.toBigDecimal() is expensive

OlegMazurov:
Eliminated extra copying. Took advantage of _new BigDecimal(long)_.

### DIFF
--- a/splice_encoding/src/main/java/com/splicemachine/encoding/BigDecimalEncoding.java
+++ b/splice_encoding/src/main/java/com/splicemachine/encoding/BigDecimalEncoding.java
@@ -185,71 +185,51 @@ class BigDecimalEncoding {
     }
 
     public static BigDecimal toBigDecimal(byte[] data, int dataOffset, int dataLength, boolean desc) {
-        byte[] dataCopy = Arrays.copyOfRange(data, dataOffset, dataOffset + dataLength);
-        if (desc) {
-            for (int i = 0; i < dataCopy.length; i++) {
-                dataCopy[i] ^= ORDER_FLIP_MASK;
-            }
-        }
-        return toBigDecimal(dataCopy, 0, dataLength - 1);
-    }
-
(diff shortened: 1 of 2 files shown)


splice_encoding/src/main/java/com/splicemachine/encoding/BigDecimalEncoding.java
```diff
@@ -185,71 +185,51 @@ public static BigDecimal toBigDecimal(byte[] data, boolean desc) {
...
-                } else {
-                    data[z] ^= ORDER_FLIP_MASK;
-                }
-            }
-        }
+        boolean negative = (h == HEADER_NEG);
+        mask = desc ^ negative ? ORDER_FLIP_MASK : 0;
 
-        long[] expOffset = ScalarEncoding.readLong(data,dataOffset,HEADER_SIZE_BITS);
+        long[] expOffset = ScalarEncoding.readLong(data, dataOffset, desc ^ negative, HEADER_SIZE_BITS);
```
yxia92: Is this logic consistent with the original? If desc=false, and negative=true, the first two bytes will remain unchanged according to the original logic:
> if (z == 0) {	
>                     data[z] ^= ORDER_FLIP_EXCLUDE_HEADER_MASK;

But with the new logic, it will be flipped. Am I right?

OlegMazurov: Yes, you are. Those two bits (not bytes) have already been processed, though, and are not looked at any more so there is no harm in changing them. The code, however, becomes simpler.

yxia92: Thanks @OlegMazurov , you are right that these two bits are not looked at anymore.

OlegMazurov:
There is no functional change here, only performance. Existing uint tests are BigDecimalEncoding_FixedTest, BigDecimalEncoding_RandomizedTest, BigDecimalEncoding_SortTest.

PR #2893, opened 2019-09-20
SPLICE-2362 Native Spark join with inequality conditions.

msirek:
Performance feature to execute joins with both equality and inequality join predicates as native spark operations, e.g.  select * from t1, t2 where t1.c1 = t2.c1 and t1.c2 <> t2.c2.  

Only broadcast join and sortmerge join are affected, as these are the only native spark joins with join predicates Splice supports.  Cross join is native spark, but in that case the join conditions are applied as a filter on the resulting DataFrame after the join is executed, so it cannot apply this optimization.  

This improves the performance of TPC-H query 21.   
See [SPLICE-2362](https://splice.atlassian.net/browse/SPLICE-2362?focusedCommentId=27913&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27913)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/JoinOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/JoinOperation.java
@@ -31,6 +31,8 @@ import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang3.SerializationUtils;
 import org.apache.log4j.Logger;
+import org.apache.spark.sql.Column;
+import org.apache.spark.sql.catalyst.parser.ParserInterface;
 import org.spark_project.guava.base.Strings;
 
 import java.io.IOException;
@@ -110,7 +112,8 @@ public abstract class JoinOperation extends SpliceBaseOperation {
         public boolean wasRightOuterJoin = false;
         public boolean isOuterJoin = false;
 		private Restriction mergeRestriction;
-		protected SparkExpressionNode sparkJoinPredicate;
+		protected SparkExpressionNode sparkJoinPredicate = null;
+		protected String sparkExpressionTreeAsString = null;
 
     public JoinOperation() {
 				super();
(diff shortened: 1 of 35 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
jenkins please test branch @cdh5.16.1

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/JoinOperation.java
```diff
@@ -208,10 +207,17 @@ public void writeExternal(ObjectOutput out) throws IOException {
...
                         out.writeObject(mergedRowTemplate);
 				}
-				boolean sparkExpressionPresent = (sparkJoinPredicate != null);
+				boolean sparkExpressionPresent = (sparkExpressionTreeAsString != null &&
+				                                  sparkExpressionTreeAsString.length() != 0);
+
 				out.writeBoolean(sparkExpressionPresent);
-				if (sparkExpressionPresent)
-					out.writeObject(sparkJoinPredicate);
+				if (sparkExpressionPresent) {
```
dgomezferro: Would it be possible to delay the decoding further to when it's actually used? For NLJoins we serialize the operation in order to clone it, so we'd be paying the overhead anyway.

msirek: @dgomezferro Better yet, we don't even need to build the SparkExpressionNode tree unless the join is broadcast join or mergesort join.  I pushed a change to this effect.  Joins on control other than nested loop join are not serialized I believe, so there should be no need to delay deserialization any further with the join strategy check in place.

msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/JoinOperation.java
dgomezferro: There might still be some performance impact because when we have a NLJ we serialize/deserialize the whole operation tree, so a broadcast join on the same tree would get serialized. I think it's ok for now and we can revisit later if NLJ becomes expensive

msirek: I see.  I pushed a change to delay deserialization of the predicates tree until it's used.  Up until then if we need to serialize/deserialize, we'll do so on the string representation of the tree.

PR #2899, opened 2019-09-25
DB-8577 Isolate compaction to its own spark queue

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

changli6:
Looks good @arnaud-splice .
Questions:
1. If we set splice.olap_server.isolated.compaction = true, but the "compaction" queue doesn't exist in YARN, what is the error?
2. Could user set the compaction queue name instead of "compaction" ?

jyuanca:
It would be more flexible if compaction queue name is configurable

arnaud-splice:
> Looks good @arnaud-splice .
> Questions:
> 
> 1. If we set splice.olap_server.isolated.compaction = true, but the "compaction" queue doesn't exist in YARN, what is the error?
> 2. Could user set the compaction queue name instead of "compaction" ?

1. If we set splice.olap_server.isolated.compaction = true, a compaction queue should exist in YARN. If that is not the case for some reason, the error message will be the same as the one which would be issued if no OlapServer exists.

2. Done

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #2927, opened 2019-10-17
DB-6168 Push down sampling into HBase scans for analyze

dgomezferro:


dgomezferro:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/mrio/api/core/SMRecordReaderImpl.java
```diff
@@ -229,10 +237,13 @@ public void restart(byte[] firstRow) throws IOException {
             Partition clientPartition;
             if (SpliceClient.isClient() && validToken()) {
             	scan.setAttribute(SIConstants.TOKEN_ACL_NAME, token);
-		ClientPartition delegate = new ClientPartition(instance.getConnection(), htable.getName(), htable, clock, driver.getPartitionInfoCache());
-		clientPartition = new AdapterPartition(delegate, instance.getConnection(), SpliceClient.getConnectionPool(debugConnections, maxConnections),htable.getName(), driver.getPartitionInfoCache());
+				ClientPartition delegate = new ClientPartition(instance.getConnection(), htable.getName(), htable, clock, driver.getPartitionInfoCache());
```
arnaud-splice: That indentation seems unnecessary

hbase_storage/src/main/java/com/splicemachine/si/impl/region/SamplingFilter.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public SamplingFilter(double rate) {
+        this.rate = rate;
+        ClassTag<String> tag = scala.reflect.ClassTag$.MODULE$.apply(String.class);
+        this.sampler = new BernoulliSampler(rate, tag);
+    }
+
+    @Override
+    public boolean filterRowKey(byte[] buffer, int offset, int length) throws IOException {
+        if (length == 0) return false; // make sure we don't remove special Memstore BEGIN / FLUSH rows
```
arnaud-splice: This corner case does not specifically appear in your tests, is it covered?

msirek: Is this sufficient to retain the special rows?  For example, a special FLUSH row could also be built with a non-zero rowkey:
```
            if (flushAlreadyReturned) {
                return new KeyValue(Bytes.toBytes(counter),ClientRegionConstants.FLUSH,ClientRegionConstants.FLUSH, 0l,ClientRegionConstants.FLUSH);
```
If the filtering were moved to filterKeyValue, then the Cell could be checked to see if it contains the special FLUSH column family, and avoid filtering it.

dgomezferro:

hbase_sql/src/main/java/com/splicemachine/mrio/api/core/SMRecordReaderImpl.java
dgomezferro: Thanks, fixed

hbase_storage/src/main/java/com/splicemachine/si/impl/region/SamplingFilter.java
dgomezferro: I added some tests that cover this through the SplitRegionScanner by running flushes concurrently with scans.

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/region/SamplingFilter.java
dgomezferro: Sorry, I missed this comment. I'd rather keep using filterRowKey because it has lower overhead and we can filter out rows without having to reconstruct the column families and all that. Those other rows that could get generated will eventually make it through (we increment the counter and regenerate them until the scanner is closed), the only case where this could be bad is if the selectivity is very very low, it might take a while until one of those rows gets picked up. However very, very low selectivities aren't much faster than low selectivities, so I don't expect this to be a big problem.

msirek: OK, thanks for the explanation. That sounds good to me.

PR #2928, opened 2019-10-17
DB-6168 Push down sampling into HBase scans for analyze (2.7)

dgomezferro:


dgomezferro:
@carolp-503 I don't think there is, I'll close it.

PR #2931, opened 2019-10-18
DB-8770 Inherit decimal scale from the input parameter in PreparedStatement.setObject

msirek:
JDBC programs that use  "void setObject(int parameterIndex, Object x, int targetSqlType)"
to specify a Decimal input parameter to a prepared statement end up getting Decimal values with the  fractional digits truncated.
The spec says a scale of 0 is implied:
https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#setObject(int,%20java.lang.Object,%20int)

This should probably be a default for data types that have no scale (such as Double).
This Jira changes the behavior to retain the scale from the Decimal source object, e.g.
        BigDecimal bd = new BigDecimal("14.51");
        statement.setObject(1, bd, Types.DECIMAL);
... will create a Decimal input parameter of 14.51, where previously the value 14 was created.
Source data types such as Double, which have no specific scale, still need to specify the scale via this method:

https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#setObject(int,%20java.lang.Object,%20int,%20int)

### DIFF
--- a/db-client/src/main/java/com/splicemachine/db/client/am/PreparedStatement.java
+++ b/db-client/src/main/java/com/splicemachine/db/client/am/PreparedStatement.java
@@ -31,6 +31,7 @@ import com.splicemachine.db.shared.common.reference.SQLState;
 import com.splicemachine.db.shared.common.sanity.SanityManager;
 import java.io.InputStream;
 import java.io.Reader;
+import java.math.BigDecimal;
 import java.sql.*;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -1537,7 +1538,10 @@ public class PreparedStatement extends Statement
                     agent_.logWriter_.traceEntry(this, "setObject", parameterIndex, x, targetJdbcType);
                 }
                 checkForClosedStatement();
-                setObjectX(parameterIndex, x, targetJdbcType, 0);
+                int scale = 0;
+                if (x instanceof BigDecimal)
+                    scale = ((BigDecimal)x).scale();
+                setObjectX(parameterIndex, x, targetJdbcType, scale);
             }
         }
         catch ( SqlException se )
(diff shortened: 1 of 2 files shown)


msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST FAILED

msirek:

spliceci:
TEST SUCCEEDED +1

arnaud-splice:
> Source data types such as Double, which have no specific scale, still need to specify the scale via this method:...

Do we have a test for that?

msirek:
> > Source data types such as Double, which have no specific scale, still need to specify the scale via this method:...
> 
> Do we have a test for that?

I've pushed a new test showing how decimal digits are truncated (as expected) when setting a Decimal from a Double without specifying scale.

jyuanca:
Should make this change that's not compatible with JDBC spec, or advise user to use a different API(setObject w/o scale)?

msirek:
> Should make this change that's not compatible with JDBC spec, or advise user to use a different API(setObject w/o scale)?

It is not entirely clear that this change is incompatible with the JDBC spec.  There's really only one sentence on it.  And the same code (found in the JIRA) run against MySQL is retaining the fractional digits:
`mysql> select * from testdb.testnumerics;
+---------------------------+---------------+-----------+
| description               | c2            | c3        |
+---------------------------+---------------+-----------+
|  supposed to be 15.51     |     15.510000 |     15.51 |
|  supposed to be 159238.51 | 159238.510000 | 159238.51 |
+---------------------------+---------------+-----------+
2 rows in set (0.01 sec)
`

I could not find a lot of examples of this online to show truncating digits is expected, so I was just going by another major db vendor's behavior... plus it seems to make sense to retain the scale (if you just constructed a BigDecimal with the exact value you want to insert into a Decimal column, it doesn't seem like chopping off the fractional digits would make a lot of sense from a user perspective).  In this case the scale is known, because a Decimal has scale.  But other data types, like Double, don't have an assigned scale.  What do you think?

carolp-503:
@dgomezferro , Any comments on what the standard says (ambiguously) vs our behavior?

PR #2937, opened 2019-10-22
DB-8821 Read-write lock contention in StripedTxnLifecycleStore

OlegMazurov:
_StripedTxnLifecycleStore_ requires a lock for each transaction. _LongStripedSynchronizer_ provides a pre-created set of read-write locks which are reused for multiple transactions  based on their id hash. It, thus, becomes possible to use the same lock by totally unrelated transactions.
The number of locks is tied to _hbase.regionserver.handler.count_ - the number of threads processing RPCs in the default thread pool. Due to hash collision, many threads will be using the same lock even if they are processing unrelated transactions.
Implement a new splice property to configure the number of locks independently. Define the default value of that property to be a multiple of _hbase.regionserver.handler.count_.


msirek:
+1

PR #2939, opened 2019-10-22
DB-8140 do not flatten non-correlated subquery that returns at most o…

yxia92:
…ne row

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
@@ -2719,7 +2719,7 @@ public class FromBaseTable extends FromTable {
                 if (!cd.isPrimaryKey())
                     continue;
                 else
-                    id = cd.getIndexDescriptor().getIndexDescriptor();
+                    id = cd.getIndexDescriptor();
             } else {
                 id=cd.getIndexDescriptor();
                 if(!id.isUnique()){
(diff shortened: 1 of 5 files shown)


yxia92:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -2709,14 +2713,18 @@ protected boolean supersetOfUniqueIndex(boolean[] eqCols) throws StandardExcepti
...
-            }
-            IndexDescriptor id=cd.getIndexDescriptor();
-
-            if(!id.isUnique()){
-                continue;
+            if(!cd.isIndex()) {
+                if (!cd.isPrimaryKey())
+                    continue;
+                else
+                    id = cd.getIndexDescriptor().getIndexDescriptor();
```
arnaud-splice: `id = cd.getIndexDescriptor();` instead of `cd.getIndexDescriptor().getIndexDescriptor()`

Both achieve the same goal, so the smaller version is better imo.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -2753,14 +2761,18 @@ protected boolean supersetOfUniqueIndex(JBitSet[] tableColMap)
...
```
arnaud-splice: Same

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
yxia92: Thanks @arnaud-splice, I've updated the code accordingly.

PR #2946, opened 2019-10-23
DB-8422 Security Audit Log (create/drop user, modify password, login)

changli6:
This PR covers DB-8422 (create / drop user), DB-8423 (login) and DB-8425 (password change)
Mainly code change are in these files:
db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java 
db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/StatementPermission.java
db-shared/src/main/java/com/splicemachine/db/shared/common/reference/AuditEventType.java

Introduction is in [DB-8422](https://splicemachine.atlassian.net/browse/DB-8422)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
@@ -50,6 +50,9 @@ import com.splicemachine.db.impl.jdbc.Util;
 import com.splicemachine.db.impl.load.Import;
 import com.splicemachine.db.impl.sql.execute.JarUtil;
 import com.splicemachine.db.jdbc.InternalDriver;
+import com.splicemachine.db.shared.common.reference.AuditEventType;
+import org.apache.log4j.Logger;
+import com.splicemachine.utils.StringUtils;
 
 import java.security.AccessController;
 import java.security.Policy;
@@ -76,6 +79,8 @@ public class SystemProcedures{
     private final static int SQL_ROWVER=2;
     private final static String DRIVER_TYPE_OPTION="DATATYPE";
     private final static String ODBC_DRIVER_OPTION="'ODBC'";
+
+    private static final Logger AUDITLOG =Logger.getLogger("splice-audit");
 
     // This token delimiter value is used to separate the tokens for multiple 
     // error messages.  This is used in DRDAConnThread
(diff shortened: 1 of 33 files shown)


db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
```diff
@@ -1710,13 +1718,19 @@ private static boolean hasTable(Connection conn,String schemaName,
...
         }catch(StandardException se){
             throw PublicAPI.wrapStandardException(se);
+        }catch (SQLException sqle){
+            status = false;
+            reason = sqle.getMessage();
+            throw sqle;
+        }
+        finally {
+            if (AUDITLOG.isInfoEnabled())
+                AUDITLOG.info(StringUtils.logSpliceAuditEvent(currentUser,AuditEventType.CREATE_USER.name(),status,ip,lcc.getStatementContext().getStatementText(),reason));
```
yxia92: By passing down the statement directly, we log the password directly in the splice-audit log. Is it appropriate? Same concern for the logging of the other two system procedures.

carolp-503: We definitely don't want to include the password in the log, even if it's the failed password.

changli6: That is the concern. Even though the splice-audit.log is permission strict, then sensitive info should not be exposed. One solution is to change the log4j configuration using MaskPatternLayout to hide sensitive info. Another choice is doing MaskPattern in internal to avoid the filtering regular expression is too complicated to customer. 
I prefer the second one.

utilities/src/main/java/com/splicemachine/utils/StringUtils.java
```diff
@@ -96,16 +107,54 @@ private static String padding(int repeat, char padChar) throws IndexOutOfBoundsE
...
+            maskedMessage.append(message, start, end+1);
+        }
+        return String.valueOf(maskedMessage);
+    }
+
+
     public static String logSpliceAuditEvent(String userid, String event, boolean status, String addr, String statement, String reason)
     {
+
+        Pattern maskPattern = Pattern.compile(maskPasswordPatternStr,Pattern.CASE_INSENSITIVE);
```
yxia92: Is it expensive to create and compile a pattern each time we log a entry? Since it is the same pattern that we are using, is it possible to let each thread share the same Pattern object?

changli6: We can move the maskPattern out of the function to be a public static member of StringUtils class, the pattern won't be compiled each time and could be shared among threads.

PR #2947, opened 2019-10-24
DB-8823 Create RegionServer endpoint to get oldest active transaction

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_protocol/src/main/protobuf/Splice.proto
```diff
@@ -160,6 +160,19 @@ message SpliceRegionSizeResponse {
...
+	rpc getOldestActiveTransaction(SpliceOldestActiveTransactionRequest)
+		returns (SpliceOldestActiveTransactionResponse);
+}
+
+message SpliceOldestActiveTransactionRequest {
+
+}
+
+message SpliceOldestActiveTransactionResponse {
+    required int64 oldestActiveTransaction = 1;
```
dgomezferro: In order to garbage collect data, you will probably need the most recent transaction for each server when there's no active transactions. I'm ok adding that later though.

yxia92:
@arnaud-splice  Could you summarize the benefit of creating an endpoint to fetch oldest active transaction? I saw that Vacuum now sends requests to region servers to get oldest active transactions in parallel as compared to the sequential way before the fix. And JDBC connection is replaced with protobuf message. Are these the main aspects? Thanks!

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/RegionServerEndpoint.java
```diff
@@ -0,0 +1,63 @@
...
+                                           RpcCallback<SpliceMessage.SpliceOldestActiveTransactionResponse> callback) {
+        if (LOG.isDebugEnabled())
+            SpliceLogUtils.debug(LOG, "getOldestActiveTransaction");
+        SpliceMessage.SpliceOldestActiveTransactionResponse.Builder writeResponse = SpliceMessage.SpliceOldestActiveTransactionResponse.newBuilder();
+
+        long oldestActiveTransaction = SIDriver.driver().getTxnStore().oldestActiveTransaction();
+        writeResponse.setOldestActiveTransaction(oldestActiveTransaction);
+        callback.run(writeResponse.build());
+    }
+}
```
jyuanca: Adding a coprocessor requires configuration changes for customers. I won't do this unless it is absolutely necessary

arnaud-splice:
> @arnaud-splice Could you summarize the benefit of creating an endpoint to fetch oldest active transaction? I saw that Vacuum now sends requests to region servers to get oldest active transactions in parallel as compared to the sequential way before the fix. And JDBC connection is replaced with protobuf message. Are these the main aspects? Thanks!

Creating an endpoint is the first step in creating a watcher -that lives at RegionServer level- which would fetch the oldestActiveTransaction every now and then and store it. To be able to delete tombstones whenever we run minor/major compaction, we need to know what is the  oldestActiveTransaction, so that we don't delete information that is still being read.
Using it in vacuum is simply a way to use the endpoint, and tangentially improve it, but that is not the main reason I created that PR. Does that make sense?

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/RegionServerEndpoint.java
arnaud-splice: See my reply to @yxia92 for the reason why I suggest adding this coprocessor.
I believe that this endpoint can be extended in the future to support all request to a region server.

arnaud-splice:

yxia92:
Thanks @arnaud-splice , Yes, it makes sense.

spliceci:
TEST SUCCEEDED +1

arnaud-splice:
> Thanks @arnaud-splice , Yes, it makes sense.
> But I will hold my approval till we reach a consensus on @jyuanca 's comment on the 3.0 branch about merging the current logic with another existing endpoint class SpliceRSRpcServices as it may incur change on master also. Thanks!

Makes sense. I will rename my RegionServerEndpoint to SpliceRSRpcServices and push a new patch. 
On 3.0, both will be merged in an upcoming patch.

PR #2952, opened 2019-10-28
DB-8498 Port splice to CDH 6.3, support HBase 2.1 and Spark 2.4

wb14123:


dgomezferro:
Is there an easy way to reduce the amount of duplicated files we have per platform?

hbase_sql/orc-spark-2.4/src/main/java/org/apache/spark/sql/execution/vectorized/ColumnarBatch.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public void setColumn(int ordinal, ColumnVector column) {
+        columns[ordinal] = column.getDelegate();
+    }
+
+    private ColumnarBatch(StructType schema, int maxRows, MemoryMode memMode) {
+        this.numRows = maxRows;
+        this.columns = new WritableColumnVector[schema.size()];
+    }
+}
```
jyuanca: The above spark 2.4 specific files look same as spark 2.3. If so, you may consider refactor to avoid duplicate files.

db-engine/src/main/java/com/splicemachine/db/iapi/types/XML.java
```diff
@@ -1148,6 +1148,11 @@ public void read(UnsafeArrayData unsafeArrayData, int ordinal) throws StandardEx
         xmlStringValue = new SQLChar(unsafeArrayData.getUTF8String(ordinal).toString());
     }
 
+    @Override
+        public int getUnsafeArrayElementSize() {
+            return 8;
+    }
+
     /**
```
jyuanca: can you document how to determine the value of getUnsafeArrayElementSize? Why it is 8 for XML and UserType?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
```diff
@@ -532,8 +533,15 @@ public void createEmptyExternalFile(StructField[] fields, int[] baseColumnMap, i
                             .mode(SaveMode.Append).parquet(location);
                 }
                 else if (storedAs.toLowerCase().equals("a")) {
+                    if (compression.equals("none")) {
+                        compression = "uncompressed";
+                    }
+                    /*
                     empty.write().option("compression",compression).partitionBy(partitionByCols.toArray(new String[partitionByCols.size()]))
                             .mode(SaveMode.Append).format("com.databricks.spark.avro").save(location);
```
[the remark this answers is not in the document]
wb14123: The new Spark API use "uncompressed" for no compress instead of "none". So convert it to make it compatible with the new API.

wb14123:
> platforms/cdh6.3.0/docs/CDH-installation.md is missing

@OlegMazurov Added installation guid

db-engine/src/main/java/com/splicemachine/db/iapi/types/XML.java
wb14123: This is the fix from @msirek related to unsaferow. I cherry-pick it from Brandon's branch about hdp 3.1 and Spark 2.4. To be honest I don't quite understand it either.

carolp-503: We need to investigate how used and why the value 8 was chosen, but that should not hold up this PR.  Is this a Spark 2.4 issue?  HBase 2.0?  Something else?

I will create a Jira for this for Q4 S3.

hbase_sql/orc-spark-2.4/src/main/java/org/apache/spark/sql/execution/vectorized/ColumnarBatch.java
carolp-503: Jira this and fix as a separate PR

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceDateFunctionsTest.java
```diff
@@ -404,6 +404,7 @@ public void testTimeStampFormats() throws Exception {
     }
 
     @Test
+    @Ignore("Failed at different time zone")
```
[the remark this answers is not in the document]
carolp-503: Will create Jira for "Cannot run ITs from different timezone with CDH 6.3/Splice 3.0" and Bin's latest commit removes the Ignore

hbase_sql/pom.xml
```diff
@@ -123,6 +123,24 @@
             <artifactId>guava</artifactId>
             <scope>provided</scope>
         </dependency>
+
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.9.1</version>
```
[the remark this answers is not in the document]
carolp-503: Resolved in latest commits from Bin.

hbase_storage/pom.xml
```diff
@@ -53,6 +57,31 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>2.9.9</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.6.7</version>
```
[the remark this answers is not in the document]
carolp-503: Resolved in latest commit from Bin

PR #2954, opened 2019-10-28
DB-8806 and oth.: improve OLTP performance

OlegMazurov:
DB-8796 High GC overhead during OLTP workload
DB-8806 CompletedTxnCache is expensive
DB-8789 Apparent deadlock during OLTP workload

Greatly reduces overhead from getTransaction() RPCs - a bottleneck for OLTP workloads like TPC-C and HTAP.
Introduces RPC priorities and a new _SpliceRpcSchedulerFactory_ to enable them (but not a Splice scheduler yet).

### DIFF
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
@@ -155,7 +155,7 @@ public class SIConfigurations implements ConfigurationDefault {
 
     // Threads processing first queue
     public static final String ROLLFORWARD_FIRST_THREADS = "splice.txn.rollforward.firstQueueThreads";
-    public static final int DEFAULT_ROLLFORWARD_FIRST_THREADS = 10;
+    public static final int DEFAULT_ROLLFORWARD_FIRST_THREADS = 25;
 
     // Threads processing second queue
     public static final String ROLLFORWARD_SECOND_THREADS = "splice.txn.rollforward.secondQueueThreads";
--- a/splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
+++ b/splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
@@ -30,7 +30,7 @@ public class SIConfigurations implements ConfigurationDefault {
     public static final byte[] CONGLOMERATE_TABLE_NAME_BYTES = Bytes.toBytes(CONGLOMERATE_TABLE_NAME);
 
     public static final String completedTxnCacheSize="splice.txn.completedTxns.cacheSize";
-    private static final int DEFAULT_COMPLETED_TRANSACTION_CACHE_SIZE=1<<17; // want to hold lots of completed transactions
+    private static final int DEFAULT_COMPLETED_TRANSACTION_CACHE_SIZE=1<<20; // want to hold lots of completed transactions
 
     public static final String completedTxnConcurrency="splice.txn.completedTxns.concurrency";
     private static final int DEFAULT_COMPLETED_TRANSACTION_CONCURRENCY=64;
--- a/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
+++ b/splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
@@ -72,7 +72,7 @@ public class SimpleTxnFilter implements TxnFilter{
                            TxnSupplier baseSupplier,
                            boolean ignoreNewerTransactions) {
         assert readResolver!=null;
-        this.transactionStore = new ActiveTxnCacheSupplier(baseSupplier,1024); //TODO -sf- configure
+        this.transactionStore = new ActiveTxnCacheSupplier(baseSupplier, 100); //TODO -sf- configure
         this.tableName=tableName;
         this.myTxn=myTxn;
         this.readResolver=readResolver;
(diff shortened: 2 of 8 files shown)


splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -273,30 +290,10 @@ private void addToTombstoneCache(DataCell data) throws IOException{
...
-			 */
-
-            long commitTs=data.valueAsLong();//dataStore.getOpFactory().getValueToLong(data);
-            TxnView toCache=new CommittedTxn(txnId,commitTs);//since we don't care about the begin timestamp, just use the TxnId
-            transactionStore.cache(toCache);
-            currentTxn=toCache;
+        long txnId = data.version();
+        if (checkLocally(txnId) == null) {
+            long commitTs = data.valueAsLong();
+            cacheLocally(new CommittedTxn(txnId, commitTs));
```
dgomezferro: This is not updating the global cache, CompletedTxnCacheSupplier is not used then.

OlegMazurov: That's the whole point. That optimization was broken for heavy workloads. From DB-8806:

> CompletedTxnCache is a global cache. An instance of SimpleTxnFilter tries to save all encountered transaction commit timestamps into that cache. When under heavy load (as it happens with DB-7009.jar), a stored CommittedTxn may not survive in the cache when it’s needed to validate the actual value recorded by that transaction, whether used data or a tombstone. There is little help in sharing commit timestamps in the global cache as other readers most likely encounter this or related commit timestamps on their own. It’s much more beneficial to keep commit timestamps in a local cache and keep CompletedTxnCache as a global cache for actual RPCs to the transaction table.

OlegMazurov:
Running TPC-C for 1 hour:

Metric          | Before             | After         
------------ | ------------- |---------|
NewOrder (cnt) | 976570        | 1059250 |
tpmC         |  16276        |   17654 |
delivGetOrderId (sec) | 658197 | 311617 |

jyuanca:
Are TPCH workload impacted positively or negatively?

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
dgomezferro: That makes sense, thanks!

splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
```diff
@@ -155,7 +155,7 @@
 
     // Threads processing first queue
     public static final String ROLLFORWARD_FIRST_THREADS = "splice.txn.rollforward.firstQueueThreads";
-    public static final int DEFAULT_ROLLFORWARD_FIRST_THREADS = 10;
+    public static final int DEFAULT_ROLLFORWARD_FIRST_THREADS = 25;
```
dgomezferro: Do we need that many roll forward threads? Have you tried decreasing the queue wait times, since we are now getting better latencies?

splice_si_api/src/main/java/com/splicemachine/si/impl/store/CompletedTxnCacheSupplier.java
```diff
@@ -34,72 +31,84 @@
...
-        this.delegate=delegate;
+
+    public CompletedTxnCacheSupplier(TxnSupplier delegate, int maxSize, int concurrencyLevel) {
+        cache = new TxnView[Integer.highestOneBit(maxSize)];    // ensure power of 2
+        this.delegate = delegate;
+    }
+
+    private int idx(long val) {
+        int hashCode = Long.hashCode(val);
+        hashCode ^= (hashCode >>> 20) ^ (hashCode >>> 12);
```
dgomezferro: Did you get hashing strategy from somewhere or did you come up with it?

OlegMazurov:
@jyuanca TPC-H 100 run shows little difference (with a notable exception of Q14) in query performance as expected (there are few write transactions which are easily cached):
http://nexus.splicemachine.com:8080/view/test_builds/job/test_onDemand/view/benchmarks/job/perf_test/21/artifact/artifacts/index.html
I expect noticeable positive impact on OLAP queries in HTAP where those are affected by OLTP transactions updating data.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
OlegMazurov: After the change, roll forward threads are 100% busy when tested with 10, 20, 40 threads. 
TPC-C benchmark performance is sensitive to their throughput (mostly for the second and third hour). The 25 value is a compromise to get better current performance before we understand the new behavior better.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/CompletedTxnCacheSupplier.java
OlegMazurov: We use the same hash function in _LongStripedSynchronizer_.
_idx2()_ is my extension to break collisions in _idx()_.

dgomezferro: I'd rather have the logic in a single place (some utils class?) with the original comment attached.

OlegMazurov: Switched to using existing HashFunctions.UtilHash (same implementation for longs).

PR #2960, opened 2019-10-29
DB-8087 function HEX implementation

changli6:


carolp-503:
I'm trying to see the original queries (as opposed to the ones that we later modified because we didn't support HEX).  For now, we can support CHAR and VARCHAR and expand once we see usage information.

carolp-503:
I took a look at the customer's DDL and the only place that HEX was used in the original PoC queries is HEX(TABLE_NAME).  Tracing back to see the data type for TABLE_NAME, it appears to be CHAR(18).  So I think this implementation is good for now, but we should update the Jira to reflect the limitations in our implementation so our documentation doesn't describe things we don't support yet.

msirek:
I'm OK with handing just char and varchar for now.

PR #2970, opened 2019-10-31
DB-8422 Security Audit Log (create/drop user, modify password, login)

changli6:


changli6:
jenkins please test branch @mem

changli6:

spliceci:
TEST SUCCEEDED +1

changli6:
Fix null message string exception.

yxia92:
@changli6 , is the following change the only difference from the reverted PR?
In StringUtils.maksMessage():
```
        if (message == null)
            return null;
        if (maskString == null)
            maskString = "";
```
Thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -3855,6 +3856,11 @@ private String formatLogStmt(String statement) {
             if (maxStatementLogLen >= 0 && maxStatementLogLen < statement.length()) {
                 subStatement = statement.substring(0, maxStatementLogLen) + " ... ";
             }
+            subStatement = StringUtils.maskMessage(subStatement, StringUtils.maskPasswordPattern,StringUtils.maskString);
```
dgomezferro: I'm a little bit worried about the cost of this function, since it's parsing the full statement, specially with very long statements. Can we run some benchmarks to make sure this is not affecting performance for OLTP workloads?

dgomezferro:
I think this is fine as a first approach, but by using regexes we are probably missing many cases ( comment just before the call for instance) and in other cases the sensitive data could be in a prepared statement variables. 

I'd like to make sure the regexes aren't too expensive to be run on every statement, otherwise we'd need to only run the regexes when we know we are running some of the sensitive operations.

changli6:
> @changli6 , is the following change the only difference from the reverted PR?
> In StringUtils.maksMessage():
> 
> ```
>         if (message == null)
>             return null;
>         if (maskString == null)
>             maskString = "";
> ```
> 
> Thanks!

Yes, it is.

changli6:
@dgomezferro

> I think this is fine as a first approach, but by using regexes we are probably missing many cases ( comment just before the call for instance) and in other cases the sensitive data could be in a prepared statement variables.
> 
> I'd like to make sure the regexes aren't too expensive to be run on every statement, otherwise we'd need to only run the regexes when we know we are running some of the sensitive operations.

It did miss some cases such as the password is a parameter in a prepared statement, and it affected the performance on matching each failed statement. 

I'd like to remove codes of masking failed statements and keep this PR only for the audit log. 
Currently, the implementation is fine with masking the sensitive info in the audit log. It could mask the password in "CALL ..." statements. If there was a prepared statement, the audit log just stored the original statement with ? as parameters without its value, so the password would not appear. 

The failed statement masking is tracked on a separated Jira DB-8828

changli6:

spliceci:
TEST FAILED

changli6:

spliceci:
TEST FAILED

changli6:

spliceci:
TEST FAILED

changli6:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
changli6: The performance concern will be covered Jira DB-8828

carolp-503:
@dgomezferro , are you good with Changli's response to your question?  If not, what more do we need to do to commit this PR?

dgomezferro:
Yep, sounds good.

PR #2974, opened 2019-10-31
SPLICE-2362 Don't compile ParserInterface.parseDataType on platforms using spark2.1.

msirek:
Fixes a compilation error on hdp2.6.1 and mapr6.0.0.  Those platforms compile against spark 2.1, so cannot use the optimization of SPLICE-2362 because one of the Spark API calls is only available in Spark 2.2 and above.


msirek:
Jenkins please test branch @cdh5.12.2

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SparkRelationalOperator.java
```diff
@@ -66,9 +68,9 @@ public SparkRelationalOperator(int relOpKind,
     @Override
     public Column getColumnExpression(Dataset<Row> leftDF,
                                       Dataset<Row> rightDF,
-                                      ParserInterface parser) throws UnsupportedOperationException {
```
dgomezferro: Could you use a more descriptive name for the function? Maybe parser, parserFunction or something like that?

msirek: OK.  Changing it to convertStringToDataTypeFunction.

PR #2978, opened 2019-10-31
DB-8085 Can also use ALIAS instead of SYNONYM

wb14123:


splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/CreateAliasIT.java
```diff
@@ -0,0 +1,57 @@
...
+    public void testCreateSynonym() throws  Exception {
+        methodWatcher.executeUpdate("create synonym t2 for t1");
+        methodWatcher.executeUpdate("drop synonym t2");
+    }
+
+    @Test
+    public void testCreateAlias() throws  Exception {
+        methodWatcher.executeUpdate("create alias t2 for t1");
+        methodWatcher.executeUpdate("drop alias t2");
+
```
dgomezferro: Can you add a few tests selecting from a synonym, etc?

wb14123: Added the test about select. While adding the test, found a bug about create synonym. Created the ticket DB-8854.

dgomezferro: Thanks @wb14123 !

OlegMazurov:
Will there be a PR for branch-3.0?

wb14123:
Created 3.0 PR #2993

PR #2979, opened 2019-11-02
DB-8132 return first available union branch first

yxia92:
For union-all operation that takes control execution, instead of always returning rows from the first union branch, return from the branch whose rows become available first.


yxia92:

spliceci:
TEST SUCCEEDED +1

PR #2985, opened 2019-11-05
DB-8856 Make testTimeFormats not dependent on system timezone.

msirek:
Fixes failing tests in SpliceDateFunctionsTest.testTimeFormats.  Apparently there is no such thing as a Time with a time zone such as PDT or PST.  The presence of the DT (daylight time) or ST (standard time) implies a DateTime value.  So, when we build a time such as "12 o'clock PM, Pacific Daylight Time", to insert into the database, for example, the only way we can construct it is to associate it with a DateTime, which is always the current system DateTime.  For this example, If the current system DateTime uses daylight savings in the Pacific time zone, then the time will be in PDT.  If the current system DateTime uses standard time in the Pacific time zone, then the time will be in PST. 

Due to all this, we can't compare a PDT or PST Time against a UTC Time, because the results will vary on different days of the year.  The solution is to only compare a PDT time with a PDT time and a PST time with a PST time.

It may also be a good idea in the future to remove support for Times that have these types of time zone specifiers because this type of behavior seems counter-intuitive.  Since our implementation relies on Java interfaces, and this is how they behave, if we want to support PDT or PST Time values not associated with a given DateTime, we would have to write our own parser to handle it.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceDateFunctionsTest.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceDateFunctionsTest.java
@@ -30,10 +30,9 @@ import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
+import java.time.*;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.TemporalAmount;
 import java.util.Calendar;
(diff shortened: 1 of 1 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

arnaud-splice:
> It may also be a good idea in the future to remove support for Times that have these types of time zone specifiers because this type of behavior seems counter-intuitive. Since our implementation relies on Java interfaces, and this is how they behave, if we want to support PDT or PST Time values not associated with a given DateTime, we would have to write our own parser to handle it.

I agree. We probably do not want to allow such a weird behaviour.

PR #2988, opened 2019-11-06
DB-8098 add doNotFlatten hint for subquery

yxia92:


yxia92:

spliceci:
TEST FAILED

yxia92:

spliceci:
TEST SUCCEEDED +1

db-shared/src/main/java/com/splicemachine/db/shared/common/reference/SQLState.java
```diff
@@ -1044,6 +1044,8 @@ The message identifier (ie. the value that matches a field in this class)
 	String LANG_INVALID_USE_OF_DEFAULT                                 = "42Y85";
 	String LANG_INVALID_SESSION_PROPERTY                               = "42Y86";
 	String LANG_INVALID_SESSION_PROPERTY_VALUE                         = "42Y87";
+	String LANG_INVALID_FORCED_NO_SUBQUERY_FLATTEN_VALUE               = "42Y88";
+	String LANG_INVALID_SUBQUYERY_PROPERTY                             = "42Y89";
```
arnaud-splice: typo: LANG_INVALID_SUBQU{Y,}ERY_PROPERTY

yxia92: Thanks @arnaud-splice I've fixed the typo.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SubqueryNode.java
```diff
@@ -220,6 +222,31 @@ public void init(
...
+            String value=(String)properties.get(key);
+
+            switch(key){
+                case "doNotFlatten":
+                    try {
+                        hintNotFlatten = Boolean.parseBoolean(StringUtil.SQLToUpperCase(value));
+                    } catch (Exception exception) {
+                        throw StandardException.newException(SQLState.LANG_INVALID_FORCED_NO_SUBQUERY_FLATTEN_VALUE, value);
+                    }
+                    break;
```
changli6: This 'break' only jumps out of the switch part, I think it's better to jump out of the while loop also.

yxia92: Thanks @changli6 . Right now, subquery only supports one hint, in the future, if we have more hints to support, we don't want to break out the while loop.

PR #3016, opened 2019-11-14
DB-8876 adjust precision for decimal 0.

yxia92:


yxia92:

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
```diff
@@ -1072,6 +1074,12 @@ public void setWidth(int desiredPrecision,
 		}
 		rawData = null;
 		setValue(value.setScale(desiredScale, BigDecimal.ROUND_HALF_UP));
+		// for decimal 0, it is possible that precision is less than scale
+		// after the call of setScale, e.g., cast(0 as decimal(14,4)). The BigDecimal 0 starts with
+		// precision 1 and scale 0, and ends with precision 1 and scale 4,
+		// so adjust the precision accordingly
+		if (value.signum() == 0)
+			setPrecision(value.scale() + 1);
```
msirek: If this is just for decimal 0, the fix looks good.  Why is there no problem with other numbers?

yxia92: For other numbers, setScale() seems to reset the precision to 0 first. When we try to fetch its precision through the function BigDecimal.decimal(), the precision is correctly re-calculated. Here is the code of BigDecimal.decimal():
```    /**
     * Returns the <i>precision</i> of this {@code BigDecimal}.  (The
     * precision is the number of digits in the unscaled value.)
     *
     * <p>The precision of a zero value is 1.
     *
     * @return the precision of this {@code BigDecimal}.
     * @since  1.5
     */
    public int precision() {
        int result = precision;
        if (result == 0) {
            long s = intCompact;
            if (s != INFLATED)
                result = longDigitLength(s);
            else
                result = bigDigitLength(intVal);
            precision = result;
        }
        return result;
    }
```

msirek: I see.  In addition to setting the precision of the SQLDecimal, it may be a good idea to set the precision of the BigDecimal value itself.  If in some processing path we call setCoreValue, passing the BigDecimal, it could using the old precision value combined with the new scale value:

	private void setCoreValue(BigDecimal theValue)
	{
		value = theValue;
		if (value !=null) {
			precision = value.precision();
			scale = value.scale();
		}
		rawData = null;
		isNull = evaluateNull();
	}

Another way to do it is construct a new BigDecimal:

value = new BigDecimal(BigInteger.ZERO, 0, scale, precision)

But then you have to make sure BigDecimal.ROUND_HALF_UP is still set.

yxia92: that is true, @msirek I had the same concern. but I didn't find a way to set the precision of BigDecimal. Let me try your way to construct a new BigDecimal. Thanks!

yxia92: Unfortunately, BigDecimal(BigInteger.ZERO, 0, scale, precision) is not public, so we cannot use it.

msirek: We could write a new function, getValuePrecision(), and use that in all places value.precision() occurs in SQLDecimal.java (and maybe add a comment not to directly call value.precision).  This function could have special logic to get the precision of a zero value.

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecimal.java
yxia92: Thanks @msirek . I've updated the fix as suggested.

yxia92:

spliceci:
TEST SUCCEEDED +1

PR #3020, opened 2019-11-15
DB-8901 Consistency check in SpliceController.batchFetch

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
```diff
@@ -168,13 +168,15 @@ public boolean batchFetch(List<RowLocation> locations, List<ExecRow> destRows, F
         return batchFetch(locations,destRows,validColumns,false);
     }
     @Override
-    public boolean batchFetch(List<RowLocation> locations, List<ExecRow> destRows,FormatableBitSet validColumns,boolean waitForLock) throws StandardException{
-        if (locations.isEmpty())
+    public boolean batchFetch(List<RowLocation> locations, List<ExecRow> destRows, FormatableBitSet validColumns, boolean waitForLock) throws StandardException{
+        if (locations.size() != destRows.size())
             return false;
```
msirek: getDescriptorViaIndexMinion calls batchFetch.  So, if there's a list size mismatch, getDescriptorViaIndexMinion might not find any tuple descriptors.  Would this error situation always be caught or should we change the 'return false' here into an assertion?

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
```diff
@@ -185,18 +187,19 @@ public boolean batchFetch(List<RowLocation> locations, List<ExecRow> destRows,Fo
...
                 row.setRowArray(destRow);
                 row.resetRowArray();
-                DataCell keyValue=result.userData();
-                rowDecoder.set(keyValue.valueArray(),keyValue.valueOffset(),keyValue.valueLength());
+                DataCell keyValue = result.userData();
+                rowDecoder.set(keyValue.valueArray(), keyValue.valueOffset(), keyValue.valueLength());
                 rowDecoder.decode(row);
                 i++;
             }
+            assert i == locations.size();
```
msirek: I believe assertions only get thrown on standalone.  If you want the error to be caught on a cluster as well, maybe an Exception would be better (I don't know this for sure about assertions, but that's the behavior I recall seeing.  You may want to test it).

jyuanca: Assertion is not enabled when splice is deployed on a cluster

arnaud-splice: This is development specific and should not happen on a production machine.

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceController.java
arnaud-splice: That's a good point, I changed the way we catch error in getDescriptorViaIndexMinion to reflect what's happening in here.

I think that there should be some more refactoring in the future in there. (legacy assertions, ..)

arnaud-splice: That's a good point, I changed the way we catch error in getDescriptorViaIndexMinion to reflect what's happening in here.

I think that there should be some more refactoring in the future in there. (legacy assertions, ..)

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -7183,8 +7183,7 @@ public final TupleDescriptor getDescriptorViaIndex(int indexId,
             rowLocations.add((RowLocation) indexRow1.getColumn(
                     indexRow1.nColumns()).cloneValue(true));
         }
-        heapCC.batchFetch(rowLocations,outRows,null);
-        if (outRows.size() != i) {
+        if (!heapCC.batchFetch(rowLocations,outRows,null)) {
             if(SanityManager.DEBUG){
```
dgomezferro: It'd be better to substitute this SanityManager block for something else that raises an error

jyuanca:
@arnaud-splice Can you please resolve change request?

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3022, opened 2019-11-16
DB-8844: fix hdp3.1.0

jyuanca:


pom.xml
```diff
@@ -854,12 +859,13 @@
             <properties>
                 <envClassifier>hdp3.1.0</envClassifier>
                 <envHbase>hbase2.0.2</envHbase>
+                <envSpark>spark2.3</envSpark>
```
OlegMazurov: We already have _spark.version_ and _spark.major.version_. Why do we need yet another one, envSpark (matching spark.major.version)?

jyuanca: Good point. I did not noticed spark.major.version is defined

db-engine/pom.xml
```diff
@@ -275,6 +275,51 @@
...
+                <executions>
+                    <execution>
+                        <id>add-platform-source</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${spark.major.version}/src/main/java</source>
```
OlegMazurov: If db-engine has to depend on Spark there is something seriously wrong in the design.

OlegMazurov: ${spark.major.version} is not defined when only core modules are built (_mvn -Pcore_).
The _spliceengine-core_ build in Jenkins will fail.

OlegMazurov:
I have removed unused code that caused the maintenance issue. All core and platform modules can be built both separately and together (-Pcore,ee or -Pcore,hdp3.1.0,ee).
Many dependencies on Spark remain in sore modules, mostly, but not limited to, shaded guava classes used explicitly from the spark-network-common jar (with the default Spark version dependency).
I believe those unwanted dependencies can be taken care of separately from this PR.

PR #3026, opened 2019-11-17
DB-8167 Stagger region load refreshes

dgomezferro:
My previous PR had a problem because the cache wasn't initialized for a long time, functions like tableLoad() could try to .put() values into a null map.

Now I initialize the cache to an empty Map and schedule the first update right away, the rest will continue to be staggered.


dgomezferro:

spliceci:
TEST SUCCEEDED +1

PR #3030, opened 2019-11-18
DB-8854 clear the corresponding nameTdCache entry after drop synonym/…

yxia92:
…alias.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropAliasConstantOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropAliasConstantOperation.java
@@ -23,14 +23,14 @@ import com.splicemachine.db.iapi.sql.depend.DependencyManager;
 import com.splicemachine.db.iapi.sql.dictionary.*;
 import com.splicemachine.db.iapi.sql.execute.ConstantAction;
 import com.splicemachine.db.iapi.store.access.TransactionController;
+import com.splicemachine.db.impl.sql.catalog.TableKey;
 import com.splicemachine.ddl.DDLMessage;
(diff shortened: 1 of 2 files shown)


yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropAliasConstantOperation.java
```diff
@@ -152,6 +152,11 @@ public static void invalidate(AliasDescriptor ad, DependencyManager dm, Language
         }
 
         dm.invalidateFor(ad, invalidationType, lcc);
+        if (ad.getAliasType() == AliasInfo.ALIAS_TYPE_SYNONYM_AS_CHAR) {
+            // invalidate nameTdCache
+            TableKey key = new TableKey(ad.getSchemaUUID(), ad.getName());
+            lcc.getDataDictionary().getDataDictionaryCache().nameTdCacheRemove(key);
```
dgomezferro: I think this won't work for other RSs, would it?

Can you add a test for that case? Create alias in RS0, use it in RS1, drop it from RS0, see if you can recreate it from RS1

yxia92: @dgomezferro , the fix works for other RSs. This function DropAliasConstantOperation.invalidate() is called by SpliceDatabase.registerDDL()->DDLUtils.preDropAlias(), so all region servers will invoke it.
I've pushed an IT to test it. Thanks!

dgomezferro: I see, thanks!

PR #3037, opened 2019-11-22
DB-8921 Make sporadic failure in test much less likely

arnaud-splice:
the failure would happen in avg once every 5! (120) execution.
It now would happen once every 20! (2,432902008E18).

### DIFF
--- a/splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceTest.scala
+++ b/splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceTest.scala
@@ -406,7 +406,7 @@ class DefaultSourceTest extends FunSuite with TestContext with BeforeAndAfter wi
   }
 
   test("partitions shuffle") {
-    val rdd = generateRows(10, 5)
+    val rdd = generateRows(40, 20)
 
     var i = 0
 


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceTest.scala
```diff
@@ -406,7 +406,7 @@ class DefaultSourceTest extends FunSuite with TestContext with BeforeAndAfter wi
   }
 
   test("partitions shuffle") {
-    val rdd = generateRows(10, 5)
+    val rdd = generateRows(40, 20)
```
yxia92: I'm fine with increasing both the number of rows and partitions, but why is the chance of failure for `generateRows(10, 5)` one out of 5!? Shouldn't it be one out of 10! where 10 is the number of rows?

arnaud-splice: I I understood correctly, we shuffle partitions, not rows. We had 5 partitions with 2 rows. We now have 20 partitions with 2 rows.
Also, if it failed once out of 10!, I would have been incredibly (un)lucky to hit this!

PR #3042, opened 2019-11-25
DB-8875 add system procedures to fake table and column stats

yxia92:


yxia92:

spliceci:
TEST FAILED

yxia92:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsMerge.java
```diff
@@ -88,20 +92,41 @@ public void init() {
...
+
+        if (!value.getType().equals(columnStatsType))
+            throw new RuntimeException("Multiple column stats type co-exists!");
+
+        if (value instanceof FakeColumnStatisticsImpl) {
+            FakeColumnStatisticsImpl fakeColumnStatistics = (FakeColumnStatisticsImpl) value;
+            nullCount += fakeColumnStatistics.nullCount();
+            totalCount += fakeColumnStatistics.totalCount();
+            if (cardinality < fakeColumnStatistics.cardinality())
+                cardinality = fakeColumnStatistics.cardinality();
```
msirek: Why update cardinality only when it is less than the faked value?

yxia92: This function merges stats from two partitions. I simply assume that the distinct values are randomly distributed in the data, so they are overlapping in the two partitions, and the merged number of distinct values will go with the bigger of the two partitions.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
```diff
@@ -0,0 +1,160 @@
...
+                             long cardinality) {
+        this.dvd = dvd;
+        this.nullCount = nullCount;
+        this.totalCount = totalCount;
+        if (totalCount < nullCount)
+            this.totalCount = nullCount;
+        this.cardinality = cardinality;
+        if (cardinality > this.totalCount - nullCount + 1)
+            this.cardinality = this.totalCount - nullCount + 1;
+        rpv = (long) (((double)(this.totalCount - nullCount))/cardinality);
```
msirek: Should catch a zero or negative cardinality and handle it somehow (use totalCount or 1 as the default?)

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public long rangeSelectivity(DataValueDescriptor start,DataValueDescriptor stop, boolean includeStart,boolean includeStop,boolean useExtrapolation) {
+        if ((start == null || start.isNull()) && (stop == null || stop.isNull()))
+            return nullCount;
+
+        if (start.equals(stop))
+            return rpv;
+
+        return (long)(0.1 * totalCount);
```
msirek: Is this a default estimation from some other location?  If so, maybe use a static constant value so if one place gets updated, so does this place.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
```diff
@@ -424,6 +435,124 @@ public static void DROP_TABLE_STATISTICS(String schema, String table) throws SQL
...
+
+            if (nullCountRatio < 0 || nullCountRatio > 1)
+                throw ErrorState.LANG_INVALID_FAKE_STATS.newException("column", "null count ratio should be in the range of [0,1]");
+
+            long nullCount = (long)(nullCountRatio * totalCount);
+
+            if (rpv > totalCount - nullCount)
+                throw ErrorState.LANG_INVALID_FAKE_STATS.newException("column", "rows per value should be larger than the total number of not-null count : " + (totalCount - nullCount));
+
+            long cardinality = (long)(((double)(totalCount- nullCount))/rpv);
```
msirek: Might want to check for zero or negative rpv here and throw a descriptive error message, like in the other cases above.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
```diff
@@ -424,6 +435,124 @@ public static void DROP_TABLE_STATISTICS(String schema, String table) throws SQL
...
+            tc.elevate("statistics");
+            dropTableStatistics(tds,dd,tc);
+            ddlNotification(tc, tds);
+
+            // compose the fake table stats row
+            ExecRow statsRow;
+            int statsType = SYSTABLESTATISTICSRowFactory.FAKE_MERGED_STATS;
+            long conglomerateId = tableDesc.getHeapConglomerateId();
+
+            statsRow = StatisticsAdmin.generateRowFromStats(conglomerateId, "-All-", rowCount, rowCount*meanRowWidth, meanRowWidth, numPartitions, statsType, 0.0d);
```
msirek: Do we maybe want sample fraction to be 1 instead of zero?  Otherwise, totalCount estimates might get adjusted down to zero in other locations?
`        if (isSampleStats)
            totalCount = (long)((double)totalCount/sampleFraction);`

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
yxia92: Done. I've added the following logic in StatisticsAdmin.FAKE_COLUMN_STATISTICS to check the eligible values for cardinality:
```
            if (rpv > totalCount - nullCount || rpv < 1)
                throw ErrorState.LANG_INVALID_FAKE_STATS.newException("column", "rows per value shouldn't be less than 1 or larger than the total number of not-null count : " + (totalCount - nullCount));

            long cardinality = (long)(((double)(totalCount- nullCount))/rpv);
```

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
yxia92: Done. I added DEFAULT_RANGE_SELECTIVITY and a few others in SelectivityUtil and make it a constant.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
yxia92: Done.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
yxia92: This is consistent with the setting in StatisticsAdmin.COLLECT_TABLE_STATISTICS(). As sample stats were added later, to be compatible with existing stats entries, the default value was set to be 0. So I prefer not to change the behavior in this JIRA. `sampleFraction` is only meaningful when `isSampleStats` is true.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        if (rpv == -1)
+            rpv = (long) (((double)notNullCount())/cardinality);
+
+        return rpv;
+    }
+
+    @Override
+    public long rangeSelectivity(DataValueDescriptor start,DataValueDescriptor stop, boolean includeStart,boolean includeStop,boolean useExtrapolation) {
+        if ((start == null || start.isNull()) && (stop == null || stop.isNull()))
+            return nullCount;
```
jyuanca: If start and stop keys  are null, should this return totoalCount?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceSystemProcedures.java
```diff
@@ -379,14 +379,40 @@ public SpliceSystemProcedures(DataDictionary dictionary) {
...
                             .ownerClass(StatisticsAdmin.class.getCanonicalName())
                             .build();
                     procedures.add(collectNonMergedSampleStatsForTable);
 
+                    Procedure fakeStatsForTable = Procedure.newBuilder().name("FAKE_TABLE_STATISTICS")
+                            .numOutputParams(0)
+                            .numResultSets(1)
+                            .modifiesSql()
+                            .varchar("schema",1024)
+                            .varchar("table",1024)
```
jyuanca: table and schema names are catalog type. See this example:
Procedure importWithBadRecords = Procedure.newBuilder().name("IMPORT_DATA")
                            .numOutputParams(0).numResultSets(1).ownerClass(HdfsImport.class.getCanonicalName())
                            **.catalog("schemaName")
                            .catalog("tableName")**
                            ...

yxia92: Thanks @jyuanca I've made the change as requested.

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
yxia92: If start and stop key are both null, this would be the scenario for predicate "col is null", that is why I used nullCount. Currently, is null takes another route to get the nullCount for selectivity directly, so it does not really exercise the above. The code above is just to make it consistent.

PR #3043, opened 2019-11-26
DB-8517 Add Maximum Concurrency parameter for Analyze Schema

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_access_api/src/main/java/com/splicemachine/access/configuration/StatsConfiguration.java
```diff
@@ -146,5 +151,6 @@ public void setDefaults(ConfigurationBuilder builder, ConfigurationSource config
 //        builder.fallbackCardinalityFraction = configurationSource.getDouble(FALLBACK_CARDINALITY_FRACTION, DEFAULT_FALLBACK_CARDINALITY_FRACTION);
 //        builder.fallbackIndexSelectivityFraction = configurationSource.getDouble(FALLBACK_INDEX_SELECTIVITY_FRACTION, DEFAULT_FALLBACK_INDEX_SELECTIVITY_FRACTION);
         builder.optimizerExtraQualifierMultiplier = configurationSource.getDouble(OPTIMIZER_EXTRA_QUALIFIER_MULTIPLIER, DEFAULT_OPTIMIZER_EXTRA_QUALIFIER_MULTIPLIER);
+        builder.collectSchemaStatisticsMaximumConcurrent = configurationSource.getInt(COLLECT_SCHEMA_STATISTICS_MAXIMUM_CONCURRENT, DEFAULT_COLLECT_SCHEMA_STATISTICS_MAXIMUM_CONCURRENT);
```
msirek: The user could break collect stats by using a negative number for this setting.  Or, getInt could return a NumberFormatException if a malicious user sets a value that can't be parsed as an int.  These cases should be caught, and we can use the default value when this happens.

arnaud-splice: Doesn't that apply to every single configuration parameter? I'd rather build something in the ConfigurationBuilder itself to restrict parameters to acceptable values. That'd be out of the scope of this PR though.

msirek: OK, we should create a Jira for it then.

arnaud-splice: I created DB-8942 to track that.

msirek: Thanks.  If we already know that analyzing a large number of tables is bad at the same time is bad, should be make the default of COLLECT_SCHEMA_STATISTICS_MAXIMUM_CONCURRENT be a smaller value than Integer.MAX_VALUE ?

arnaud-splice: That's a good point, but I quite frankly don't know what kind of value would be appropriate. We would not want to take small machines as a common denominator. What would be a good value?

msirek: Just imagine what we think would be the maximum number of simultaneous requests we could expect a Splice system to be able to handle, without excessively slowing down.  I don't know if we've ever tested this.  I would think 1000 might be at the upper end of reasonable, but still be better than Integer.MAX_VALUE.  My guess is that 100 might be a good default value, even on large clusters.  We could error on the side of having the number be larger than we think the best number is, just so we don't artificially limit collect statistics performance.  Do you know if analyze could execute in control, or do we try to always execute it in spark?

dgomezferro: I agree 100 sounds reasonable, we could always increase it if customers need more parallelism.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
```diff
@@ -837,13 +826,41 @@ public static ExecRow generateOutputRow(String schemaName, String tableName, Exe
...
+        Iterable<StatisticsOperation> movingExecutionWindow = () -> new Iterator<StatisticsOperation>() {
+            int i = 0;
+            @Override
+            public boolean hasNext() {
+                return i < collectOps.size();
+            }
+
+            @Override
+            public StatisticsOperation next() {
+                if ((long)i + (long)maximumConcurrent < (long)collectOps.size()) {
```
dgomezferro: Why are we casting to long? If it's to avoid integer overflow we'd have it anyway on collectOps.get(i + maximumConcurrent)

dgomezferro:
Just small requests: set a reasonable default and remove unnecessary casts.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
arnaud-splice: I disagree, collectOps.size() returns an int, so if we go in the loop, we know that i+maximumConcurrent does not overflow.
If i+maximumConcurrent were to overflow line 849, casting would prevent the overflow and we would not enter the if.

arnaud-splice: Oh, and by the way, if someone sets maximumConcurrent to MAXINT, we **will** overflow.

splice_access_api/src/main/java/com/splicemachine/access/configuration/StatsConfiguration.java
arnaud-splice: Done.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
dgomezferro: I see, that makes sense, thanks!

PR #3049, opened 2019-11-28
DB-8915 Fix Kerberos on Olap Server

dgomezferro:


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerMaster.java
```diff
@@ -96,6 +97,11 @@ private void run() throws Exception {
         // Initialize clients to ResourceManager and NodeManagers
         Configuration conf = HConfiguration.unwrapDelegate();
 
+        UserGroupInformation.isSecurityEnabled();
```
arnaud-splice: Why check if security is enabled but not look at the result?

dgomezferro: That method has some side effects, that's what we care about.

PR #3056, opened 2019-11-29
DB-8942 Allow restricting config parameter to a certain range (2.7)

arnaud-splice:


arnaud-splice:
We probably do not want to bring this to 2.7 @carolp-503 what do you think?

carolp-503:
> We probably do not want to bring this to 2.7 @carolp-503 what do you think?

At this point, we're  strongly restricting what goes into 2.7, so no, we shouldn't bring this to 2.7 unless there's a strong customer demand or data integrity issue.

PR #3073, opened 2019-12-10
DB-8933 Support Syntax OPTIMIZE FOR n ROWS

changli6:
Note:
DB2 supports the syntax for ROW and ROWS both and does not care n (n > 0) is singular or plural.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -2391,6 +2391,31 @@ public class SQLParser
            return buf.toString();
    }
 
+    /**
+     * Determine if we are seeing an Optimize For Rows clause or the identifier OPTIMIZE
+     *
+     * @return true if it is an Optimize For Rows clause.
+     */
+    private boolean seeingOptimizeForRowsClause()
+    {
+        int nesting = 0;
+
+        // Token number, i == 1: OPTIMIZE
+        int i = 2;
+
+        int tokKind = getToken(i).kind;
+
+        if (tokKind == FOR) {
+            tokKind = getToken(++i).kind;
+            if (tokKind == EXACT_NUMERIC) {
+                tokKind = getToken(++i).kind;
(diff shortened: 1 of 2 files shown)


changli6:

spliceci:
TEST SUCCEEDED +1

carolp-503:
Can you submit PRs for 2.8 and 3.0 for this change?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -3004,6 +3004,7 @@ TOKEN [IGNORE_CASE] :
 |	<PARQUET: "parquet">
 |   <AVRO: "avro">
 |	<ORC: "orc">
+|   <OPTIMIZE: "optimize">
```
msirek: Is there a need to make this a reserved keyword instead of a non-reserved keyword?  If by any chance a customer is using optimize as a column name or table name this change would break use of that object.

changli6:
jenkins please test branch @mem

spliceci:
TEST SUCCEEDED +1

changli6:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
changli6: Set OPTIMIZE to the non-reserved keyword. Thanks @msirek pointed out the potential upgrade issue.

spliceci:
TEST FAILED

changli6:

spliceci:
TEST SUCCEEDED +1

changli6:
> Looks good. I'm OK with the reserved keyword approach if using a nonreserved keyword proves problematic. We can always change it to unreserved later if we have a customer issue.

@msirek The latest commit set OPTIMIZE as the non-reserved keyword with a few revised on parsing logic of OPTIMIZE FOR ROWS.  It avoids the defect if customers using OPTIMIZE as table / column name in case.

PR #3078, opened 2019-12-10
DB-8980 Unit test SICompactionState.mutate

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

pom.xml
```diff
@@ -121,6 +121,12 @@
             <version>1.9.5</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+            <version>1.3</version>
+            <scope>test</scope>
+        </dependency>
```
OlegMazurov: This is a duplicate and most likely redundant dependency at the topmost level.

OlegMazurov: If you want _hamcrest-library_ to be available to all modules for unit tests, this is the right location to specify the dependency (along with junit itself). _hamcrest-core_ should also be moved here then.
The _dependencyManagement_ section is good when we want to specify a dependency selectively in a subset of modules but manage the attributes of that dependency (version, scope) in one place.
If _hamcrest_ dependencies are specified as global (spliceengine-parent dependencies) there is no need to add them to the _dependencyManagement_ section.

arnaud-splice:

pom.xml
arnaud-splice: I removed the hamcrest reference in _dependencyManagement_ and added hamcrest (without -core or -library) along with junit.

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -0,0 +1,125 @@
...
+    private final boolean purgeDeletedRows;
+    private long maxTombstoneTimestamp;
+
+    SICompactionStateMutate(boolean purgeDeletedRows) {
+        this.purgeDeletedRows = purgeDeletedRows;
+        this.dataToReturn = new TreeSet<>(KeyValue.COMPARATOR);
+        this.maxTombstoneTimestamp = 0;
+    }
+
+    private boolean isSorted(List<Cell> list) {
```
dgomezferro: You could use Guava here: `Ordering.from(KeyValue.COMPARATOR).isOrdered(list);`

hbase_storage/src/test/java/com/splicemachine/si/impl/server/SICompactionStateMutateTest.java
```diff
@@ -0,0 +1,251 @@
...
+        assertThat(outputCells, equalTo(Arrays.asList(
+                SITestUtils.getMockCommitCell(0x300),
+                SITestUtils.getMockAntiTombstoneCell(0x300),
+                SITestUtils.getMockValueCell(0x300))));
+    }
+
+    @Test
+    public void committed() {
+        TxnView committedWithRootParent = TxnTestUtils.getMockCommittedTxn(0x100, 0x200);
+        assertTrue(SICompactionStateMutate.committed(committedWithRootParent));
```
dgomezferro: Can you rename `SICompactionStateMutate.committed` to `isCommitted` ? I think that would be clearer.

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
arnaud-splice: I wanted to, but I could not make it work without adding a new dependency and was wondering if it was worth it.

hbase_storage/src/test/java/com/splicemachine/si/impl/server/SICompactionStateMutateTest.java
arnaud-splice: done

PR #3086, opened 2019-12-12
DB-8793 Remove deleted rows during compaction

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/hbase/TransactionsWatcher.java
```diff
@@ -34,17 +34,23 @@
...
+    private static AtomicLong lowWatermarkTransaction = new AtomicLong(Long.MAX_VALUE);
 
     public static TransactionsWatcher INSTANCE = new TransactionsWatcher();
 
     private static ScheduledExecutorService updateService =
             MoreExecutors.namedSingleThreadScheduledExecutor("hbase-transactions-watcher-%d");
 
     private static final Runnable updater = () -> {
-        oldestActiveTransaction.set(fetchOldestActiveTransaction());
+        long fetchTimestamp = SIDriver.driver().getTimestampSource().nextTimestamp();
```
jyuanca: This will increase timestamp. Does retriveTimestamp work?

jyuanca:
Please resolve conflicts

hbase_storage/src/main/java/com/splicemachine/hbase/TransactionsWatcher.java
arnaud-splice: I'm not quite sure that retrieveTimestamp works here, especially since it does not seem to using an atomic operation underneath. Unless this is blocking, I would address that in a separate change.

dgomezferro:
@arnaud-splice this is included in DB-9001 right? Should we close it?

PR #3101, opened 2019-12-23
DB-9024 Subquery with IN or NOT IN hits AnalysisException when native spark execution is used.

msirek:
The Derby optimizer projects inner table columns of semijoin or antijoin sometimes.
When using native Spark execution, the inner table columns are not included in the data frame, so attempting to project out nonexistent inner table columns results in an AnalysisException.
The fix is to modify the join data frame to add nulls for the inner table columns when we have a subquery.  

See description in [DB-9024](https://splicemachine.atlassian.net/browse/DB-9024).

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
@@ -851,8 +851,14 @@ public class NativeSparkDataSet<V> implements DataSet<V> {
     // ExecRowDefinition.
     private Dataset<Row> fixupColumnNames(JoinOperation op, JoinType joinType,
                                           Dataset<Row> leftDF, Dataset<Row> rightDF,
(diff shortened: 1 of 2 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -851,8 +851,14 @@ public static boolean joinProjectsOnlyLeftTableColumns(JoinType joinType) {
     // ExecRowDefinition.
     private Dataset<Row> fixupColumnNames(JoinOperation op, JoinType joinType,
                                           Dataset<Row> leftDF, Dataset<Row> rightDF,
-                                          Dataset<Row> joinedDF) {
+                                          Dataset<Row> joinedDF,
+                                          SpliceOperation leftOp,
```
changli6: The leftOp is not used?

PR #3120, opened 2020-01-07
DB-9035 Let SQLVarchar share SQLChar's hashCode() function

yxia92:


yxia92:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLVarchar.java
```diff
@@ -416,25 +416,7 @@ public BooleanDataValue greaterOrEquals(DataValueDescriptor left,
...
-				return 0;
-			} else {
-				return str.hashCode();
-			}
-		}
-		catch (StandardException se)
-		{
-			throw new RuntimeException(se);
-		}
+		return super.hashCode();
```
arnaud-splice: Why override it at all and not let it be inherited?

changli6:
@yxia92 There are conflicts on merging, could you resolve it?

PR #3129, opened 2020-01-10
DB-8915 Fix Kerberos on Olap Server

dgomezferro:
Logic stays the same for non kerberized clusters, for kerberized clusters we recover the old logic (two different users, one to interact with Yarn and another to submit the Spark app)


dgomezferro:

spliceci:
TEST SUCCEEDED +1

PR #3149, opened 2020-01-19
SPLICE-1806 NullPointerException in PipingCallBuffer (3.0)

OlegMazurov:
Extensively tested in bespoke/AWS with TPC-C. Runs are much cleaner and there is also some noticeable performance improvement (up to 10%; tpmC ~= 32K in a 4 node cluster).

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/storage/ClientPartition.java
+++ b/hbase_storage/src/main/java/com/splicemachine/storage/ClientPartition.java
@@ -14,8 +14,6 @@
 
 package com.splicemachine.storage;
 
-import com.splicemachine.si.api.data.TxnOperationFactory;
-import com.splicemachine.si.impl.driver.SIDriver;
 import org.spark_project.guava.base.Function;
 import com.splicemachine.si.data.HExceptionFactory;
 import com.splicemachine.si.impl.HNotServingRegion;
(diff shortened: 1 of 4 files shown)


hbase_storage/src/main/java/com/splicemachine/storage/ClientPartition.java
```diff
@@ -155,32 +152,26 @@ public boolean checkAndPut(byte[] key,byte[] family,byte[] qualifier,byte[] expe
...
             }
-            partitions = formatPartitions(getAllRegionLocations(true));
-            partitionInfoCache.invalidate(tableName);
-            assert partitions!=null:"partitions are null";
-            partitionInfoCache.put(tableName,partitions);
+            List<HRegionLocation> tableLocations = ((ClusterConnection) connection).locateRegions(tableName, !refresh, false);
+            partitions = new ArrayList<>(tableLocations.size());
+            for (HRegionLocation location : tableLocations){
+                if (location.getServerName() == null) {     // ensure no offline regions
+                    continue;
```
msirek: Is it OK to ignore offline regions?  The old code kept retrying until all regions were online.  Could we potentially end up not scanning all rows in the table?

OlegMazurov: _HRegionLocation.getServerName() == null_ causes the NPE. I believe it's an issue in HBase itself as serverName is used in HRegionLocation.hashCode() and throws NPE when null (making such HRegionLocation instances poorly defined and unusable).
In fact, _ClusterConnection.locateRegions(..., false)_ already ensures that such locations are filtered out. I make that condition explicit..

msirek: I see, thanks.

PR #3162, opened 2020-01-27
DB-9012 Flatten half outer joins(3.0)

yxia92:


assembly/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>spliceengine-parent</artifactId>
         <groupId>com.splicemachine</groupId>
-        <version>3.0.0.1944-SNAPSHOT</version>
+        <version>3.0.0.1945-SNAPSHOT</version>
```
carolp-503: Not sure why these are a part of this PR.  Is this an intentional change?

hbase_storage/src/main/java/com/splicemachine/si/impl/CoprocessorTxnStore.java
```diff
@@ -395,12 +397,21 @@ private TxnView decode(long queryId, TxnMessage.Txn message) throws IOException{
             // we didn't find it
             if (ignoreMissingTransactions) {
                 // return committed mock transaction
-                return new InheritingTxnView(Txn.ROOT_TRANSACTION,queryId,queryId,
+                return new InheritingTxnView(Txn.ROOT_TRANSACTION, queryId, queryId,
```
carolp-503: Another set of changes that I don't recognize from the 2.8 and Master versions of this PR. Are these supposed to be here in 3.0?

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/CaseExpressionIT.java
```diff
@@ -0,0 +1,116 @@
...
+import com.splicemachine.homeless.TestUtils;
+import org.junit.*;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+
+public class CaseExpressionIT extends SpliceUnitTest {
```
carolp-503: Another set of changes that don't seem to be related to DB-9012

assembly/pom.xml
yxia92: Thanks @carolp-503 I don't know how these changes get into this PR. To make it simple, I'm going to close this PR and open a new one.

PR #3189, opened 2020-02-04
DB-8883 Fix IT failures on mem platform.

msirek:
Reverts a couple last-minute changes to DB-8883 which cause ITs to fail in mem.
The changes were:
1. Addition of synchronization in ContextManager.
2. Sharing the same session number between all LanguageConnectionContexts used by parallel triggers (so the logs would nicely show all triggers for a given SQL fired in the same "session".


msirek:

spliceci:
TEST SUCCEEDED +1

PR #3197, opened 2020-02-05
DB-9139 Print reason to choose engine in explain

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/impl/ast/PlanPrinter.java
```diff
@@ -93,10 +93,10 @@ public Visitable defaultVisit(Visitable node) throws StandardException {
             m.put(query,orderedNodes);
             if (LOG.isDebugEnabled()){
                 DataSetProcessorType currentType = rsn.getLanguageConnectionContext().getDataSetProcessorType();
-                currentType = currentType.combine(queryHintedForcedType(orderedNodes));
-                boolean useSpark = currentType.isSpark() || currentType.isDefaultControl() && PlanPrinter.shouldUseSpark(orderedNodes, true);
+                currentType = currentType.combine(rsn.getCompilerContext().getDataSetProcessorType());
+                currentType = currentType.combine(determineSpark(orderedNodes));
```
yxia92: Do we really this line? I think the dataSetProcessorType obtained from rsn.getCompilerContext().getDataSetProcessorType() is the final decision, and also the one that is passed to activation, so we don't need to go through every node in orderedNodes.

spliceci:
TEST FAILED

db-engine/src/main/java/com/splicemachine/db/impl/ast/PlanPrinter.java
arnaud-splice: You're right, I merely refactored that function while keeping the behaviour intact, but we don't need that anymore. We also don't need the languagecontext.type as this one was put in compilercontext.type during the optimisation phase.

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3198, opened 2020-02-05
DB-9057 Separate Control and Spark costing (2.8)

arnaud-splice:


db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/DataSetProcessorType.java
```diff
@@ -0,0 +1,105 @@
+package com.splicemachine.db.iapi.sql.compile;
```
carolp-503: Not a blocker, but shouldn't this file have a copyright?

PR #3200, opened 2020-02-05
DB-8930: retry split until the region is online(3.0)

jyuanca:


jyuanca:
So far this issue only happened on hbase 2.x, so fix branch-3.0

PR #3205, opened 2020-02-05
DB-9054: retrieve replication latency

jyuanca:


jyuanca:
Please also review ee branch

splice_machine/src/main/java/com/splicemachine/replication/ReplicationSystemProcedure.java
```diff
@@ -61,6 +62,19 @@
...
+    public static void GET_REPLICATION_PROGRESS(ResultSet[] resultSets) throws StandardException, SQLException {
+        ReplicationManager replicationManager = EngineDriver.driver().manager().getReplicationManager();
+        long ts = replicationManager.getReplicationProgress();
+        if (ts > 0) {
+            DateTime dateTime = new DateTime(ts);
+            resultSets[0] = ProcedureUtils.generateResult(
+                    "Replication progress", dateTime.toString());
+        }
+        else {
+            resultSets[0] = ProcedureUtils.generateResult("Error", "Information for available");
```
yxia92: typo in the result string, I guess it should be "Information not available".

splice_machine/src/main/java/com/splicemachine/replication/ReplicationSystemProcedure.java
```diff
@@ -61,6 +62,19 @@
 
     private static Logger LOG = Logger.getLogger(ReplicationSystemProcedure.class);
 
+    public static void GET_REPLICATION_PROGRESS(ResultSet[] resultSets) throws StandardException, SQLException {
+        ReplicationManager replicationManager = EngineDriver.driver().manager().getReplicationManager();
+        long ts = replicationManager.getReplicationProgress();
```
yxia92: It seems that we assume only one replication slave cluster. Can this logic be extended in the future if we have  multiple slave clusters?

hbase_sql/src/main/java/com/splicemachine/hbase/SpliceReplicationSinkChore.java
```diff
@@ -260,19 +264,29 @@ private void updateProgress() throws IOException, KeeperException, InterruptedEx
...
-                        long appliedPosition = replicationProgress.get(walName);
+                    byte[] colName = CellUtil.cloneQualifier(cell);
+                    if(Arrays.equals(colName, HBaseConfiguration.REPLICATION_SNAPSHOT_TSCOL_BYTES)){
+                        ts = Bytes.toLong(CellUtil.cloneValue(cell));
                         //if (LOG.isDebugEnabled()) {
+                        SpliceLogUtils.info(LOG, "Process snapshot take at %s", new DateTime(ts).toString());
+                        //}
+                    }
+                    else {
+                        String walName = Bytes.toString(CellUtil.cloneQualifier(cell));
```
yxia92: The call to CellUtil.cloneQualifier(cell) here seems redundant, should we just use colName?

splice_machine/src/main/java/com/splicemachine/replication/ReplicationSystemProcedure.java
jyuanca: This system procedure is run on a slave cluster

yxia92: I see. Thanks!

platform_it/src/test/java/com/splicemachine/test/SpliceTestPlatformConfig.java
```diff
@@ -321,8 +321,8 @@ public static Configuration create(String hbaseRootDirUri,
         // Replication
         //
         config.setBoolean("replication.source.eof.autorecovery", true);
-        config.set("hbase.replication.source.service", "com.splicemachine.replication.SpliceReplication");
-        config.set("hbase.replication.sink.service", "com.splicemachine.replication.SpliceReplication");
+        //config.set("hbase.replication.source.service", "com.splicemachine.replication.SpliceReplication");
+        //config.set("hbase.replication.sink.service", "com.splicemachine.replication.SpliceReplication");
```
arnaud-splice: Why is that necessary?

jyuanca: Comment them out so that splice can be started w/o building ee branch, because SpliceReplication is only available on ee branch. Without commenting them out, the following command will fail
./start-splice-cluster -pcdh6.3.0

You have to always build with ee branch
start-splice-cluster -pcdh6.3.0,ee

PR #3209, opened 2020-02-06
DB-9058 (DB-9144 Subselect without alias + DB-9143 right of union derives name from left)

arnaud-splice:


genedavis:
For completeness, can we have both UNION and UNION ALL IT test cases?

dgomezferro:
Should this be extended to other set operations like EXCEPT and INTERSECT?

arnaud-splice:
> For completeness, can we have both UNION and UNION ALL IT test cases?

Done

arnaud-splice:
> Should this be extended to other set operations like EXCEPT and INTERSECT?

Good point. I would do it in a separate change once build servers are back online though. Unless we think it's blocking.

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3210, opened 2020-02-06
DB-9116 Handle NPE in SYSCS_PERFORM_MAJOR_COMPACTION_ON_SCHEMA() (3.0)

yxia92:


hbase_storage/src/main/java/com/splicemachine/storage/ClientPartition.java
```diff
@@ -247,7 +247,8 @@ public void compact(boolean isMajor) throws IOException{
                     IOException ioe = exceptionFactory.processRemoteException(e);
                     if (ioe instanceof HNotServingRegion ||
                         ioe instanceof HWrongRegion ||
-                        ioe instanceof HRegionTooBusy) {
+                        ioe instanceof HRegionTooBusy ||
+                        e instanceof NullPointerException) {
```
yxia92: Since we use `e` here, we should pass `e` also in the debug message in the next line if it is an instace of NullPointerException.

PR #3214, opened 2020-02-10
DB-9140 Support tuples in predicates

arnaud-splice:


splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/PredicateTupleIT.java
```diff
@@ -0,0 +1,170 @@
...
+                " 1 |10 |100 |";
+
+        testQuery(query, expected);
+    }
+
+    @Ignore("Find a way to pass nulls inside a tuple")
+    @Test
+    public void testTupleWithNulls() throws Exception {
+        String query = format("select count(*) from D --splice-properties useSpark=%s\n" +
+                "where (d1,d2,d3) = (null,null,null)", useSpark);
```
yxia92: You may use the expression `case when 1=0 then 1 end` to generate a null value.

arnaud-splice: Removing this test as it does not make much sense to compare to null

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -6821,13 +6890,14 @@ multiplicativeOperator() throws StandardException :
 /*
  * <A NAME="unaryExpression">unaryExpression</A>
  */
-ValueNode
```
dgomezferro: I don't like changing all these return types from ValueNode to QueryTreeNode, I feel like a tuple should also be a ValueNode, either by making a ValueNodeList extend a ValueNode or by introducing another node type for tuples.

arnaud-splice:
> It would also be good to support queries such as
> `select * from a where (i, j) in (select j, i from a);`
> and
> `select * from a where (i, j) = (select j, i from a);`
> Not sure how much effort would that be though.
Created by DB-9169


> There's also a problem with inequality comparisons, tuples are compared lexicographically ( compare first element, if they are equal second element...) however this PR compares them all at the same time, the following query should produce 1 row:
> 
> ```
> create table a ( i int, j int);
> insert into a values (1,2), (2,1);
> select * from a where (i, j) < (j, i);
> ```
Since the purpose of this PR was only to support = and !=, I will remove lt/le/gt/ge comparison from this PR and track it with DB-9170

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
arnaud-splice: Introduced ValueTupleNode that derives ValueNode

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -8003,6 +8063,35 @@ valueExpressionPrimary() throws StandardException :
...
+    {
+        if (tupleOrFirstValue instanceof ValueTupleNode) {
+            valueTuple = (ValueTupleNode) tupleOrFirstValue;
+        } else {
+            valueTuple = (ValueTupleNode) nodeFactory.getNode(
+                                    C_NodeTypes.VALUE_TUPLE_NODE,
+                                    getContextManager());
+            valueTuple.addValueNode((ValueNode) tupleOrFirstValue);
+        }
+        newValue = valueExpression();
```
yxia92: @arnaud-splice Can you move the call of valueExpression() up. valueExpression() is part of the parser rule, although eventually it will be compiled into a function of the same name, I think it is better to keep it as a parser rule and put explicitly at the top as shown below, following the tradition of all other parser rules:
```
ValueTupleNode
valueTupleElement(ValueNode tupleOrFirstValue) throws StandardException:
{
    ValueTupleNode  valueTuple;
    ValueNode       newValue;
}
{
    newValue = valueExpression()
    {
        if (tupleOrFirstValue instanceof ValueTupleNode) {
            valueTuple = (ValueTupleNode) tupleOrFirstValue;
        } else {
            valueTuple = (ValueTupleNode) nodeFactory.getNode(
                                    C_NodeTypes.VALUE_TUPLE_NODE,
                                    getContextManager());
            valueTuple.addValueNode((ValueNode) tupleOrFirstValue);
        }

        if (newValue instanceof ValueTupleNode) {
            throw StandardException.newException(SQLState.LANG_SYNTAX_ERROR,
                "Nested tuple not supported");
        }
        valueTuple.addValueNode(newValue);
        return valueTuple;
    }
}
```

arnaud-splice: Sure, that's a good idea!

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3243, opened 2020-02-21
DB-9124 Large concat expression cause long time hanging

changli6:


changli6:

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceStringFunctionsIT.java
```diff
@@ -713,6 +713,27 @@ public void testHEX() throws Exception {
         rs.close();
     }
 
+    @Test
+    public void testLongConcatenateString() throws Exception {
+        String sqlText = "SELECT COUNT(*) FROM (" + "SELECT CAST(CAST(c AS DECIMAL(9,4)) AS CHAR(50))||','"
```
dgomezferro: It seems this query used to work (slowly) before the fix, right? In that case we'd need to add a timeout for instance, to make sure the test fails if there's a regression. Or make the query bigger so that it crashes without the fix.

changli6: Agreed. I set the test case failed after 1 second. Without the fix, it took more than 10 seconds.

changli6:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST FAILED

changli6:

spliceci:
TEST SUCCEEDED +1

PR #3247, opened 2020-02-24
DB-9176 Implement nested ContextManagers + DB-8883

OlegMazurov:
ContextManagers can now be nested: if a ContextManager is created with a parent ContextManager, all pushContext/popContext operations will still be performed on the current ContextManager but getContext will delegate to parent ContextManager (potentially owned by another thread but assumed unmodified until all parallel operations finish).
SpliceTransactionManager.startNestedUserTransaction() is the only place where the new mechanism is used (ensuring that parallel trigger execution introduced by DB-8883 is thread safe).


OlegMazurov:

spliceci:
TEST SUCCEEDED +1

OlegMazurov:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerEventActivator.java
```diff
@@ -31,49 +31,78 @@
...
+    // at the time they were pushed.  Saving the original "LCC" at the time of
+    // the push ensures that we pop from the correct LCC.  We sometimes need to
+    // save two versions LCCs because we may push to the LCC in the activation,
+    // and also to the LCC stored in the current ContextManager, if they happen
+    // to be different.  Trigger execution requires that both the activation and
+    // the current ContextManager both have a triggerExecutionContext.
+    private LanguageConnectionContext esvLCC1;
+    private LanguageConnectionContext esvLCC2;
+    private LanguageConnectionContext tecLCC1;
+    private LanguageConnectionContext tecLCC2;
```
dgomezferro: @OlegMazurov Do you think there's a chance of simplifying this now that we support nested context managers? Here's Mark's comment from the previous PR:

Yes it is messy. The ContextManager doesn't seem well designed for threading. The ExecutionStmtValidator is required to detect if we're doing DDL on a trigger table. The TriggerExecutionContext is required to be in both the LCC of the Activation and of the active ContextManager for proper trigger execution. A lot of the time the Activation and the ContextManager share the same LCC, which is what we want. But sometimes the ContextManager's LCC is pushed in someplace unexpected and we end up with a mismatch. I think the whole ContextManager logic should probably be revisited and maybe rewritten. I had tried some things to add more synchronization or use different data structures, e.g. ConcurrentHashMap in place of HashMap, but that either caused the database to run too slowly or caused hangs or other failures.

OlegMazurov: I have filed _DB-9282: Rework DB-8883 to rely on nested ContextManager's_ to deal with this separately.

PR #3253, opened 2020-02-27
DB-9177 Use Guava's cache instead of LongKeyedCache for active txns

dgomezferro:


dgomezferro:

spliceci:
TEST FAILED

dgomezferro:

dgomezferro:

spliceci:
TEST SUCCEEDED +1

splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveTxnCacheSupplier.java
```diff
@@ -33,16 +33,15 @@
...
-        this(delegate, maxSize, false);
+    public ActiveTxnCacheSupplier(TxnSupplier delegate, int initialSize, int maxSize) {
+        this(delegate, initialSize, maxSize, false);
     }
 
-    public ActiveTxnCacheSupplier(TxnSupplier delegate, int maxSize, boolean threadSafe) {
-        this.cache = LongKeyedCache.<TxnView>newBuilder().maxEntries(maxSize)
-                .withHashFunction(HashFunctions.murmur3(0)).threadSafe(threadSafe).build();
+    public ActiveTxnCacheSupplier(TxnSupplier delegate, int initialSize, int maxSize, boolean threadSafe) {
+        this.cache = CacheBuilder.newBuilder().initialCapacity(initialSize).maximumSize(maxSize).concurrencyLevel(threadSafe ? 32 : 1).build();
```
arnaud-splice: Where does 32 come from?

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -184,7 +184,7 @@ private MutationStatus getCorrectStatus(MutationStatus status,MutationStatus old
             constraintState = new SimpleTxnFilter(null, txn, NoOpReadResolver.INSTANCE, txnSupplier);
             supplier = constraintState.getTxnSupplier();
         } else {
-            supplier = new ActiveTxnCacheSupplier(txnSupplier, 100);
+            supplier = new ActiveTxnCacheSupplier(txnSupplier, 100, 4000);
```
arnaud-splice: Why 4000?

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
```diff
@@ -89,13 +89,18 @@ public SimpleTxnFilter(String tableName,
...
         assert readResolver!=null;
-        this.transactionStore = new ActiveTxnCacheSupplier(baseSupplier, 100); //TODO -sf- configure
         this.myTxn=myTxn;
         this.readResolver=readResolver;
         this.ignoreNewerTransactions = ignoreNewerTransactions;
         SIDriver driver = SIDriver.driver();
-        if (driver != null) {
+        if (driver == null) {
+            // only happens during testing
+            this.transactionStore = new ActiveTxnCacheSupplier(baseSupplier, 128, 2048);
```
arnaud-splice: If this only happens during testing, do we have a way to throw an exception if we're not in test mode?

splice_si_api/src/main/java/com/splicemachine/si/impl/store/ActiveTxnCacheSupplier.java
dgomezferro: It was arbitrary, I substituted it for Guava's default of 4 with an explicitly defined constant.

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
dgomezferro: It was inline with the defaults of 128/4096 , now I get them from the Config if it's available or set it to the defaults if not.

splice_si_api/src/main/java/com/splicemachine/si/impl/SimpleTxnFilter.java
dgomezferro: I couldn't find a good way to do that.

PR #3256, opened 2020-02-27
DB-9198 Add a cast node for CHAR(n) CHAR(m) comparison on Spark

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

dgomezferro:
@arnaud-splice does this only happen when running in "native" spark ? Is it possible to add the cast node at the spark level instead of adding it to our Splice plan? We already deal with datatype conversions and other stuff that way.

PR #3289, opened 2020-03-04
DB-9228 Fix MapR executor configuration

bklo94:
One line change to switch the order of how the jars are located for MapR 6.1 executors.


assembly/mapr/src/main/resources/scripts/splice/hbase-env.sh
```diff
@@ -93,7 +93,7 @@ SPLICE_HBASE_MASTER_OPTS="$SPLICE_HBASE_MASTER_OPTS -Dsplice.spark.driver.extraL
 SPLICE_HBASE_MASTER_OPTS="$SPLICE_HBASE_MASTER_OPTS -Dsplice.spark.driver.extraClassPath=/opt/mapr/hbase/hbase1.1.8-splice/conf:/opt/mapr/hbase/hbase1.1.8-splice/lib/*:/opt/splice/default/lib/*"
```
changli6: Do we need to modify the driver classPath also?

OlegMazurov: Yes, we do

PR #3298, opened 2020-03-05
DB-9028 Fix ClassCastException when query contains round()

changli6:


changli6:
jenkins please teste branch @cdh5.12.2

changli6:

spliceci:
TEST FAILED

changli6:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -968,7 +968,7 @@ else if (operand instanceof ConditionalNode) {
                             methodName.equals("ADD_MONTHS")    ||
                             methodName.equals("ADD_YEARS")     ||
                             methodName.equals("ADD_DAYS")      ||
-                            methodName.equals("ROUND"))
+                            methodName.equalsIgnoreCase("ROUND"))
```
arnaud-splice: Shouldn't that apply to every other item in this list?

changli6: Won't need to do that. Other functions were "made in splice" and their method names were uppercase. The round function came from Splice package (uppercase) and java.lang.StrictMath (lowercase), we need to handle this one especially.

arnaud-splice: All right! Could you add a comment in the code to explain that?

changli6: Done

PR #3306, opened 2020-03-09
DB-8848: fix flatten.sh for cdh6.3.0

jyuanca:


assembly/cdh6.3.0/src/main/resources/scripts/flatten.sh
```diff
@@ -4,7 +4,7 @@ cm_host="localhost"
 cm_port="7180"
 cm_user="admin"
 cm_pass="admin"
-cm_api_ver="api/v9"
+cm_api_ver="api/v32"
```
yxia92: Should the version be v33 according to https://archive.cloudera.com/cm6/6.3.0/generic/jar/cm_api/swagger-html-sdk-docs/python/README.html ?

jyuanca: Good catch! Thanks @yxia92

PR #3308, opened 2020-03-11
DB-8836: truncate date for import_data

jyuanca:


jyuanca:

spliceci:
TEST FAILED

jyuanca:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/stream/function/AbstractFileFunction.java
```diff
@@ -268,6 +270,16 @@ else if (values.size() < returnRow.nColumns()) {
         }
     }
 
+    private static String truncateDateIfNecessary(String value) {
+        if (value != null) {
+            String[] s = value.split(" ");
```
dgomezferro: There are multiple date formats that use spaces as separators, this would break them. I think we need an alternative approach here
https://en.wikipedia.org/wiki/Date_format_by_country

We shouldn't truncate it if we have an explicit date formatter, but even if we don't I think we can't assume a space is not a separator, we make an effort to try to parse Localised dates in SQLDate.parseDate() when we don't have an explicit formatter.

PR #3331, opened 2020-03-20
DB-9291 Switch spotbugs to cdh6.3.0 on master

arnaud-splice:


arnaud-splice:
> This commit is fine. While I saw there were some sub-objects in the ticket, e.g. A-INT to a K8s internal cluster. Are there following commits, or separate PRs?

The title of the PR was wrong. This is about DB-9291 and not DB-9251. Sorry about the confusion

PR #3332, opened 2020-03-20
DB-8949 Decouple OlapServer from HMaster

dgomezferro:
Allowing for Kubernetes-based deployment

Also:
- Retry Olap Server submission if failure is a ConnectionException 
- Add diagnostics if region server can't contact OlapServer (both for a Yarn or Spark failure)


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerZNode.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public String getQueueName() {
+        return queueName;
+    }
+
+    public long getSequence() {
+        return sequence;
+    }
+
+    public String toZNode() {
```
martinrupp, line 22: looks good, although i don't understand everything. are server names always of the form xxxxNUM, where NUM is 10 digits?

dgomezferro: To avoid duplicates each OlapServer registers itself in ZooKeeper using a sequence node, which appends a unique identifier at the end to guarantee uniqueness.

PR #3334, opened 2020-03-23
DB-9259 Fix spotbugs errors in DataDictionaryImpl.java (2.8)

arnaud-splice:
See also spliceengine-ee

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
@@ -5094,7 +5094,7 @@ public abstract class DataDictionaryImpl extends BaseDataDictionary{
                         bArray[2]=true;
                         break;
                     default:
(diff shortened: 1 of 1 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -5107,6 +5093,8 @@ public void updateConstraintDescriptor(ConstraintDescriptor cd,
                     case SYSCONSTRAINTSRowFactory.SYSCONSTRAINTS_TABLEID:
                         bArray[2]=true;
                         break;
+                    default:
+                        assert false: "Unexpected colToSet: " + colToSet;
```
changli6: There are other possible vaule:
```
In SpliceDataDictionary.java 
 protected void addSubKeyConstraint(
colsToSet[0]=SYSCONSTRAINTSRowFactory.SYSCONSTRAINTS_REFERENCECOUNT
```

PR #3338, opened 2020-03-24
DB-9283 Join should not be considered between a query and its non fla…

arnaud-splice:
…ttened subquery


arnaud-splice:
(Still in progress, some tests fail)

PR #3341, opened 2020-03-25
DB-9309 Reduce memory pressure on RegionServers

dgomezferro:


start-splice-cluster
```diff
@@ -48,7 +48,7 @@ wait_for () {
 }
 
 BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-DEFAULT_PROFILE="hdp3.1.0"  # default hbase platform profile
+DEFAULT_PROFILE="cdh6.3.0"  # default hbase platform profile
```
arnaud-splice: Why?

dgomezferro: We use cdh more often and it's a pain to switch profiles back and forth in intellij, if people don't mind I'd rather have it be the default profile.

dgomezferro:

hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
```diff
@@ -75,13 +78,15 @@ public boolean next(List<Cell> list) throws IOException{
...
             List<TxnView> txns = waitFor(entry.txns);
             compactionState.mutate(entry.cells, txns, list, purgeConfig);
             if (!more) {
                 timer.cancel();
-                    context.close();
+                context.close();
             }
+            permits.release(toRelease);
             return more;
         } catch (Exception e) {
```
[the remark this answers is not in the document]
dgomezferro: That's a good point, I added some code to explicitly stop the Reading thread and unblock it if there's any exception here.

spliceci:
TEST FAILED

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
```diff
@@ -230,7 +229,7 @@ public InternalScanner preCompact(ObserverContext<RegionCoprocessorEnvironment>
                 }
                 SICompactionScanner siScanner = new SICompactionScanner(
                         state, scanner, purgeConfig,
-                        conf.getOlapCompactionResolutionShare(), conf.getOlapCompactionResolutionBufferSize(), context);
+                        conf.getOlapCompactionResolutionShare(), conf.getLocalCompactionResolutionBufferSize(), context);
                 siScanner.start();
```
jyuanca, line 233: Why the buffer size is set to LocalCompactionResolutionBufferSize for precompact()? Should it be used for preFlush() instead?

dgomezferro: This affects compactions that couldn't be run in Spark and fallback to run in HBase. I think flushes are fine because we are dealing with less data and the data itself is already in the MemStore, we are not bringing new data from a file into memory, so having a larger buffer shouldn't be a problem.

PR #3354, opened 2020-03-27
DB-8887 Fixed nullpointer exception in vacuum

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:
> Any time this ran and raised an NPE it probably left some rows in DROPPED_CONGLOMERATES, could you add a pass at the end to remove any rows that no longer should be there? Basically any conglomerate there that doesn't exist in HBase should be removed.
> 
> Thanks!

Done

PR #3357, opened 2020-03-27
DB-9272 enable testcases in SparkExplainIT

yxia92:


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SparkExplainIT.java
```diff
@@ -34,6 +36,7 @@
 import static com.splicemachine.test_tools.Rows.row;
 import static com.splicemachine.test_tools.Rows.rows;
 
+@Category(HBaseTest.class)
```
OlegMazurov: This category makes the IT ineligible for remote mode. Is it justified? If yes, please update remote_it/pom.xml.

yxia92: Thanks @OlegMazurov The reason I put this test in the HBaseTest category is to prevent it from running on the mem profile. There are two reasons: 1) sparkexplain under mem profile does not really return spark explain, it returns the control execution plan. 2) the plan from control path is a bit different from the spark plan. 
However, I don't intend to make the IT ineligible for remote mode. Is there a way that I can make it eligible for remote test but still not triggered for mem profile?

OlegMazurov: I was wrong assuming that the category brings dependencies on HBase. I don't think we should exclude the test from _remote_it_.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
```diff
@@ -290,9 +290,28 @@ protected void rowContainsQuery(int level, String query, String contains, Splice
...
-        ResultSet resultSet = methodWatcher.executeQuery(query);
-        while (resultSet.next())
-            Assert.assertFalse("failed query: " + query + " -> " + resultSet.getString(1), resultSet.getString(1).contains(notContains));
+        queryDoesNotContainString(query,new String[] {notContains}, methodWatcher, false);
+    }
+
+    protected void queryDoesNotContainString(String query, String[] notContains,SpliceWatcher methodWatcher, boolean caseInsensitive) throws Exception {
+        ResultSet rs = null;
+        try {
+            rs = methodWatcher.executeQuery(query);
```
OlegMazurov, line 299: ResultSet is AutoCloseable. You can use try-with-resources here.

PR #3362, opened 2020-03-30
Fix wrong spliceengine-ee checkout

bklo94:
Fix checking out of spliceengine-ee branches to fix failing spotbugs jenkins job

### DIFF
--- a/pipelines/spot-bugs/Jenkinsfile
+++ b/pipelines/spot-bugs/Jenkinsfile
@@ -36,7 +36,7 @@ node('splice-standalone') {
         try{
             checkout([  
                 $class: 'GitSCM', 
-                branches: [[name: 'refs/remotes/origin/${BRANCH_NAME}']], 
+                branches: [[name: 'refs/remotes/origin/'+source_branch]],
                 doGenerateSubmoduleConfigurations: false, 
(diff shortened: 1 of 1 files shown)


pipelines/spot-bugs/Jenkinsfile
```diff
@@ -36,7 +36,7 @@ node('splice-standalone') {
         try{
             checkout([  
                 $class: 'GitSCM', 
-                branches: [[name: 'refs/remotes/origin/${BRANCH_NAME}']], 
+                branches: [[name: 'refs/remotes/origin/'+source_branch]],
```
OlegMazurov: Why this inconsistency across the file? Is it OK?

> branches: [[name: source_branch]],
> branches: [[name: 'refs/remotes/origin/'+source_branch]],
> branches: [[name: 'refs/remotes/origin/'+branch]],

bklo94: I took out the extra `refs/remotes/origin/` since it is not needed anymore.
 
The `branch` variable remains since it's only initialized when there is no source_branch to checkout from spliceengine-ee, thus the pipeline can find the branch to checkout to run the spotbugs against.

PR #3376, opened 2020-04-02
DB-6542: create missing index rows

jyuanca:
Implement syscs_util.fix_table to make base table and index consistent.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
@@ -14,6 +14,17 @@
 
 package com.splicemachine.derby.stream.control;
 
+import com.splicemachine.ddl.DDLMessage;
+import com.splicemachine.derby.stream.function.IndexTransformFunction;
+import com.splicemachine.derby.stream.function.KVPairFunction;
+import com.splicemachine.derby.stream.output.DataSetWriter;
+import com.splicemachine.kvpair.KVPair;
+import com.splicemachine.pipeline.PipelineDriver;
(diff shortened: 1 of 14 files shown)


splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
```diff
@@ -157,7 +178,7 @@ private void populateData(PairDataSet table, PairDataSet index) {
 
     private List<String> checkInvalidIndexes() throws StandardException {
```
yxia92, line 179: For checkDuplicateIndexes(), shouldn't we also handle the fix=true path, and remove duplicate indexes?

jyuanca: This does not happen for system tables. I'll file another Jira to address this issue.

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
```diff
@@ -167,57 +188,150 @@ private void populateData(PairDataSet table, PairDataSet index) {
...
             for (String baseRowId : result.keySet()) {
                 List<byte[]> keys = result.get(baseRowId);
                 for (byte[] key : keys) {
-                    if (i >= maxCheckTableErrors) {
+                    if (i == maxCheckTableErrors) {
                         messages.add("...");
                         return messages;
                     }
+                    writeBuffer.add(new KVPair(key, new byte[0], KVPair.Type.DELETE));
+                    if (i > maxCheckTableErrors) {
```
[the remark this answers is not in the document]
jyuanca: Let me restructure the code to make it more readable.

PR #3380, opened 2020-04-03
DB-9336 Allow 'default null' in create table

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UserTypeConstantNode.java
```diff
@@ -54,194 +54,199 @@
...
+
+            assert maxWidth != null;
+
+            super.init(
+                typeId,
+                (arg1 == null) ? Boolean.TRUE : Boolean.FALSE,
+                maxWidth);
+
+            if( dvd != null) {
+                setValue(dvd);
```
yxia92, line 152: With the lines added after this line to set the value, this line become redundant and can be removed.

arnaud-splice: Weirdly enough, no. `setValue` sets the attribute of the base class, whereas `value=` sets the value of this derived class. The whole thing is a bit messy and explained in a comment above `UserTypeConstantNode.value` declaration:

` /*
    ** This value field hides the value in the super-type.  It is here
    ** Because user-type constants work differently from built-in constants.
    ** User-type constant values are stored as Objects, while built-in
    ** constants are stored as StorableDataValues.
    **
    ** RESOLVE: This is a bit of a mess, and should be fixed.  All constants
    ** should be represented the same way.
    */`

PR #3384, opened 2020-04-04
DB-9327 allow SSQ in update statement to be flattened, and disable BatchOnce optimization

yxia92:
Also include the fix:
DB-9343 fix wrong result for flattened SSQ where correlation is in a single table condition

Please see fix notes in DB-9327 and DB-9343.


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -591,6 +591,20 @@ else if (nativeSparkAggregationModeString.equals("forced"))
...
+
+            String ssqFlatteningForUpdateDisabledString =
+                    PropertyUtil.getCachedDatabaseProperty(lcc, Property.SSQ_FLATTENING_FOR_UPDATE_DISABLED);
+            boolean ssqFlatteningForUpdateDisabled = CompilerContext.DEFAULT_SSQ_FLATTENING_FOR_UPDATE_DISABLED;
+            try {
+                if (ssqFlatteningForUpdateDisabledString != null)
+                    ssqFlatteningForUpdateDisabled =
+                            Boolean.valueOf(ssqFlatteningForUpdateDisabledString);
+            } catch (Exception e) {
+                // If the property value failed to convert to a boolean, don't throw an error,
```
arnaud-splice, line 603: An warning could be logged here.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ResultSetFactory.java
```diff
@@ -2105,60 +2105,62 @@ NoPutResultSet getRowCountResultSet(
...
-	                                  GeneratedMethod errorTextGenerator) throws StandardException;
+    NoPutResultSet getBatchOnceResultSet(NoPutResultSet source,
+                                         Activation activation,
+                                         int resultSetNumber,
+                                         NoPutResultSet subqueryResultSet,
+                                         String updateResultSetFieldName,
+                                         int sourceCorrelatedColumnItem,
+                                         int subqueryCorrelatedColumnItem,
+                                         int sourceRowLoationColumnPosition,
+                                         int cardinalityCheck) throws StandardException;
```
dgomezferro, line 2141: Are we forcing the recompilation of stored prepared statements? Otherwise we might break some with this change (triggers for instance)

splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
```diff
@@ -129,7 +128,7 @@ private void setupASTVisitors(LanguageConnectionContext lctx) {
         afterOptVisitors.add(RowLocationColumnVisitor.class);
         afterOptVisitors.add(FixSubqueryColRefs.class);
         afterOptVisitors.add(JoinConditionVisitor.class);
-        afterOptVisitors.add(BatchOnceVisitor.class);
+    //    afterOptVisitors.add(BatchOnceVisitor.class);
```
dgomezferro, line 131: If we remove it from here we won't be using it anymore, right? I'd remove it altogether (BOVisitor, BONode and BOOperation)

db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ResultSetFactory.java
yxia92: Thanks Daniel! Yes, we need to force a re=compilation of stored prepared statement. I've added an upgrade script to do that and tested upgrade with a trigger case.

splice_machine/src/main/java/com/splicemachine/derby/impl/db/SpliceDatabase.java
yxia92: They are removed.

PR #3387, opened 2020-04-06
DB-9029 Support conditional nodes in broadcast join's predicate

arnaud-splice:


arnaud-splice:
> +1
> Compared with the PR of DB-9363, I think the extra code added in this PR is the test cases in BroadcastJoinIT. Did I miss anything else?

That's correct! Sorry about the late reply.

PR #3402, opened 2020-04-08
DB-9331 Display result set numbers in plan

arnaud-splice:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnionNode.java
```diff
@@ -755,14 +755,6 @@ public boolean getIsRecursive() {
         return isRecursive;
     }
 
-    public void setStepNumInExplain(int stepNum) {
-        stepNumInExplain = stepNum;
```
yxia92, line 759: Can you remove the variable stepNumInExplain as it is no longer used?

arnaud-splice: Done

PR #3417, opened 2020-04-11
DB-9392 Update remote tests

OlegMazurov:


remote_it/pom.xml
```diff
@@ -214,6 +218,8 @@
                                 <exclude>com/splicemachine/mrio/**</exclude>
                                 <exclude>**/CallableTransactionIT.java</exclude>
                                 <exclude>**/VacuumIT.java</exclude>
+                                <!-- EE ITs failing in remote mode -->
```
arnaud-splice, line 221: Do we have a jira to track all those disabled tests?

OlegMazurov: No specific JIRA. Currently, it would be too generic without a specific plan what to do about them. Most failing tests were developed to rely on files in the local file system. We can't enable that with just a JDBC connection and no access to remote HDFS.

PR #3419, opened 2020-04-13
DB-8828 Set default mask string of sensitive message to asterisks

changli6:


mem_sql/src/main/resources/log4j.properties
```diff
@@ -17,7 +17,10 @@ log4j.rootLogger=INFO, File1
 log4j.appender.File1=org.apache.log4j.FileAppender
 log4j.appender.File1.file=spliceMem.log
 log4j.appender.File1.layout=org.apache.log4j.EnhancedPatternLayout
-log4j.appender.File1.layout.ConversionPattern=%d{HH:mm:ss,SSS} (%t) %-5p [%c{1.}] - %m%n
+#log4j.appender.File1.layout=com.splicemachine.utils.logging.MaskPatternLayout
+log4j.appender.File1.layout.ConversionPattern=%d{HH:mm:ss,SSS} (%t) %-5p [%c{1}] - %m%n
+#log4j.appender.File1.layout.MaskPattern=insert (?:.*) ([0-9]+),([0-9]+),([0-9]+)
+#log4j.appender.File1.layout.MaskString="********"
```
arnaud-splice, line 23: Should those lines remain commented out?

changli6: These lines were a kind of example template and commented out intended. The User could modify them based on real cases.

PR #3425, opened 2020-04-13
SPLICE-1806 NullPointerException in the PipingCallBuffer (2.8)

OlegMazurov:
Backporting from 3.0

### DIFF
--- a/hbase_storage/hbase1.1.2.2.5/src/main/java/com/splicemachine/storage/ClientPartition.java
+++ b/hbase_storage/hbase1.1.2.2.5/src/main/java/com/splicemachine/storage/ClientPartition.java
@@ -54,13 +54,13 @@ public class ClientPartition extends SkeletonHBaseClientPartition{
     private final Table table;
     private final Connection connection;
     private final Clock clock;
-    private PartitionInfoCache partitionInfoCache;
+    private PartitionInfoCache<TableName> partitionInfoCache;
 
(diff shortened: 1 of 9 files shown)


hbase_storage/hbase1.1.2.2.5/src/main/java/com/splicemachine/storage/ClientPartition.java
```diff
@@ -153,32 +153,26 @@ public boolean checkAndPut(byte[] key,byte[] family,byte[] qualifier,byte[] expe
...
-            partitions = formatPartitions(getAllRegionLocations(true));
-            partitionInfoCache.invalidate(tableName);
-            assert partitions!=null:"partitions are null";
-            partitionInfoCache.put(tableName,partitions);
+            List<HRegionLocation> tableLocations = ((ClusterConnection) connection).locateRegions(tableName, !refresh, false);
+            partitions = new ArrayList<>(tableLocations.size());
+            for (HRegionLocation location : tableLocations){
+                if (location.getServerName() == null) {     // ensure no offline regions
+                    continue;
+                }
```
carolp-503, line 165: Does skipping offline regions ever result in a case where we miss data? That is, is tableLocations ever used in a way that ignores a region that's temporarily down (due to an RS failure, say) such that the missing region doesn't get detected and reported later?

PR #3437, opened 2020-04-17
DB-9401 avoid costing control execution path in the presence of cross…

yxia92:
… join hint

### DIFF
--- a/db-engine/pom.xml
+++ b/db-engine/pom.xml
@@ -107,6 +107,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>annotations</artifactId>
+        </dependency>
     </dependencies>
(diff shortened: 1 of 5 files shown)


yxia92:
@arnaud-splice @dgomezferro Thanks for the review. I've fixed the spotbugs. Also I added a new commit to allow derived table to be the right table of a cross join when hinted. Please take a look, thanks!

dgomezferro:
Looks good to me @yxia92 , one question though, why allowing only crossjoin when hinted for derived tables? Are we worried we don't support well all cases?

yxia92:
@dgomezferro There are some costing issues related to cross join, and other join strategies, they are fixed in DB-9078 which is still under code review. I'm concerned that if I relax the cross join criteria without the fix of DB-9078, the chance of cross join being picked when it shouldn't will be increased.

dgomezferro:
That makes sense @yxia92 , thanks!

PR #3440, opened 2020-04-17
DB-9245 Added alias ANALYSE

ipraznik-splice:


arnaud-splice:
Could you merge the latest master in this to get build checks to pass?

ipraznik-splice:
I

> Please add a test case.

I've added integration tests for ANALYSE alias

PR #3458, opened 2020-04-21
DB-9423: fix foreign key constraint uniqueness

jyuanca:


jyuanca:
This statement is expected to fail because table b is empty so it violates FK constraints
insert into a values (1,1);

jyuanca:
This statement is expected to fail because table b is empty so it violates FK constraints
insert into a values (1,1);

jyuanca:
This statement is expected to fail because table b is empty so it violates FK constraints
insert into a values (1,1);

jyuanca:
This statement is expected to fail because table b is empty so it violates FK constraints
insert into a values (1,1);

dgomezferro:
You are right @jyuanca , sorry, I was trying to reproduce the issue without your change and got confused.

I meant to paste this:
```
drop table a;
drop table b; 
create table b(i int, j int, primary key (i));
create table a (i int, j int);
insert into b values (1,1);
insert into a values (1,1);
create unique index aj on a(j);
alter table a add CONSTRAINT fc FOREIGN KEY(j) REFERENCES b(i) ON UPDATE NO ACTION ON DELETE restrict;
drop index aj;
delete from a where i = 1;
delete from b;
```
That last delete should be successful but fails. If I force the creation of a new conglomerate for the FK it works.

yxia92:
@jyuanca If I switch the order the index and FK are created, the two indexes are actually created using different conglomerate ids, so I tend to agree with Daniel, so that the behavior is more consistent. And when the unique index is dropped and only the FK constraint is left, we can insert rows with duplicate values on the column j.
```
create table b(i int, j int, primary key (i));
create table a (i int, j int);
insert into b values (1,1);
insert into a values (1,1);
alter table a add CONSTRAINT fc FOREIGN KEY(j) REFERENCES b(i) ON UPDATE NO ACTION ON DELETE restrict;
create unique index aj on a(j);
splice> show indexes from a;
TABLE_NAME                                        |INDEX_NAME                                        |COLUMN_NAME         |ORDINAL&|NON_UNIQUE|TYPE |ASC&|CONGLOM_NO
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
A                                                 |AJ                                                |J                   |1       |false     |BTREE|A   |1697      
A                                                 |SQL200421165039490                                |J                   |1       |true      |BTREE|A   |1681      

2 rows selected
```
So I also think we should make

jyuanca:
There was a bug for sharing an index with a foreign key. If the index is dropped from the table, it is dropped from write pipeline. I'm fixing that and hope sharing works because it has performance and space benefits

jyuanca:
@dgomezferro @yxia92 Can you take another look?

yxia92:
@jyuanca One problem I see with sharing the same conglomerate for a unique index and FK constraint is that after the unique index is dropped, the FK constraint still shows as unique. For example:
```
drop table a;
drop table b; 
create table b(i int, j int, primary key (i));
create table a (i int, j int);
insert into b values (1,1);
insert into a values (1,1);
create unique index aj on a(j);
alter table a add CONSTRAINT fc FOREIGN KEY(j) REFERENCES b(i) ON UPDATE NO ACTION ON DELETE restrict;
splice> show indexes from a;
TABLE_NAME                                        |INDEX_NAME                                        |COLUMN_NAME         |ORDINAL&|NON_UNIQUE|TYPE |ASC&|CONGLOM_NO
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
A                                                 |AJ                                                |J                   |1       |false     |BTREE|A   |1809      
A                                                 |SQL200512120915900                                |J                   |1       |false     |BTREE|A   |1809      

2 rows selected

splice> drop index aj;
splice> show indexes from a;
TABLE_NAME                                        |INDEX_NAME                                        |COLUMN_NAME         |ORDINAL&|NON_UNIQUE|TYPE |ASC&|CONGLOM_NO
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
A                                                 |SQL200512120915900                                |J                   |1       |false     |BTREE|A   |1809      

1 row selected
```
Because the FK constraint is still marked as unique, we cannot insert a duplicate row into A:
```
splice> insert into a values (1,1);
ERROR 23505: The statement was aborted because it would have caused a duplicate key value in a unique or primary key constraint or unique index identified by 'AJ' defined on 'A'.
splice> 
```

jyuanca:
The foreign key constraints is non-unique, so it should not share an existing unique sonstraint. Thanks @yxia92 for testing this!

jyuanca:

spliceci:
TEST FAILED

jyuanca:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/CreateIndexConstantOperation.java
```diff
@@ -452,6 +452,9 @@ public void executeConstantAction( Activation activation ) throws StandardExcept
                 }
 
                 IndexRowGenerator irg = cd.getIndexDescriptor();
+                if (irg.isUnique() != unique) {
```
yxia92, line 455: Could you also add check so that if  excludeNulls or excludeDefaults do not match, the index cannot be shared either?
Here is a test case:
```
create table b(i int, j int, primary key (i));
create table a (i int, j int);
insert into b values (1,1);
insert into a values (1,1);
create index aj on a(j) exclude null keys;
alter table a add CONSTRAINT fc FOREIGN KEY(j) REFERENCES b(i) ON UPDATE NO ACTION ON DELETE restrict;
splice> show indexes from a;
TABLE_NAME                                        |INDEX_NAME                                        |COLUMN_NAME         |ORDINAL&|NON_UNIQUE|TYPE |ASC&|CONGLOM_NO
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
A                                                 |AJ                                                |J                   |1       |true      |BTREE|A   |1665      
A                                                 |SQL200701155616910                                |J                   |1       |true      |BTREE|A   |1665      

2 rows selected
ELAPSED TIME = 1531 milliseconds
```

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/969/

jyuanca:

spliceci:
TEST SUCCEEDED +1

PR #3461, opened 2020-04-21
DB-9109 Fix sporadics in jenkins due to overwrite of git data in tests

martinrupp:


hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -54,19 +54,13 @@
...
     public static void cleanoutDirectory() {
-        try {
-            File file = new File(getExternalResourceDirectory());
-            if (file.exists())
-                FileUtils.deleteDirectory(new File(getExternalResourceDirectory()));
-            file.mkdir();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
-        }
+        SpliceUnitTest.clearDirectory( getExternalResourceDirectory() );
```
yxia92, line 61: There are currently four ITs putting the test files in this directory: ExternalTableIT, ExternalTablePartitionIT, ShowCreateTableIT, ViewsInSysibmIT. When these tests are running in parallel, would this call accidentally delete the files that are generated by other tests?  Is this the source of intermittent test failure or wrong result?

martinrupp: yes, this is problematic. didn't knew they are running in parallel.
a better solution is to use Files.createTempDirectory, copy files/directories there, then delete after use, e.g. with java.lang.AutoCloseable. A small downside of this is that in the case of errors we don't see the failed files at the end, but i think this is acceptable. If people really need to investigate output files, one can uncomment the FileUtils.deleteDirectory line.

martinrupp: will do another change

PR #3463, opened 2020-04-21
DB-9361: remove duplicate indexes

jyuanca:


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -148,30 +148,58 @@ public void setTableDataSet(DataSet tableDataSet) {
...
+            } else {
+                return reportDuplicateIndexes(it);
             }
+        }catch (Exception e) {
+            throw StandardException.plainWrapException(e);
         }
-        if (count >= maxCheckTableError) {
-            count = duplicateIndexRdd.count();
-            messages.add("...");
+    }
```
martinrupp, line 174: re-add SpliceSpark.popScope()

(this function does SpliceSpark.pushScope , but SpliceSpark.popScope(); was moved (unintentionally i guess) to line 204))

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -148,30 +148,58 @@ public void setTableDataSet(DataSet tableDataSet) {
...
+            Tuple2<String, Tuple2<byte[], ExecRow>> t = (Tuple2<String, Tuple2<byte[], ExecRow>>)it.next();
+            messages.add(t._2._2 + "=>" + t._1);
+            if (count >= maxCheckTableError) {
+                messages.add("...");
+                break;
+            }
+            count++;
+        }
         messages.add(0, String.format("The following %d indexes are duplicates:", count));
         SpliceSpark.popScope();
```
martinrupp, line 204: remove popScope

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -148,30 +148,58 @@ public void setTableDataSet(DataSet tableDataSet) {
...
+    private List<String> reportDuplicateIndexes(Iterator it) throws StandardException, InterruptedException, ExecutionException {
+        List<String> messages = Lists.newLinkedList();
 
+        long count = 0;
+        while (it.hasNext()) {
+            Tuple2<String, Tuple2<byte[], ExecRow>> t = (Tuple2<String, Tuple2<byte[], ExecRow>>)it.next();
+            messages.add(t._2._2 + "=>" + t._1);
+            if (count >= maxCheckTableError) {
+                messages.add("...");
+                break;
```
martinrupp, line 199: the only difference i see between these two functions is 
if(count >= maxCheckTableError) . I don't see how fixDuplicateIndexes fixes anything.
In general, we could do 

`private List<String> reportDuplicateIndexes(Iterator it, int _maxCheckTableError)`
then have 
`if (_maxCheckTableError >= 0 || count >= _maxCheckTableError)`

and then do
`return reportDuplicateIndexes(it, fix ? -1 : maxCheckTableError);`

martinrupp: i think fixDuplicateIndexes is missing the lines https://github.com/splicemachine/spliceengine/pull/3463/files#diff-49eff1a8a3acfe993feb4e304ff615ceL161-L167 , no?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
jyuanca: Good catch!

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
jyuanca: Depending on whether fix is true
JavaRDD duplicateIndex = joinedRdd
                    .mapPartitions(new SparkFlatMapFunction<>(new DeleteDuplicateIndexFunction<>(conglomerate, txn, tentativeIndex, baseColumnMap, **fix**)));

DeleteDuplicateIndexFunction deletes (or ignores)  extra indexes

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
```diff
@@ -134,41 +136,100 @@ private void populateData(PairDataSet table, PairDataSet index) {
...
+                ExecRow baseRow = tableData.get(baseRowId);
+                List<Tuple2<byte[], ExecRow>> indexRows = result.get(baseRowId);
+                for (Tuple2<byte[], ExecRow> indexRow : indexRows) {
+                    boolean duplicate = false;
+                    DataValueDescriptor[] dvds = indexRow._2.getRowArray();
+                    for (int i = 0; i < dvds.length - 1; ++i) {
+                        int col = baseColumnMap[indexToMain.get(i) - 1];
+                        if (!dvds[i].equals(baseRow.getColumn(col + 1))) {
+                            duplicate = true;
+                            break;
```
yxia92, line 186: We seem to assume that if there are duplicate index rows, the extra ones must have different values than the base row. Is it possible that the extra index row also has the same values as the base row?

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -148,32 +148,62 @@ public void setTableDataSet(DataSet tableDataSet) {
...
+    private List<String> fixDuplicateIndexes(Iterator it) {
+        List<String> messages = Lists.newLinkedList();
+
+        long count = 0;
+        while (it.hasNext()) {
+            Tuple2<String, Tuple2<byte[], ExecRow>> t = (Tuple2<String, Tuple2<byte[], ExecRow>>)it.next();
+            messages.add(t._2._2 + "=>" + t._1);
+            count++;
         }
+        messages.add(0,String.format("Deleted the following %d duplicate indexes", count));
```
martinrupp, line 189: i still don't understand how this function is fixing any duplicate indices.
it does pretty much the same as reportDuplicateIndexes when maxCheckTableError = infinity (well, except for the "Deleted the following %d duplicate indexes" / "The following %d indexes are duplicates" ).

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlTableChecker.java
jyuanca: The index that has some column values as base row is the correct one and should be kept.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
```diff
@@ -148,32 +148,62 @@ public void setTableDataSet(DataSet tableDataSet) {
...
+    private List<String> reportDuplicateIndexes(Iterator it) throws StandardException, InterruptedException, ExecutionException {
+        List<String> messages = Lists.newLinkedList();
 
+        long count = 0;
+        while (it.hasNext()) {
+            Tuple2<String, Tuple2<byte[], ExecRow>> t = (Tuple2<String, Tuple2<byte[], ExecRow>>)it.next();
+            messages.add(t._2._2 + "=>" + t._1);
+            if (count >= maxCheckTableError) {
+                messages.add("...");
+                break;
```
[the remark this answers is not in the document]
jyuanca: This was intentional to avoid counting exact number. However, my second thought is that it is rarely a very expensive operations, so I will revert back

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkTableChecker.java
jyuanca: The index is fixed by DeleteDuplicateIndexFunction. This function reports fixed indexes.

jyuanca: Let me restructure the code around here so not to cause confusion

PR #3475, opened 2020-04-23
DB-9439 Handle carried bytes correctly when computing cutpoints

dgomezferro:


hbase_storage/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionUtil.java
```diff
@@ -98,7 +98,7 @@ public static void updateReadRequests(HRegion region,long numReads){
                     totalStoreFileInBytes += ((HStoreFile) file).getFileInfo().getFileStatus().getLen();
                 }
             }
-            numSplits = totalStoreFileInBytes / bytesPerSplit;
+            numSplits = (totalStoreFileInBytes + store.getMemStoreSize()) / bytesPerSplit;
```
yxia92, line 101: Why do we need to add store.getMemStoreSize() in the calculation? How do we compute the cutpoint for content in memStore? I don't see any logic about it. Will all content in memStore go into one split?
A few lines below computes bytesPerSplitEvenDistribution which is the value eventually used as the splitBlockSize. And it uses this formula:
```
            long bytesPerSplitEvenDistribution = totalStoreFileInBytes / numSplits;
```
This does not consider memStoreSize. is it intended?

dgomezferro: Yes, it is intended. I added some comments to clarify since it can be confusing.
Basically, we take into account the memstore size when computing bytesPerSplit but we weren't taking it into account when computing the splits, so we always were creating less splits than intended.
By taking into account the memstore size when computing the right numSplits, we make sure that at least we end up with the right amount of splits, even though some might be skewed if the data in the MemStore doesn't follow the same distribution as the data in the HFiles.

yxia92: Thanks for the explanation!

PR #3481, opened 2020-04-23
DB-9404: make sure timestamp never decrease

jyuanca:


OlegMazurov:
Do we understand the scenario that leads to decreased timestamp stored in Zookeeper? (I don't see how TransactionWatcher could be responsible for that).

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampBlockManager.java
```diff
@@ -43,11 +43,17 @@ public ZkTimestampBlockManager(RecoverableZooKeeper rzk,String blockNode) {
...
     }
 
-
-    public void reserveNextBlock(long nextMax) throws TimestampIOException{
-        byte[] data = Bytes.toBytes(nextMax);
+    @Override
+    public long reserveNextBlock(long nextMax) throws TimestampIOException{
         try {
-            rzk.setData(blockNode, data, -1 /* version */); // durably reserve the next block
+            byte[] bs = rzk.getData(blockNode, false, null);
```
dgomezferro, line 49: Could we use the version parameter in .setData() to make sure we only update it if it matches the expected version?

I don't think we should be playing catch up here, if somebody else updated the stored timestamp we should bail out, either somebody else is the new TimestampOracle or we are in the middle of a restore were we shouldn't be creating new transactions.

dgomezferro:
@OlegMazurov we experienced this during a Restore that restored a larger timestamp, but then the ZkTimestampBlockManager overwrote it with a lower timestamp, possibly due to TransactionWatcher consuming lots of transactions and forcing ZkTimestampBlockManager to reserve a new block.

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampBlockManager.java
jyuanca: It is possible that a timestamp is increased on a replica. We may want to catch up in this case.

dgomezferro: I thought replicas were read only and we were not increasing timestamps there, how would that happen?

jyuanca: The replica applies WALs and compare the progress with the snapshot of master cluster. If it has replicated all WALs for a snapshot, the replica bumps up timestamp to the snapshot's timestamp.

carolp-503: When we got the timezone correct (matching the timezone on the backup) on the cluster where we were testing this, we did not need to apply the workaround.  Seems to be timezone related.  So, to Oleg's point, is this PR addressing a symptom rather than the problem (backup restoring to a different timezone)?

OlegMazurov: I see that replication reads/updates the same Zookeeper node (_/transactions/maxReservedTimestamp_)  directly, bypassing TimestampOracle. That violates many assumptions made in TimestampOracle code.
TimestampOracle itself is not free from subtle data races. _timestampCounter.set() can not be reliably used in multithreaded environment (it's atomic all right, but can't maintain integrity) and should be replaced with compare-and-swap.
I believe replication approach should be revised.

OlegMazurov:
I see that replication reads/updates the same Zookeeper node (/transactions/maxReservedTimestamp) directly, bypassing TimestampOracle. That violates many assumptions made in TimestampOracle code.
TimestampOracle itself is not free from subtle data races. _timestampCounter.set() can not be reliably used in multithreaded environment (it's atomic all right, but can't maintain integrity) and should be replaced with compare-and-swap.
I believe replication approach should be revised.

OlegMazurov:
From performance perspective, another call to Zookeeper to read data before we update it doubles the cost of updates increasing the latency of an arbitrary operation creating a new timestamp (we had a customer who was very sensitive to that). We should try hard to avoid that cost increase.

hbase_storage/src/main/java/com/splicemachine/timestamp/hbase/ZkTimestampBlockManager.java
dgomezferro: I don't think it's TZ related, I think it's mostly dependent on how long the cluster is running after the restore finishes, and there's a bit of (bad) luck involved. Some attempts the other day with the incorrect TZ didn't hit this problem either, I think it's just coincidence that we didn't hit it now.

jyuanca:
I have a simpler fix for this. Closing this PR

PR #3486, opened 2020-04-24
DB-9109 fix problems with parallel runs of tests all writing to xxx/target/external

martinrupp:
every test that needs a temporary output dir will now on-demand create a temp directory,
that will be deleted automatically after the test has run, using the @After annotation of JUnit.


hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
```diff
@@ -572,7 +562,7 @@ public void testWriteReadNullValues() throws Exception {
     @Test
     public void testWriteReadNullValuesAvro() throws Exception {
 
-        String tablePath = getExternalResourceDirectory()+"null_test_location_avro";
+        String tablePath = getTempOutputDirectory() + "null_test_location_avro";
```
yxia92, line 565: I think "null_test_location_avro" should be "/null_test_location_avro", otherwise, the temporary directory used would be something like "SpliceUnitTest_temp_xxxxnull_test_location_avro" instead of "SpliceUnitTest_temp_xxxxn/ull_test_location_avro". And the after test call deleteTempDirectory() wouldn't be able to delete this directory.  Everything else looks good to me.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
```diff
@@ -0,0 +1,72 @@
...
+            System.out.println( "Not deleting test temporary directory " + tempDir );
+        else
+            FileUtils.deleteDirectory( tempDir.toFile() );
+    }
+
+    /// this will return the temp directory, that is created on demand once for each test
+    public String getTempOutputDirectory() throws Exception
+    {
+        if( tempDir == null)
+            tempDir = Files.createTempDirectory( "SpliceUnitTest_temp_" );
```
OlegMazurov, line 37: Better create temp directories in "/target" as previously - build is self-contained, more control, less surprises.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/// this class extends SpliceUnitTest so that each test
+/// can access a temporary output directory that is removed after the test is run
+/// the temporary output directory is created only on demand.
+/// might be also doable like SpliceSchemaWatcher extends TestWatcher
+public class SpliceUnitTestWithTempDirectory extends SpliceUnitTest
```
OlegMazurov, line 14: Why not merge the content of this class into SpliceUnitTest. We already have _createBadLogDirectory_ there, f.e.
Developers creating new unit tests won't need to differentiate between SpliceUnitTest and SpliceUnitTestWithTempDirectory: either you use the temporary directory or not. Passive overhead is negligible.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
```diff
@@ -17,7 +17,9 @@
 import com.splicemachine.homeless.TestUtils;
 import com.splicemachine.utils.Pair;
 import org.apache.commons.io.FileUtils;
+import org.junit.AfterClass;
 import org.junit.Assert;
+import org.junit.BeforeClass;
```
OlegMazurov, line 22: Not used?

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
martinrupp: ok i didn't do much java/maven before, for me it was a bit surprising that target is used for these kind of temporary directories, test output, current database state, logs, and not only as a target for binaries. actually a reason we have to advice to do mvn **clean** every time, to clean off the target dir - if i understood correctly.
But also works for me. With the current solution target dir should be cleaned after a test ran.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
martinrupp: if preferred, also possible.

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/ExternalTableIT.java
martinrupp: actually a lot of tests forgot the /, so i added it to _getTempOutputDirectory_

martinrupp: done

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
martinrupp: done

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTestWithTempDirectory.java
martinrupp: done

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
martinrupp: done

PR #3487, opened 2020-04-25
DB-9404: do not increment timestamp during restore

jyuanca:


jyuanca:
@dgomezferro I filed DB-9461 to address data version problem

PR #3497, opened 2020-04-29
DB-9434 ORDER BY can use the aliases defined in SELECT AS

martinrupp:
e.g. as in
  select C as C2+1 from A order by -C2;

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
@@ -605,7 +605,7 @@ public class ResultColumnList extends QueryTreeNodeVector<ResultColumn>{
                     retVal=resultColumn;
                 }else if(!retVal.isEquivalent(resultColumn)){
                     throw StandardException.newException(SQLState.LANG_DUPLICATE_COLUMN_FOR_ORDER_BY,columnName);
-                }else if(index>=size-orderBySelect){// remove the column due to pullup of orderby item
+                }else if( index >= getFirstOrderByIndex() ){// remove the column due to pullup of orderby item
                     removeElement(resultColumn);
                     decOrderBySelect();
                     obc.clearAddedColumnOffset();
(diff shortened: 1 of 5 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumnList.java
```diff
@@ -3622,7 +3640,7 @@ private static void updateArrays(int[] mapArray,
     }
 
 
-    public class ColumnMapping{
+    public static class ColumnMapping{
```
martinrupp, line 3643: from spotbugs

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/OrderByIT.java
```diff
@@ -377,4 +382,48 @@ public void testOrderByClauseWithNullInWindowFunctionThroughSparkPath() throws E
...
+    public void testUsingAliasOnlyInOrderBy() throws Exception {
+        String sqlTexts[] =
+                {       "select a1+1 as ALIAS from t2 WHERE ALIAS > 0 ORDER BY -ALIAS",
+                        "select a1+1 as ALIAS from t2 GROUP BY ALIAS ORDER BY -ALIAS",
+                        "select a1, sum(b1+1) as ALIAS from t1 GROUP BY a1 HAVING ALIAS > 0 ORDER BY -ALIAS"
+                        };
+        for( String sqlText : sqlTexts ) {
+            SqlExpectException( methodWatcher, sqlText, "42X04" );
+        }
+    }
```
martinrupp, line 423: As the comment says, this makes sure we disable access to alias for WHERE, HAVING and GROUP BY.
Currently, this is not supported. If we remove the line with [clearAliases](https://github.com/splicemachine/spliceengine/pull/3497/files#diff-516cd97dea19fe70597a7399aec1534cR1424) we can enable this. @yxia92 and i agreed on it might be safe currently to not change the behavior here.

PR #3498, opened 2020-04-29
DB-9448 Fix resurfacing rows issue in Spark

arnaud-splice:
Migration to hbase2.0 created two issues:
1. We did not use memStoreAware's lock properly anymore
2. We did not know about compacted files when creating a HRegion from
Spark's side

This commit addresses these two issues in the following way:

1.
SpliceCompactionRequest.setOffPeak is now hijacked to properly
release the lock on memStoreAware once a compaction has completed.
Short of modifying HBase's code to be able to use a custom
CompactionLifeCycleTracker (only DUMMY is used outside of their tests),
this seems to be the only reliable solution.

2.
Spark side now RPC calls the relevant region to get a list of compacted
files so that regions created on Spark side can properly exclude
compacted store files that pertains to a specific region.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
+++ b/hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
@@ -18,6 +18,8 @@ import com.splicemachine.access.client.MemstoreAware;
 import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.HStoreFile;
+import org.apache.hadoop.hbase.regionserver.Store;
+import org.apache.hadoop.hbase.regionserver.compactions.CompactionLifeCycleTracker;
 import org.apache.hadoop.hbase.regionserver.compactions.CompactionRequestImpl;
 import org.apache.log4j.Logger;
 
@@ -33,6 +35,7 @@ public class SpliceCompactionRequest extends CompactionRequestImpl {
     private static final Logger LOG = Logger.getLogger(SpliceCompactionRequest.class);
     private AtomicReference<MemstoreAware> memstoreAware;
     private HRegion region;
+    private boolean compactionCountIncremented = false;
 
     public SpliceCompactionRequest(Collection<HStoreFile> files) {
         super(files);
@@ -42,6 +45,7 @@ public class SpliceCompactionRequest extends CompactionRequestImpl {
         assert memstoreAware != null;
         while (true) {
             MemstoreAware latest = memstoreAware.get();
+            assert latest.currentCompactionCount >= 0;
             if (latest.currentScannerCount>0) {
                 SpliceLogUtils.warn(LOG,"compaction Delayed waiting for scanners to complete scannersRemaining=%d",latest.currentScannerCount);
                 try {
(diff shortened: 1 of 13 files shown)


arnaud-splice:

spliceci:
TEST FAILED

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
```diff
@@ -82,4 +102,26 @@ public void setRegion(HRegion region) {
...
+        // At those points, the value passed is irrelevant and is "only" used for logging
+        // purpose, so we can hijack it.
+        super.setOffPeak(value);
+        if (!value) {
+            afterExecute();
+        }
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
dgomezferro, line 119: These two methods can be removed since they don't change the superclass implementation.

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactionPolicy.java
```diff
@@ -48,7 +48,9 @@ public CompactionRequestImpl selectCompaction(Collection<HStoreFile> candidateFi
         HRegion region = store.getHRegion();
         String storeName = store.getColumnFamilyName();
         scr.setIsMajor(cr.isMajor(), cr.isAllFiles());
-        scr.setOffPeak(cr.isOffPeak());
+        // Ignoring cr.isOffPeak() and hardcoding true because SpliceCompactionRequest.setOffPeak is hijacked to run
+        // SpliceCompactionRequest.afterExecute if isOffPeak(false) is called.
+        scr.setOffPeak(true);
```
dgomezferro, line 53: Are there any interactions with real off-peak compactions scheduled from HBase?

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
arnaud-splice: Those are there as documentation to satisfy spotbugs.

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
```diff
@@ -51,19 +55,35 @@ public void preStorefilesRename() throws IOException {
...
+                }
+                assert !compactionCountIncremented;
+                compactionCountIncremented = true;
                 break;
+            }
         }
     }
     public void afterExecute(){
-        if (memstoreAware == null) {
+        if (memstoreAware == null || !compactionCountIncremented) {
```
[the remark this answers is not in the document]
arnaud-splice: I don't think so. Without `compactionCountIncremented`, we end up decrementing compactionCount below 0.
That's because we setMemstoreAware in preCompact but we call preStorefilesRename after the actual execution in SpliceDefaultCompactor.compact()

arnaud-splice:

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceCompactionRequest.java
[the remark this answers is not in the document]
dgomezferro: That's right, I got confused and thought we were setting it in preStorefilesRename. Looks good!

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactionPolicy.java
arnaud-splice: Based on what I could see, this parameter was ignored anyway, so it should not have any impact.

spliceci:
TEST FAILED

PR #3505, opened 2020-04-30
DB-9109 fix jenkins problem with testCollectStats

martinrupp:
in the previous version the temporary directory gets deleted after every test,
however the table gets only deleted after the whole class shuts down,
so e.g. analyze schema fails as it can't access those tables anymore

what will happen now:
- before instantiating a test class like, e.g. ExternalTableIT would go to /platform_it/target/external + /SCHEMA_NAME,
e.g. /platform_it/target/external/EXTERNALTABLEIT, wipe that directory (with @BeforeClass)
- we then run all tests, create external tables in that subdirectory (e.g. .../external/EXTNERALTABLEIT/avro_simple_file_test etc.)
- at the end of the whole class @AfterClass, delete the temporary directory (target/external/EXTERNALTABLEIT), but not all of target/external

with this
- allows parallel runs of ExternalTableIT while running ExternalTablePartitionIT
- temporary files are delete when running normally
- even if VM crashes or similar, the directory will be cleaned the next time, reducing trash on developers system

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
@@ -37,6 +37,7 @@ import com.splicemachine.test_tools.TableCreator;
 import org.junit.*;
 import org.junit.experimental.categories.Category;
 
+import java.io.File;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.regex.Matcher;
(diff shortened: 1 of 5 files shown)


splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
```diff
@@ -101,6 +102,26 @@ public static void cleanup()
         classWatcher.closeAll();
     }
 
+    static File tempDir;
+
+    @BeforeClass
+    public static void createTempDirectory() throws Exception
```
dgomezferro, line 108: Can you add these methods to SpliceUnitTest ?

If you don't want to create a temp directory for every class even if it doesn't use it we can create the directory on demand in getExternalResourceDirectory() (if not already created) and remove it if exists in deleteTempDirectory

martinrupp: I know this looks like copied code but putting this in SpliceUnitTest would prevent ExternalTableIT, ExternalTablePartitionIT to run parallel:

Since **static File tempDir** is static, it is shared by all instantiations of SpliceUnitTest, especially between ExternalTableIT, ExternalTablePartitionIT, etc. . However, @AfterClass is called after ExternalTableIT is done, then would delete the (shared) tempDir, which might be still used by ExternalTablePartitionIT when running in parallel.

So **static File tempDir** can be static, but static in ExternalTableIT, so that it is not shared with the one in ExternalTablePartitionIT . Another solution would be to have a TestWatcher.

dgomezferro: Hadn't thought of that, makes sense. Thanks!

jyuanca:
@martinrupp Can you also fix branch-2.8 and branch-3.0?

PR #3514, opened 2020-05-03
DB-9471: flush index rows after base row is successfully written

jyuanca:


pipeline_api/src/main/java/com/splicemachine/pipeline/callbuffer/ServerCallBuffer.java
```diff
@@ -158,8 +159,12 @@ public BulkWrites getBulkWrites() throws Exception {
         for (Entry<byte[], PartitionBuffer> regionEntry : this.buffers.entrySet()) {
             PartitionBuffer value = regionEntry.getValue();
             if (value.isEmpty()) continue;
-            bws.add(value.getBulkWrite());
-            value.clear(); // zero out
+            BulkWrite bulkWrite = value.getBulkWrite();
+            bws.add(bulkWrite);
+            Collection<KVPair> kvPairs =  bulkWrite.getMutations();
+            for (KVPair kvPair:kvPairs) {
+                value.remove(kvPair); // zero out
```
arnaud-splice, line 166: Is that called a lot? Because we now have a o(n*m) instead of o(n) given that value.buffer is an ArrayList

PR #3521, opened 2020-05-04
DB-9471: do not auto flush index buffer

jyuanca:


dgomezferro:
When is autoFlush true and when is it false? Wouldn't it always be false in practice?

jyuanca:
It is true by default, so that writes from client are flushed when buffer is full or size exceeds a threshold. For index buffers in write pipeline, autoFlush is set to false, because index should be always flushed after base row is written. Disabling auto flush for index buffer should not cause memory issues, because number of index is always equal to number of base rows and size of index is always smaller than a base row.

dgomezferro:
Thanks for the explanation @jyuanca , that makes sense! Do you think it would be possible to add an IT that exercises this problem so that there are no regressions?

jyuanca:
@dgomezferro Yes, that's good suggestion. Thanks!

PR #3531, opened 2020-05-05
DB-9447 Relabel Engine in EXPLAIN plans

martinrupp:
DB-9447 Relable Control->OLTP, Spark->OLAP in explain and get_running_operations
DB-9447 fix spotbugs

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/RunningOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/RunningOperation.java
@@ -4,6 +4,7 @@ import java.util.Date;
 import java.util.UUID;
 
 import com.splicemachine.derby.stream.iapi.DataSetProcessor;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  *  Entity for storing running operations' info
@@ -22,6 +23,7 @@ public class RunningOperation{
     private SpliceOperation operation;
     private Thread thread;
 
+    @SuppressFBWarnings("EI_EXPOSE_REP2") // mutable Date problem
     public RunningOperation(SpliceOperation operation,
                             Thread thread,
                             Date submittedTime,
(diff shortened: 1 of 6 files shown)


splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/RunningOperation.java
```diff
@@ -29,14 +29,15 @@ public RunningOperation(SpliceOperation operation,
                             UUID uuid, String rdbIntTkn) {
         this.operation = operation;
         this.thread =thread;
-        this.submittedTime = submittedTime;
+        this.submittedTime = (Date)submittedTime.clone();
```
dgomezferro, line 32: I would rather remove this clone, it shouldn't be too costly but for OLTP queries it could add up since it's in the hot path.

martinrupp: this is a spotbugs issue, complaining about assigning a value that could be modified later. i can ignore it though.

martinrupp:
> +1, Another change that we need to make, maybe not in this JIRA, is the engine type we write to the log file, for example, in SpliceBaseOperation.logExecutionStart().

yes i saw a couple of those as well, but chose not to do it here, not sure if customers rely on certain log messages, will have to double-check this with someone.
Other changes needed are to **add** (not replace) useOLAP=true in query and session hint

splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/RunningOperation.java
martinrupp: reverted and added ignores for Spotbugs EI_EXPOSE_REP and EI_EXPOSE_REP2:
EI_EXPOSE_REP2: May expose internal representation by incorporating
reference to mutable object

We might want to replace java.util.Date with java.time.Instance, see
https://stackoverflow.com/questions/36639154/convert-java-util-date-to-what-java-time-type

PR #3538, opened 2020-05-05
DB-9504 Add NativeSpark support for NotNode

dgomezferro:


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
```diff
@@ -356,6 +356,9 @@ else if (operand instanceof SimpleStringOperatorNode) {
                     SimpleStringOperatorNode sso = (SimpleStringOperatorNode) operand;
                     return format("%s(%s) ", operatorString, opToString2(sso.getOperand(), vars));
                 }
+                else if (operand instanceof NotNode) {
```
martinrupp, line 359: i think this is effectively the same code as lines 355-357. SimpleStringOperatorNode doesn't override getOperand, and even if - if i understand java inheritance correctly - i think uop.getOperand() would call the same as (((SimpleStringOperatorNode)uop).getOperand() .
so i think lines 355-361 can be replaced by
```
else if (operand instanceof SimpleStringOperatorNode || operand instanceof NotNode) {
    return format("%s(%s) ", operatorString, opToString2(uop.getOperand(), vars));
}
```

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsOperation.java
```diff
@@ -126,6 +126,9 @@ else if (storedAs.equals("O"))
             } else {
                 statsDataSet = scanSetBuilder.buildDataSet(scope).map(new CloneFunction<>(operationContext));
             }
+            if(statsDataSet.partitions() == 0) {
```
martinrupp, line 129: do we test this case?

dgomezferro: Yes, ExternalTableIT.collectStats() was failing due to this since DB-9369 was merged

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
dgomezferro: Good eye, I applied that change.

PR #3539, opened 2020-05-05
DB-9388 write complete update row when updating the sys.sysstatements…

yxia92:
… table

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
@@ -3472,12 +3472,11 @@ public abstract class DataDictionaryImpl extends BaseDataDictionary{
         FormatableBitSet columnToReadSet=new FormatableBitSet(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT);
         FormatableBitSet columnToUpdateSet=new FormatableBitSet(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT);
         for(int i=0;i<SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT;i++){
-            if (i+1==SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE) {
-                columnToUpdateSet.set(i);
-            } else {
+            // we do not want to read the saved serialized plan
+            if (i+1 != SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE) {
                 columnToReadSet.set(i);
-                columnToUpdateSet.set(i);
             }
+            columnToUpdateSet.set(i);
         }
         /* Set up a couple of row templates for fetching CHARS */
         DataValueDescriptor[] rowTemplate =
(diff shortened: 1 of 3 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -3471,20 +3471,21 @@ public void clearSPSPlans() throws StandardException{
 
         FormatableBitSet columnToReadSet=new FormatableBitSet(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT);
         FormatableBitSet columnToUpdateSet=new FormatableBitSet(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT);
-        columnToUpdateSet.set(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_VALID-1);
-        columnToUpdateSet.set(SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE-1);
-
+        for(int i=0;i<SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT;i++){
+            if (i+1==SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE) {
+                columnToUpdateSet.set(i);
```
martinrupp, line 3476: why is `columnToUpdateSet.set(i);` done in if and in else?
shouldn't it rather be

```
       for(int i=0;i<SYSSTATEMENTSRowFactory.SYSSTATEMENTS_COLUMN_COUNT;i++){
            if (i+1 != SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE) {
                columnToReadSet.set(i);
            }
            columnToUpdateSet.set(i);
        }
```

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -3502,6 +3503,14 @@ public void clearSPSPlans() throws StandardException{
 
         while(sc.fetchNext(rowTemplate)){
             /* Replace the column in the table */
+            for (int i=0; i<rowTemplate.length; i++) {
+                if (i  == SYSSTATEMENTSRowFactory.SYSSTATEMENTS_VALID-1)
+                    replaceRow[i] = new SQLBoolean(false);
+                else if (i == SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE-1)
```
martinrupp, line 3509: rest of file uses style `i+1 == SYSSTATEMENTSRowFactory.SYSSTATEMENTS_CONSTANTSTATE`, i think we should stick to one style (same line 3507)

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -96,7 +96,7 @@ public void run() throws StandardException{
         NavigableSet<Splice_DD_Version> keys=scripts.navigableKeySet();
         for(Splice_DD_Version version : keys){
             if(currentVersion!=null){
-                if(ddComparator.compare(version,currentVersion)<0){
+                if(ddComparator.compare(version,currentVersion)<=0){
```
martinrupp, line 99: i'm new, how is this related to the change in sys.sysstatements ?

yxia92: @martinrupp You are right, this change is not related to the update for the sys.sysstatements table. It is a fix for a minor issue with our upgrade process in general. 
Each upgrade script is associated with a version (let's call it S) which is usually the database version this script is introduced. 
In principal, if the currently installed database version (let's call it C) is already equal or above the version S, we don't need to call this upgrade script as the change should have already been applied. 
Before the fix, the upgrade script is still possible to be executed when C is the same as S.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
yxia92: I made the change to make the style consistent. Thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
yxia92: Done!

PR #3552, opened 2020-05-08
DB-9465 external tables shouldn't write when only reading

martinrupp:
SparkDataSetProcessor.getExternalFileSchema was copying one file out of the location so that the schema infer is only looking at one file (performance optimization). In the case that we can't access that file, we fall back to the normal behavior.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkDataSetProcessor.java
@@ -465,19 +465,42 @@ public class SparkDataSetProcessor implements DistributedDataSetProcessor, Seria
             if (!mergeSchema) {
                 fs = FileSystem.get(URI.create(location), conf);
                 String fileName = getFile(fs, location);
-                if (fileName != null) {
-                    temp = new Path(location, TEMP_DIR_PREFIX + "_" + UUID.randomUUID().toString().replaceAll("-",""));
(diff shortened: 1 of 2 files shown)


jpanko1:

spliceci:
TEST FAILED

martinrupp:

spliceci:
TEST FAILED

martinrupp:

spliceci:
TEST SUCCEEDED +1

liprais:
testing result by comparing strings is easily breakable especially when the header of column name is involved, just my 2 cents

martinrupp:
> testing result by comparing strings is easily breakable especially when the header of column name is involved, just my 2 cents

i'm partly on your side. We should have better compare methods like e.g.

```
result->AssertIntCol( /*colid* 0, new int[]{ 1, 2, 3 } );
result->AssertStringCol( /*colid* 1, new String[]{ "one", "two", "three" } );

result->AssertCount( 3 );

result->AssertHashCol( /*colid*/ 0, 0xAF878945 );
result->AssertHashCol( /*colid*/ 1, 0x3893EFDC );

result->AssertHash( 0x89403FAE ); // combined hash of all columns
```

We'll check to fix that in the future. Same is also a bit true for creating of parquet/orc/csv files, inserting into that, that should be done in a much more parameterized way, like check all types and type combinations, null / non null, etc.. In general, tests should be parameterized by filetype, and not have a test testWriteParquet and then testWriteOrc. So, some things to do, will gradually improve over time.

PR #3558, opened 2020-05-11
DB-9540 Make RegionLoads use bytes instead of MBs

dgomezferro:
Fix RegionSplitsIT and don't require /BAD directory to exist for
SpliceRegionAdminIT

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/storage/HPartitionLoad.java
+++ b/hbase_storage/src/main/java/com/splicemachine/storage/HPartitionLoad.java
@@ -19,15 +19,13 @@ package com.splicemachine.storage;
  *         Date: 1/7/16
  */
 public class HPartitionLoad implements PartitionLoad{
-    private final int storefileSizeMB;
-    private final int memStoreSizeMB;
-    private final int storefileIndexSizeMB;
+    private final long storefileSize;
+    private final long memStoreSize;
     private final String name;
 
-    public HPartitionLoad(String name,int storefileSizeMB,int memStoreSizeMB,int storefileIndexSizeMB){
-        this.storefileSizeMB=storefileSizeMB;
-        this.memStoreSizeMB=memStoreSizeMB;
-        this.storefileIndexSizeMB=storefileIndexSizeMB;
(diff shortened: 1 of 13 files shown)


hbase_storage/src/main/java/com/splicemachine/storage/HPartitionLoad.java
```diff
@@ -19,15 +19,13 @@
  *         Date: 1/7/16
  */
 public class HPartitionLoad implements PartitionLoad{
-    private final int storefileSizeMB;
-    private final int memStoreSizeMB;
-    private final int storefileIndexSizeMB;
+    private final long storefileSize;
+    private final long memStoreSize;
```
jyuanca, line 23: Why storefileIndexSize is removed?

hbase_storage/src/main/java/com/splicemachine/storage/HServerLoad.java
```diff
@@ -60,7 +60,7 @@ public long totalRequests(){
         for(Map.Entry<byte[],RegionLoad> regionLoad:regionsLoad.entrySet()){
             String name = Bytes.toString(regionLoad.getKey());
             RegionLoad rl = regionLoad.getValue();
-            PartitionLoad pl = new HPartitionLoad(name,rl.getStorefileSizeMB(),rl.getMemStoreSizeMB(),rl.getStorefileIndexSizeMB());
+            PartitionLoad pl = new HPartitionLoad(name,rl.getStorefileSizeMB()*1024,rl.getMemStoreSizeMB()*1024);
```
yxia92, line 63: Should we multiple 1024*1024 to get the number in bytes?

hbase_storage/src/main/java/com/splicemachine/storage/HPartitionLoad.java
dgomezferro: It wasn't used

hbase_storage/src/main/java/com/splicemachine/storage/HServerLoad.java
dgomezferro: Good catch @yxia92 , addressed in a new commit

PR #3577, opened 2020-05-15
DB-9374: refetch region locations if cache is stale

jyuanca:


hbase_storage/src/main/java/com/splicemachine/storage/ClientPartition.java
```diff
@@ -157,6 +157,9 @@ public boolean checkAndPut(byte[] key,byte[] family,byte[] qualifier,byte[] expe
                 }
             }
             List<HRegionLocation> tableLocations = ((ClusterConnection) connection).locateRegions(tableName, !refresh, false);
+            if (outdated(tableLocations)) {
+                tableLocations = ((ClusterConnection) connection).locateRegions(tableName, refresh, false);
```
yxia92, line 161: Is it better to use "false" explicitly than the variable "refresh" to force not to use the cache?

PR #3579, opened 2020-05-16
DB-9562 NSDS Update and Delete should handle Partial Rows and Empty Inputs

jpanko1:


jpanko1:

spliceci:
TEST FAILED

jpanko1:

jpanko1:

spliceci:
TEST FAILED

spliceci:
TEST FAILED

splice_spark/src/main/spark2.1/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
```diff
@@ -39,10 +39,16 @@ import org.apache.hadoop.hbase.util.Bytes
...
 import org.apache.spark.scheduler.{SparkListener, SparkListenerApplicationEnd}
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 
+@SerialVersionUID(20200513211L)
 object Holder extends Serializable {
   @transient lazy val log = Logger.getLogger(getClass.getName)
 }
+@SerialVersionUID(20200513212L)
+@SuppressFBWarnings(value = Array("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD"), justification = "Need to set SpliceClient.connectionString")
+@SuppressFBWarnings(value = Array("NP_ALWAYS_NULL"), justification = "These fields usually are not null")
```
dgomezferro, line 50: What fields was it complaining about?

splice_spark2/src/main/spark2.2/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
```diff
@@ -595,11 +595,13 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
...
   
   /** Schema string built from JDBC metadata. */
-  def schemaString(schemaTableName: String): String =
+  def schemaString(schemaTableName: String, schema: StructType = new StructType()): String =
     SpliceJDBCUtil.retrieveColumnInfo(
       new JDBCOptions(Map(
         JDBCOptions.JDBC_URL -> url,
         JDBCOptions.JDBC_TABLE_NAME -> schemaTableName))
+    ).filter(
+      col => schema.isEmpty || schema.exists( field => field.name.toUpperCase.equals( col(0).toUpperCase ) )
```
dgomezferro, line 618: Does this work for quoted columns? 

`create table a ("notEQUAL" int, "NOTequal" int);`

splice_spark2/src/main/spark2.2/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
```diff
@@ -577,7 +577,7 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
    * @param schema
    * @param schemaTableName table to delete from
    */
-  def delete(rdd: JavaRDD[Row], schema: StructType, schemaTableName: String): Unit = {
+  def delete(rdd: JavaRDD[Row], schema: StructType, schemaTableName: String): Unit = if( !rdd.isEmpty ) {
```
dgomezferro, line 594: This could force a computation of the RDD, ideally we'd want to make sure the RDD is only computed once. 
What's the error if the RDD is empty?

splice_spark2/src/main/spark2.2/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
jpanko1: Verified now that it works for unquoted columns but not quoted columns.  I’ll file a jira.

splice_spark2/src/main/spark2.2/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
jpanko1: The delete function (and it's the same for update) calls modifyOnKeys() which passes the rdd to sendData() which would put the data into Kafka, if there was data.  But in this case sendData() returns after having no data to pass, and then the sql is created and it hangs at 
`statement.executeUpdate(sql)`.  The code in the DB is probably polling Kafka when no data was put on the topic.  It looks like KafkaReadFunction would go into an infinite loop polling Kafka:
``` 
                    ConsumerRecords<Integer, Externalizable> records = null;
                    while (records == null || records.isEmpty()) {
                        records = consumer.poll( java.time.Duration.ofMillis(1000) );
                        if (TaskContext.get().isInterrupted()) {
                            consumer.close();
                            throw new TaskKilledException();
                        }
                    }
```

splice_spark/src/main/spark2.1/com/splicemachine/spark/splicemachine/SplicemachineContext.scala
jpanko1: Wrote https://splicemachine.atlassian.net/browse/DB-9580 and listed the issues in the jira.

splice_spark2/src/main/spark2.2/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
jpanko1: Wrote https://splicemachine.atlassian.net/browse/DB-9581

PR #3586, opened 2020-05-19
DB-9558 Fix NPE caused by RemoteQueryClientImpl.hasLOBs

arnaud-splice:


arnaud-splice:
This is causing some AssertError because getResultDescription sometimes returns null. Let's not merge that just yet.

PR #3600, opened 2020-05-23
DB-9597 NSDS v2 Build Jar

jpanko1:
Requires -Dnsds2jar to be added to the mvn command so that the jar will be built.

Example: mvn -P core,cdh6.3.0,ee -DskipTests -Dnsds2jar clean install

The jar will be at splice_spark2/target/splice_spark2-3.1.0.1959-SNAPSHOT-shaded.jar

The jar size in my env was about 120 MB.

### DIFF
--- a/splice_spark/pom.xml
+++ b/splice_spark/pom.xml
@@ -312,7 +312,7 @@
     </build>
     <profiles>
         <profile>
-            <id>dbaas</id>
+            <id>nsdsjar</id>
             <build>
                 <plugins>
                     <plugin>
--- a/splice_spark2/pom.xml
+++ b/splice_spark2/pom.xml
@@ -269,12 +269,7 @@
     </build>
     <profiles>
         <profile>
-            <id>dbaas</id>
-            <activation>
-                <property>
-                    <name>nsds2jar</name>
-                </property>
-            </activation>
+            <id>nsds2jar</id>
             <build>
                 <plugins>
                     <plugin>
--- a/splice_spark2/pom.xml
+++ b/splice_spark2/pom.xml
@@ -270,6 +270,11 @@
     <profiles>
         <profile>
             <id>dbaas</id>
+            <activation>
+                <property>
+                    <name>nsds2jar</name>
+                </property>
+            </activation>
             <build>
                 <plugins>
                     <plugin>


OlegMazurov:
What value does the flag add? How is it different from enabling the dbaas profile with -Pdbaas?

jpanko1:
@OlegMazurov The flag is independent of the profile, so that the jar can be built with other profiles like cdh6.3.0 or hdp3.1.0

OlegMazurov:
All the flag does is activate profile _dbaas_. My point was that the profile can already be activated with _-Pdbaas_. (It can be combined with other profiles, f.e.: _-Pcdh6.3.0,dbaas_).
Presumably, the _dbaas_ profile does more than just build the NSDS2 jar. The name of the property is misleading in that regard.

jpanko1:
In my standalone, running
`mvn -Pcore,cdh6.3.0,dbaas,ee -DskipTests clean install`
or
`mvn -Pcore,dbaas,ee -DskipTests clean install`
fails with compilation errors in hbase_storage.

The dbaas profile in spliceengine/pom.xml has older versions of hbase, hadoop, spark, kafka, etc, than the cdh6.3.0 and hdp3.1.0 profiles.

Meanwhile, the mvn command in the first comment of this PR builds successfully.

OlegMazurov:
I see. So, we use the _dbaas_ profile in the parent pom.xml, where it's activated by _-Denv=dbaas_ and is totally broken as it specifies dependencies on CDH 5.14.0 (that's the reason the build fails). We also use the _dbaas_ profile in _splice_spark/pom.xml_ and _splice_spark2/pom.xml_ to build a shaded version of the jar file.
Now the question is: why is _splice_spark2_ different or special? If the shaded version is needed for DBaaS, shouldn't the profile be fixed in the parent _pom.xml_ before anything else? You seem to be introducing a back door for one particular feature leaving the whole thing broken.

I'm probably lacking context here. From the current broken state, I tend to conclude that nobody actually uses the _dbaas_ profile. Maybe we should remove it and if the shaded splice-spark2 jar is still needed then build it properly without hijacking irrelevant profiles.

jpanko1:
Maybe dbaas removal can be covered in a separate Jira and PR, as dbaas appears in several pom files.

In splice_spark2/pom.xml, I can change _dbaas_ to _nsds2jar_ and remove the activation block for _nsds2jar_.

Would that be an appropriate solution?

OlegMazurov:
That would be fine, yes.

jpanko1:
Updated splice_spark2/pom.xml

Now it can be built like:
`mvn -P core,cdh6.3.0,ee,nsds2jar -DskipTests clean install`

Also made equivalent change in splice_spark/pom.xml so that it could be built if needed.

jpanko1:
For dbaas removal, added https://splicemachine.atlassian.net/browse/DB-9611 -- as discussed, it's not required for this PR.

jpanko1:

spliceci:
TEST FAILED

spliceci:
TEST FAILED

PR #3612, opened 2020-05-27
DB-9610 fix start-splice-cluster on linux, add docker build&start scripts

martinrupp:
- fixed issue with nc in start-splice-cluster on linux

used base image for docker is maven:3.6.3-jdk-8 .

both docker_run.sh and docker_start-splice-cluster.sh do:

- to be able to sqlshell.sh into the container, we are exposing the containers' port 1527
as port 2527 for the host, so you can connect from host e.g. with
  ./sqlshell.sh -h localhost -p 2527

- to re-use the $HOME/.m2/settings.xml and the already downloaded
maven dependencies, we map the path $HOME/.m2 of host to
 /root/.m2 inside of the container (since $HOME=/root, this mimicks $HOME/.m2 in the container)

- to have access to the source, we map $(pwd) to /usr/src/
- we cd into /usr/src in the container.

With these preparation steps, you can then just instantiate mvn,
short scripts are:

build:
./docker_run.sh mvn install -Pcdh6.3.0 -DskipTests

start splice cluster + sqlshell:
./docker_start-splice_cluster.sh -b

To get a normal bash inside the container:
./docker_run.sh /bin/bash

ATTENTION: Because of the path mapping of $HOME/.m2 and $(pwd), this means the docker
container will MODIFY these directories on the host.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,8 @@
+FROM maven:3.6.3-jdk-8
+
+# nc is used in start-splice-cluster to check if connection is opened -> service is ready
+RUN apt-get update && apt-get install -y netcat
+
+COPY entrypoint.sh /usr/local/bin/se-entrypoint.sh
+RUN chmod +x /usr/local/bin/se-entrypoint.sh
+ENTRYPOINT [ "/usr/local/bin/se-entrypoint.sh" ]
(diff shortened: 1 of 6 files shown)


martinrupp:

spliceci:
TEST FAILED

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #3619, opened 2020-05-28
DB-9613 DefaultSourceIT Insertion with Sampling Fix

jpanko1:
Setting hbase.rpc.timeout to avoid rpc timeouts in Jenkins.

The reason for setting hbase.rootdir as a system property: the sqlContext.read calls returning the dataframes are implemented as calls to SplicemachineContext.df() in SpliceRelation.buildScan().  The df function uses an EmbeddedConnection to query the data from the db.  Somewhere in the call stack, SplitRegionScanner.createAndRegisterClientSideRegionScanner() is called and it looks for hbase.rootdir in config files or in system properties.  It wasn’t getting the correct value for hbase.rootdir.  So the code was probably reading rows from the memstore, but didn’t have the right path to HBase to be able to read the additional rows stored in HFiles, and so it wasn't returning all of the rows.  The solution for DefaultSourceIT was to set the value in system properties at the beginning of the test.  In normal environments, hbase-site.xml in the driver/executor extraClassPath would provide hbase.rootdir.  (Thanks to Jun for troubleshooting the cause of not all rows being retrieved.)

### DIFF
--- a/splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceIT.scala
+++ b/splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceIT.scala
@@ -118,6 +118,10 @@ class DefaultSourceIT extends FunSuite with TestContext with BeforeAndAfter with
   }
 
   test("insertion with sampling") {
+    val userDir: String = System.getProperty("user.dir")
+    val hbaseDir = new File(userDir).getParent + "/platform_it/target/hbase"
+    System.setProperty("hbase.rootdir", hbaseDir)
+
     val conn = JdbcUtils.createConnectionFactory(internalJDBCOptions)()
     conn.createStatement().execute("create table TestContext.T(id INTEGER NOT NULL GENERATED ALWAYS AS IDENTITY (START WITH 1, INCREMENT BY 1), c1 double, c2 double, c3 double, primary key(id))")
     conn.createStatement().execute("insert into TestContext.T(c1,c2,c3) values (100, 100, 100), (200, 200, 200), (300, 300, 300), (400, 400, 400)");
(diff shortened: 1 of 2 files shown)


jpanko1:
For the test setup issue that caused the failure, entered https://splicemachine.atlassian.net/browse/DB-9616

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceIT.scala
```diff
@@ -117,9 +117,14 @@ class DefaultSourceIT extends FunSuite with TestContext with BeforeAndAfter with
     assert(newDF.count == 20)
   }
 
-  test("insertion with sampling") {  // DB-9395
+  test("insertion with sampling") {
     val userDir: String = System.getProperty("user.dir")
-    val dataDir = userDir+"/src/test/data/lineitem.csv";
+    val hbaseDir = new File(userDir).getParent + "/platform_it/target/hbase"
```
dgomezferro, line 122: Would it be possible to make hbase-site.xml be part of the classpath of the ITs? I'm worried about other tests failing in the same way
We write hbase-site.xml to platform_it/target when standing up standalone, I think it would be a matter of adding that file to the ITs classpath

jpanko1:
Now addressing hbase.rpc.timeout in DB-9675 https://github.com/splicemachine/spliceengine/pull/3674

This PR will now be for only the hbase-site.xml fix.

jpanko1:

spliceci:
TEST FAILED

jpanko1:

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/DefaultSourceIT.scala
jpanko1: Done in commit [9fb55e9](https://github.com/splicemachine/spliceengine/pull/3619/commits/9fb55e9202962c914d4ab8aca4884771f49ae943)

spliceci:
TEST SUCCEEDED +1

PR #3622, opened 2020-05-29
DB-9606 Kafka Topic Deletion in SplicemachineContext.

jpanko1:


jpanko1:

spliceci:
TEST SUCCEEDED +1

splice_spark2/src/main/scala/com/splicemachine/nsds/kafka/KafkaUtils.scala
```diff
@@ -13,14 +28,14 @@ import scala.collection.JavaConverters._
 object KafkaUtils {
   def messageCount(bootstrapServers: String, topicName: String): Long = {
     val props = new Properties
-    val groupId = "spark-consumer-s2s-ku"
+    val groupId = "spark-consumer-nsdsk-ku"
```
martinrupp, line 31: this should be a const that is used in both messageCount functions. Or have a helper function to remove the copied code from both

jpanko1: Addressed in commit [e0977a2](https://github.com/splicemachine/spliceengine/pull/3622/commits/e0977a26b345440115cc0c9d97236d05412ac59b).

jpanko1:
> i don't really understand where these kafka consumerIds come from and why they changed, but looks ok to me

In "spark-consumer-nsdsk-ku", the _nsdsk_ is based on the package name and the _ku_ is based on the class name.  I changed it because I moved the class to a different package.  Unique ids help identify the consumer in logs.

PR #3627, opened 2020-06-02
DB-9572 add useOLAP=X as query and session config/hint

martinrupp:
Note that useOLAP is *added*. useSpark=X still works.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
@@ -41,7 +41,7 @@ import java.sql.SQLException;
  */
 public interface SessionProperties {
     enum PROPERTYNAME{
-        USESPARK(0),
+        USEOLAP(0),
         DEFAULTSELECTIVITYFACTOR(1),
         SKIPSTATS(2),
         RECURSIVEQUERYITERATIONLIMIT(3),
@@ -75,11 +75,13 @@ public interface SessionProperties {
     static Pair<PROPERTYNAME, String> validatePropertyAndValue(Pair<String, String> pair) throws StandardException {
         String propertyNameString = pair.getFirst();
         SessionProperties.PROPERTYNAME property;
+        if( propertyNameString.equalsIgnoreCase("useSpark" ) )
+            propertyNameString = "USEOLAP";
         try {
             property = SessionProperties.PROPERTYNAME.valueOf(propertyNameString);
         } catch (IllegalArgumentException e) {
             throw StandardException.newException(SQLState.LANG_INVALID_SESSION_PROPERTY,propertyNameString,
-                    "useSpark, defaultSelectivityFactor, skipStats, olapQueue, recursiveQueryIterationLimit");
+                    "useOLAP, useSpark (deprecated), defaultSelectivityFactor, skipStats, olapQueue, recursiveQueryIterationLimit");
         }
 
         String valString = pair.getSecond();
@@ -87,7 +89,7 @@ public interface SessionProperties {
             return new Pair(property, "null");
 
         switch (property) {
-            case USESPARK:
+            case USEOLAP:
             case SKIPSTATS:
             case DISABLE_TC_PUSHED_DOWN_INTO_VIEWS:
                 try {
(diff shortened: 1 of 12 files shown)


yxia92:
Please add "useOLAP" to the error message which lists all the possible hints at below exception: https://github.com/splicemachine/spliceengine/blob/76a891c9eeb7ac04d8afab4080ad8c769a7dddfb/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java#L723

martinrupp:
> Please add "useOLAP" to the error message which lists all the possible hints at below exception:
> 
> https://github.com/splicemachine/spliceengine/blob/76a891c9eeb7ac04d8afab4080ad8c769a7dddfb/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java#L723



> Please add "useOLAP" to the error message which lists all the possible hints at below exception:
> 
> https://github.com/splicemachine/spliceengine/blob/76a891c9eeb7ac04d8afab4080ad8c769a7dddfb/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java#L723

Done, thanks!

martinrupp:

spliceci:
TEST FAILED

martinrupp:

spliceci:
TEST FAILED

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
```diff
@@ -42,6 +42,7 @@
 public interface SessionProperties {
     enum PROPERTYNAME{
         USESPARK(0),
+        USEOLAP(0), // alias for USESPARK
```
yxia92, line 45: @martinrupp `useOLAP` needs to be added to SessionPropertiesImpl.setProperty() also, otherwise, `set session_property useOLAP=true` does not have any effect. Also, is it better to use different id for `useOLAP` and `useSpark`?  Currently the output of `values current session_property` returns both properties, which is a bit confusing:
```
splice> values current session_property;
1                                                                                                                                                                                                                                                               
USESPARK=true; USEOLAP=true;    
```

martinrupp: > useOLAP needs to be added to SessionPropertiesImpl.setProperty()

since both USESPARK and USEOLAP are defined to 0, i think it would also work like this, but maybe would more consistent to add it to the switch in SessionPropertiesImpl.setProperty.

I thought that having both the same id is better, because USESPARK and USEOLAP are really the same thing, so giving them different ids we need more caution in implementing. 

> Also, is it better to use different id for useOLAP and useSpark? 

I thought about this and actually think the way it's displayed in **values current session_property** now is good - given we're having now two values meaning the same. We could also remove one of the outputs USESPARK or USEOLAP, but as said, it's the same thing, which of those would we remove? If we had different values, people would assume different values mean different things, then  might start adding USESPARK=true **and** USEOLAP=true just to be sure to have it correctly.

martinrupp: There's other options for this as well:

1. useOLAP=useSpark like above, show both in session_property
2. useOLAP=useSpark like above, but only show USEOLAP=true; in session_property
3. have useOLAP a different config value than useSpark, results then in different outputs for session_property (when using useOlap, shows USEOLAP=true; in session_property, using useSpark, shows USESPARK=true; in session_property, when using both, shows both)

The reason i didn't choose 3) is that i think it might be users could be confused if useOlap is another thing as useSpark, resulting in people setting both useOLAP and useSPARK just to be sure (like it could be useSpark for OLAP, is there another OLAP executor?)
Reason i didn't choose 2) is that users might config useSpark and then it doesn't show up. Actually i think that is more preferable than 3).

There was a small discussion on Slack, and if i understand correctly, both @yxia92 and @dgomezferro would like to not have both be displayed and rather deprecate useSpark, that is: display useOLAP, but allow both useOLAP and still useSpark (option 2) ).

yxia92: @martinrupp yes, I prefer option 2 like you described(display useOLAP only, but allow both to support backward compatibility). Thanks!

martinrupp:

spliceci:
TEST FAILED

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #3628, opened 2020-06-02
DB-9624 sqlshell.cmd for Windows.

hatyo:


martinrupp:
hm there's multiple sqlshell.sh. the sqlshell for the sqlshell package is in spliceengine/assembly/common/src/main/resources/bin/sqlshell.sh and doesn't require mvn (using java only). That would be also cool to have.

hatyo:
Thanks for the comments Martin! I will work on addressing them ASAP.

hatyo:
Superseded by #3654

PR #3629, opened 2020-06-02
DB-9601 apply TC on predicates pushed from outside view/dt, also skip…

yxia92:
… broadcast join memory limit check if hinted

DB-9601 introduce session property


dgomezferro:
@yxia92 is this disabled by default? Since we have an easy way to disable it I think we should have it enabled by default.

yxia92:
@dgomezferro Yes, it make sense to turn it on by default. I'll make the code change. Thanks!

yxia92:

yxia92:
@dgomezferro I've pushed a commit to enable the logic by default.

spliceci:
TEST FAILED

yxia92:

spliceci:
TEST SUCCEEDED +1

PR #3636, opened 2020-06-03
DB-9449 Fix spark scan issue that missed a row during flush

arnaud-splice:
If a flush happened during a Spark scan, the flush token that is
returned by MemStoreFlushAwareScanner would be marked with timestamp 0
and would therefore be read before any other cell.
However, when the scanner fetches a row, it first reads from the mem
store, then from hfiles. The flush token would prevent us from reading
from hfiles for that row and we would resume the scan at the next row.

We addressed the issue by making sure that the flush token will be
sorted exactly after the last row that was being scanned, so that we can
properly merge data from MemStore and HFiles before addressing the flush
and resetting scanners

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
+++ b/hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
@@ -176,18 +176,12 @@ public class MemStoreFlushAwareScanner extends StoreScanner implements RegionSca
             return HBasePlatformUtils.scannerEndReached(scannerContext);
         }
         if (super.next(outResult,scannerContext)) {
-            if (didWeFlush()) {
-                SpliceLogUtils.warn(LOG, "Flush happened while scanning the row that we return. We might return incomplete results: %s", outResult);
-            }
             if (LOG.isTraceEnabled()) {
                 SpliceLogUtils.trace(LOG, "Next: returning " + outResult.size() +
                         ". mayHaveMoreCellsInARow=" + HRegionUtil.mayHaveMoreCellsInARow(scannerContext));
                 SpliceLogUtils.trace(LOG, "Next: actual output: %s" + outResult);
             }
             return true;
-        }
-        if (didWeFlush()) {
-            SpliceLogUtils.warn(LOG, "Flush happened after didWeFlush returned false");
         }
 
         if (LOG.isTraceEnabled())
--- a/hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
+++ b/hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
@@ -14,13 +14,9 @@
 
 package com.splicemachine.access.client;
 
-import java.io.IOException;
-import java.util.List;
-import java.util.NavigableSet;
-import java.util.concurrent.atomic.AtomicReference;
+import com.splicemachine.utils.SpliceLogUtils;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.HRegionInfo;
 import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.client.RegionInfo;
 import org.apache.hadoop.hbase.client.Scan;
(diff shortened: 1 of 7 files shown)


arnaud-splice:

hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -176,10 +176,19 @@ public boolean internalNext(List<Cell> outResult,ScannerContext scannerContext)
             return HBasePlatformUtils.scannerEndReached(scannerContext);
         }
         if (super.next(outResult,scannerContext)) {
-            if (LOG.isTraceEnabled())
-                SpliceLogUtils.trace(LOG, "Next: returning " + outResult.size());
+            if (didWeFlush()) {
+                SpliceLogUtils.warn(LOG, "Flush happened while scanning the row that we return. We might return incomplete results: %s", outResult);
+            }
```
dgomezferro, line 181: I think we can remove this check

hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
```diff
@@ -176,10 +176,19 @@ public boolean internalNext(List<Cell> outResult,ScannerContext scannerContext)
...
+            if (LOG.isTraceEnabled()) {
+                SpliceLogUtils.trace(LOG, "Next: returning " + outResult.size() +
+                        ". mayHaveMoreCellsInARow=" + HRegionUtil.mayHaveMoreCellsInARow(scannerContext));
+                SpliceLogUtils.trace(LOG, "Next: actual output: %s" + outResult);
+            }
             return true;
         }
+        if (didWeFlush()) {
+            SpliceLogUtils.warn(LOG, "Flush happened after didWeFlush returned false");
+        }
```
dgomezferro, line 191: This one too.

hbase_storage/src/main/java/com/splicemachine/access/client/MemstoreKeyValueScanner.java
```diff
@@ -59,16 +63,35 @@ public boolean advance(){
...
+            assert currentResult.rawCells().length == 1;
+            if (previousResult != null && previousResult.rawCells() != null) {
+                assert previousResult.rawCells().length > 0;
+                KeyValue previousCell = (KeyValue) previousResult.rawCells()[previousResult.rawCells().length - 1];
+                if (LOG.isTraceEnabled())
+                    LOG.trace("PREVIOUS CELL: " + previousCell);
+                if (CellUtil.matchingRows(currentResult.rawCells()[0], HConstants.EMPTY_START_ROW)) {
+                    byte[] rowId = Bytes.add(
+                            CellUtil.cloneRow(previousCell),
+                            CellUtil.cloneRow(currentResult.rawCells()[0]));
```
dgomezferro, line 84: We should append a NULL byte, not the RowId of the Flush cell.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -155,29 +155,29 @@ public boolean nextRaw(List<Cell> result) throws IOException {
...
-                    SpliceLogUtils.debug(LOG, "Flush occurred");
-                if (this.topCell != null) {
-                    if (LOG.isDebugEnabled())
-                        SpliceLogUtils.debug(LOG, "setting start row to %s", topCell);
-                    //noinspection deprecation
-                    scan.setStartRow(Bytes.add(CellUtil.cloneRow(topCell), new byte[]{0}));
-                }
+                    SpliceLogUtils.debug(LOG, "setting start row to %s", topCell);
+                //noinspection deprecation
+                scan.setStartRow(Bytes.add(CellUtil.cloneRow(topCell), Bytes.toBytes(0)));
```
dgomezferro, line 170: This is not correct, this appends 4 NULL bytes instead of a single NULL byte.

hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
arnaud-splice: Done

hbase_storage/src/main/java/com/splicemachine/access/client/MemStoreFlushAwareScanner.java
arnaud-splice: Done

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
arnaud-splice: Done

hbase_storage/src/main/java/com/splicemachine/access/client/MemstoreKeyValueScanner.java
arnaud-splice: Done

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3646, opened 2020-06-03
DB-9449 Fix spark scan issue that missed a row during flush (2.8)

arnaud-splice:
If a flush happened during a Spark scan, the flush token that is
returned by MemStoreFlushAwareScanner would be marked with timestamp 0
and would therefore be read before any other cell.
However, when the scanner fetches a row, it first reads from the mem
store, then from hfiles. The flush token would prevent us from reading
from hfiles for that row and we would resume the scan at the next row.

We addressed the issue by making sure that the flush token will be
sorted exactly after the last row that was being scanned, so that we can
properly merge data from MemStore and HFiles before addressing the flush
and resetting scanners


arnaud-splice:

spliceci:
TEST FAILED

PR #3653, opened 2020-06-05
DB-9651 Fix spark scan issue with partial rows during flush

arnaud-splice:
During Spark scan, flushing the mem store would prevent us from
completing the scan of a specific row.
That could be one of the following reasons::
 - the row was partly in the mem store, partly in hfiles and we only
    read the memstore part
 - the memstore scan returned a partial row

This fix introduces a buffer row, that is only returned once we are sure
that it is complete. If a flush happens, we ditch the buffer and we
scan this row from scratch

This PR reverts DB-9449 as the new solution fixes both DB-9449 and DB-9651


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3654, opened 2020-06-05
DB-9624 Ensure sqlshell.sh works on Windows

hatyo:
Ready for CR.


assembly/pom.xml
```diff
@@ -240,6 +240,9 @@
                                         <copy todir="target/SPLICEMACHINE-${project.version}.${envClassifier}.${parcel.patch}/bin">
                                             <fileset dir="common/src/main/resources/bin" includes="sqlshell.sh"/>
                                         </copy>
+                                        <copy todir="target/SPLICEMACHINE-${project.version}.${envClassifier}.${parcel.patch}/bin">
```
OlegMazurov, line 243: I don't think we should include _winsqlshell.cmd_ into Cloudera parcel.

assembly/pom.xml
```diff
@@ -644,6 +647,9 @@
                                                 <source>
                                                     <location>common/src/main/resources/bin/sqlshell.sh</location>
                                                 </source>
+                                                <source>
```
OlegMazurov, line 650: ditto

assembly/pom.xml
```diff
@@ -1188,6 +1194,9 @@
...
```
OlegMazurov, line 1197: ditto

assembly/pom.xml
hatyo: You're right, I thought we would ship it with all supported deliverables, but this is incorrect, we will just add it to the `sqlshell` profile. Thank you for the hint.

assembly/pom.xml
hatyo: ok :)

GETTING-STARTED.md
```diff
@@ -66,3 +66,7 @@ This will compile everything and start the database.
 Then to connect to the database:
 
 `./sqlshell.sh`
+
+Or if you are on Windows:
+
+`call sqlshell.cmd`
```
OlegMazurov, line 72: See comment regarding sqshell.sh below.

sqlshell.cmd
```diff
@@ -0,0 +1,14 @@
+@echo off
```
OlegMazurov, line 1: That would be a Splice developer sqlshell on Windows. Only there are no Splice developers on Windows.
I think this file should be removed for now.

hatyo: You're right, I will remove it (and revert the comment in GETTING-STARTED.md

GETTING-STARTED.md
hatyo: ok

hatyo:
OK Oleg, thanks for the comments, I am pushing a new commit addressing them

hatyo:
I made a stupid mistake in the branch name (DB9624 instead of DB-9624). If necessary, I can definitely create a new PR using the correct naming format.

hatyo:
Thanks a lot @OlegMazurov , can I bring in this change or should I ask Jenkins to verify it first? (although this is probably not necessary since no business logic code is touched)

assembly/common/src/main/resources/bin/winsqlshell.cmd
```diff
@@ -0,0 +1,280 @@
...
+   REM error, print help and exit
+   CALL :help
+   EXIT /B 1
+)
+
+REM set splice lib dir, for Windows we assume it is always local
+
+set SCRIPT_DIR=%SCRIPT_DIR:"=%
+SET SPLICE_LIB_DIR=%SCRIPT_DIR%lib
+echo SPLICE_LIB_DIR=%SPLICE_LIB_DIR%
```
martinrupp, line 145: needed?

assembly/common/src/main/resources/bin/winsqlshell.cmd
```diff
@@ -0,0 +1,280 @@
...
+    -Djavax.net.ssl.trustStore.ssl.trustStorePassword^=%CLIENT_SSL_TRUSTSTOREPASSWD%
+)
+
+call :message Running Splice Machine SQL shell
+call :message For help%: "splice> help;"
+REM todo: catch stderr messages and filter out Picked up JAVA_TOOL_OPTIONS   2>nul
+java -classpath %CLASSPATH% %GEN_SYS_ARGS% %SSL_ARGS% %IJ_SYS_ARGS% com.splicemachine.db.tools.ij %SCRIPT% 2>nul
+
+CALL :eof
+
```
martinrupp, line 245: this is returning and then printing ECHO %* again, resulting e.g. in
```
splice> exit;
-h srv131 -u bla -s password
```
at the end.

assembly/common/src/main/resources/bin/winsqlshell.cmd
hatyo: First two lines are needed, last echo can be removed. I removed it.

assembly/common/src/main/resources/bin/winsqlshell.cmd
hatyo: Right, thanks, fixed it.

hatyo:
> 
> 
> additionally it would be cool to have a super basic tutorial how to use (can be also a wiki page).
> like
> 
>     * make sure you have java installed (Java JRE is enough), e.g. go Command Prompt, enter `java`
> 
>     * extract file
> 
>     * go to command shell, cd sqlshell
>       `winsqlshell.cmd -h yourhost -u splice -P`

There is already a How-To page explaining how to use this interactive SQL shell in Windows. Moreover, similar to `sqlshell.sh`, there is a check for not only the existence of Java in PATH but also that its version >=1.8.

hatyo:

spliceci:
TEST FAILED

PR #3662, opened 2020-06-05
DB-9552: fix JarLoader

jyuanca:


jyuanca:

spliceci:
TEST FAILED

jyuanca:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/services/reflect/UpdateLoader.java
```diff
@@ -211,6 +205,7 @@ public Object run(){
 	Class loadClass(String className, boolean resolve) 
 		throws ClassNotFoundException {
 
+		boolean isRS = Thread.currentThread().getName().contains("DRDAConnThread");
```
dgomezferro, line 208: If running in a NLJ, union operation or other contexts they could be running in a different thread, we need a different way to identify RS vs Olap/Spark

jyuanca: Is it possible nested loop join may try loading jars?

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/968/

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedConnection.java
```diff
@@ -112,7 +112,9 @@
      */
     public static final SQLException NO_MEM =
         Util.generateCsSQLException(SQLState.LOGIN_FAILED, "java.lang.OutOfMemoryError");
-    
+
+	public static final ThreadLocal<Boolean> isHBaseJVM = new ThreadLocal<>();
```
dgomezferro, line 116: I don't think this should be a ThreadLocal, are we setting this property for every Thread that could access this field? If  a NLJ runs in a different thread that might not have executed bootDatabase() it will think it's not in a regionServer. This should be either a static volatile boolean / static AtomicBoolean or we'd need a different mechanism.

jyuanca: This field is only set and used during updateLoader initialization, and set a flag for the singleton UpdaterLoader whether it is running on a HBase. If any thread needs to load jar, it should use the shared singleton UpdateLoader instance, which always know where it is running. No thread needs to access the ThreadLocal variable and decide whether it is running on HBase

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/998/

jyuanca:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedConnection.java
dgomezferro: I see, thanks @jyuanca !

PR #3663, opened 2020-06-08
DB-9615 Add file logging to SQL shell.

hatyo:
- Add commands logto, logstop, and logclear which give the user
  the possibility to log SQL statements to a local file.
- logto requires quoting the file path, i.e. logto '/tmp/log.txt'

This is still WIP, but I still wanted to take your input on this change, specially the way I extended the ij parser rules for the new logging commands.


hatyo:
Hi Oleg, sure it makes sense to reuse logging, could you maybe provide me with a couple of pointers about Derby's client-side logging that you mentioned so I could have a look?

hatyo:
thanks for the explanation, I am not sure the JIRA ticket requirement aligns with normal logging, I thought it is e.g. similar to GDB's [file logging](https://sourceware.org/gdb/onlinedocs/gdb/Logging-Output.html). i.e. we want only to allow record user interactions with interactive shell and save it to a file.
To me this is different from normal (log4j) logging although we might argue that we could also leverage (log4j) logging to do the same, i.e. by having an explicit log-level for SQL statements and their output written somewhere so it can be picked up later using some log-aggregation framework.
What do you think?

dgomezferro:
@hatyo I agree this is different from "execution tracing" logging, and I think a separate feature independent of log4j makes sense (one less dependency too for such a simple feature). Other DBs' shells have similar features:
- postgresql: \out command
- oracle: SPOOL command https://docs.oracle.com/cd/B19306_01/server.102/b14357/ch12043.htm

hatyo:
I updated the PR and changed the commands as the following:
* `logto <filename>` --> `spool <filename>`
* `logstop` --> `spool stop`
* `logclear` --> `spool clear`

hatyo:

spliceci:
TEST SUCCEEDED +1

db-tools-ij/src/main/grammar/ij.jj
```diff
@@ -3374,6 +3378,7 @@ HelpStatement() :
 		while (st.hasMoreTokens()) {
 		    v.addElement(st.nextToken());
 		}
+		v.addElement("Something");
```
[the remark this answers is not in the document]
hatyo: I was experimenting with something and this slipped through, thanks for finding it out.

db-tools-ij/src/main/grammar/ij.jj
```diff
@@ -4043,7 +4061,42 @@ ijResult CP_DisconnectStatement() throws SQLException
 
 }
 
+ijResult LogToStatement() throws IOException
```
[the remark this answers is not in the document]
hatyo: That makes sense, thanks.

db-drda/src/main/java/com/splicemachine/db/impl/tools/sysinfo/Main.java
```diff
@@ -1027,52 +1014,42 @@ private static ZipInfoProperties checkDirectory(String dirname)
...
+            for (String infoName : infoNames) {
+                thisEntry = zf.getEntry(infoName);
+                if (thisEntry != null) {
                     break;
                 }
             }
-
             if (thisEntry == null)
             {
+                zf.close();
```
arnaud-splice, line 1036: You don't need this zf.close anymore

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
```diff
@@ -845,4 +820,26 @@ int getCurrentRowNumber(ResultSet rs)
...
+		logFileName = path;
+		File f = new File(logFileName);
+		f.createNewFile(); // no-op if file exists.
+		this.out = new LocalizedOutput(new ForkOutputStream(new FileOutputStream(f)));
+	}
+
+	public void stopSpooling() {
+		doSpool = false;
+		this.out.flush();
+		this.out = new LocalizedOutput(System.out);
```
dgomezferro, line 835: We should close the FileOutputStream, relying on finalizers is very finicky

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
```diff
@@ -320,6 +295,11 @@ private int runScriptGuts() {
 				out.flush();
 				command = commandGrabber[currCE].nextStatement();
 
+				if(doSpool) {
```
dgomezferro, line 298: Does this echo the commands to the terminal too? Is that what we want?

dgomezferro:
This is probably not the place for the discussion, but it seems one use case for Oracle's SPOOL is to create SQL scripts dynamically, which we won't be able to do because we are writing the original SQL query too.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
hatyo: ok

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
hatyo: From the JIRA ticket:
> they asked for a function that would allow them to log sqlshell commands AND results to a file on demand

Let's discuss improvements in a different JIRA maybe?

hatyo:
> This is probably not the place for the discussion, but it seems one use case for Oracle's SPOOL is to create SQL scripts dynamically, which we won't be able to do because we are writing the original SQL query too.

Yeah it makes sense to have more control about what's reported. I think Michael had some ideas and maybe we could address them in a another improvement ticket.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
dgomezferro: I mean this code seems to be echoing the typed command back to the terminal, so the users would see what they typed twice (once when they type it, once when it's echoed) and then the result. The spooled file would have the command once and the result once, which is correct. I might be wrong and we might not be echoing the command to the terminal again, but I wanted to make sure whether that's the case.

hatyo: Hi Daniel,
sorry I misread you comment, you're right, we're writing the command twice to tty, I will adapt the change accordingly.

dgomezferro: No worries! Change looks good, thanks!

PR #3668, opened 2020-06-10
DB-8190 DB-9604 Make Spark fetch and shuffle partitions configurable

dgomezferro:


dgomezferro:

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -70,14 +71,23 @@ private SpliceSpark() {} // private constructor forbids creating instances
...
     public static synchronized SparkSession getSession() {
         String threadName = Thread.currentThread().getName();
         if (!SpliceClient.isClient() && !threadName.startsWith("olap-worker-")) {
              // Not running on the Olap Server... raise exception. Use getSessionUnsafe() if you know what you are doing.
             throw new RuntimeException("Trying to get a SparkSession from outside the OlapServer");
         }
-        return getSessionUnsafe();
+        SparkSession result = sessions.get();
+        if (result == null) {
+            result = session.newSession();
```
jyuanca, line 87: Is it possible session is not initialized and null when getSession s called first time?

hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
```diff
@@ -153,6 +156,8 @@ private void submitBatch(int batch, int batchSize, int numPartitions, final Java
             public Object call() throws Exception {
                 SparkContext sc = SpliceSpark.getContextUnsafe().sc();
                 sc.setLocalProperties(properties);
+                if (shufflePartitions != null)
+                    sc.setLocalProperty(SQLConf.SHUFFLE_PARTITIONS().key(), shufflePartitions.toString());
```
yxia92, line 160: Why do we need to set shuffle partition property both here and in QueryJob?

dgomezferro: This wasn't needed, it was an initial attempt at fixing the problem that didn't work out, nice catch!

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
dgomezferro: Nice catch, I'm calling getSessionUnsafe() now to guarantee it's properly initialized

dgomezferro:

spliceci:
TEST FAILED

PR #3683, opened 2020-06-15
DB-9665 Clear ZNode for OlapServer app watcher

dgomezferro:
Restart OlapServer if software version changed

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
@@ -105,6 +105,7 @@ public class OlapServerSubmitter implements Runnable {
     private Path appStagingBaseDir;
     private String amKeytabFileName = null;
     private Thread keepAlive;
+    private OlapMessage.KeepAlive keepAlivePrototype;
 
     private Configuration conf;
 
@@ -119,6 +120,8 @@ public class OlapServerSubmitter implements Runnable {
         try {
             // Create yarnClient
             conf = HConfiguration.unwrapDelegate();
+
+            initKeepAlivePrototype();
 
             YarnClient yarnClient = YarnClient.createYarnClient();
             yarnClient.init(conf);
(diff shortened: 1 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -244,6 +244,24 @@ public void run() {
 
     }
 
+    private byte[] getKeepAlivePayload() {
+        DatabaseVersion version;
+        try {
+            version = JMXUtils.getLocalMBeanProxy(JMXUtils.SPLICEMACHINE_VERSION, DatabaseVersion.class);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
hatyo, line 253: Is it needed to wrap the exception in a `RuntimeException`, can we just remove the catch?

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -244,6 +244,24 @@ public void run() {
 
     }
 
+    private byte[] getKeepAlivePayload() {
+        DatabaseVersion version;
```
OlegMazurov, line 248: Does _version_ need to be re-read on every call? Can it be an instance, or even static, variable similar to _OlapServerMaster_?

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
```diff
@@ -166,10 +163,13 @@ public void run() {
...
+                            String keepAlivePath = root + HBaseConfiguration.OLAP_SERVER_PATH + HBaseConfiguration.OLAP_SERVER_KEEP_ALIVE_PATH + "/" + appName;
+
+                            byte[] payload = getKeepAlivePayload();
                             try {
-                                zk.create(path, payload, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL);
+                                // clean up node if it exists
+                                ZkUtils.safeDelete(keepAlivePath, -1);
+                                zk.create(keepAlivePath, payload, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.EPHEMERAL);
                             } catch (KeeperException ke) {
                                 if (ke.code().equals(NODEEXISTS)) {
```
OlegMazurov, line 174: Is this code possible after the change? Should we still ignore it?

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
dgomezferro: I removed that as part of the refactoring to address other comments

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
dgomezferro: I initialize a keepAlive message prototype once in the beginning with the version information and then use that to generate the payload with the time, so we only access the version once per OlapServer lifetime.

hbase_sql/src/main/java/com/splicemachine/olap/OlapServerSubmitter.java
dgomezferro: I removed it, good point

dgomezferro:

spliceci:
TEST SUCCEEDED +1

PR #3693, opened 2020-06-17
DB-9714 Support time travel SELECT statements with tx id

hatyo:
- add parser extensions to support AS OF clause.
- only the tx id is supported.
- propagate the transaction id down the execution tree to
  TableScanOperation.
- adapt code generation to pass the tx id to TableScanOperation
  at runtime.
- add tests

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
@@ -196,36 +196,30 @@ public class FromBaseTable extends FromTable {
     private AggregateNode aggrForSpecialMaxScan;
 
     private boolean isBulkDelete = false;
+
+    private long pastTxId = -1;
+
     @Override
     public boolean isParallelizable(){
         return false;
     }
 
     /**
-     * Initializer for a table in a FROM list. Parameters are as follows:
-     * <p/>
-     * <ul>
-     * <li>tableName            The name of the table</li>
-     * <li>correlationName    The correlation name</li>
-     * <li>derivedRCL        The derived column list</li>
-     * <li>tableProperties    The Properties list associated with the table.</li>
-     * </ul>
-     * <p/>
-     * <p>
-     * - OR -
-     * </p>
-     * <p/>
-     * <ul>
-     * <li>tableName            The name of the table</li>
-     * <li>correlationName    The correlation name</li>
-     * <li>updateOrDelete    Table is being updated/deleted from. </li>
-     * <li>derivedRCL        The derived column list</li>
-     * </ul>
+     * Initializer for a table in a FROM list.
+     * @param tableName The name of the table
+     * @param correlationName The correlation name
+     * @param rclOrUD update/delete flag or result column list
+     * @param propsOrRcl properties or result column list
+     * @param isBulkDeleteOrTxId bulk delete flag or past tx id.
      */
     @Override
-    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDelete){
+    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDeleteOrTxId){
+        if(isBulkDeleteOrTxId instanceof Boolean) {
+            this.isBulkDelete = (Boolean) isBulkDeleteOrTxId;
+        }else if(isBulkDeleteOrTxId instanceof Long){
+            this.pastTxId = (Long) isBulkDeleteOrTxId;
+        }
         init(tableName, correlationName, rclOrUD, propsOrRcl);
-        this.isBulkDelete = (Boolean) isBulkDelete;
     }
 
     @Override
@@ -2357,6 +2351,10 @@ public class FromBaseTable extends FromTable {
         // compute the default row
         numArgs += generateDefaultRow((ActivationClassBuilder)acb, mb);
 
+        // also add the past transaction id
+        mb.push(pastTxId);
+        numArgs++;
+
         return numArgs;
     }
 
@@ -3410,7 +3408,11 @@ public class FromBaseTable extends FromTable {
         if(niceIndexName!=null){
             cName = "IndexScan["+niceIndexName+"]";
         }else{
-            cName = "TableScan["+getPrettyTableName()+"]";
+            cName = "TableScan["+getPrettyTableName();
+            if(pastTxId >= 0){
+                cName += "timeTravelTx(" + pastTxId + ")";
+            }
+            cName += "]";
         }
         if(isMultiProbing())
             cName = "MultiProbe"+cName;
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
@@ -221,10 +221,11 @@ public class SQLParser
     private static final int    USING_CLAUSE = 1;
 
     // Defines for optional table clauses
-    private static final int    OPTIONAL_TABLE_CLAUSES_SIZE = 3;
+    private static final int    OPTIONAL_TABLE_CLAUSES_SIZE = 4;
     private static final int    OPTIONAL_TABLE_CLAUSES_TABLE_PROPERTIES = 0;
     private static final int    OPTIONAL_TABLE_CLAUSES_DERIVED_RCL = 1;
     private static final int    OPTIONAL_TABLE_CLAUSES_CORRELATION_NAME = 2;
+    private static final int    OPTIONAL_TABLE_CLAUSES_TXN_ID = 3;
 
     // Define for UTF8 max
     private static final int    MAX_UTF8_LENGTH = 65535;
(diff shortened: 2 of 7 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -12276,6 +12295,18 @@ windowDefinition(WindowList wl) throws StandardException :
     }
 }
 
+Long
+asOfClause() throws StandardException :
+{
+    Long txId = null;
+}
+{
+    LOOKAHEAD({(getToken(1).kind == AS && getToken(2).kind == OF)}) <AS> <OF> txId = exactNumber()
```
arnaud-splice, line 12304: Is this lookahead necessary?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -196,36 +196,30 @@
...
+     * @param tableName The name of the table
+     * @param correlationName The correlation name
+     * @param rclOrUD update/delete flag or result column list
+     * @param propsOrRcl properties or result column list
+     * @param isBulkDeleteOrTxId bulk delete flag or past tx id.
      */
     @Override
-    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDelete){
+    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDeleteOrTxId){
+        if(isBulkDeleteOrTxId instanceof Boolean) {
```
arnaud-splice, line 217: I think a dedicated parameter would be cleaner instead of shoehorning a long in a boolean parameter.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -3410,7 +3408,11 @@ private String getClassName(String niceIndexName) throws StandardException {
         if(niceIndexName!=null){
             cName = "IndexScan["+niceIndexName+"]";
         }else{
-            cName = "TableScan["+getPrettyTableName()+"]";
+            cName = "TableScan["+getPrettyTableName();
+            if(pastTxId >= 0){
+                cName += "timeTravelTx(" + pastTxId + ")";
```
arnaud-splice, line 3413: Please add a space before timeTravelTx.
Also, you could simply write "as of" instead, to echo the query, but that's unimportant.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -10914,28 +10915,45 @@ tableReferenceTypes(boolean nestedInParens) throws StandardException :
...
     otc = optionalTableProperties()
     {
         otc[OPTIONAL_TABLE_CLAUSES_DERIVED_RCL] = derivedRCL;
         otc[OPTIONAL_TABLE_CLAUSES_CORRELATION_NAME] = correlationName;
+        otc[OPTIONAL_TABLE_CLAUSES_TXN_ID] = txnId;
         return otc;
     }
 |
+    LOOKAHEAD( { getToken(1).kind != AS || getToken(2).kind != OF })
     [ [ <AS> ]
```
dgomezferro, line 10934: Could we remove the lookahead with something like

`[ [ <AS> [ <OF> txnId = asOfClause() [ AS ] ] ...`

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: I think the general term for this feature is time travel. Therefore I would like to keep using this term for consistency.

hatyo: I will fix the extra-space.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: I wanted to follow the convention, that's all. Anyway, I will look into rewriting this.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
hatyo: I tried but I couldn't find a good alternative, I mean with your proposal we would bind `AS OF` with having an alias (line 10936) which is not necessary.
Ideally we should allow the following cases:
* `<ALIAS>`
* `AS <ALIAS>`
* `<ALIAS> AS OF ...`
* `AS <ALIAS> AS OF ...`
* `AS OF ...`
I couldn't find single rule for all these cases that why I broke it in two.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
hatyo: I am afraid it is, in order to resolve ambiguity for statements like:
`SELECT * FROM T AS x`
`SELECT * FROM T AS OF <something>`
Without the lookahead, I think the first rule (line 10934) will be matched causing an error. Therefor the lookahead is relevant.

hatyo:
@dgomezferro could you please check my comment?

hatyo:
I just saw extra comments from Daniel, I will address them in another commit.

hatyo:
> Some comments:
> 
> * We should raise an error if "as of" is specified for a view or an external table
Done, please check last commit. I am wondering whether we should actually only allow the time travel clause for `BASE_TABLE_TYPE` only and reject everything else.
> * I think changing the result set factory signatures could create upgrade issues. Isn't that right @yxia92 @jyuanca ?
@yxia92 , @jyuanca could you please check?

dgomezferro:
@hatyo see this for how to invalidate SPSs to avoid upgrade issues: https://github.com/splicemachine/spliceengine/pull/3739/files?diff=split&w=1#diff-dd25f9aa08c0583a38bc8e12207ae300R86

hatyo:
Thanks for the Information @dgomezferro , I will add a similar upgrade logic to my change to invalidate cached statements and SPS.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
yxia92: Yes @dgomezferro , you are right, the we need upgrade script to invalidate the saved plans in sys.systatements.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1042,6 +1038,17 @@ public ResultSetNode bindNonVTITables(DataDictionary dataDictionary,
                                           FromList fromListParam)throws StandardException{
         TableDescriptor tableDescriptor=bindTableDescriptor();
 
+        int tableType = tableDescriptor.getTableType();
+        if(pastTxId >= 0)
+        {
+            if(tableType==TableDescriptor.VIEW_TYPE) {
```
yxia92, line 1044: Can you add check for `tableDescriptor.getTableType()==TableDescriptor.WITH_TYPE` also?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -3411,7 +3422,11 @@ private String getClassName(String niceIndexName) throws StandardException {
         if(niceIndexName!=null){
             cName = "IndexScan["+niceIndexName+"]";
```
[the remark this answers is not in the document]
hatyo: Right, thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: Sure, I added a check and a test as well.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -10914,28 +10916,45 @@ tableReferenceTypes(boolean nestedInParens) throws StandardException :
...
     {
         otc = new Object[OPTIONAL_TABLE_CLAUSES_SIZE];
         otc[OPTIONAL_TABLE_CLAUSES_TABLE_PROPERTIES] = tableProperties;
         otc[OPTIONAL_TABLE_CLAUSES_DERIVED_RCL] = derivedRCL;
         otc[OPTIONAL_TABLE_CLAUSES_CORRELATION_NAME] = correlationName;
+        otc[OPTIONAL_TABLE_CLAUSES_TXN_ID] = txnId;
+        return otc;
+    }
+|
+    [txnId = asOfClause()]
```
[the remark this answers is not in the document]
hatyo: You're right, I removed it.

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #3702, opened 2020-06-19
DB-8781 Purge old updates

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -116,6 +120,27 @@ private void mutate(Cell element, TxnView txn) throws IOException {
                     assert deleteRightAfterFirstWriteTimestamp == 0;
                     deleteRightAfterFirstWriteTimestamp = element.getTimestamp();
                     break;
+                case USER_DATA: {
+                    long t = element.getTimestamp();
```
dgomezferro, line 124: Here we are considering the begin timestamp of the transaction that wrote the data, but we need to consider the commit timestamp instead (there's the same problem with deletes too, https://splicemachine.atlassian.net/browse/DB-9640 )

Consider the lowWaterMark is 1000 (and there is a transaction with timestamp 1000) 
If we are processing an update with timestamp = 900 and commit timestamp 1200, we are adding it to the map. Then we remove the baseline insert @ ts 100. 

When the lowWatermark transaction @ 1000 goes to read, there's no baseline anymore and the row is gone.

I think the race condition is pretty hard to hit but better to get it fixed.

arnaud-splice: Very good point. 
I created separate PRs for DB-9640 against 2.8/3.0/master that address the deleted rows issue.
I also updated this (DB-8781) to include the fix and extend it to updates.

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/server/PurgeConfigBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+package com.splicemachine.si.impl.server;
+
+public class PurgeConfigBuilder {
+    private PurgeConfig.PurgeLatestTombstone purgeLatestTombstone = null;
+    private Boolean respectActiveTransactions = null;
+    private Boolean purgeDeletes = null;
+    private Boolean purgeUpdates = null;
+
+    public PurgeConfigBuilder purgeDeletes(boolean b) {
+        assert purgeDeletes == null;
```
hatyo, line 24: why is it not allowed to tolerate situations like:
builder.purgeDeletes(true);
... // oops business logic shows that we can't do that
builder.purgeDeletes(false); // this will throw

hbase_storage/src/main/java/com/splicemachine/si/impl/server/PurgeConfig.java
```diff
@@ -23,34 +23,16 @@
 
     private final PurgeLatestTombstone purgeLatestTombstone;
     private final boolean respectActiveTransactions;
-    private final boolean purge;
+    private final boolean purgeDeletes;
+    private final boolean purgeUpdates;
 
-    public PurgeConfig(boolean purge, PurgeLatestTombstone purgeLatestTombstone, boolean respectActiveTransactions) {
-        this.purge = purge;
+    public PurgeConfig(boolean purgeDeletes, PurgeLatestTombstone purgeLatestTombstone, boolean purgeUpdates, boolean respectActiveTransactions) {
```
hatyo, line 29: respectActiveTransactions I find the name a bit weird :) but it is ok

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -52,7 +56,7 @@ public void mutate(List<Cell> rawList, List<TxnView> txns, List<Cell> results) t
         assert dataToReturn.isEmpty();
         assert results.isEmpty();
         assert maxTombstoneTimestamp == 0;
-        assert isSorted(rawList): "CompactionStateMutate: rawList not sorted";
+        //assert isSorted(rawList): "CompactionStateMutate: rawList not sorted";
```
hatyo, line 59: remove

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -61,15 +65,17 @@ public void mutate(List<Cell> rawList, List<TxnView> txns, List<Cell> results) t
...
-            }
-            results.addAll(dataToReturn);
-            assert isSorted(results) : "CompactionStateMutate: results not sorted";
+            Stream<Cell> stream = dataToReturn.stream();
+            if (shouldPurgeDeletes())
+                stream = stream.filter(not(this::purgeableDeletedRow));
+            if (shouldPurgeUpdates())
+                stream = stream.filter(not(this::purgeableOldUpdate));
+            stream.forEachOrdered(results::add);
+            //assert isSorted(results) : "CompactionStateMutate: results not sorted";
```
hatyo, line 74: dito

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
arnaud-splice: I would rather leave them there as they are quite handy when debugging

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
arnaud-splice: same

hbase_storage/src/main/java/com/splicemachine/si/impl/server/PurgeConfigBuilder.java
arnaud-splice: At the moment, we set all of that in the same place, so I'd rather be defensive about it and force users to only call that function once.

PR #3706, opened 2020-06-22
DB-9455 Refactor TimestampOracle to expose three distinct functions

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,78 +29,89 @@
...
-		long maxTS = _maxReservedTimestamp; // avoid the double volatile read
-		if (nextTS > maxTS) {
-			reserveNextBlock(maxTS);
-		}
-		_numTimestampsCreated.incrementAndGet(); // JMX metric
-		return nextTS;
-	}
+    public long getNextTimestamp(boolean refresh, boolean increment) throws TimestampIOException {
+        if (refresh) {
+            synchronized (this) {
```
OlegMazurov, line 92: **synchronized** does not address the issue described in the bug.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,78 +29,89 @@
...
+    }
+
+    private void reserveNextBlockIfNecessary(long nextTimestamp) throws TimestampIOException {
+        long maxTS = _maxReservedTimestamp; // avoid the double volatile read
+        if (nextTimestamp > maxTS) {
+            reserveNextBlock(maxTS);
+        }
+        // Check if reserving next block was enough
+        if (nextTimestamp > _maxReservedTimestamp) {
+            reserveNextBlockIfNecessary(nextTimestamp);
```
OlegMazurov, line 112: Recursion is not justified. A simple loop is clearer and can be used in the calling method:
`while (nextTimestamp  > _maxReservedTimestamp ) {
    reserveNextBlock(_maxReservedTimestamp);
}
`
I wouldn't worry about "double volatile read": it's not on the hot path and synchronization cost along with ZK read make its extra cost negligible.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,78 +29,89 @@
...
+        if (refresh) {
+            synchronized (this) {
+                _maxReservedTimestamp = timestampBlockManager.initialize();
+                _timestampCounter.set(_maxReservedTimestamp);
+            }
+        }
+
+        long nextTS = increment ? _timestampCounter.addAndGet(TIMESTAMP_INCREMENT): _timestampCounter.get();
+        long maxTS = _maxReservedTimestamp; // avoid the double volatile read
+        reserveNextBlockIfNecessary(maxTS);
```
OlegMazurov, line 100: nextTS ?

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,114 @@
...
+        long maxTS = _maxReservedTimestamp; // avoid the double volatile read
+        if (nextTS > maxTS) {
+            reserveNextBlock(nextTS);
+        }
+        _numTimestampsCreated.incrementAndGet(); // JMX metric
+        return nextTS;
+    }
+
+    private synchronized void reserveNextBlock(long nextTS) throws TimestampIOException {
+        while (nextTS > _maxReservedTimestamp) {
```
OlegMazurov, line 113: If **nextTS** is larger than **_maxReservedTimestamp** by several **blockSize**, we'll go to ZK several times.
Why do that when we can do the right thing in one take?

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,114 @@
...
+        if (timestamp > _timestampCounter.get()) {
+            timestampBlockManager.reserveNextBlock(timestamp);
+            _maxReservedTimestamp = timestamp;
+            _timestampCounter.set(timestamp);
+        }
+    }
+
+    public long getNextTimestamp() throws TimestampIOException {
+        long nextTS = _timestampCounter.addAndGet(TIMESTAMP_INCREMENT);
+        long maxTS = _maxReservedTimestamp; // avoid the double volatile read
```
OlegMazurov, line 104: Double volatile read is gone. Can get rid of maxTS.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,114 @@
...
-        synchronized(this)  {
-            if (_maxReservedTimestamp > priorMaxReservedTimestamp) return; // some other thread got there first
+    public long getCurrentTimestamp() {
+        return _timestampCounter.get();
+    }
+
+    public synchronized void bumpTimestamp(long timestamp) throws TimestampIOException {
+        if (timestamp > _timestampCounter.get()) {
+            timestampBlockManager.reserveNextBlock(timestamp);
+            _maxReservedTimestamp = timestamp;
```
OlegMazurov, line 97: By calling **reserveNextBlock(timestamp)** we have guaranteed that **timestamp <= _maxReservedTimestamp**. Reassigning **_maxReservedTimestamp** to  **timestamp** doesn't look right.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,114 @@
...
+            _timestampCounter.set(timestamp);
```
OlegMazurov, line 98: **synchronized** doesn't guard **_timestampCounter** . I'd still like to see **compareAndSet** here.

splice_timestamp_api/pom.xml
```diff
@@ -40,5 +40,11 @@
             <version>${spark.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.splicemachine</groupId>
+            <artifactId>splice_protocol</artifactId>
+            <version>3.1.0.1962-SNAPSHOT</version>
```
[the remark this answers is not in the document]
arnaud-splice: done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: Done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: Fixed

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: I'm not quite sure what else to put instead. @jyuanca, can you weigh in?

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
OlegMazurov: blockCount is not computed right (can be zero or even negative).

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,115 @@
...
+    }
+
+    private synchronized void reserveNextBlock(long nextTS) throws TimestampIOException {
+        while (nextTS > _maxReservedTimestamp) {
+            long blockCount = (_maxReservedTimestamp - nextTS) / blockSize + 1;
+            long nextMax = _maxReservedTimestamp + blockSize * blockCount;
             timestampBlockManager.reserveNextBlock(nextMax);
             _maxReservedTimestamp = nextMax;
-            _numBlocksReserved.incrementAndGet(); // JMX metric
+            for (int i = 0; i < blockCount; ++i)
```
OlegMazurov, line 117: Use _addAndGet()_

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,115 @@
...
-            if (_maxReservedTimestamp > priorMaxReservedTimestamp) return; // some other thread got there first
-            long nextMax = _maxReservedTimestamp + blockSize;
+    public long getCurrentTimestamp() {
+        return _timestampCounter.get();
+    }
+
+    public synchronized void bumpTimestamp(long newTimestamp) throws TimestampIOException {
+        if (newTimestamp > _timestampCounter.get()) {
+            timestampBlockManager.reserveNextBlock(newTimestamp);
+            _maxReservedTimestamp = newTimestamp;
```
OlegMazurov, line 97: _reserveNextBlock_ assigns a new value to _maxReservedTimestamp, here it's overwritten.
updateAndGet ensures assignment (it has a loop), which is more than needed here; a simple _compareAndSet_ with the value from the initial _timestampCounter.get() would suffice.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: Fixed

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: Done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,115 @@
...
+        if (nextTS > _maxReservedTimestamp) {
+            reserveNextBlock(nextTS);
+        }
+        _numTimestampsCreated.incrementAndGet(); // JMX metric
+        return nextTS;
+    }
+
+    private synchronized void reserveNextBlock(long nextTS) throws TimestampIOException {
+        while (nextTS > _maxReservedTimestamp) {
+            long blockCount = (_maxReservedTimestamp - nextTS) / blockSize + 1;
```
jyuanca, line 113: Change to nextTS  - _maxReservedTimestamp to make it positive? Just in case next Ts is much larger than _maxReservedTimestamp and blockCount becomes negative.

arnaud-splice: Done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: Done

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1408/

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,128 @@
...
+        long nextTS = _timestampCounter.addAndGet(TIMESTAMP_INCREMENT);
+        if (nextTS > _maxReservedTimestamp) {
+            reserveNextBlock(nextTS);
+        }
+        _numTimestampsCreated.incrementAndGet(); // JMX metric
+        return nextTS;
+    }
+
+    private synchronized void reserveNextBlock(long nextTS) throws TimestampIOException {
+        while (nextTS > _maxReservedTimestamp) {
```
OlegMazurov, line 126: "while" is redundant here, "if" would suffice

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
```diff
@@ -29,108 +29,128 @@
...
+
+    public synchronized void bumpTimestamp(long newTimestamp) throws TimestampIOException {
+        newTimestamp = roundUp(newTimestamp);
+        if (newTimestamp > _timestampCounter.get()) {
+            if (newTimestamp> _maxReservedTimestamp) {
+                reserveNextBlock(newTimestamp);
+            }
+            long currentTimestampCounter;
+            while ((currentTimestampCounter = _timestampCounter.get()) < newTimestamp) {
+                _timestampCounter.compareAndSet(currentTimestampCounter, newTimestamp);
```
OlegMazurov, line 111: _compareAndSet_ returns _true_ if successful, which also guarantees the condition is satisfied.
Consider:

	public synchronized void bumpTimestamp(long newTimestamp) throws TimestampIOException {
		newTimestamp = roundUp(newTimestamp);
		if (newTimestamp> _maxReservedTimestamp) {
			reserveNextBlock(newTimestamp);
		}
		long currentTimestampCounter;
		while ((currentTimestampCounter = _timestampCounter.get()) < newTimestamp) {
			if (_timestampCounter.compareAndSet(currentTimestampCounter, newTimestamp)) break;
		}
	}

arnaud-splice:
The error in IT seems to be the one that DB-10025 fixes. I'll wait until it's merged and will trigger ITs again.

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: done

splice_timestamp_api/src/main/java/com/splicemachine/timestamp/impl/TimestampOracle.java
arnaud-splice: done

dgomezferro (approved):
Nice cleanup!

PR #3752, opened 2020-07-01
DB-8286 Fix case sensitivity issue in SET_PURGE_DELETED_ROWS

arnaud-splice:


splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -1383,6 +1383,8 @@ public static void SYSCS_SAVE_SOURCECODE(String schemaName, String objectName, S
     }
 
     public static void SET_PURGE_DELETED_ROWS (String schemaName, String tableName, String enable) throws Exception{
+        schemaName = schemaName.toUpperCase();
```
dgomezferro, line 1386: Use EngineUtils.validateSchemat()/validateTable() instead, they deal with quoted names. This would also be needed in many other procedures here.

arnaud-splice:

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
arnaud-splice: Done. Short question though, validate{Schema, Table} do not seem to be idempotent. Are they?

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/995/

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
dgomezferro: No, they are not.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -693,9 +693,12 @@ public static void SYSCS_PERFORM_MAJOR_COMPACTION_ON_SCHEMA(String schemaName) t
      */
     public static void SYSCS_PERFORM_MAJOR_COMPACTION_ON_TABLE(String schemaName,String tableName) throws SQLException {
         LanguageConnectionContext lcc = (LanguageConnectionContext) ContextService.getContext(LanguageConnectionContext.CONTEXT_ID);
+        assert lcc != null;
         DataDictionary dd = lcc.getDataDictionary();
         // sys query for table conglomerate for in schema
         PartitionFactory tableFactory = SIDriver.driver().getTableFactory();
+        schemaName = EngineUtils.validateSchema(schemaName);
+        tableName = EngineUtils.validateTable(tableName);
```
dgomezferro, line 701: IT failed due to this, SYSCS_PERFORM_MAJOR_COMPACTION_ON_SCHEMA calls this function with tableName set to null. We could perform the validate here and move all the logic into a separate function that gets called both by SYSCS_PERFORM_MAJOR_COMPACTION_ON_TABLE and SYSCS_PERFORM_MAJOR_COMPACTION_ON_SCHEMA

arnaud-splice: Fixed!

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3770, opened 2020-07-06
DB-9764 Add IF NOT EXISTS option to CREATE TABLE statement

ascend1:


ascend1:

ascend1:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -4785,13 +4787,25 @@ declareTemporaryTableDeclaration() throws StandardException :
...
+    Token ifTok = null;
+    Token notTok = null;
+    Token existsTok = null;
+    Token ifTok2 = null;
+    Token notTok2 = null;
+    Token existsTok2 = null;
 }
 {
-    <DECLARE> <GLOBAL> <TEMPORARY> <TABLE> tableName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH)
+    <DECLARE> <GLOBAL> <TEMPORARY> <TABLE> [ ifTok = <IF> notTok = <NOT> existsTok = <EXISTS> ] tableName = qualifiedName(Limits.MAX_IDENTIFIER_LENGTH) [ ifTok2 = <IF> notTok2 = <NOT> existsTok2 = <EXISTS> ]
```
dgomezferro, line 4798: Is it possible to simplify this pattern a little bit? At least remove the "notTok" and "existsTok" variables, I don't think those are needed, and then have`boolean ifNotExists = ifTok != null || ifTok2 != null;`

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -1333,6 +1333,10 @@ public void removeStatement(GenericStatement statement) throws StandardException
     public PreparedStatement lookupStatement(GenericStatement statement) throws StandardException{
         GenericStorablePreparedStatement ps = getDataDictionary().getDataDictionaryCache().statementCacheFind(statement);
         if (ps==null) {
+            String sqlText = statement.getSource().toUpperCase();
+            if ((sqlText.startsWith("CREATE") || sqlText.startsWith("DECLARE")) && sqlText.contains("IF NOT EXISTS")){
+                return null;
+            }
```
dgomezferro, line 1339: Why do we need this?

ascend1: CREATE TABLE statements are cached if they have run through once. So if we run `CREATE TABLE IF NOT EXISTS T ...` twice, the second execution will bypass all checks in bind phase, including the newly added "if not exists" check. As a result, the execution returns an error. That should not be the case when "if not exists" presents.

It seems that CREATE SCHEMA statement is never cached, that's why "if not exists" works just fine there.

An alternative would be to pass this createBehavior all the way down to the engine and check there. Not sure that's a better way, though.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
ascend1: I'll give it a try. I'm not yet familiar with javacc grammar and got this from `CREATE SCHEMA IF NOT EXISTS ...` implementation.

ascend1: Done.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1002/

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
ascend1: I just found that it's better to remove the hack in statement cache and add the check in its constant action. Will improve it with a new commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1014/

spliceci:
TEST SUCCEEDED +1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1073/

PR #3771, opened 2020-07-06
DB-9715 AS OF should also accept a timestamp expression

hatyo:


db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ResultSetFactory.java
```diff
@@ -1012,7 +1013,8 @@ NoPutResultSet getTableScanResultSet(
 								String location,
 								int partitionByRefItem,
 								GeneratedMethod defaultRowFunc,
-								int defaultValueMapItem
+								int defaultValueMapItem,
+								GeneratedMethod pastTxFunctor
```
dgomezferro, line 1017: @yxia92 @jyuanca do we need to do something here regarding upgrades?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -196,36 +196,30 @@
...
+     * @param tableName The name of the table
+     * @param correlationName The correlation name
+     * @param rclOrUD update/delete flag or result column list
+     * @param propsOrRcl properties or result column list
+     * @param isBulkDeleteOrTxId bulk delete flag or past tx id expression.
      */
     @Override
-    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDelete){
+    public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDeleteOrTxId){
+        if(isBulkDeleteOrTxId instanceof Boolean) {
```
dgomezferro, line 217: I'm not a fan of the `isBulkDeleteOrTxId` parameter, would it be possible to add an extra parameter for the transaction id ?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -3410,7 +3431,11 @@ private String getClassName(String niceIndexName) throws StandardException {
         if(niceIndexName!=null){
             cName = "IndexScan["+niceIndexName+"]";
         }else{
-            cName = "TableScan["+getPrettyTableName()+"]";
+            cName = "TableScan["+getPrettyTableName();
+//            if(pastTxIdExpression >= 0){
```
dgomezferro, line 3435: Remove comments

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -10914,28 +10915,45 @@ tableReferenceTypes(boolean nestedInParens) throws StandardException :
...
 {
     otc = optionalTableProperties()
     {
         otc[OPTIONAL_TABLE_CLAUSES_DERIVED_RCL] = derivedRCL;
         otc[OPTIONAL_TABLE_CLAUSES_CORRELATION_NAME] = correlationName;
+        otc[OPTIONAL_TABLE_CLAUSES_TXN_ID] = txnIdExpression;
         return otc;
     }
 |
+    LOOKAHEAD( { getToken(1).kind != AS || getToken(2).kind != OF })
```
dgomezferro, line 10933: Is it possible to remove the look ahead here?

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
```diff
@@ -0,0 +1,101 @@
...
+            for(Cell kv : keyValues){
+                if(CellUtils.singleMatchingColumn(kv,V2TxnDecoder.FAMILY,V2TxnDecoder.DATA_QUALIFIER_BYTES)) {
+                    dataKv=kv;
+                }
+                if(ts <  kv.getTimestamp()) {
+                    ts = kv.getTimestamp(); // get last update
+                }
+            }
+            if(dataKv==null) return null;
+            long txnId= TxnUtils.txnIdFromRowKey(dataKv.getRowArray(),dataKv.getRowOffset(),dataKv.getRowLength());
```
dgomezferro, line 45: If I'm understanding this correctly, this is mapping the Txn begin timestamp to the instant when the transaction was last updated. This would be wrong for long running transactions:
1. txn1 is created at instant1
2. a long time passes
3. txn1 is committed at instant500
5. this function maps txnId 1 to instant 500 (it should map txnId 1 to instant 1)

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
```diff
@@ -0,0 +1,101 @@
...
+
+        @Override
+        public void close() throws IOException{
+            regionScanner.close();
+        }
+    }
+
+    @Override
+    public Pair<Long, Long> find(byte bucket, byte[] begin, boolean reverse) throws IOException {
+        Scan s = new Scan();
```
dgomezferro, line 87: You can specify the column you are interested in directly in the scan, so you don't have to check for the right KV.

mem_sql/pom.xml
```diff
@@ -268,12 +268,10 @@
                             <key>com.splicemachine.enableLegacyAsserts</key>
                             <value>true</value>
                         </systemProperty>
-                     <!-- Uncomment the following to dump activation class files for mem platform to the mem_sql directory.
```
dgomezferro, line 271: Recomment this

platform_it/pom.xml
```diff
@@ -1055,6 +1055,7 @@
                                 <argument>-Dderby.language.updateSystemProcs=false</argument>
                                 <!-- Setting the logStatementText option to true enables logging of all statements. -->
                                 <argument>-Dderby.infolog.append=true</argument>
+                                <argument>-Dderby.debug.true=DumpClassFile</argument>
```
dgomezferro, line 1058: comment this by default

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/TableScanOperation.java
```diff
@@ -175,10 +190,34 @@ public TableScanOperation(long conglomId,
         this.indexColItem=indexColItem;
         this.indexName=indexName;
         init();
+        if(pastTxFunctor != null) {
+            this.pastTx = MapToTxId((DataValueDescriptor)pastTxFunctor.invoke(activation));
+            if(pastTx == -1) {
+                pastTx = 0; // force going back to oldest transaction instead of ignoring it.
```
dgomezferro, line 196: Txn id 0 is used for the ROOT_TRANSACTION, which is a special transaction parent of all user level transactions. I don't think this is going to be problematic but I'd rather use a different transaction id other than 0, in case we think at some point this is actually the root transaction.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/TableScanOperation.java
```diff
@@ -175,10 +190,34 @@ public TableScanOperation(long conglomId,
...
+            }
+        } else {
+            this.pastTx = -1; // nothing is set, go ahead and use latest transaction.
+        }
         if(LOG.isTraceEnabled())
             SpliceLogUtils.trace(LOG,"isTopResultSet=%s,optimizerEstimatedCost=%f,optimizerEstimatedRowCount=%f",isTopResultSet,optimizerEstimatedCost,optimizerEstimatedRowCount);
     }
 
+    private long MapToTxId(DataValueDescriptor dataValue) throws StandardException {
```
dgomezferro, line 205: Method names should start with lowercase

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SelectTimeTravelIT.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private static final String SCHEMA = SelectTimeTravelIT.class.getSimpleName().toUpperCase();
+
+    @ClassRule
+    public static SpliceWatcher watcher = new SpliceWatcher(SCHEMA);
+
+    @ClassRule
+    public static SpliceSchemaWatcher spliceSchemaWatcher = new SpliceSchemaWatcher(SCHEMA);
+
+    private long GetTxId() throws SQLException {
```
dgomezferro, line 21: Method names start lowercase

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SelectTimeTravelIT.java
```diff
@@ -0,0 +1,253 @@
...
+        long result = -1;
+        ResultSet rs =  watcher.executeQuery("CALL SYSCS_UTIL.SYSCS_GET_CURRENT_TRANSACTION()");
+        Assert.assertTrue(rs.next());
+        result = rs.getLong(1);
+        Assert.assertFalse(rs.next());
+        return result;
+    }
+
+    static int counter = 0;
+    private static String GenerateTableName() {
```
dgomezferro, line 31: ditto

splice_si_api/src/main/java/com/splicemachine/si/impl/TxnUtils.java
```diff
@@ -30,10 +30,23 @@ private TxnUtils(){}
 		long beginTS = txnId & SIConstants.TRANSANCTION_ID_MASK;
 		byte[] rowKey = new byte[9];
 		rowKey[0] = (byte)((beginTS / SIConstants.TRASANCTION_INCREMENT) & (TRANSACTION_TABLE_BUCKET_COUNT-1));
+		rowKey[0] = (byte)(2);
```
dgomezferro, line 33: Remove

hbase_storage/src/test/java/com/splicemachine/si/impl/TimeTravelTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        MockTxnFinder AddTxn(int bucket, long txnId, String ts) throws ParseException {
+            assert bucket >= 0 && bucket < SIConstants.TRANSACTION_TABLE_BUCKET_COUNT;
+            txnTimeline.get(bucket).put(txnId, toTs(ts));
+            return this;
+        }
+
+        @Override
+        public Pair<Long, Long> find(byte bucket, byte[] begin, boolean reverse) throws IOException {
+            if(begin == null) {
```
martinrupp, line 45: `SortedMap<Long, Long> m = txnTimeline.get(bucket);` and `if(m.size() == 0) { return null`
can be done independent of if(begin == null) and if(reverse).

I think you can add at the beginning of this function
```
SortedMap<Long, Long> m = txnTimeline.get(bucket);
if(m.size() == 0) {
    return null;
}
```
and remove in other lines.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/SelectTimeTravelIT.java
```diff
@@ -0,0 +1,253 @@
...
+        return "T" + counter++;
+    }
+
+    @Test
+    public void testTimeTravelWorks() throws Exception {
+        String tbl = GenerateTableName();
```
martinrupp, line 37: do we really need this? maybe sth like "testTimeTravelWorksT" would be more intuitive? the SpliceWatcher will drop these tables automatically, so you can run the test multiple times.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: Right, I did it in the base change (DB-9714). The new signature looks like this:
`public void init(Object tableName,Object correlationName,Object rclOrUD,Object propsOrRcl, Object isBulkDelete, Object pastTxIdExpr)`

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: Done

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
hatyo: Would you please check my comment [here](https://github.com/splicemachine/spliceengine/pull/3693#discussion_r450398057)? I am not sure I can remove it actually but maybe I missed something.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
hatyo: You're right, what's a better way to correlate? maybe parse and check the timestamp of `COMMIT_QUALIFIER_BYTES` or `STATE_QUALIFIER_BYTES` (sorry maybe we already went through this once, but I just want to make sure).

hbase_storage/src/main/java/com/splicemachine/si/impl/region/TxnTimeTraveler.java
```diff
@@ -0,0 +1,153 @@
...
+                            TxnUtils.rowKeytoHbaseEscaped(toRowKeyWithBucket(before.getFirst(), i)),
+                            new Timestamp(before.getSecond()).toString(),
+                            before.getFirst().toString(),
+                            new Timestamp(ts).toString());
+                    return new Pair<>(slideToNextTx(closestTx), closestTS); // nothing, needle is too early: <needle> ....... [before ...... after]
+            }
+            after = last(i);
+            if(after.getSecond() < ts) {
+                if(LOG.isTraceEnabled())
+                    SpliceLogUtils.trace(LOG,"in bucket %d the last txn [%s (TS: %s), id %s] happened before the needle (%s) exiting!",
```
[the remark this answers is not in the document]
hatyo: Right, I don't know why I chose TRACE and not DEBUG, I will change it on DEBUG level then.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/TxnTimeTraveler.java
```diff
@@ -0,0 +1,153 @@
...
+                    return null;
+                }
+            }
+            changed = false;
+            middle = middle(left, right);
+            Pair<Long, Long> middleTx;
+            middleTx = txnFinder.find(bucket, middle, reversed);
+            reversed = false;
+            if(LOG.isTraceEnabled())
+                SpliceLogUtils.trace(LOG,"middle of range: [%s (TS: %s) id %s, %s (TS: %s)] id %s is %s (TS: %s) id %s",
```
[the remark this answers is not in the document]
hatyo: Yeah, I changed all of them back to DEBUG.

hbase_storage/src/main/java/com/splicemachine/si/impl/CoprocessorTxnStore.java
```diff
@@ -268,6 +268,17 @@ public int compare(TxnView o1,TxnView o2){
         }
     }
 
+    @Override
+    public long getTxnAt(long ts) throws IOException {
+        final TxnMessage.TxnAtRequest request=TxnMessage.TxnAtRequest.newBuilder().setTs(ts).build();
+        try(TxnNetworkLayer table = tableFactory.accessTxnNetwork()){
+            TxnMessage.TxnAtResponse result = table.getTxnAt(request);
+            return result.getTxnId();
+        }catch(Throwable throwable){
```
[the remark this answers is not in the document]
hatyo: Done.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
dgomezferro: I think the best option would be to scan for STATE_QUALIFIER_BYTES with ACTIVE value (0x00), since that's what's written when the transaction is first created. In principle there should only be 2 states at most (ACTIVE, superseded by COMMITTED or ROLLEDBACK) but I think I saw some cases where ROLLEDBACK was written multiple times. If there's such a case and the ACTIVE keyvalue is not available I'd ignore that row.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/TableScanOperation.java
hatyo: Done. I created a constant: `public static final long OLDEST_TIME_TRAVEL_TX = TRASANCTION_INCREMENT;` and I am using it here instead of simply using zero.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/TxnTimeTraveler.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    // When we use a past transaction P whose timestamp is T, we can read data which is committed before T
+    // i.e. all data with time < T will be visible, therefore, to be able to read the data of the past
+    // transaction itself, we slide to the next tx.
+    private long slideToNextTx(long tx) throws IOException {
+        if(tx == -1) { // special case for timestamp that happened before the first transaction.
+            return -1;
+        }
+        return tx + SIConstants.TRASANCTION_INCREMENT;
```
carolp-503, line 109: While  misspellings generally aren't noteworthy and this isn't a blocker for merging today, the literal here is spelled wrong. Could make searching for it harder in the future, so can we fix this in a future release?

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1274/

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #3777, opened 2020-07-08
DB-9171 Allow defining unique and PK constraint on same column(s)

hatyo:
- Some SQL generation tools (e.g. SQLAlchemy) generate a DDL 
  statement against SpliceMachine containing PK and UQ constraints
  on the same set of columns.
- The fix allow this syntax, but under the hood it ignores the UQ
  constraint since it is redundant.
- If a user attempts to add a redundant UQ constraint via ALTER table
  then we still throw an exception.
- Add tests and address SpotBug reported issues on affected files.


splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/CreateTableIT.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @ClassRule
+    public static SpliceSchemaWatcher spliceSchemaWatcher = new SpliceSchemaWatcher(SCHEMA);
+
+    static int counter = 0;
+    private static String GenerateTableName() {
+        return "T" + counter++;
+    }
+
+    private void ShouldContains(String tableName, List<List<Integer>> rows) throws SQLException {
```
dgomezferro, line 33: Please use lower case method names

hatyo: Thanks, I fixed them.

PR #3785, opened 2020-07-09
DB-9589 rely on schema name for checking sys tables.

hatyo:
- previously, the conglomerate ID was used to check whether a table
  is a system table being smaller than FIRST_USER_TABLE_NUMBER.
- this assumption becomes invalid after upgrade where sys tables
  are recreated with potentially larger conglomerate ID than
  FIRST_USER_TABLE_NUMBER.
- since the check fails, the wrong serializer is used with TIMESTAMP
  columns causing incorrect data to be shown.
- the fix makes the check look at the schema of the table, if it is
  "SYS" then we know for sure this is a system table regardless of
  its conglomerate ID.
- add unit test.


hatyo:
I think we have other potential issues related to this assumption:
- in `Vacuum.java` we do again this check:
```
if(tableConglom < DataDictionary.FIRST_USER_TABLE_NUMBER) {
                        if (LOG.isTraceEnabled()) {
                            LOG.trace("Ignoring system table: " + table.getTableName());
                        }
                        continue; //ignore system tables
                    }
```
which could fail after an upgrade? I could open another ticket for this.
Also, in `DataDictionaryCache` we have:
```
if (!dd.canReadCache(xactMgr) && conglomId>=DataDictionary.FIRST_USER_TABLE_NUMBER)
            // Use cache even if dd says we can't as long as it's a system table (conglomID is < FIRST_USER_TABLE_NUMBER)
            return null;
        if (LOG.isDebugEnabled())
            LOG.debug("conglomerateCacheFind " + conglomId);
```

yxia92:
For the scenario in Vacuum.java I think we lose some opportunity to skip additional checks for the stats table(with conglomerate id larger than DataDictionary.FIRST_USER_TABLE_NUMBER), but it is not critical as the stats table cannot be dropped by users. And for the scenario in DataDictionaryCache.java, we again miss an opportunity to use the cache under certain scenarios, but it is not a big concern to me either.

hatyo:
Thanks Yi for the comments, also thank you and Daniel for the CR.

hatyo:
> +1
> 
> Need PRs for 2.8 and 3.0

Done, [3.0](https://github.com/splicemachine/spliceengine/pull/3842) and [2.8](https://github.com/splicemachine/spliceengine/pull/3841)

PR #3788, opened 2020-07-09
DB-9820 Set default purge old updates to false

arnaud-splice:


jyuanca:
Can you explain why the default value changes from true to false

dgomezferro:
@jyuanca the idea is to be more conservative and test this internally for some time before enabling it by default on customers that get a recent build. We can still enable it for customers that want the feature even with some risk.

PR #3807, opened 2020-07-10
DB-9825 add SQL shell command to show clock in prompt

hatyo:
- Add a new SQL shell commands: `prompt clock on` and
  `prompt clock off` which shows a clock in the sqlshell
  prompt in client's local time.


db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
```diff
@@ -845,4 +846,8 @@ int getCurrentRowNumber(ResultSet rs)
 	public final Object run() {
 		return  getClass().getResourceAsStream(ProductGenusNames.TOOLS_INFO);
 	}
+
+	public void setPromptClock(boolean showPromptClock) {
```
arnaud-splice, line 850: Shouldn't that be static? Setting a static field from a nonstatic method is weird.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
```diff
@@ -664,13 +661,17 @@ void newResourceInput(String resourceName) {
 	 * used at each carriage return to show that it is still "live"
 	 * when it is reading multi-line input.
 	 */
-	static void doPrompt(boolean newStatement, LocalizedOutput out, String tag) 
+	 static void doPrompt(boolean newStatement, LocalizedOutput out, String tag)
 	 {
+	 	final SimpleDateFormat sdf = new SimpleDateFormat("yyyy.MM.dd HH:mm:ss");
```
arnaud-splice, line 666: Please use ISO 8601 delimiters for the date: yyyy-MM-dd

hatyo: Good point! thanks, I will fix it.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
hatyo: Fixed.

PR #3808, opened 2020-07-11
DB-9797 Restore DB-9651 with consistency fix

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

arnaud-splice:
Sure, we can do that!
In the meantime, I believe that current ITs take care of that:
DB-9651 transformed `updateTopCell` and replaced `if (response)  topCell = results.get(results.size() - 1);` with `if (response && !results.isEmpty())  topCell = results.get(results.size() - 1);` because ITs would throw an `ArrayOutOfBoundException` when trying to access `results.get(-1)` in case `results` was empty.
This PR does not relax the condition anymore, and the ITs do not throw that exception.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -141,10 +143,44 @@ public boolean isFilterDone() throws IOException{
...
+        boolean response;
+        if (firstCall) {
+            rowBuffer = new ArrayList<>();
+            response = nextMerged(rowBuffer);
+            if (!response) {
+                result.addAll(rowBuffer);
+                rowBuffer = null;
+            } else {
+                fetchNextAndGiveBuffer(result);
+            }
```
hatyo, line 174: So if there are no more rows, we consume the rowBuffer and return, otherwise, we consume the rowBuffer and pre-fetch the next row into rowBuffer, right?

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -68,6 +68,8 @@
     private boolean flushed;
     private long numberOfRows = 0;
     private FileSystem customFilesystem;
+    private List<Cell> rowBuffer = null;
+    private boolean noMoreRecords = false;
```
hatyo, line 72: Is this field necessary? could you please consider removing it, it makes understanding the code unnecessarily more complicated IMHO, i.e.:
make `fetchNextAndGiveBuffer()` return boolean indicating whether more rows are to come  (similar to `RegionScanner.nextRaw`) and assign it to a local variable in `nextRaw` call (maybe with some documentation)

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -141,10 +143,44 @@ public boolean isFilterDone() throws IOException{
...
+            }
+        }
+    }
+
     @Override
     public boolean nextRaw(List<Cell> result) throws IOException {
-        boolean res = nextMerged(result);
-        boolean returnValue = updateTopCell(res,result);
+        boolean firstCall = rowBuffer == null;
+        boolean response;
```
hatyo, line 165: consider renaming this to something like: `moreRowsFollow`.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -189,8 +232,7 @@ public HRegion getRegion(){
     /*private helper methods*/
 
     private boolean updateTopCell(boolean response, List<Cell> results) throws IOException {
-        if (!results.isEmpty() &&
-                CellUtil.matchingFamily(results.get(0),ClientRegionConstants.FLUSH)){
+        if (matchingFamily(results, ClientRegionConstants.FLUSH)) {
             if (LOG.isDebugEnabled())
                 SpliceLogUtils.debug(LOG,"received flush message " + results.get(0));
             flushed = true;
```
hatyo, line 238: I think this is neither thread-safe nor exception-safe.
Instead maybe we should:
1. Moved the `flushed` flag into the `updateScanner` method and update it there with care.
2. Mark `updateScanner` method as `synchronized`.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -155,7 +191,7 @@ public boolean nextRaw(List<Cell> result) throws IOException {
      * refresh underlying RegionScanner we call this when new store file gets
```
hatyo, line 191: could you please rephrase the documentation.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -155,7 +191,7 @@ public boolean nextRaw(List<Cell> result) throws IOException {
...
      * created by MemStore flushes or current scanner fails due to compaction
      */
-    public void updateScanner() throws IOException {
+    void updateScanner() throws IOException {
             if (LOG.isDebugEnabled()) {
                 SpliceLogUtils.debug(LOG,
                         "updateScanner with hregionInfo=%s, tableName=%s, rootDir=%s, scan=%s",
```
hatyo, line 197: optional: add: `assert htd != null && htd.getTableName() != null;`

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -164,11 +200,18 @@ public void updateScanner() throws IOException {
...
+                if (rowBuffer != null && !rowBuffer.isEmpty()) {
+                    restartRow = CellUtil.cloneRow(rowBuffer.get(0));
+                    rowBuffer = null;
+                } else if (this.topCell != null) {
+                    restartRow = Bytes.add(CellUtil.cloneRow(topCell), new byte[]{0});
+                }
+                if (restartRow != null) {
                     if (LOG.isDebugEnabled())
-                        SpliceLogUtils.debug(LOG, "setting start row to %s", topCell);
+                        SpliceLogUtils.debug(LOG, "setting start row to %s", Hex.encodeHexString(restartRow));
```
hatyo, line 212: nice! 👍

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
```diff
@@ -141,10 +143,44 @@ public boolean isFilterDone() throws IOException{
         return scanner.isFilterDone();
     }
 
+    private void fetchNextAndGiveBuffer(List<Cell> result) throws IOException {
+        assert result.isEmpty();
+        List<Cell> nextResult = new ArrayList<>();
+        boolean response = nextMerged(nextResult);
+        if (matchingFamily(nextResult, ClientRegionConstants.FLUSH)) {
+            // A flush should be returned before a potential partial result in the buffer
```
hatyo, line 151: I don't understand this comment, does it mean that we have the full nextResult or could it be partial?

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
arnaud-splice: That's not possible. The attribute would be read only the next time that nextRaw is being called, and that happens through some hbase code.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
arnaud-splice: Typically, when we return from this function, we have managed to put the buffer in `result` and we've gotten a new buffer.
If we get a flush token, we cannot return the current buffer, because it may contain partial results. Those partial results would never be completed since nextRaw would return FLUSH.
So we need to return the flush before.

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
arnaud-splice: yes

hbase_storage/src/main/java/com/splicemachine/access/client/SkeletonClientSideRegionScanner.java
arnaud-splice: It does not need to be thread safe, only one thread is using this scanner.

PR #3823, opened 2020-07-14
DB-9826 Fix NPE when ROWID is used in condition

hatyo:


hatyo:
Already brought by merging DB-9826 to /master since the commits in this PR were [merged into](https://github.com/splicemachine/spliceengine/pull/3828/commits) DB-9826 PR.

PR #3828, opened 2020-07-15
DB-9620 Add ROWID SQL functions

hatyo:
- Add test to TO_INSTANT SQL function.
- Add TO_HBASE_ESCAPED function.
- Add tests and address SpotBug issues.


db-engine/src/main/java/com/splicemachine/db/iapi/util/RowIdUtil.java
```diff
@@ -0,0 +1,33 @@
...
+
+    static DateTimeDataValue toInstant(DataValueDescriptor s) throws StandardException {
+        if(s == null) {
+            return null;
+        }
+        String hex = toHBaseEscaped(s.getString());
+        long value = Bytes.toLong(Bytes.toBytesBinary(hex));
+        long ts = System.currentTimeMillis() & (~TIMESTAMP_MASK);
+        ts |= (value >> TIMESTAMP_SHIFT) & TIMESTAMP_MASK;
+        return new SQLTimestamp(new Timestamp(ts));
```
martinrupp: this could be a static function. then you could have a unit test for toLongTimestamp

```
public static long toLongTimestamp( String str, long time )
{
	String hex = toHBaseEscaped( str );
    long value = Bytes.toLong(Bytes.toBytesBinary(hex));
    long ts = time & (~TIMESTAMP_MASK);
    ts |= (value >> TIMESTAMP_SHIFT) & TIMESTAMP_MASK;
    return ts;
}

static DateTimeDataValue toInstant(DataValueDescriptor s) throws StandardException {
    if(s == null) {
        return null;
    }

    long time = System.currentTimeMillis();
    return new SQLTimestamp(new Timestamp( toLongTimestamp(s.getString(), time)));
}
```

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/RowIdIT.java
```diff
@@ -450,4 +451,30 @@ public void testRowIdToInstantFunctionInvalidInput() {
...
+    public void testRowIdToHbaseEscapedFunction() throws Exception {
+        ResultSet resultSet  = methodWatcher.executeQuery(String.format("select cast(rowid as varchar(30)) from %s",
+                this.getTableReference(TABLE7_NAME)));
+        Assert.assertTrue(resultSet.next());
+        String rowId = resultSet.getString(1);
+        Assert.assertFalse(resultSet.next());
+        StringDataValue result = RowIdUtil.toHBaseEscaped(new SQLChar(rowId));
+        resultSet = methodWatcher.executeQuery(String.format("select to_hbase_escaped(rowid) from %s", this.getTableReference(TABLE7_NAME)));
+        Assert.assertTrue(resultSet.next());
+        Assert.assertEquals(result.getString(), resultSet.getString(1));
```
martinrupp, line 465: this looks like you're comparing the result against the result.
maybe instead of result use expectedEscaped.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/RowIdIT.java
```diff
@@ -450,4 +451,30 @@ public void testRowIdToInstantFunctionInvalidInput() {
...
+        Assert.assertFalse(resultSet.next());
+    }
```
martinrupp, line 467: can you add a test where you have input and output as strings? e.g. 

`Assert.equals( RowIdUtil.toHBaseEscaped("HelloWorld\x33asdf\n\t") , "HelloWorld\\x33asdf\\x0D" ));`

Otherwise it somehow feels like you're testing toHBaseEscaped by using toHBaseEscaped.

PR #3830, opened 2020-07-15
DB-9844 Spark 3.0 Remove org.spark_project.* references (3.0)

ipraznik-splice:


ipraznik-splice:
Another implementation is prepared for review.

PR #3836, opened 2020-07-15
DB-9853 Low watermark for transactions should never be MAX_VALUE

arnaud-splice:


dgomezferro:
@arnaud-splice can we get a fix for the misconfigured case as well? When fetchOldestActiveTransaction fails we return Long.MAX_VALUE too and use the current timestamp.

arnaud-splice:
Good point, I addressed that in `28b49fb`

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1068/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/hbase/TransactionsWatcher.java
```diff
@@ -48,14 +48,19 @@
...
-            lowWatermarkTransaction.set(fetchTimestamp);
-        } else {
-            lowWatermarkTransaction.set(oldestActiveTransaction);
-        }
-        if (LOG.isDebugEnabled()) {
-            LOG.debug(String.format("lowWatermarkTransaction set: %d", lowWatermarkTransaction.get()));
+        try {
+            long oldestActiveTransaction = fetchOldestActiveTransaction();
+            if (oldestActiveTransaction == Long.MAX_VALUE) {
+                lowWatermarkTransaction.set(fetchTimestamp);
```
martinrupp, line 54: fetchTimestamp can never bee Long.MAX_VALUE? Or do we assume then we can't fix this here anyhow?

arnaud-splice: I don't think that fetchTimestamp can ever reach MAX_VALUE, otherwise, our monotonically increasing timestamp would be broken. @dgomezferro, do you agree?

PR #3870, opened 2020-07-21
DB-9807 use info in tabledescriptor to colllect the ddl info for "sho…

yxia92:
…w create table"

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
@@ -54,7 +54,6 @@ import com.splicemachine.derby.ddl.DDLUtils;
 import com.splicemachine.derby.iapi.sql.execute.RunningOperation;
 import com.splicemachine.derby.impl.store.access.SpliceTransactionManager;
 import com.splicemachine.derby.stream.ActivationHolder;
-import com.splicemachine.derby.stream.iapi.DataSetProcessor;
 import com.splicemachine.hbase.JMXThreadPool;
 import com.splicemachine.hbase.jmx.JMXUtils;
 import com.splicemachine.pipeline.ErrorState;
@@ -89,7 +88,9 @@ import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.stream.Collectors;
 
+import static com.splicemachine.db.iapi.sql.StatementType.*;
 import static com.splicemachine.db.shared.common.reference.SQLState.*;
 
 /**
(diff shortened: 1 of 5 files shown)


yxia92:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1112/

yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2148,276 +2149,218 @@ public static void SHOW_CREATE_TABLE(String schemaName, String tableName, Result
     {
```
ascend1, line 2149: I happened to touch system procedures too when implementing local temporary tables, so I think maybe I should to put the questions and observations I had here. None of them are really critical and you can just ignore if you think it's not worth implementing.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2148,276 +2149,218 @@ public static void SHOW_CREATE_TABLE(String schemaName, String tableName, Result
...
-
-            String getTableId = "SELECT T.TABLEID, T.TABLETYPE, T.COMPRESSION, T.DELIMITED, " +
-                    "T.ESCAPED, T.LINES, T.STORED, T.LOCATION" +
-                    " FROM SYSVW.SYSTABLESVIEW T " +
-                    "WHERE T.TABLETYPE IN ('T','E','S','V') " +
-                    "AND T.TABLENAME LIKE '" + tableName + "' AND T.SCHEMANAME = '" + schemaName + "'";
-
-            String tableId = "" ;
-            boolean firstCol = true;
+            TableDescriptor td = EngineUtils.verifyTableExists(connection, schemaName, tableName);
```
ascend1, line 2152: I found some system procedures call `schemaName = EngineUtils.validateSchema(schemaName)` and `tableName = EngineUtils.validateTable(tableName)` before doing the actual logic but some don't, likely a 50/50 percent distribution.

These validations kind of normalize the names and I think it's good to call them. Most observable effect is that without calling them, we have to provide schema name and table name always in upper case. Otherwise they cannot be resolved.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2148,276 +2149,218 @@ public static void SHOW_CREATE_TABLE(String schemaName, String tableName, Result
...
+                }
+            }//End External Table
+            else if (td.getTableType() == TableDescriptor.VIEW_TYPE) {
+                //Target table is a View
+                throw ErrorState.LANG_INVALID_OPERATION_ON_VIEW.newException("SHOW CREATE TABLE", "\"" + schemaName + "\".\"" + tableName + "\"");
+            } else if (td.getTableType() == TableDescriptor.SYSTEM_TABLE_TYPE) {
+                //Target table is a system table
+                throw ErrorState.LANG_NO_USER_DDL_IN_SYSTEM_SCHEMA.newException("SHOW CREATE TABLE", schemaName);
+            }
+            // Get column list, and write DDL for each column.
```
ascend1, line 2224: The new code here might introduce a behavior change. It's tricky. In the old query, predicate on table type is `T.TABLETYPE IN ('T','E','S','V')`, and that actually excludes temporary tables whose value in `TABLETYPE` column is 'X'. As a result, if we call show_create_table on a global temporary table, it runs through but return no SQL create statement.

The new code would get the `TableDescriptor` of a global temporary table just as a base table and eventually reaches line 2225 and below. But `CREATE (LOCAL|GLOBAL) TEMPORARY TABLE` statement has two special terms `[NOLOGGING | ON COMMIT PRESERVE ROWS]`. So maybe we should have an extra branch for temporary tables and check these two properties. Or we can keep the old behavior and silently skip for temporary tables.

yxia92: Thanks @ascend1 ! I think it makes sense to support show create table for temp table also. I'll add the code to handle the options.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
yxia92: The system procedure call SHOW_CREATE_TABLE() takes quoted schemaname and tablename, so the names are expected to be in the exact form matching that in the data dictionary table. For example:
```
create table "t1" (a1 int, b1 int, c1 int);
splice> call syscs_util.show_create_table('SPLICE','t1');
DDL                                                                      
-------------------------------------------------------------------------
CREATE TABLE "SPLICE"."t1" (
"A1" INTEGER
,"B1" INTEGER
,"C1" INTEGER
) ;

1 row selected
```
For the ij command "show create table", it actually does some normalization(fill in scheme name, changing the name to upper case) in the below code in ij.jj:
```
|   <CREATE> <TABLE> ( i = caIdentifier() ( <PERIOD> i2 = caIdentifier() )?


		{
			if (i2 == null) {
				schema = null;
				tblname = i;
			} else {
				schema = i;
				tblname = i2;
			}
            if (schema == null) {
                schema = util.getSelectedSchema(theConnection);
            }
			return showCreateTable(schema, tblname);
```
However, there seems a general problem ij command issue that it cannot handle quoted names for the case-sensitive names. I've opened DB-9928 to track it.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
yxia92: I've added the code to support temporary table.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
ascend1: Got it. Thanks for the reply.

yxia92:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1205/

yxia92:

spliceci:
TEST SUCCEEDED +1

PR #3882, opened 2020-07-23
DB-9895: remove first write token during wal replay

jyuanca:


jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1124/

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
```diff
@@ -508,4 +510,22 @@ protected boolean aclCheck(Scan scan) throws IOException, StandardException {
...
+
+        SConfiguration config = HConfiguration.getConfiguration();
+        String namespace = logKey.getTableName().getNamespaceAsString();
+        if (namespace.equals(config.getNamespace())) {
+            ArrayList<Cell> cells = logEdit.getCells();
+            Iterator<Cell> it = cells.iterator();
+            while (it.hasNext()) {
+                Cell cell = it.next();
+                CellType cellType = CellUtils.getKeyValueType(cell);
+                if (cellType == CellType.FIRST_WRITE_TOKEN) {
```
hatyo, line 525: So, if I understand correctly, we will remove the `FIRST_WRITE_TOKEN` cell from the list of WAL edits, therefore it will be written to the HFile causing a proper deletion to happen when compaction kicks in, right?

jyuanca: +1

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
```diff
@@ -508,4 +510,22 @@ protected boolean aclCheck(Scan scan) throws IOException, StandardException {
             }
         }
     }
+
+    @Override
+    public void preWALRestore(ObserverContext<? extends RegionCoprocessorEnvironment> ctx, RegionInfo info, WALKey logKey, WALEdit logEdit) throws IOException {
+
```
dgomezferro, line 516: Please add a comment here explaining why this is needed, or at least a pointer to 
DB-9895 and add the full explanation there.

jyuanca: I had a hard time reproducing this manually. I can only verify that the token is removed and rows are not purged during WAL replay. Will prepare a parcel to run Stephanie's test on cloud

PR #3892, opened 2020-07-27
DB-9425 Reduce scale of a result decimal type when necessary

ascend1:


ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1150/

ascend1:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DecimalIT.java
```diff
@@ -677,4 +684,17 @@ public void testPreparedStatementRetainsFractionalDigits() throws Exception {
...
+    public void testReduceFractionalDigitsForIntegralPart() throws Exception {
+        String sqlText = String.format("select cast(COALESCE(((quantity/units) * ( price / units)* cast(factor AS decimal(10,2)) ),0 ) AS NUMERIC(31,10))\n" +
+                "from DB_9425_TBL --splice-properties useSpark=%s", useSpark);
+
+        String expected = "1              |\n" +
+                "-----------------------------\n" +
+                "3112011525395705.4789182174 |";
+
+        testQuery(sqlText, expected, methodWatcher);
+    }
```
martinrupp, line 699: i can't see immediately if this covers all the cases above. what i do see is that you don't use + or -, so that code is not tested.
can we have, additional to this, something for sum, for times, for divide separated?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
```diff
@@ -364,6 +364,53 @@ else if (typeId.isBigIntTypeId() &&
...
+				{
+					int lprec = leftType.getPrecision();
+					int lscale = leftType.getScale();
+					int rprec = rightType.getPrecision();
+					int rscale = rightType.getScale();
+					int integralNumDigits;
+
+					switch (operator) {
+						case TypeCompiler.SUM_OP:
+						case TypeCompiler.MINUS_OP:
```
yxia92, line 380: I think PLUS_OP should be added here similar to MINUS_OP.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
```diff
@@ -364,6 +364,53 @@ else if (typeId.isBigIntTypeId() &&
...
+							if (integralNumDigits <= precision &&
+								integralNumDigits + scale > precision && scale > MIN_DECIMAL_DIVIDE_SCALE &&
+							    precision - integralNumDigits >= MIN_DECIMAL_DIVIDE_SCALE) {
+								scale = precision - integralNumDigits;
+							}
+							break;
+						}
+						case TypeCompiler.DIVIDE_OP: {
+							if (rightTypeId.isDecimalTypeId()) {
+								integralNumDigits = lprec - lscale + rprec;
```
yxia92, line 400: should it be rscale instead of rprec?

yxia92: It seems that for the division precision and scale computation, the logic in getScale() and getPrecision() are not consistent.
In getPrecision():
```
			val = Math.min(maxPrecision,
						   this.getScale(operator, leftType, rightType) + lprec - lscale + rprec);
```
In getScale():
```
				val = Math.max(maxPrecision - lprec + lscale - rscale,
						MIN_DECIMAL_DIVIDE_SCALE);
```
I think the formula in getPrecision should be changed to 
```
			val = Math.min(maxPrecision,
						   this.getScale(operator, leftType, rightType) + lprec - lscale + rscale);
```
to be consistent with getScale().

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DecimalIT.java
ascend1: When trying to build tests for individual arithmetic operators, I actually found something interesting in our way of deducing division result precision and scale. I document it in the code.

But because of this special thing, I have to follow SQL Server's rules of reducing scale and cannot make it more strict. As a result, we lose scale digits more often and small error accumulates. A lot more tests fail now and I'm trying to get the results from SQL Server for those queries.

Implementation is almost fully changed. Next push would be a force one since delta doesn't make any sense.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
ascend1: Yes, it was a mistake. It should be PLUS_OP, not SUM_OP.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
ascend1: This is the mystery part and I think it was done like this on purpose. The idea seems to be always use `maxPrecision` for division and leave all free digits to scale. Let's take a closer look:

The way we get scale for division, is to calculate the number of digits needed for the integral part (`lprec - lscale + rscale`), and subtract it from `maxPrecision`. When getting precision, we add to scale a term `lprec - lscale + rprec` that is always GE than the integral part, so it will be always `maxPrecision` in the end. This change was done in implementing Decimal 38. The logic of getting scale for division was not like this, and we can see that from the old comment.

Changing the logic back to the old comment or the SQL server way causes quite many tests to fail, including TPCH ITs. I'm examining the failing tests one by one.

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DecimalIT.java
ascend1: I end up with implementing SQL Server's scale reduction rules and fixing our result type precision/scale of division. In this way, TPC-H results are correct and only a few tests need to be adapted. Seems to be the best approach so far.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
ascend1: I finally get the root cause of test failures: SQL Server's MIN_DECIMAL_DIVIDE_SCALE = 6, but we set it to 4, and I was using 4 in multiplication case, too. That causes TPC-H queries return wrong results.

I fixed this part and also revert the logic of getPrecision() for division according to the old comment. Now only a few tests need to be adapted.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NumericTypeCompiler.java
```diff
@@ -364,6 +364,57 @@ else if (typeId.isBigIntTypeId() &&
...
+						case TypeCompiler.TIMES_OP: {
+							integralNumDigits = lprec + rprec + 1 - lscale - rscale;
+							/* In SQL Server, MIN_DECIMAL_DIVIDE_SCALE = 6. In splice, it's 4.
+							 * We have to use 6 in multiplication case, otherwise TPC-H results would be
+							 * off too much because we round intermediate results too early.
+							 */
+							if (integralNumDigits < maxPrecision - MIN_DECIMAL_MULTIPLICATION_SCALE) {
+								scale = Math.min(scale, maxPrecision - integralNumDigits);
+							}
+							else if (scale > MIN_DECIMAL_MULTIPLICATION_SCALE) {
```
yxia92, line 398: Now that you introduced MIN_DECIMAL_MULTIPLICATION_SCALE, would there be an impact on the below code which looks at MIN_DECIMAL_DIVIDE_SCALE only. Could you double check the effect of your change on it:
```
				// Use high precision for decimals, so we don't overflow.
				// But don't use max precision to leave room for 4 digits to
				// the right of the decimal place for operations like AVG.
				// For the old client, use the old size of 31 for consistent behavior.
				int scaleAllowance = Math.max(0, MIN_DECIMAL_DIVIDE_SCALE - Math.max(scale, 0));
				int maxImplicitCastPrecision = precision > OLD_MAX_DECIMAL_PRECISION_SCALE ||
				                               supportsDecimal38()  ?
				      MAX_DECIMAL_PRECISION_WITH_RESERVE_FOR_SCALE + scaleAllowance :
				      OLD_MAX_DECIMAL_PRECISION_SCALE;
				if (precision < maxImplicitCastPrecision)
				    precision = maxImplicitCastPrecision;
```
However, I have some reservation about this code to use the highest precision possible for an operation on decimals(See DB-9639).

ascend1: Right, this part basically promotes the precision to at least 34. If the determined `scale` is smaller than 4, then there would be some digits allowance for the integral part, thus total precision is further increased.

`MIN_DECIMAL_MULTIPLICATION_SCALE` should have no impact here. If the initial scale was smaller than 6, say 3, then there will be no scale reduction in case of an multiplication. This part sees that there is 4-3=1 digit allowance for the integral part, and promote the type to (35,3).

If for any reason we reduce the scale to 6, then there will be no scale allowance. So if the initial precision was smaller than 34, final type would be (34,6), otherwise it's (precision,6).

Key idea is, when we reach this part, we know that we either have enough digits for integral part (scale reduced) or we cannot prevent overflow (scale untouched). In both cases, increasing precision do not affect correctness.

BUT!!!!!

This logic seems very strange to me. In my opinion, we should not blindly increase precision but to deduce it properly. If (20,8) is enough, then we should not increase it to (34,8).

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1204/

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1215/

spliceci:
TEST SUCCEEDED +1

PR #3905, opened 2020-07-29
DB-9883 spliceck - HBase/Splice Support Tool

hatyo:
Also DB-9901 and DB-9903 tasks.


hatyo:
To use:
* build.
* from the root folder and run `./sck.sh`.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCOLUMNSRowFactory.java
```diff
@@ -60,17 +60,17 @@
...
     protected static final int		SYSCOLUMNS_TABLEID = 1;
-    protected static final int		SYSCOLUMNS_REFERENCEID = 1;
-    protected static final int		SYSCOLUMNS_COLUMNNAME = 2;
-    protected static final int		SYSCOLUMNS_COLUMNNUMBER = 3;
+    public static final int		SYSCOLUMNS_REFERENCEID = 1;
+    public static final int		SYSCOLUMNS_COLUMNNAME = 2;
+    public static final int		SYSCOLUMNS_COLUMNNUMBER = 3;
     protected static final int		SYSCOLUMNS_STORAGECOLUMNNUMBER = 4;
-    protected static final int		SYSCOLUMNS_COLUMNDATATYPE = 5;
+    public static final int		SYSCOLUMNS_COLUMNDATATYPE = 5;
```
martinrupp, line 73: (minor comment) tab/space alignment looks off

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCOLUMNSRowFactory.java
```diff
@@ -208,27 +208,33 @@ public ExecRow makeRow(TupleDescriptor td, TupleDescriptor parent) throws Standa
...
 		    /* Build the row to insert  */
         row = getExecutionFactory().getValueRow(SYSCOLUMNS_COLUMN_COUNT);
+        setRowColumns(row, colName, defaultID, tabID, colID, storageNumber, typeDesc, defaultSerializable, autoincStart,
+                autoincInc, autoincValue, partitionPosition, autoinc_create_or_modify_Start_Increment, collectStats, useExtrapolation);
 
-		    /* 1st column is REFERENCEID (UUID - char(36)) */
+        return row;
+    }
+
+    public static void setRowColumns(ExecRow row, String colName, String defaultID, String tabID, Integer colID, Integer storageNumber, TypeDescriptor typeDesc, Object defaultSerializable, long autoincStart, long autoincInc, long autoincValue, int partitionPosition, long autoinc_create_or_modify_Start_Increment, boolean collectStats, byte useExtrapolation) {
```
martinrupp, line 217: please use line breaks (80-100 columns max i would say)

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
```diff
@@ -257,6 +257,27 @@ public ExecRow makeRow(TupleDescriptor td,
 		/* Build the row to insert  */
 		row = getExecutionFactory().getValueRow(SYSTABLES_COLUMN_COUNT);
 
+		setRowColumns(row, tableID, tableName, tabSType, schemaID, lockGranularity, tableVersion, columnSequence,
+				delimited, escaped, lines, storedAs, location, compression, isPinned, purgeDeletedRows);
+		return row;
+	}
+
+	public static void setRowColumns(ExecRow row,
+									 String tableID,
```
martinrupp, line 266: are these... tabs? ;-)

sck.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+mvn -q -Pcore,cdh6.3.0 -f ./splice_ck/pom.xml exec:java -Dexec.args="$*"
```
martinrupp, line 3: maybe we should be able to use this also for cdh5.12.0, or is this all the same?
```
PLATFORM=$1
shift # ignore $1
mvn -q -Pcore,${PLATFORM} -f ./splice_ck/pom.xml exec:java -Dexec.args="$*"
```

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
```diff
@@ -0,0 +1,176 @@
...
+                        CheckNull(td.getValue(SIConstants.TABLE_DISPLAY_NAME_ATTR)),
+                        CheckNull(td.getValue(SIConstants.INDEX_DISPLAY_NAME_ATTR)),
+                        CheckNull(td.getValue(SIConstants.TRANSACTION_ID_ATTR)));
+            }
+            return tabular;
+        }
+    }
+
+    private String CheckNull(String value) {
+        return value == null ? "NULL" : value;
```
martinrupp, line 85: maybe "\"" + value + "\"", to differentiate the string "NULL" from NULL ?
And maybe also escape strings? Not sure how "pro" this needs to be

```
if( value == null )
  return "NULL";

return "\"" + value.replace("\\", "\\\\").replace("\"", "\\\"") + "\"";
```

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                }
+                return schema.get();
+            }
+        }
+    }
+
+    public String getHbaseRegionOf(String tableName) throws Exception {
+        Utils.Tabular results = getListTables();
+        Utils.Tabular.Row r = results.rows.stream().filter(result -> result.cols.get(2 /*name*/).equals(tableName)
```
martinrupp, line 156: tableName

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
```diff
@@ -0,0 +1,176 @@
...
+                && result.cols.get(3 /*index*/).equals("NULL")).findAny().orElse(null);
```
martinrupp, line 157: index? isn't that transactionId?

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
```diff
@@ -0,0 +1,176 @@
...
+        if (r == null) {
+            throw new TableNotFoundException();
+        } else {
+            return r.cols.get(0);
+        }
+    }
+
+    public String getSpliceTableNameOf(String regionName) throws Exception {
+        Utils.Tabular results = getListTables();
+        Utils.Tabular.Row r = results.rows.stream().filter(result -> result.cols.get(0 /*name*/).equals(regionName)
```
martinrupp, line 167: schemaName?

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
```diff
@@ -0,0 +1,176 @@
...
+                && result.cols.get(3 /*index*/).equals("NULL")).findAny().orElse(null);
```
martinrupp, line 168: transactionId?

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
```diff
@@ -0,0 +1,206 @@
...
+            formats.add("%-" + (length + 5) + "s");
+        }
+
+        StringBuilder stringBuilder = new StringBuilder();
+        for(int i = 0; i < tabular.headers.size(); ++i) {
+            stringBuilder.append(Colored.boldWhite(String.format(formats.get(i), tabular.headers.get(i))));
+        }
+        stringBuilder.append("\n");
+        int totalLength = lengths.stream().reduce(0, Integer::sum);
+        totalLength += 5 * lengths.size();
```
martinrupp, line 91: make 5 a descriptive constant

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
```diff
@@ -0,0 +1,206 @@
...
+    public static SQLType[] toSQLTypeArray(List<String> values) {
+        SQLType[] result = new SQLType[values.size()];
+        int cnt = 0;
+        for(String value : values) {
+            result[cnt++] = typeMap.get(value);
+        }
+        return result;
+    }
+
+    public static class Colored {
```
martinrupp, line 152: woah these are some sweet util classes! i'm not sure maybe we would want a way to disable coloring for some weird terminals / file forwarding

splice_ck/src/main/java/com/splicemachine/ck/visitor/CellPrinter.java
```diff
@@ -0,0 +1,101 @@
...
+    protected void postVisit(Cell cell) {
+        Long key = cell.getTimestamp();
+        events.computeIfAbsent(key, k -> new ArrayList<>());
+        events.get(key).add(stringBuilder.toString());
+    }
+
+    @Override
+    public void visitCommitTimestamp(Cell cell) {
+        stringBuilder.append(Utils.Colored.green("commit timestamp "));
+        stringBuilder.append(Utils.Colored.green(Long.toString(com.splicemachine.primitives.Bytes.toLong(cell.getValueArray(),cell.getValueOffset(),cell.getValueLength()))));
```
martinrupp, line 52: too long line

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/Main.java
```diff
@@ -55,220 +49,192 @@
...
+        }
+
+        // load the property file if specified
+        util.getPropertyArg(args);
+
+        // readjust output to db.ui.locale and db.ui.codeset if
+        // they were loaded from a property file.
+        langUtil.init();
+        out = langUtil.getNewOutput(System.out);
+        main.initAppUI();
```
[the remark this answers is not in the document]
hatyo: I reverted the changes done to this file since it is not relevant to this PR.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCOLUMNSRowFactory.java
hatyo: I wouldn't fix it here since it would make this change even bigger, I can do it later in a more focused refactoring change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCOLUMNSRowFactory.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
hatyo: :) yeah, the whole file uses tabs and my IDE is smart enough to follow whatever whitespace convention the file has, I would leave it as-is for now. Should do a major refactoring to all these files to follow a consistent style.

sck.sh
hatyo: thanks! I wrote this small script for myself. I will make it accept CDH parameter as you described.
it is not yet clear to me how we want to ship spliceck.

hatyo: I modified the script s.t. it defaults to cdh6.3.0 (since I hate typing too much :) ):

```
#!/bin/bash

platform=cdh6.3.0

function help() {
  echo "sck.sh [--platform=<platform>] command"
  echo "<platform> The cluster platform, default is cdh6.3.0"
}

if [ $# -eq 0 ]
then
  help
else
  if [ $1 = "platform" ]
  then
     platform=$2
     shift 2
  fi
  mvn -q -Pcore,${platform} -f ./splice_ck/pom.xml exec:java -Dexec.args="$*"
fi
```

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
hatyo: that could get more complicated for its purpose, also we could easily make things more confusing if we tamper with the original values therefore I wouldn't do any escaping myself. Instead maybe we can give the user the option to tell us how nulls should look like. I would rather leave it as is at the moment.

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
hatyo: done. I use instead of index number a properly named integer constant.

```
public String regionOf(String tableName) throws Exception {
        Utils.Tabular results = getListTables();
        Utils.Tabular.Row row = results.rows.stream().filter(result -> result.cols.get(TBL_TABLES_NAME_IDX).equals(tableName)
                && result.cols.get(TBL_TABLES_INDEX_IDX).equals(NULL)).min((l, r) -> Long.compare(Long.parseLong(r.cols.get(TBL_TABLES_CREATE_TXN_IDX)),
                Long.parseLong(l.cols.get(TBL_TABLES_CREATE_TXN_IDX)))).orElseThrow(TableNotFoundException::new);
        assert row.cols.size() > 0;
        return row.cols.get(TBL_TABLES_HBASE_NAME_IDX);
    }
```

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
hatyo: Actually yes, it is the `index`, in hbase one could find many tables sharing the same table name (as far as I understand, base table and its indices share the same name), one way to tell which one is the base table is by checking that it doesn't have an index name.

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
hatyo: dito.

```
    public String tableOf(String regionName) throws Exception {
        Utils.Tabular tables = getListTables();
        Utils.Tabular.Row r = tables.rows.stream().filter(result -> result.cols.get(TBL_TABLES_HBASE_NAME_IDX).equals(regionName)
                && result.cols.get(TBL_TABLES_INDEX_IDX).equals(NULL)).findAny().orElseThrow(TableNotFoundException::new);
        return r.cols.get(TBL_TABLES_NAME_IDX);
    }
```

splice_ck/src/main/java/com/splicemachine/ck/HBaseInspector.java
hatyo: dito

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
hatyo: done.

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
hatyo: done, check more recent commits, there is a new option: `"-l"` resp. `"--colors"` which enables/disables coloring.

splice_ck/src/main/java/com/splicemachine/ck/visitor/CellPrinter.java
hatyo: done.

hatyo:
many thanks @martinrupp for code review, I addressed you comments.

splice_machine/src/main/java/com/splicemachine/derby/utils/marshall/EntryDataDecoder.java
```diff
@@ -26,6 +26,7 @@
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.types.DataValueDescriptor;
+import org.apache.spark.sql.Row;
```
arnaud-splice, line 29: Remove

sck.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+platform=cdh6.3.0
+
+function help() {
+  echo "sck.sh [--platform=<platform>] command"
```
arnaud-splice, line 6: This help message does not seem to match what's happening: running `./sck.sh --platform=cdh6.3.0` gives me `Unknown option: '--platform=cdh6.3.0'`

splice_ck/src/main/java/com/splicemachine/ck/command/RPutCommand.java
```diff
@@ -0,0 +1,112 @@
...
+import com.splicemachine.ck.Utils;
+import com.splicemachine.ck.encoder.RPutConfigBuilder;
+import org.apache.commons.lang3.StringUtils;
+import picocli.CommandLine;
+
+import java.util.concurrent.Callable;
+
+
+@CommandLine.Command(name = "rput",
+        description = "modify HBase row",
```
arnaud-splice, line 27: I would add quite the disclaimer here, informing that doing so may compromise data integrity, break our MVCC model, etc.

splice_ck/src/main/java/com/splicemachine/ck/command/RGetCommand.java
```diff
@@ -0,0 +1,81 @@
+/*
```
arnaud-splice, line 1: If a table exists in multiple schemas, this will always select the latest one. It'd be nice to be able to specify the schema. We'd need to find a way to distinguish between the schema in which the table resides and the schema you already use here (boolean, int, etc.)
Also, table names and schema names should be normalised (`EngineUtils.validateTable`, `EngineUtils.validateSchema`) so that the end user can write `t2` instead of `T2` for example.

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
arnaud-splice: The helper message lets the user believe that `-l` prints coloured output whereas it disabled colours on my side.

arnaud-splice: The same goes for verbose. It seems enabled by default and `-v` disables it.
I'd suggest to have colours enabled, verbose disabled, and the respective options to toggle that.

hatyo: Yes! thanks, I was messing around with these parameters locally to test some stuff, I will set them properly.

sck.sh
hatyo: Sorry, I will fix that.

splice_ck/src/main/java/com/splicemachine/ck/command/RPutCommand.java
hatyo: Good idea, I will do that.

splice_ck/src/main/java/com/splicemachine/ck/command/RGetCommand.java
hatyo: Good point(s) :)

splice_machine/src/main/java/com/splicemachine/derby/utils/marshall/EntryDataDecoder.java
hatyo: done.

splice_ck/src/main/java/com/splicemachine/ck/Utils.java
hatyo: Now the default values are also printed to the output making it easier to know the default behavior.

hatyo: > The helper message lets the user believe that -l prints coloured output whereas it disabled colours on my side.

the default "fallback" values for boolean flags is false by default, I flipped it so now it is more what we expect from setting `-l` or `-v`.

hatyo:
Thanks @arnaud-splice for the valuable hints and remarks. I addressed them in the latest commit.

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1304/

hatyo:
tests failed because of intermittent failure in ShowCreateTableIT which is addressed in DB-9957. Merging master to consume Yi's fix for this test before we run the tests again.

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #3910, opened 2020-07-31
DB-9890 Skip column if column statistics object is too big

ascend1:


splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
```diff
@@ -1045,8 +1045,19 @@ public boolean hasNext() {
...
                                                 ExecRow statsRow = StatisticsAdmin.generateRowFromStats(conglomId, "-All-", columnId, (ColumnStatisticsImpl) ois.readObject());
-                                                dataDictionary.addColumnStatistics(statsRow, tc);
-                                                bais.close();
+                                                try {
+                                                    dataDictionary.addColumnStatistics(statsRow, tc);
+                                                } catch (StandardException e) {
+                                                    // DB-9890 Skip a column if its statistics object doesn't fit into HBase cell.
+                                                    if (e.getCause().getMessage().contains("KeyValue size too large")) {
+                                                        SpliceLogUtils.warn(LOG, "Statistics object of [ConglomID=%d, ColumnID=%d] exceeds max KeyValue size. Try increase hbase.client.keyvalue.maxsize.",
+                                                                conglomId, columnId);
```
yxia92, line 1054: Is it possible to return this warning as part of the analyze output in addition to the warning in the log? so that user is aware that some column stats are skipped.

ascend1: I was thinking of something similar but not sure what's the correct way of doing it. It seems the standard way is to throw an exception and set the severity to warning. But we cannot throw here since we want to continue. Building the warning message into the result table would misuse the a char column, unless we add an extra column to the result table schema and set the message there. Not sure we should do that, either.

Do you have any suggestions?

yxia92: I'm thinking of adding an extra column in the analyze output to indicate whether some column stats were not saved due to space limitation. Thanks!

ascend1: Got it, will add another commit.

ascend1: Done.

splice_machine/src/main/java/com/splicemachine/derby/utils/StatisticsAdmin.java
```diff
@@ -1135,9 +1146,11 @@ public boolean hasNext() {
                                             } catch (StandardException e) {
                                                 // DB-9890 Skip a column if its statistics object doesn't fit into HBase cell.
                                                 if (e.getCause().getMessage().contains("KeyValue size too large")) {
+                                                    int columnId = nextRow.getColumn(SYSCOLUMNSTATISTICSRowFactory.COLUMNID).getInt();
                                                     SpliceLogUtils.warn(LOG, "Statistics object of [ConglomID=%d, ColumnID=%d] exceeds max KeyValue size. Try increase hbase.client.keyvalue.maxsize.",
                                                             nextRow.getColumn(SYSCOLUMNSTATISTICSRowFactory.CONGLOMID).getInt(),
-                                                            nextRow.getColumn(SYSCOLUMNSTATISTICSRowFactory.COLUMNID).getInt());
+                                                            columnId);
+                                                    skippedColIds.add(columnId);
```
yxia92, line 1153: For non-merged stats (that is, we output one row in the table stats table for each partition, so are the column stats), that means, we may add the same columnId multiple times to the `skippedColIds` list. Non-merged stats is a mode that we are not using currently, still I prefer to maintain it (in consideration in the future to support incremental stats) and add duplicate checks for this path. You may collect non-merged stats by calling the system procedure `COLLECT_NONMERGED_TABLE_STATISTICS`.

ascend1: I see. I pushed a new commit and use HashSet instead of ArrayList.

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #3931, opened 2020-08-04
DB-9213 Log UUID in EndFetching event

ascend1:


db-client/src/main/java/com/splicemachine/db/client/am/ResultSet.java
```diff
@@ -153,7 +156,7 @@
     public boolean isAfterLast_ = false;
     public boolean isFirst_ = false;
     public boolean isLast_ = false;
-    public boolean rowsetContainsLastRow_ = false;
+    //public boolean rowsetContainsLastRow_ = false;
```
dgomezferro, line 159: Can you remove these lines rather than commenting them out?

ascend1: OK, for the lines in `getRef()` and `getArray()`, should I remove them, too? Not sure if we would need them later.

ascend1: Done.

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #3947, opened 2020-08-07
DB-9904 Abort compaction if driver is not yet available 

arnaud-splice:


hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -152,7 +152,11 @@ public SpliceDefaultCompactor(final Configuration conf, final Store store, long
         CompactionResult result = null;
         SpliceCompactionRequest scr = (SpliceCompactionRequest) request;
 
-        Future<CompactionResult> futureResult = EngineDriver.driver().getOlapClient().submit(jobRequest, getCompactionQueue());
+        EngineDriver driver = EngineDriver.driver();
+        if (driver == null) {
+            throw new IOException("EngineDriver not available, compaction aborted");
+        }
```
jyuanca, line 158: Finally we get rid of this NullPointerException during startup :-)

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

PR #3950, opened 2020-08-07
DB-9982 Fix NPE in MemstoreAwareObserver 

arnaud-splice:


OlegMazurov:
We need to investigate whether returning a new MemStoreFlushAwareScanner when the input s is null is acceptable (or maybe even desirable). Anyway, the proposed fix does not change that behavior so we can merge it and attend to the question separately.

martinrupp:

PR #3964, opened 2020-08-11
DB-9883 spliceck - HBase/Splice Support Tool (2.8)

hatyo:


splice_ck/pom.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+// Licensed under the Apache License, Version 2.0 (the "License");
```
carolp-503, line 3: This should be GNU Affero license, not ASLv2.0.  See our 3.0 spliceengine pom.xml.

hatyo: ok, I will fix it.

PR #3969, opened 2020-08-11
DB-10025 Fix issues that cause a full IT run to fail.

msirek:
The following issues are fixed:

1.  The hbase balancer, which moves regions between region servers to balance out the load incurs a lot of memory pressure and heavy network traffic because it is configured to run every 5 seconds.  Sometimes it is too much for the colo network and servers to handle, resulting in OOM or network timeouts to be hit, which can bring down splice if it's in a critical region like getting a timestamp.  The solution is to bump up hbase.balancer.period and hbase.balancer.statusPeriod to a high value for standalone only, to eliminate this overhead.  To a lesser degree status messages between master and region server add overhead, so hbase.regionserver.msginterval is increased from 1 second to 15 seconds.

2.  Creating a new object via Optional.of(this) in Splice coprocessors and endpoints every time getRegionObserver is called is expensive, so this is changed to only create one of these objects.

3.  SpliceSpark.getSessionUnsafe would sometimes try to re-use an already closed spark session or open more than one spark context, leading to errors.  The solution is to use thread-local sessions like the 'getSession' method, and detect when session re-initialization is required (if a previous user closed it).  If spark.driver.allowMultipleContexts is true, then a new spark context is created.   

4.  SICompactionStateMutate, which purges tombstones during compaction, can produce more than one FIRST_WRITE_TOKEN when --splice-properties skipConflictDetection=true is used, or more than one DELETE_RIGHT_AFTER_FIRST_WRITE_TOKEN when the first write is actually a DELETE.  This can cause a DroppedSnapshotException during compaction of the bad region, forcing WAL replay and region server shutdown.  The solution is to avoid creating a FIRST_WRITE_TOKEN in either of these cases, and it will now only be created for insert or upsert.  As a safety precaution, for all of the failed assertion cases, code is added to bypass the tombstone purge for that key, since assertions tend to be disabled on clusters (we don't want to silently ignore the problem since purging the wrong thing could corrupt data).  Also, in SITransactor, possibleConflicts isn't always null when there are no key matches.  Sometimes it is returned as an empty DataResult.  This also indicates a FIRST_WRITE scenario, so the logic is update to check if (possibleConflicts == null || possibleConflicts.isEmpty()).

5.  splicemachineContext.tableExists is very expensive and seems to sometimes hang.  DefaultSourceIT and NativeTransformationsIT are updated to avoid calling it so many times.  Perhaps a Jira should be opened to look into it.

6. TestContext.scala is coded to open a different spark context on every instantiation, which sporadically hits an error message that only one spark context is allowed (use of sc = new SparkContext(conf) is deprecated.  Instead we're supposed to use SparkSession.newSession, passing the pre-existing SparkContext, if we want a new session.  This is handled in SpliceSpark.getSessionUnsafe.

7.  NativeSparkJoinWithInequalityPredsIT has too many tests defined, and can take up to an hour and a half to run.  Many of the tests aren't testing the new join functionality, but predicate evaluation, and are redundant.  The test is greatly pared down so that it runs in about a minute.

8.  DDLWatchRefresher.assignDDLDemarcationPoint runs at the completion of a DDL statement, including a restore statement, and saves the transaction id of the DDL statement as a cutoff point to tell which transaction can use the DDL cache, and which need to be looked up in the data dictionary.  This method has the assertion: 'assert txn.allowsWrites()', which looks up the transaction in splice:SPLICE_TXN if it does not exist in cache.  A full database restore may wipe out and restore all transactions in splice:SPLICE_TXN, so if the DDL txnId can't be found in cache, we try to look it up in plice:SPLICE_TXN.  Since it has been erased by the restore, we crash.  The solution is to avoid this assertion for restore statements.


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1311/

msirek:
jenkins please test branch @hdp3.1.0

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1328/

msirek:
jenkins please test branch @hdp3.1.0

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1329/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1331/

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -108,11 +111,30 @@ public static synchronized SparkSession getSessionUnsafe() {
...
+                System.exit(0);
+            sessionToUse = session = initializeSparkSession();
             ctx =  new JavaSparkContext(session.sparkContext());
         }
-        return session;
+        else {
+            if (sessionToUse == null || needsReinitialization) {
+                String allowMultipleContextsString = null;
+                try {
+                    allowMultipleContextsString = session.conf().get("spark.driver.allowMultipleContexts");
```
dgomezferro, line 123: Can we get rid of spark.driver.allowMultipleContexts in tests and remove this special case?

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
```diff
@@ -108,11 +111,30 @@ public static synchronized SparkSession getSessionUnsafe() {
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
-            System.exit(0);
-            session = initializeSparkSession();
+            if (isOlapWorker)
+                System.exit(0);
```
dgomezferro, line 115: Please add a LOG here so it's clear why we are shutting down

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLWatchRefresher.java
```diff
@@ -214,7 +216,30 @@ private void clearFinishedChanges(Collection<String> children,Collection<DDLWatc
     private void assignDDLDemarcationPoint(DDLChange ddlChange) {
         try {
             TxnView txn = new LazyTxnView(ddlChange.getTxnId(),txnSupplier,exceptionFactory);
-            assert txn.allowsWrites(): "DDLChange "+ddlChange+" does not have a writable transaction";
+            // A full Restore operation overwrite SPLICE_TXN, so the transaction used by the restore
+            // may not be found.  Avoid the assertion to avoid crashing, e.g.,
+            // 2020-08-11 18:28:11,785 (ZooKeeperDDLWatcherRefresher) ERROR [c.s.d.d.AsynchronousDDLWatcher] - Failed to refresh ddl
+            //java.lang.RuntimeException: com.splicemachine.si.api.txn.TransactionMissing: Couldn't resolve transaction with id 81981952
```
dgomezferro, line 222: Can you remove the stack trace from the comment? I think it makes sense to have it on the JIRA but not here (feel free to keep the explanation and a pointer to the JIRA for instance)

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/TestContext.scala
```diff
@@ -169,9 +170,14 @@ trait TestContext extends BeforeAndAfterAll { self: Suite =>
...
   override def beforeAll() {
-    sc = new SparkContext(conf)
-    SpliceSpark.setContext(sc)
-    spark = SparkSession.builder.config(conf).getOrCreate
+    spark = SpliceSpark.getSessionUnsafe
+    sqlContext = spark.sqlContext
+    spark.conf.set("spark.master", "local[*]")
+    spark.conf.set("spark.app.name", "test")
+    spark.conf.set("spark.ui.enabled", "false")
+    spark.conf.set("spark.app.id", appID)
```
dgomezferro, line 178: Does this make sense? We'd have to set those before instantiating the session, and this is already after that point, right?

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -71,11 +72,36 @@ private boolean isSorted(List<Cell> list) {
...
+        }
+        if (!rawListAndTxnListSameSize) {
+            bypassPurge = true;
+            LOG.warn("Skipping tombstone purge.  rawList and txn list not the same length.");
+        }
+        assert dataToReturnIsEmpty;
+        assert resultsIsEmpty;
+        assert maxTombstoneIsNull;
+        assert debugSortCheck : "CompactionStateMutate: rawList not sorted";
+        assert rawListAndTxnListSameSize;
```
dgomezferro, line 104: Can you move all these checks to a separate method to bring into focus the logic of the method?

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -71,11 +72,36 @@ private boolean isSorted(List<Cell> list) {
...
+            LOG.warn("Skipping tombstone purge.  rawList is not sorted.");
+        }
+        if (!dataToReturnIsEmpty) {
+            bypassPurge = true;
+            LOG.warn("Skipping tombstone purge.  dataToReturn is not properly initialized.");
+        }
+        if (!resultsIsEmpty) {
+            bypassPurge = true;
+            LOG.warn("Skipping tombstone purge.  results list not properly initialized.");
+        }
```
dgomezferro, line 91: Is this something that happens sometimes or just defensive programming?

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
msirek: Done.  spark.driver.allowMultipleContexts doesn't exist in the tests.  This is just a spark setting I was experimenting with.  I've removed this option.

hbase_sql/src/main/java/com/splicemachine/derby/impl/SpliceSpark.java
msirek: Done.

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLWatchRefresher.java
msirek: Done.

splice_spark/src/test/scala/com/splicemachine/spark/splicemachine/TestContext.scala
msirek: This isn't a SparkConf, as in initializeSparkSession().  This is a RuntimeConfig:
`  /**
   * Runtime configuration interface for Spark.
   *
   * This is the interface through which the user can get and set all Spark and Hadoop
   * configurations that are relevant to Spark SQL. When getting the value of a config,
   * this defaults to the value set in the underlying `SparkContext`, if any.
   *
   * @since 2.0.0
   */
  @transient lazy val conf: RuntimeConfig = new RuntimeConfig(sessionState.conf)`

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
msirek: Done.

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
msirek: Yes, this is mainly just defensive programming.  I only ran across 2 of the assertions being hit, but I decided to handle all of them, just in case.  An alternative would be to convert the assert statements into checks which throw IOExceptions.  These would be active on a cluster as well, whether or not assertions were enabled on it.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1348/

msirek:
jenkins please test branch @hdp3.1.0

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1352/

msirek:
jenkins please test branch @hdp3.1.0

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1353/

msirek:
jenkins please test branch @hdp3.1.0

spliceci:
TEST SUCCEEDED +1

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1355/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1356/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1357/

msirek:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -161,10 +206,24 @@ private void mutate(Cell element, TxnView txn) throws IOException {
                 lastSeenAntiTombstone = element;
                 break;
             case FIRST_WRITE_TOKEN:
+                // Assertions are only thrown on standalone.
```
arnaud-splice, line 209: More accurately, assertions are thrown during our IT runs as well.

dgomezferro (approved):

hbase_storage/src/main/java/com/splicemachine/si/impl/server/SICompactionStateMutate.java
```diff
@@ -71,15 +72,43 @@ private boolean isSorted(List<Cell> list) {
...
+
     /***
      * @return the size of all cells in the `rawList` parameter.
      */
     public long mutate(List<Cell> rawList, List<TxnView> txns, List<Cell> results) throws IOException {
-        assert dataToReturn.isEmpty();
-        assert results.isEmpty();
-        assert maxTombstone == null;
-        assert !LOG.isDebugEnabled() || isSorted(rawList) : "CompactionStateMutate: rawList not sorted";
-        assert rawList.size() == txns.size();
```
jyuanca, line 82: We may want to keep assertions because they are fired during ITs and can catch bugs. In case they are disables on a cluster, handleSanity Checks can safely bypass purge.

jpanko1: Looks like the same assertions were moved into the handleSanityChecks method, so they are being kept.

jyuanca: They becomes warning messages and does not cause a failure for IT

jpanko1: Feel free to write a Jira.  I'm merging now so it can get into this week's build.

PR #3972, opened 2020-08-12
DB-9048 Support all types for create/alter table with default 

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
```diff
@@ -1756,6 +1756,10 @@ public void setChildren(DataTypeDescriptor[] children) {
      }
 
     public DataValueDescriptor getDefault() throws StandardException {
+        if (typeId.getTypeFormatId() == StoredFormatIds.BIT_TYPE_ID) {
+            return new SQLBit(new byte[typeDescriptor.getMaximumWidth()]);
```
dgomezferro, line 1760: Why do we need to create it with maximumWidth instead of 0?

arnaud-splice: We mimic what DB2 is doing. For char(n) for bit data, their default is a byte array of size n, filled with 0x0

PR #3983, opened 2020-08-13
DB-10006 Reduce memory footprint of CSV reader

hatyo:


hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1339/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1343/

hatyo:

splice_machine/src/main/java/com/splicemachine/derby/impl/load/SpliceCsvReader.java
```diff
@@ -20,13 +20,16 @@
...
 import com.splicemachine.derby.stream.utils.BooleanList;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.supercsv.prefs.CsvPreference;
 
 /**
  * SpliceCsvReader is a simple reader that reads a row from a CSV file into an array of Strings.
  *
  * @author dwinters
  */
+@SuppressFBWarnings(value = "IT_NO_SUCH_ELEMENT", justification = "this is be design, the caller is responsible for " +
```
ascend1, line 31: "by design"?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -69,211 +140,309 @@
...
+    private void handleNormalState(final List<String> columns, final BooleanList quotedColumns) {
+        if (charIndex == getCurrentLine().length()) { // Newline. Add any required spaces (if surrounding spaces don't need quotes) and return (we've read a line!).
+            finishCol(columns, quotedColumns);
+            state = TokenizerState.FINISHED;
+        } else {
+            final char c = getCurrentLine().charAt(charIndex);
+            if (c == delimeterChar) { // Delimiter. Save the column (trim trailing space if required) then continue to next character.
+                finishCol(columns, quotedColumns);
+                potentialSpaces = 0;
+                allocateColumnMemory();
```
dgomezferro, line 343: I think this call is not needed, we already allocated a StringBuilder big enough for the largest column, the only thing we need is to "clear" the StringBuilder (`.setLength(0)`) right?

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -69,211 +140,309 @@
...
-        do{
-            line=readLine();
-            if(line==null){
-                return false; // EOF
+        if (!readFirstLine()) return false; // EOF
+
+        // now, if some column value is larger than scanThreshold we perform the dry run to avoid extra allocation of StringBuilder.
+        checkExactSizeFirst = false;
+        if (valueSizeHints != null) {
+            if (valueSizeHints.stream().max(Integer::compare).get() > scanThresold) {
```
dgomezferro, line 213: >> Could we add another check here? If we are parsing with oneLineRecords we could check the length of the current line, if it's less than the scanThreshold we avoid the dryRun too

> I think at this point of execution we don't know the actual size of the current CSV line, or maybe I missed some info in your comment?

We can get the size of the current line with `getCurrentLine().length()` don't we?
What we don't know is whether we are parsing in `oneLineRecords` mode, we'd need to pass down that flag into the QuoteTrackingTokenizer, or have a helper method in MutableCSVTokenizer (that one only gets used in `oneLineRecords` mode)

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -30,18 +33,86 @@
...
+    private boolean wasQuoted;
+    // the line number where a potential multi-line cell starts
+    private int quoteScopeStartingLine;
+    private TokenizerState state;
+    private int colIdx;
+    private boolean firstRun;
+    private boolean checkExactSizeFirst = false;
+    private int rowIdx = 0;
+    private int[] exactColumnSizes = null;
+    final static int SCAN_THRESHOLD = 10000;
```
dgomezferro, line 55: If you want to add a configuration option:

1. You'd get the config option with SIDriver.driver().getConfiguration().getCsvScanThreshold() (or whatever name you come up with)
2. You need to add that method to the interface, implementation and ConfigurationBuilder
3. Add the configuration to a *Configuration class (possibly SQLConfiguration, that's where `importMaxQuotedColumnLines` is defined)

spliceci:
TEST SUCCEEDED +1

martinrupp:
Looks good. Can you maybe check for double quote in strings? e.g. "hello ""World""!" should result in **hello "World"!**

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
hatyo: You're right, no need for resizing.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
hatyo: done. Thanks for the explanation.

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
hatyo: ... actually we do need this to reset the `StringBuilder` for the next column, otherwise we keep adding to the `StringBuilder`.

hatyo:
> Looks good. Can you maybe check for double quote in strings? e.g. "hello ""World""!" should result in **hello "World"!**

thanks, double-quoted (and n-quoted) strings are already tested in `QuoteTrackingTokenizerTest`; there you find a CSV generator (called `CSVBuilder`) that generates quotes inside quotes and we make sure that the output matches the behavior you mentioned. Check `QuoteTrackingTokenizer.java` around lines 226-231 for more information (maybe also run the test to see it in action).

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
hatyo: done, it made the parsing logic slightly more complicated but it is totally worth it. Thanks!

splice_machine/src/main/java/com/splicemachine/derby/impl/load/SpliceCsvReader.java
hatyo: fixed. thanks!

splice_machine/src/main/java/com/splicemachine/derby/stream/function/QuoteTrackingTokenizer.java
```diff
@@ -30,36 +34,94 @@
...
+            this.bufferList = new ArrayList<String>(1000);
+        }
 
-    private final int quoteChar;
+        public int length() {
+            if (sizeCached) {
+                return size;
+            }
+            size = 0;
+            for (String s : bufferList) {
```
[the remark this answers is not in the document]
hatyo: Thank you, yes I forgot that, I will fix it 👍

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1392/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1396/

hatyo:

dgomezferro (approved):
Looks great, thanks for being so thorough!

spliceci:
TEST SUCCEEDED +1

PR #3986, opened 2020-08-14
DB-9980 Report tables and columns missing statistics in explain statement

ascend1:


ascend1:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
```diff
@@ -393,7 +417,8 @@ public static int getPartitions(String table, List<Partition> partitions, boolea
             throw StandardException.plainWrapException(ioe);
         } finally {
             try {
-                root.close();
+                if (root != null)
+                    root.close();
             } catch (IOException e) {
                 // ignore
             }
```
jyuanca, line 424: Same here

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
```diff
@@ -372,7 +395,8 @@ public static int getPartitions(byte[] table, List<Partition> partitions, boolea
             throw StandardException.plainWrapException(ioe);
         } finally {
             try {
-                root.close();
+                if (root != null)
+                    root.close();
```
jyuanca, line 399: You can rewrite this with try-with-resources block

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
```diff
@@ -173,8 +178,26 @@ public StoreCostControllerImpl(TableDescriptor td, ConglomerateDescriptor conglo
                     throw StandardException.plainWrapException(e);
                 }
             }
+            useRealTableStatistics = false;
         } else {
             tableStatistics = new TableStatisticsImpl(tableId, partitionStats,fallbackNullFraction,extraQualifierMultiplier);
+            useRealTableStatistics = true;
+            assert !partitionStatistics.isEmpty();
+
+            List<ColumnStatisticsDescriptor> columnStats = partitionStatistics.get(0).getColumnStatsDescriptors();
```
yxia92, line 187: Ideally, missing column stats should not be all the columns in the table that do not have stats. It should only be the columns that are used by optimizer to estimate selectivity or cost. For example, 
```
create table t2(a2 int, b2 int, c2 int);

call syscs_util.disable_column_statistics('SPLICE', 'T1', 'C1');
analyze table t1;
analyze table t2;
explain
select * from t1, t2 where a1=a2 and b1=1;
```
For this query, we don't care whether there is stats on c1 and b2, c2, we only care if there is stats on b1, a1, and a2 as they are the ones used by optimizer. So we should only report missing stats on these columns, but not on c1,  b2, and c2.

yxia92: So ideally, the missing column or stats info is specific to a query, and the info should be stored in the FromBaseTable instances instead of StoreCostControllerImpl which is shared among all the instances of the same table with in the same session.

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
ascend1: Done.

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
ascend1: Done.

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1437/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1449/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
```diff
@@ -147,6 +150,25 @@ private void addSelectivity(SelectivityHolder holder) {
         return holders;
     }
 
+    private void checkInListPredColumnStats(Predicate p) throws StandardException {
+        if (p.getSourceInList() != null) {
+            for (Object o : p.getSourceInList().leftOperandList) {
+                ColumnReference cr = (ColumnReference) o;
```
yxia92, line 156: Object from the leftOperandList of an inlist may not be a ColumnReference, for example: `a1+1 in (1,2,3)`. So it is better to be conservative to check the type before casting. The above code does not have problem now as currently, only inlist with column can come here.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByNode.java
```diff
@@ -1445,4 +1445,24 @@ public String printExplainInformation(String attrDelim) throws StandardException
...
+    public HashSet<String> getNoStatsColumns() throws StandardException {
+        if (costEstimate == null)
+            throw new RuntimeException("Should not be null");
+
+        HashSet<String> noStatsColumns = new HashSet<>();
+        CollectNodesVisitor cnv = new CollectNodesVisitor(ColumnReference.class);
+        for (OrderedColumn oc : groupingList) {
+            oc.accept(cnv);
+        }
+        for (AggregateNode an : aggregateVector) {
```
yxia92, line 1457: We do not need the stats on columns inside the aggregate function. So there is no need to visit the aggregate function nodes.

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1469/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
ascend1: I see. Added a check before casting.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByNode.java
ascend1: Got it, removed in new commit.

ascend1:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

dgomezferro:
Should we append the missing table stats unconditionally? I thought the idea was to make users more aware of missing stats, and I feel like most of the times they'll be running plain "explain" rather than the extended version.

ascend1:
> Should we append the missing table stats unconditionally? I thought the idea was to make users more aware of missing stats, and I feel like most of the times they'll be running plain "explain" rather than the extended version.

I was doing that in the first few commits, and Yi suggested to have this switch, document it, and by default not reporting missing stats. I think her concern is valid that for big queries, the missing stats report could be long and potentially spam the console, making it difficult to scroll back and see the plan. Of course, we could have further discussions on this.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
```diff
@@ -1513,13 +1517,19 @@ public double joinSelectivity(Optimizable optTable,
 
         if (rightOperand instanceof ColumnReference && ((ColumnReference) rightOperand).getSource().getTableColumnDescriptor() != null) {
             ColumnReference right = (ColumnReference) rightOperand;
+            if (!right.useRealColumnStatistics()) {
+                noStatsColumns.add(right.getSource().getSchemaName() + "." + right.getSource().getFullName());
```
martinrupp, line 1521: since we're using it here multiple times, maybe add function 

```
public String getSchemaTableName() throws StandardException
{
   return getSchemaName() + "." + getFullName();
}
```

to ResultColumn.java

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExplainNode.java
```diff
@@ -104,6 +115,49 @@ public void optimizeStatement() throws StandardException {
             getCompilerContext().setDataSetProcessorType(DataSetProcessorType.FORCED_SPARK);
         }
         node.optimizeStatement();
+
+        // collect tables and columns that are missing statistics only for splice explain
+        // showNoStatsObjects == false for all kinds of spark explain
+        if (!showNoStatsObjects)
+            return;
```
martinrupp, line 122: this "early out" somehow looks dangerous to me. someone might miss that and add something to the end of the function, not seeing most of this function is only executed if showNoStatsObjects.
i'd rather do
```
if( showNoStatsObjects )
{ 
  // code
}
```

or better - extract to a function

if( showNoStatsObjects )
  noStatsColumns = collectMissingStatisticsColumns();

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
```diff
@@ -147,6 +150,27 @@ private void addSelectivity(SelectivityHolder holder) {
         return holders;
     }
 
+    private void checkInListPredColumnStats(Predicate p) throws StandardException {
```
martinrupp, line 153: a "check" function for me is either throwing an error or returning bool true/false = check succeeded / check failed. these two functions are filling the usedNoStatsColumnIds, so i would say a better name would be

checkInListPredColumnStats -> addNoStatsColsFromInListPred(usedNoStatsColumnIds)
checkUnaryAndBinaryPredColumnStats -> addNoStatsColsFromUnaryAndBinaryPred(usedNoStatsColumnIds)

martinrupp (changes requested):
when i'm not doing an ANALYZE SCHEMA, should this also be visible in `explain get no statistics` ? if yes, do we have a test for this?

martinrupp (approved):
i'm fine with submitting it as but consider my comments maybe in a follow up cleanup PR

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
```diff
@@ -31,24 +31,25 @@
...
+ import com.splicemachine.db.iapi.types.DataValueDescriptor;
+ import com.splicemachine.db.iapi.types.Orderable;
+ import com.splicemachine.db.iapi.types.TypeId;
+ import com.splicemachine.db.iapi.util.JBitSet;
+
+ import java.sql.Types;
+ import java.util.HashSet;
+ import java.util.List;
+
+ import static com.splicemachine.db.impl.sql.compile.SelectivityUtil.*;
```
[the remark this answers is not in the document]
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByNode.java
```diff
@@ -1445,4 +1445,22 @@ public String printExplainInformation(String attrDelim) throws StandardException
         return sb.toString();
     }
 
+    public HashSet<String> getNoStatsColumns() throws StandardException {
+        if (costEstimate == null)
+            throw new RuntimeException("Should not be null");
```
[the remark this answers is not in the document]
ascend1: Ah, right, I forget to clean that up. Fixed in the latest commit.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExplainNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
ascend1: Done.

jyuanca (approved):

ascend1:

ascend1:
> i'm fine with submitting it as but consider my comments maybe in a follow up cleanup PR

Comments addressed. :)

ascend1:
> when i'm not doing an ANALYZE SCHEMA, should this also be visible in `explain get no statistics` ? if yes, do we have a test for this?

Yes. If a table never got analyzed, then it's also reported as missing stats. `testReportMissingTableStatistics()` is this case.

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

carolp-503:
See Jira comments/questions.  They don't block this PR though.

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #3994, opened 2020-08-17
DB-9554 check for cancellation requests during long-running compactions.

hatyo:


hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -453,19 +458,20 @@ protected boolean performCompaction(Compactor.FileDetails fd, InternalScanner sc
...
                     bytesWritten += len;
                     if (bytesWritten > closeCheckInterval) {
-                        bytesWritten = 0;
-//                        if (!store.areWritesEnabled()) {
-//                            progress.cancel();
-//                            return false;
-//                        }
+                        bytesWritten = 0; // reset so check whether cancellation is requested in the next
+                                          // <closeCheckInterval>-bytes mark
+                        if (!store.areWritesEnabled()) { // this call could be expensive.
```
dgomezferro, line 463: This might always return false when running in Spark, can you verify that?

hatyo: Sure, I am on it.

hatyo: You're right, when `performCompaction` is called from Spark executor, the status of regions is always false causing compactions to be cancelled all the time.
To solve this, I am adding a `isSpark` flag that is set when running compactions from Spark, we use this flag to make sure we detect region status only when `performCompaction` is called from HBase and not Spark.
Thanks for the hint.

hatyo: ... I added a branch for Spark jobs, if task is interrupted we perform similar cancellation preventing Spark executors from pointlessly running the compaction job.

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -115,14 +116,16 @@ public SpliceDefaultCompactor(final Configuration conf, final Store store, long
...
     public List<Path> compact(CompactionRequestImpl request, ThroughputController throughputController, User user) throws IOException {
         assert request instanceof SpliceCompactionRequest;
+        SpliceCompactionRequest spliceRequest = (SpliceCompactionRequest)request;
         // Used if we cannot compact in Spark
-        ((SpliceCompactionRequest) request).setPurgeConfig(buildPurgeConfig(request));
+        spliceRequest.setPurgeConfig(buildPurgeConfig(request));
 
-        if(!allowSpark || store.getRegionInfo().getTable().isSystemTable())
+        if(!allowSpark || store.getRegionInfo().getTable().isSystemTable()) {
+            isSpark = false;
```
arnaud-splice, line 124: Setting isSpark here is redundant

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -453,19 +462,19 @@ protected boolean performCompaction(Compactor.FileDetails fd, InternalScanner sc
...
                     }
                 }
             }
-            // Log the progress of long running compactions every minute if
-            // logging at DEBUG level
+            // Log the progress of long-running compactions every minute if logging at DEBUG level
             if (LOG.isDebugEnabled()) {
                 if ((now - lastMillis) >= 60 * 1000) {
-                    LOG.debug("Compaction progress: " + progress + String.format(", rate=%.2f kB/sec",
+                    LOG.debug(String.format("Compaction progress: %s, rate=%.2f kB/sec", progress.toString(),
```
arnaud-splice, line 477: Call to toString is redundant, isn't it?

hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
```diff
@@ -107,6 +109,10 @@ public boolean next(List<Cell> list) throws IOException {
         }
     }
 
+    public long getSize() {
```
arnaud-splice, line 112: I suggest renaming size to numberOfScannedBytes or something like this. The size of a scanner is a bit vague.

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
hatyo: Right, but I would still leave it there to protected against e.g. accidentally initializing the member variable to `true`

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
hatyo: fixed.

hbase_storage/src/main/java/com/splicemachine/si/impl/server/AbstractSICompactionScanner.java
hatyo: done.

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1393/

hatyo:

dgomezferro (approved):
Nice fix!

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1397/

hatyo:
Failure seems to be related to DB-9855.

PR #3995, opened 2020-08-17
DB-10027 Fix grouping by out-of-bound positions

ascend1:


ascend1:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByList.java
```diff
@@ -110,6 +110,10 @@ public void bindGroupByColumns(SelectNode select,List<AggregateNode> aggregateVe
             if (OrderByColumn.isReferedColByNum(groupByCol.getColumnExpression())) {
                 ResultColumnList targetCols = select.getResultColumns();
                 int columnPosition = (Integer) groupByCol.getColumnExpression().getConstantValueAsObject();
+                if (columnPosition <= 0) {
+                    throw StandardException.newException(SQLState.LANG_INVALID_GROUP_POSITION,
```
yxia92, line 114: Can we throw the same error message like the one in the below code which comes a few lines after this line:
```
			    /* Column is out of range if either a) resultCol is null, OR
			     * b) resultCol points to a column that is not visible to the
			     * user (i.e. it was generated internally).
			    */
                if ((resultCol == null) ||
                        (resultCol.getColumnPosition() > targetCols.visibleSize()))
                {
                    throw StandardException.newException(SQLState.LANG_COLUMN_OUT_OF_RANGE,
                            String.valueOf(columnPosition));
                }
```
I think they are of similar nature, so it would be better to return consistent error message.

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/GroupByList.java
ascend1: Oh yes, sorry that I didn't see the exception below. I was searching only using the term "GROUP". Fixed in the new commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1374/

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #4001, opened 2020-08-18
DB-9897 Change the default schema of CTEs from current to SESSION

ascend1:


ascend1:

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1382/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1388/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateViewNode.java
```diff
@@ -291,9 +291,6 @@ public void bindStatement() throws StandardException
 			//cannot define views on temporary tables
 			if (queryExpr instanceof SelectNode)
 			{
-				//If attempting to reference a SESSION schema table (temporary or permanent) in the view, throw an exception
```
yxia92, line 294: I think we should not remove this check completely, we allow dynamic view to be in session schema, but not regular views. Right now with the change, regular view will be created successfully using the session schema:
```
splice> create view session.v1 as select a1 from t1;
ELAPSED TIME = 129 milliseconds
splice> select * from session.v1;
ERROR XJ001: DERBY SQL error: SQLCODE: -1, SQLSTATE: XJ001, SQLERRMC: java.lang.NullPointerExceptionXJ001.U
splice> quit;
```

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/WithStatementIT.java
```diff
@@ -209,5 +231,34 @@ public void testInsertContainingWith() throws Exception {
...
+            assertEquals(expectedSchema, TestUtils.FormattedResult.ResultFactory.toString(rs));
+        }
 
+        // user with no write privilege, dynamic view in SESSION schema
+        try (ResultSet rs = testUserConn.createStatement().executeQuery(format("with dt as (select * from t12) select * from dt"))) {
+            assertEquals(expected, TestUtils.FormattedResult.ResultFactory.toString(rs));
+        }
+        // user with no write privilege, dynamic view in current schema
+        assertFailed(testUserConn, format("with %s.dt as (select * from t12) select * from dt", SCHEMA), SQLState.AUTH_NO_ACCESS_NOT_OWNER);
+    }
```
yxia92, line 263: Could you add a test case to test that in two concurrent connections, we can create two dynamic views using the same name under the session schema, and they will not interfere with each other? Thanks!

ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateViewNode.java
ascend1: Hmm, I agree that removing the check is not correct. Basically, we want to allow a CTE to reference another CTE, which could be in SESSION schema. But we don't want to allow a normal view to reference any objects in SESSION. I fix this part in the new commit by adding an `isDynamic` flag to `CreateViewNode`.

However, the example you showed is also very interesting. Since the check is `queryExpr.referencesSessionSchema()`, we only check if the view definition references any objects in SESSION schema, not if the view is created under SESSION schema. I tried your example on the latest master, and it causes the same error.

I think we should definitely disallow this case, too. So in the latest commit, I also improved the condition to make sure we throw if users want to create a view under SESSION. Please take a look if it is good enough.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1400/

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateViewNode.java
yxia92: The change looks good to me. A minor comment is that the call with the updated version of `this.referencesSessionSchema()` completely covers `queryExpr.referencesSessionSchema()`, so there is some redundancy. Maybe you can leave unchanged, and directly add the check `getObjectName().schemaName.equals(SchemaDescriptor.STD_DECLARED_GLOBAL_TEMPORARY_TABLES_SCHEMA_NAME)` here?

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/WithStatementIT.java
yxia92: Thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateViewNode.java
ascend1: Sure, done in the latest commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1418/

arnaud-splice (approved):

PR #4007, opened 2020-08-18
DB-10045 add MIN_RETAINED_VERSIONS column to SYSTABLES

hatyo:
This is part of the ongoing effort to implement minimum version retention for tables. For more information please refer to the parent JIRA (DB-9722)


jyuanca:
From Jira
"Since we want to have minimum retained versions on table-granularity we should save for each table its own minimum retained versions. SYSTABLES system table is the right place to save this information."

Can you explain what is minimum retained versions, and why it is needed?

hatyo:
@jyuanca The minimum retained version is the minimum amount of versions the users wants to keep for a certain table. This is needed to make the behavior of time travel consistent since flushing / compaction could effectively change how table looks like in the past.

For example, if row `r1` in table `tbl` was updated several times in the past as the following (time here is loosely implying tx id, just for illustration):

| Time          | Value of `r1` |
| ------------- | ------------- |
| 1             | a             |
| 2             | b             |
| 3             | c             |
| 4             | d             |
| 5             | e             |

Before compaction, doing the following time travel query:
`SELECT * FROM tbl AS OF 3;`
we get back:
`c`

After compaction, we remove all the past updates, therefore running the same query:
`SELECT * FROM tbl AS OF 3;`
we get back nothing.

So, the user in this case gets the wrong past data.

To overcome this, we need to give the user the ability to set control how many versions of a specific table we should keep and prevent from flushing / compaction. Therefore we need to this column `minRetentionPeriod`, the field is given as a duration in seconds (more about that later). Coming back to our example, if the user sets `minRetentionPeriod` for table `tbl` to e.g. `3` then we can compact only until `3`. i.e. it will look like this:

| Time          | Value of `r1` |
| ------------- | ------------- |
| 3             | c             |
| 4             | d             |
| 5             | e             |

This allows the user to get consistent past results up until point `3` of time.

**How do we use `minRetentionPeriod`?**

Currently, when doing flushing and compaction, we consider the `lowTransactionWatermark` which is the begin timestamp of the oldest active transaction as a threshold.
In the future, we will do the following:

1. Read the `minRetentionPeriod` for the table we are about to flush / compact.
2. Translate this period into a transaction ID (please look at DB-9715 for more information).
3. Instead of having `lowTransactionWatermark` as a threshold, we now choose `Math.min(txIdOf(minRetentionPeriod), lowTransactionWatermark)`.

hatyo:
Thanks for the comment, @jyuanca will do that.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
```diff
@@ -52,661 +52,664 @@
...
+                        new ColumnDescriptor(COMPRESSION, 13, 13,
+                                DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.VARCHAR, true),
+                                null, null, view, viewId, 0, 0, 0),
+                        new ColumnDescriptor(IS_PINNED, 14, 14,
+                                DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.BOOLEAN, false),
+                                null, null, view, viewId, 0, 0, 0),
+                        new ColumnDescriptor(PURGE_DELETED_ROWS, 15, 15,
+                                DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.BOOLEAN, false),
+                                null, null, view, viewId, 0, 0, 0),
+                        new ColumnDescriptor(MIN_RETENTION_PERIOD, 16, 16,
```
yxia92, line 638: As @jyuanca mentioned, we need an upgrade script to add the new column to the table for existing system. The upgrade script should also cover the update of the view. Please refer to SpliceDataDictionary.updateColumnViewInSys() for the logic to update the sysvw.systablesview.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
```diff
@@ -52,661 +52,664 @@
...
+                SystemColumnImpl.getColumn(COLSEQUENCE, Types.INTEGER, false),
+                SystemColumnImpl.getColumn(DELIMITED, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(ESCAPED, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(LINES, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(STORED, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(LOCATION, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(COMPRESSION, Types.VARCHAR, true),
+                SystemColumnImpl.getColumn(IS_PINNED, Types.BOOLEAN, false),
                 SystemColumnImpl.getColumn(PURGE_DELETED_ROWS, Types.BOOLEAN, false),
+                SystemColumnImpl.getColumn(MIN_RETENTION_PERIOD, Types.BIGINT, false),
```
yxia92, line 586: We may need to define MIN_RETENTION_PERIOD as nullable, since the existing rows upgraded from lower versions won't have this field, so it's value is null.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
```diff
@@ -52,661 +52,664 @@
...
+                row.getColumn(SYSTABLES_COLUMN_SEQUENCE).getInt(),
+                delimitedDVD != null ? delimitedDVD.getString() : null,
+                escapedDVD != null ? escapedDVD.getString() : null,
+                linesDVD != null ? linesDVD.getString() : null,
+                storedDVD != null ? storedDVD.getString() : null,
+                locationDVD != null ? locationDVD.getString() : null,
+                compressionDVD != null ? compressionDVD.getString() : null,
+                isPinnedDVD.getBoolean(),
+                purgeDeletedRowsDVD.getBoolean(),
+                minRetentionPeriodDVD.getLong()
```
yxia92, line 536: We may consider minRetentionPeriodDVD as NULL here for rows upgrade from lower versions.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSTABLESRowFactory.java
hatyo: Thanks, I addressed your comment, but not sure if it is exactly correctly.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1645,4 +1645,47 @@ public void addCatalogVersion(TransactionController tc) throws StandardException
...
+
+            // now add the column to the table's column descriptor list.
+            td.getColumnDescriptorList().add(columnDescriptor);
+            updateSYSCOLPERMSforAddColumnToUserTable(td.getUUID(), tc);
+
+            SpliceLogUtils.info(LOG, String.format("%s upgraded: added a column: %s.", SYSTABLESRowFactory.TABLENAME_STRING,
+                    SYSTABLESRowFactory.MIN_RETENTION_PERIOD));
+
+            // finally, update the SYSTABLESVIEW
+            ViewDescriptor vd=getViewDescriptor(td);
```
yxia92, line 1682: We should not use the table's table descriptor `td` to get the view descriptor. The SYSTABLESVIEW is inside the schema sysvw. So we should use
```
        SchemaDescriptor sysVWSchema=sysViewSchemaDesc;
        TableDescriptor td1 = getTableDescriptor("SYSTABLESVIEW", sysVWSchema, tc);
        ViewDescriptor vd=getViewDescriptor(td1);
```

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1645,4 +1645,47 @@ public void addCatalogVersion(TransactionController tc) throws StandardException
...
+            dropViewDescriptor(vd, tc);
+            DataDescriptorGenerator ddg=getDataDescriptorGenerator();
+            vd=ddg.newViewDescriptor(td.getUUID(),SYSTABLESRowFactory.SYSTABLE_VIEW_NAME, SYSTABLESRowFactory.SYSTABLE_VIEW_SQL,0,sd.getUUID());
+            addDescriptor(vd,sd,DataDictionary.SYSVIEWS_CATALOG_NUM,true, tc,false);
```
yxia92, line 1686: Also since the new column you add to the view is not the last column in the view, it is better to drop all the columns in the view together with the view and re-create the view, similar to what we do in SpliceDataDictionary.updateColumnViewInSys().

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: done.

hatyo:
I think now it is works as expected.

I performed a manual upgrade from 1971:

before upgrade:

```
splice> select * from sysvw.systablesview;
REFERENCEID|COLUMNNAME|COLUMNNUMBER|STORAGENUMBER|COLUMNDATATYPE|COLUMNDEFAULT|COLUMNDEFAULTID|AUTOINCREMENTVALUE|AUTOINCREMENTSTART|AUTOINCREMENTINC|COLLECTSTATS|PARTITIONPOS
ITION|USEEXTRAPOLATION|TABLENAME|SCHEMANAME
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-------------------------------------------
601a400f-0174-3601-99a7-000017f24530|SYSCOLUMNS|V|c013800d-00f8-5b53-28a9-00000019ed88|R|4.0|-1|NULL|NULL|NULL|NULL|NULL|NULL|false|false
...
```

during upgrade, we see these messages in the logs:

```
7053 2020-08-28 19:21:32,276 (SpliceDatabaseLifecycleManager) INFO  [c.s.d.i.s.c.u.UpgradeScriptBase] - Adding minRetentionPeriod column to system table SYSTABLES and updating      SYSTABLESVIEW
7054 2020-08-28 19:21:32,294 (RpcServer.default.FPBQ.Fifo.handler=48,queue=3,port=60020) INFO  [o.a.h.h.r.HRegion] - writing data to region splice:80,,1598634766141.e778edf309     7866bdb9732ecdbd31c8ad. with WAL disabled. Data may be lost in the event of a crash.
7055 2020-08-28 19:21:32,309 (SpliceDatabaseLifecycleManager) INFO  [c.s.d.i.s.c.SpliceDataDictionary] - SYSTABLES upgraded: added a column: MIN_RETENTION_PERIOD.
7056 2020-08-28 19:21:32,323 (RpcServer.default.FPBQ.Fifo.handler=48,queue=3,port=60020) INFO  [o.a.h.h.r.HRegion] - writing data to region splice:64,,1598634766140.3a2a554788     d0402c8a72aeb62fdfe59e. with WAL disabled. Data may be lost in the event of a crash.
7057 2020-08-28 19:21:32,370 (RpcServer.default.FPBQ.Fifo.handler=48,queue=3,port=60020) INFO  [o.a.h.h.r.HRegion] - writing data to region splice:368,,1598634774083.21b9f5be9     c62bb315b51a144ba2e9bde. with WAL disabled. Data may be lost in the event of a crash.
7058 2020-08-28 19:21:32,397 (SpliceDatabaseLifecycleManager) INFO  [c.s.d.i.s.c.SpliceDataDictionary] - View: SYSTABLESVIEW in SYSVW is created!
7059 2020-08-28 19:21:32,397 (SpliceDatabaseLifecycleManager) INFO  [c.s.d.i.s.c.SpliceDataDictionary] - SYSTABLESVIEW upgraded: added a column: MIN_RETENTION_PERIOD.
```

after upgrade:

```
splice> select * from sysvw.systablesview;
TABLEID|TABLENAME|TABLETYPE|SCHEMAID|LOCKGRANULARITY|VERSION|COLSEQUENCE|DELIMITED|ESCAPED|LINES|STORED|LOCATION|COMPRESSION|IS_PINNED|PURGE_DELETED_ROWS|MIN_RETENTION_PERIOD|
SCHEMANAME
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
----------
601a400f-0174-3615-f5b0-000008198608|SYSCOLUMNS|V|c013800d-00f8-5b53-28a9-00000019ed88|R|4.0|-1|NULL|NULL|NULL|NULL|NULL|NULL|false|false|NULL|SYSIBM
...
```

```
splice> select * From sys.systables;
TABLEID|TABLENAME|TABLETYPE|SCHEMAID|LOCKGRANULARITY|VERSION|COLSEQUENCE|DELIMITED|ESCAPED|LINES|STORED|LOCATION|COMPRESSION|IS_PINNED|PURGE_DELETED_ROWS|MIN_RETENTION_PERIOD
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
601a400f-0174-3615-f5b0-000008198608|SYSCOLUMNS|V|c013800d-00f8-5b53-28a9-00000019ed88|R|4.0|-1|NULL|NULL|NULL|NULL|NULL|NULL|false|false|NULL
```

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: done.

hatyo:

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1484/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1485/

hatyo:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #4012, opened 2020-08-20
 DB-8256 allow nestedloop join condition to be pushed down to derived table

yxia92:


yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/NLJPredicatePushedToDT.java
```diff
@@ -0,0 +1,593 @@
...
+
+import static com.splicemachine.test_tools.Rows.row;
+import static com.splicemachine.test_tools.Rows.rows;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by yxia on 8/19/20.
+ */
+@RunWith(Parameterized.class)
+public class NLJPredicatePushedToDT extends SpliceUnitTest {
```
dgomezferro, line 45: Nitpick: can you end the name in *IT to follow the convention?

dgomezferro (approved):

yxia92:
@dgomezferro I pushed a new commit to prevent nestedloop join from pushing down for window function as it is not semantically equivalent and could lead to wrong result. Please take a look at the last commit. More detail is in the JIRA. Thanks!

PR #4046, opened 2020-08-25
DB-9749 sync user roles before running remote queries.

hatyo:
Previously, certain code paths that run remote jobs such as ASQ where getting the user roles as-is from the current SQL session context. Problem is, the retrieved list of roles is not kept up to date, meaning if the DBA revoked some roles from the user running ASQ, it wouldn't have any effect since the list of roles is not updated and the revoked role is not removed from it.

With this fix, we always refresh the list of roles before retrieving it.

I chose not to make refresh logic part of `LanguageConnectionContext.getCurrentRoles` as it might affect performance since it is used in many other locations that does proper cleansing of the role list already.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
+++ b/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
@@ -22,6 +22,7 @@ import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.sql.Activation;
 import com.splicemachine.db.iapi.sql.ResultColumnDescriptor;
+import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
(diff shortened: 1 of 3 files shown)


dgomezferro (approved):

arnaud-splice (approved):

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
```diff
@@ -122,6 +122,8 @@ public void submit() throws StandardException {
                     streamingBatches, streamingBatchSize, parallelPartitions, shufflePartitionsProperty);
 
             String requestedQueue = (String) activation.getLanguageConnectionContext().getSessionProperties().getProperty(SessionProperties.PROPERTYNAME.OLAPQUEUE);
+            // remove any stale revoked roles (DB-9749)
```
yxia92, line 125: We can skip the check of roles if user is database owner, and use the requestedQueue directly.

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #4061, opened 2020-08-30
DB-8049 handle begin or end rows for window function properly (2.8)

yxia92:
DB-8049 fix LEAD/LAG


carolp-503 (approved):
Approved for 2.8

PR #4063, opened 2020-08-31
DB-10067 Support creating indexes on expressions

ascend1:


ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1489/

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -269,20 +322,35 @@ else if (isUniqueWithDuplicateNulls)
...
         isUnique = fh.getBoolean("isUnique");
-        int bcpLength = fh.getInt("keyLength");
-        baseColumnPositions = new int[bcpLength];
-        isAscending = new boolean[bcpLength];
-        for (int i = 0; i < bcpLength; i++) {
-            baseColumnPositions[i] = fh.getInt("bcp" + i);
-            isAscending[i] = fh.getBoolean("isAsc" + i);
-        }
+        int keyLength = fh.getInt("keyLength");
+		isAscending = new boolean[keyLength];
```
martinrupp, line 341: is that tab/spaces mixed?

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -269,20 +322,35 @@ else if (isUniqueWithDuplicateNulls)
...
+		for (int i = 0; i < keyLength; i++) {
+			isAscending[i] = fh.getBoolean("isAsc" + i);
+		}
+		int numBaseColumns = fh.containsKey("numBaseColumns") ? fh.getInt("numBaseColumns") : keyLength;
```
martinrupp, line 345: we should maybe use sth like `static final String numBaseColumnsKey = "numBaseColumns"` for this (also other keys)

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
```diff
@@ -150,17 +185,23 @@ public ExecIndexRow getNullIndexRow(ColumnDescriptorList columnList,
...
+		if (columnTypes.length == 0) {
+			int[] baseColumnPositions = id.baseColumnPositions();
+			for (int i = 0; i < baseColumnPositions.length; i++) {
+				DataTypeDescriptor dtd =
+						columnList.elementAt(baseColumnPositions[i] - 1).getType();
+				indexRow.setColumn(i + 1, dtd.getNull());
+			}
+			indexRow.setColumn(baseColumnPositions.length + 1, rowLocation);
+		} else {
+			for (int i = 0; i < columnTypes.length; i++) {
```
martinrupp, line 202: it's not immediately clear why you don't use id.baseColumnPositions(); in this else branch, can you maybe add a comment in the code?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
```diff
@@ -230,49 +271,69 @@ private void getIndexRowHelper(ExecRow baseRow,
...
-			for (int i = 0; i < colCount ; i++)
-			{
-				indexRow.setColumn(i + 1,
-						baseRow.getColumn(baseColumnPositions[i]));
+		int colCount;
+		if (isOnExpression()) {
+			colCount = isAscending().length;
+			ExecRow expandedRow;
+
+			if (bitSet == null) {
```
martinrupp, line 281: maybe extract function getExpandedRow() ?

martinrupp: why is this code so different from the code in lines 326-336? it somehow has the feel it's doing the same, it should maybe look the same, or maybe use same code (e.g. extract a function which returns an array of pairs expandedRowIndex -> column, so that i could write e.g. code like

```
expandedRow = new ValueRow(maxNumCols);
auto mapping = getRowColMapping();
for( auto pair : mapping )
{
  expandedRow.setColumn(pair.first, pair.second);
}
```

and the code below would be

```
auto mapping = getRowColMapping();
for( auto pair : mapping )
{
  indexRow.setColumn(pair.first, pair.second);
}
```

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
```diff
@@ -489,4 +553,33 @@ public int getTypeFormatId()
...
+	/** @see IndexDescriptor#getExecutableIndexExpression */
+	@Override
+	public BaseExecutableIndexExpression getExecutableIndexExpression(int indexColumnPosition)
+			throws StandardException
+	{
+    	return id.getExecutableIndexExpression(indexColumnPosition);
+	}
+
+	private int getMaxBaseColumnPosition() {
+    	int maxPosition = 1;  // base column positions are 1-based
```
martinrupp, line 584: tabs/spaces mix?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
```diff
@@ -64,8 +68,9 @@
 	String				indexType;
 	TableName			indexName;
 	TableName			tableName;
-	Vector				columnNameList;
-	String[]			columnNames = null;
+	Vector<IndexExpression> expressionList;
```
martinrupp, line 70: maybe align like other members? (also below)

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
```diff
@@ -264,6 +280,55 @@ public void bindStatement() throws StandardException
 //  												 sd.getSchemaName());
 //  		}
 
+		if (onExpression) {
```
martinrupp, line 280: extract into function

martinrupp (changes requested):

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
ascend1: Yes I also found my Intellij is confused about mixing tabs and spaces after pushing these commits. I try to fix them in the following commits.

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
ascend1: Good point, would make it less error-prone. Done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
ascend1: The code in lines 326 - 336 (in latest commit 329 - 339) doesn't actually expand `baseRow`. Consider a partial `baseRow` that has the original columns 3, 5, 7. The row looks like this:

```
original col pos    :  3   |   5   |   7
row column pos   :  0   |   1    |   2
```
`bitSet` should have bits for 3, 5, and 7 set. Let's suppose the index column to base column mapping (i.e., `baseColumnPositions`) is 5, 3, 7. That is, the first index column is full base row column 5, the second index column is full base row column 3, and so on. Btw, column position is always 1-based.

The code tries to assign columns in `indexRow` directly from the partial row. To achieve that, it needs to get the position of the full base row column in this partial row. Take the first index column as an example:

`fullColumnNumber = baseColumnPositions[0]`, which is 5. Now we count the bitSet from 1 to 5 inclusive. 3 is set, 5 is set, so we get `partialColumnNumber = 2`. That is, the full base row column 5 is at position 2 in this partial row.

The new code doesn't do this. Instead, it creates a new row, trying to be wide enough to hold the columns at their full base row positions. Example of an expanded row:

```
original column pos :  x | x | x | 3 | x | 5 | x | 7
expanded row pos   :  0 | 1 | 2 | 3 | 4 | 5 | 6 | 7
```

We need to expand the row as a way to normalize the input for generated code. During code generation, we only know full base row column position, not with any mapping. So for generated code, it assumes the row passed in is a full base row.

We can unify the two code paths by always expanding the row. But then it would be unnecessary extra cost in case of normal index creation. That's one point I mentioned in the design discussion the other day.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
ascend1: Fixed in following commits.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
ascend1: Yes, fixed in the following commits.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
```diff
@@ -371,32 +439,79 @@ public ConstantAction	makeConstantAction() throws StandardException
...
+			columnNameSet.toArray(involvedColumnNames);
+		}
+		else {
+			Hashtable	ht = new Hashtable(size + 2, (float) .999);
+			involvedColumnNames = new String[size];
+
+			for (int index = 0; index < size; index++) {
+				/* Verify that this column's name is unique within the list
+				 * Having a space at the end meaning descending on the column
+				 */
```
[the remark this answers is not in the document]
ascend1: Good catch! Removed.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1502/

spliceci:
TEST SUCCEEDED +1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1526/

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
```diff
@@ -140,170 +150,262 @@ public void init(
...
+//              throw StandardException.newException(SQLState.LANG_OBJECT_ALREADY_EXISTS_IN_OBJECT,
+//                                                   "Index",
+//                                                   indexName.getTableName(),
+//                                                   "schema",
+//                                                   sd.getSchemaName());
+//          }
+
+        if (onExpression) {
+            bindIndexExpressions();
+            generateExecutableIndexExpression();
```
yxia92, line 280: It seems more appropriate to put  generateExecutableIndexExpression() in makeConstantAction() at code generation phase.

yxia92 (changes requested):
Code change looks good. I just have a couple of minor comments.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
ascend1: Right, that makes more sense. Fixed in new commit.

splice_machine/src/main/java/com/splicemachine/protobuf/ProtoUtil.java
```diff
@@ -300,6 +305,19 @@ public static Index createIndex(long conglomerate, IndexDescriptor indexDescript
...
+        for (ByteArray bc : bytecodeArray) {
+            builder = builder.addBytecodeExprs(ByteString.copyFrom(bc.getArray(), bc.getOffset(), bc.getLength()));
+        }
+        String[] classNames = indexDescriptor.getGeneratedClassNames();
+        for (String cn : classNames) {
+            builder = builder.addGeneratedClassNames(cn);
+        }
+        DataTypeDescriptor[] indexColumnTypes = indexDescriptor.getIndexColumnTypes();
+        for (DataTypeDescriptor dtd : indexColumnTypes) {
+            builder = builder.addIndexColumnFormatIds(dtd.getNull().getTypeFormatId());
```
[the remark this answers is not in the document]
ascend1: Bingo! Now I see it's really a trap for people. I was doing `dtd.getTypeId().getTypeFormatId()` at the beginning and then spent about half an hour debugging what's wrong.

Problem is that `getTypeFormatId()` call from `TypeId` class returns something like `BOOLEAN_TYPE_ID`, `CHAR_TYPE_ID`, etc., while `getTypeFormatId()` call from a concrete SQL type class, say `SQLBoolean`, returns `SQL_BOOLEAN_ID`. Since we need to encode/decode concrete SQL values later, we have to get the latter format IDs. Otherwise it would complain something about no suitable serializer.

For normal index, the DDL message contains a list of all base column format IDs in the table message. There we call `getFormatIds` from `TableDescriptor`, which does `type.getNull().getTypeFormatId()`, too.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
martinrupp: ah, ok

martinrupp (approved):

splice_machine/src/main/java/com/splicemachine/protobuf/ProtoUtil.java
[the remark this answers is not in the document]
yxia92: I see, thanks for the explanation!

yxia92 (approved):
+1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1542/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1548/

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

carolp-503 (approved):
+1

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #4077, opened 2020-09-02
DB-8357 no implicit schema creation for queries

martinrupp:


ipraznik-splice (approved):

ascend1 (commented):
Would this change alter the behavior of auto-creating schema when a user connects? After reading the thread in JIRA I understand that it is needed by some customer.

PR #4093, opened 2020-09-07
DB-10194 Create new database through JDBC connect

arnaud-splice:


yxia92 (commented):
I understand this PR is WIP. Here are some thoughts/questions, please treat them as todo suggestions if not already considered or things to clarify:

1.  It is not clear from the design the relationship between users and databases. Is a user's scope within a specific database or it is across multiple databases? How about roles? Right now, username and role name are unique, does that still hold with multi-database support?
Do we have one super user for all the databases or each database will have its own database owner. If the latter, would there be a super super user for the whole system?
The answer for the above could also affect the authorization logic. 
2. Muti-database support would impact ranger authorization too. For example, we now have a new object `database` in addition to `schema`, `table`, `column`, ..., do we want it to be visible to ranger? I guess the answer is YES.
3. There is a system prodecure to add a database, do we also need one to delete a database?
4. We may consider to add schema id and database id to more system tables and its corresponding indexes to help improve the metadata query performance.

arnaud-splice:
Thanks a lot for all these comments @yxia92!

> I understand this PR is WIP. Here are some thoughts/questions, please treat them as todo suggestions if not already considered or things to clarify:
> 
> 1. It is not clear from the design the relationship between users and databases. Is a user's scope within a specific database or it is across multiple databases? How about roles? Right now, username and role name are unique, does that still hold with multi-database support?

username and role names will be associated to a given database ID and the pairs <DBID, rolename> or <DBID, username> will be unique. This is not covered by this PR though and will be addressed by DB-10199.

>    Do we have one super user for all the databases or each database will have its own database owner. If the latter, would there be a super super user for the whole system?

Each database will have its own database owner. Introducing a super super user might be necessary, but hasn't been decided yet. This is also part of DB-10199.

>    The answer for the above could also affect the authorization logic.
> 2. Muti-database support would impact ranger authorization too. For example, we now have a new object `database` in addition to `schema`, `table`, `column`, ..., do we want it to be visible to ranger? I guess the answer is YES.

I am not familiar with ranger authorization yet, so I can't answer that. I will get back to you on that one once I've read more about it.

> 3. There is a system prodecure to add a database, do we also need one to delete a database?

Covered by DB-10222


> 4. We may consider to add schema id and database id to more system tables and its corresponding indexes to help improve the metadata query performance.

Agreed. I will address the different system tables in the following jiras.

db-tools-ij/src/main/grammar/ij.jj
```diff
@@ -983,168 +983,197 @@ class ij {
...
+    /**
+       Return a resultset of databases from database metadata
+     */
+    public ijResult showDatabases() throws SQLException {
+        ResultSet rs = null;
+        try {
+            haveConnection();
+
+            rs = theConnection.createStatement().executeQuery
+                ("SELECT DATABASENAME FROM SYS.SYSDATABASES");
```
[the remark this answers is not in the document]
arnaud-splice: Done!

yxia92 (commented):
+1

db-client/src/main/java/com/splicemachine/db/client/am/PreparedStatement.java
```diff
@@ -941,8 +944,8 @@ public void setRowId(int parameterIndex, java.sql.RowId x) throws SQLException{
                     agent_.logWriter_.traceEntry(this, "setRowId", parameterIndex, x);
                 }
 
-                final int paramType =
-                        getColumnMetaDataX().getColumnType(parameterIndex);
+                getColumnMetaDataX().getColumnType(parameterIndex);
```
ipraznik-splice, line 947: this call is not assign to any variable, usually paramType. Looks like not needed any more and can be removed.

db-client/src/main/java/com/splicemachine/db/client/net/NetConnectionReply.java
```diff
@@ -296,8 +289,9 @@ private void parseSECCHKreply(NetConnection netConnection) throws DisconnectExce
             // incorrect but consider it a conversation protocol error
             // 0x03 - OBJDSS sent when not allowed.
             //parseSECTKN (true);
-            boolean done = false;
-            byte[] bytes = parseSECTKN(false);
+            //boolean done = false;
+            //byte[] bytes = parseSECTKN(false);
+            parseSECTKN(false);
```
ipraznik-splice, line 294: this call is not assign to any variable

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedConnection.java
```diff
@@ -101,7 +102,8 @@
  */
 public abstract class EmbedConnection implements EngineConnection
 {
-	protected static final StandardException exceptionClose = StandardException.closeException();
+    public static final ThreadLocal<Boolean> isCreate = new ThreadLocal<>();
```
ipraznik-splice, line 105: is it used somewhere in subclasses? in EmbedConnection I see only .set and .remove

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
```diff
@@ -51,14 +51,18 @@
     protected LanguageConnectionContext lcc;
     protected String ipAddress;
 
-    public SpliceTransactionResourceImpl() throws SQLException{
-        this(CONNECTION_STRING, new Properties());
+    public SpliceTransactionResourceImpl() throws SQLException{ // XXX Remove this one and properly pass the right DB
```
ipraznik-splice, line 54: Should be removed?:)

ipraznik-splice (approved):
Please look at my comments, if it's needed to be changed

db-client/src/main/java/com/splicemachine/db/client/am/PreparedStatement.java
arnaud-splice: I would leave this like this since `getColumnType` can throw. We never used `paramType`, but if we want to keep the current behaviour, then we should continue calling this function.

db-client/src/main/java/com/splicemachine/db/client/net/NetConnectionReply.java
arnaud-splice: (spotbug) It simply moves the parser forward. We don't do anything with the result.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/EmbedConnection.java
arnaud-splice: It is used in `SpliceDatabase.bootStore`
It is quite similar to `EngineLifecycleService.isCreate` that we use when creating the very first `splicedb`. However, when we create a new database, we don't go all the way to `EngineLifecycleService.start()` so we cannot use that one class variable.

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
arnaud-splice: This will be addressed in the following jiras

PR #4099, opened 2020-09-08
DB-10026 Make derbyclient dependency test only (redux)

OlegMazurov:
Don’t rerun splice_machine unit tests from mem_sql: that is redundant and the classpath is not right.


jyuanca (approved):

ipraznik-splice (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #4105, opened 2020-09-09
DB-10195 Support 2 schemas with the same name in 2 different DBs

arnaud-splice:


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -994,6 +994,9 @@ public SchemaDescriptor getSchemaDescriptor(String schemaName,
         if(tc==null){
             tc=getTransactionCompile();
         }
+        if (dbId == null) {
+            dbId = getLCC().getDatabaseId();
+        }
 
         if(getSystemSchemaDescriptor().getSchemaName().equals(schemaName)){
```
yxia92, line 1001: We should check not only SysSchemaDescriptor and SysIBMSchemaDescriptor here, but also the below list of system schemas:
```
    protected SchemaDescriptor sysIBMADMSchemaDesc;
    protected SchemaDescriptor systemUtilSchemaDesc;
    protected SchemaDescriptor sysFunSchemaDesc;
    protected SchemaDescriptor sysViewSchemaDesc;
```
Otherwise, we cannot query objects in these schemas from other databases.

yxia92 (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
arnaud-splice: Done

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
```diff
@@ -376,9 +377,23 @@ public static void preAlterStats(DDLMessage.DDLChange change, DataDictionary dd,
...
-        dd.getDataDictionaryCache().schemaCacheRemove(change.getDropSchema().getSchemaName());
-        dd.getDataDictionaryCache().oidSchemaCacheRemove(ProtoUtil.getDerbyUUID(change.getDropSchema().getSchemaUUID()));
-
+        SpliceTransactionResourceImpl transactionResource = null;
+        boolean prepared = false;
+        try {
+            TxnView txn = DDLUtils.getLazyTransaction(change.getTxnId());
+            transactionResource = new SpliceTransactionResourceImpl();
+            prepared = transactionResource.marshallTransaction(txn);
+            LanguageConnectionContext lcc = transactionResource.getLcc();
```
martinrupp, line 386: code to get lcc.getDatabaseId()

martinrupp (approved):

yxia92 (approved):
+1

PR #4107, opened 2020-09-10
DB-10018 skip outer join conditions that are released from flattened …(3.0)

yxia92:
…outer join to where clause when checking for unssatisfiable block


carolp-503 (approved):
Approved for 3.0

PR #4110, opened 2020-09-10
DB-10175 Qualify expression-based indexes

ascend1:


ascend1:

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1558/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -436,18 +435,20 @@ public boolean nextAccessPath(Optimizer optimizer,
...
-                        rowOrdering.nextOrderPosition(rowOrderDirection);
-                        pos = rowOrdering.addOrderedColumn(rowOrderDirection,getTableNumber(),baseColumnPositions[i]);
-                    }
-                    // check if the column has a constant predicate like "col=constant" defined on it,
-                    // if so, we can treat it as sorted as it has only one value
-                    if (pos >=0 &&    /* a column ordering is added or exists */
-                        hasConstantPredicate(getTableNumber(), baseColumnPositions[i], predList)) {
-                        ColumnOrdering co = rowOrdering.getOrderedColumn(pos);
-                        co.setBoundByConstant(true);
+                    if (!irg.isOnExpression()) {
```
ascend1, line 438: Here we set proper ordering to base table columns according to index column ordering. For expression-based indexes, however, I'm not sure what to do here since whether an index column is ascending or descending doesn't really mean anything to the ordering of base table columns.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -516,8 +517,11 @@ public boolean isCoveringIndex(ConglomerateDescriptor cd) throws StandardExcepti
             return false;
 
         IndexRowGenerator irg=cd.getIndexDescriptor();
-        int[] baseCols=irg.baseColumnPositions();
 
+        if (irg.isOnExpression())
+            return false;
```
ascend1, line 522: For now, simply take all expression-based indexes as non-covering indexes. Doing so limits the scope of this change to a manageable size for code review. Treating an expression-based index as a covering index is DB-10236 and will have its own PR.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1952,43 +1980,61 @@ private ResultColumnList newResultColumns(
...
 
-        for(int basePosition : baseCols){
-            ResultColumn oldCol=oldColumns.getResultColumn(basePosition);
-            ResultColumn newCol;
+        if (irg.isOnExpression()) {
+            // When building new ResultColumn instances, we don't need to set expression
+            // or virtual column number as they are not needed. In case of a scan on an
+            // expression-based index, these are just placeholders. All we care about is
+            // that they should all be referenced for now so that we can build the
+            // template row, then we will clear reference status and set them properly.
```
ascend1, line 1993: This comment is true only for this change because all expression-based indexes are non-covering indexes. In next step, to rewrite an index expression to a column reference, we do the following:
```
rc.setIndexExpression(exprAst);
rc.setReferenced();
rc.setVirtualColumnId(i + 1);  // virtual column IDs are 1-based
rc.setName(idxCD.getConglomerateName() + "_col" + rc.getColumnPosition());
```

ascend1:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
```diff
@@ -81,6 +81,8 @@
     // getPredScopedForResultSet() method of this class for more.
     private boolean scoped;
 
+    private boolean matchIndexExpression;
```
yxia92, line 84: Can you reset the flag matchIndexExpression in clearScanFlags()?

yxia92 (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
ascend1: OK, I created a Jira ticket to track this issue. I need to investigate a bit more on how we can respect the sort order defined on index expressions.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
```diff
@@ -502,12 +523,27 @@ public boolean selfComparison(ColumnReference cr)
...
-        }else {
+        if(lcr != null && !lcr.isEmpty()){
+            int bound = forIndexExpression ? lcr.size() : 1;
+            for (int i = 0; i < bound; i++) {
+                if (lcr.get(i) == cr) {
+                    otherSide = rightOperand;
+                    break;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ascend1: You are right, behavior of this function is changed and it's not intended. It's fixed in next commit.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -183,7 +200,7 @@ public boolean useful(Optimizable optTable,ConglomerateDescriptor cd) throws Sta
             ** nor a useful stop key for this table, it is not useful
             ** for limiting an index scan.
             */
-            if((!isIn) && (!relop.usefulStartKey(optTable)) && (!relop.usefulStopKey(optTable))){
+            if((!isIn) && (!relop.usefulStartKey(optTable, id)) && (!relop.usefulStopKey(optTable, id))){
```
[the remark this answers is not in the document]
ascend1: Oh, that's a good catch! In next commit, I have moved this check down below, after `matchIndexExpression` is called.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -564,6 +612,58 @@ public static Integer isIndexUseful(Predicate pred,
...
+                }
+            } else if (relOp instanceof IsNullNode) {
+                IsNullNode isNull = (IsNullNode) relOp;
+                if (isNull.getOperand().equals(indexExprAst)) {
+                    isNull.getOperand().accept(cnv);
+                    // No need to set any matchIndexExpr flag since it won't be used in code generation.
+                }
+            }
+        }
+        return cnv.getList();
```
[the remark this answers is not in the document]
ascend1: Right, the list is not useful. It can be optimized a bit. I made it that way just to reuse the call site logic with minimal change. In next commit, I turn the return value to boolean and tuned the caller a bit to handle that.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -522,21 +549,42 @@ public static Integer isIndexUseful(Predicate pred,
...
-                                || (indexCol.getColumnNumber() != baseColumnPositions[indexPosition])
-                                || inNode.selfReference(indexCol))
-                            indexCol = null;
+        if (isIndexOnExpr) {
+            LanguageConnectionContext lcc = pred.getLanguageConnectionContext();
+            CompilerContext newCC = lcc.pushCompilerContext();
+            Parser p = newCC.getParser();
+
+            String[] exprTexts = indexDescriptor.getExprTexts();
+            for (indexPosition = 0; indexPosition < exprTexts.length; indexPosition++) {
```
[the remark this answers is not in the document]
ascend1: Yes, we definitely need to cache them. I have been thinking about it and created a Jira ticket to track things that could be improved: https://splicemachine.atlassian.net/browse/DB-10312.

I didn't do it in this PR because I was not sure about where to cache these ASTs. I was thinking that they would be required in many other places not touched in this change, and if I implement something back then, it probably needs to be changed later.

Now that I'm almost in the last major piece of this whole story, my idea is to cache the index expression ASTs in `ConglomerateDescriptor`. They will not be serialized or deserialized with other fields. Instead, we have always an empty `Valuenode` array when getting an instance either by constructing or deserialzing. Index expression will be parsed and stored in the array in the first time we use them, and later we just get from the array.

I think it's a good place because we set `ConglomerateDescriptor` in access path once and almost every following operations on this access path get this instance. It's global enough, yet much better scoped than the context.

I'd like to have your opinions on this before I start the implementation.

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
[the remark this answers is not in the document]
yxia92: Thanks @ascend1 ! It makes sense to me to cache it in the ConglomerateDescriptor. I couldn't think of any side effect of this approach, but it would be great to run through @dgomezferro and @jyuanca also and see if they see any downside about this approach.

yxia92 (approved):
+1

ascend1:

spliceci:
TEST SUCCEEDED +1

ascend1:

ascend1:

spliceci:
TEST SUCCEEDED +1

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
```diff
@@ -110,9 +110,11 @@ public ValueNode getLeftOperand() {
...
     }
     
-    public boolean allLeftOperandsColumnReferences() {
-        for (Object obj:leftOperandList) {
-            if (!(obj instanceof  ColumnReference))
+    public boolean allLeftOperandsContainColumnReferences() throws StandardException {
+        CollectNodesVisitor cnv = new CollectNodesVisitor(ColumnReference.class);
+        for (int i = 0; i < leftOperandList.size(); i++) {
+            leftOperandList.elementAt(i).accept(cnv);
+            if (cnv.getList().isEmpty())
```
yxia92, line 117: The variable declaration of `cnv` should be moved inside the for loop, otherwise, we won't be able to detect empty column references correctly for second element and upper.

yxia92 (changes requested):
See comment in BinaryListOperatorNode.java, thanks!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
ascend1: Ah, yes, this is buggy. I add `cnv.getList().clear()` in the loop. It should clear the vector and avoid constructing new object for each iteration.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -564,6 +613,53 @@ public static Integer isIndexUseful(Predicate pred,
...
+                if (isProbe) {
+                    assert relOp instanceof BinaryOperatorNode;
+                    BinaryOperatorNode binOp = (BinaryOperatorNode) relOp;
+                    binOp.setMatchIndexExpr(tableNumber, true);
+                }
+            }
+        } else {
+            if (relOp instanceof BinaryOperatorNode) {
+                BinaryOperatorNode binOp = (BinaryOperatorNode) relOp;
+                if (binOp.getLeftOperand().equals(indexExprAst)) {
```
msirek, line 633: The equals method will only match 2 different JavaToSQLValueNodes if they are the exact same object:
>     protected boolean isEquivalent(ValueNode o)
>     {
>         // anything in the java domain is not equiavlent.
>         return false;//this == o;
>     }

Therefore, if you have something like:
> create index t1_idx on t1(ln(a));
> explain select * from t1 --splice-properties index=t1_idx
> where ln(a) > 1;

It cannot take advantage of the index.  It may be better to write a new method for comparing expressions, that defaults to using the equals method for all nodes, but then for JavaToSQLValueNode we could override the implementation and do the proper expression comparison.  Note that non-deterministic functions should not ever be considered the same expression.

msirek (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
```diff
@@ -527,19 +565,34 @@ public boolean selfComparison(ColumnReference cr)
...
+          */
+         boolean isIndexOnExpr = id != null && id.isOnExpression();
+         int columnSide = isIndexOnExpr ? indexExprOnOneSide(optTable) : columnOnOneSide(optTable);
+
+         return columnSide != NEITHER && usefulStopKey(columnSide == LEFT);
+     }
+
     /**
      * Return true if a key column for the given table is found on the
      * left side of this operator, false if it is found on the right
```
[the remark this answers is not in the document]
ascend1: Yes, I encountered this problem and adapted this method in a later change: https://github.com/splicemachine/spliceengine/pull/4302/files#diff-1a44e74c70698678a5e81411dcd41091a5b75308c80a5717a30f842b31256dfc. But still, I think it's better to apply the fix here and I make a new test for a case like `a+b`.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
ascend1: Absolutely. I just found this yesterday when implementing `hashCode()` and it made my eyes open. There are also some other classes compare using `==`. I'm working on a fix.

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -553,10 +564,90 @@ public BaseExecutableIndexExpression getExecutableIndexExpression(int indexColum
...
+        LanguageConnectionContext lcc = context;
+        if (lcc == null) {
+            lcc = (LanguageConnectionContext) ContextService.getContext(LanguageConnectionContext.CONTEXT_ID);
+        }
+        assert lcc != null;
+
+        if (exprAsts != null) {
+            for (ValueNode ast : exprAsts) {
+                setTableNumberToIndexExpr(ast, optTable);
+                bindNecessaryNodesInIndexExpr(ast, optTable, lcc);
```
yxia92, line 589: @ascend1 Could you check if IndexDescriptorImpl is cached as part of the TableDescriptor in dictionary cache? If so, it is shared among all the queries and the these code could cause the data structure to be modified concurrently, and that could be problematic.

yxia92 (changes requested):

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
ascend1: @yxia92 I looked into the code again and I think `IndexDescriptorImpl` will be effectively cached because it's part of `ConglomerateDescriptor`, and `TableDescriptor` has a `conglomerateDescriptorList`. Thanks for pointing this out, it's indeed a serious problem.

For this change, I can remove the cache and let the method always parse. I mean, at this point, it doesn't seem to be a bad idea because it's correct and simple. Options to improve it could be:

1. Cache the ASTs in a statement-level cache so that different queries do not step on each other's feet. I'm not sure which objects' life cycle align with a statement and are always accessible during optimization. Maybe `CompilerContext`? It's managed in a stack, doesn't seem to be per statement. `LCC`? It's probably alive for the whole session?
2. Implement `clone()` method to `ValueNode` class hierarchy and always return a deep copy of ASTs. IMHO, code would be more readable and maintainable in this way. Also, `clone()` is needed (I encountered two places) by some logic and currently we cannot do that for index expressions. But then back to the beginning, this probably has only marginal performance improvement comparing to always parsing since in both way we need to construct the same amount of objects. Only the grammar matching of parsing is saved, which is probably not much anyway, since javacc doesn't do backtracking and index expressions are usually short.

ascend1:

ascend1:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1828/

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
yxia92: Thanks @ascend1 for checking! I'm good with moving this change out from the current PR. CompilerContext as I understand is per-query, and it should be like a global structure visible during the query compilation phase for a particular query.

yxia92 (approved):

msirek (changes requested):
Hi @ascend1 , the updates look good.  There are maybe a couple issues to look at though.
examples:

> drop table t1;
> create table t1 (a int, b int, c int, d int);
> create index t1_idx on t1(b+1, c+1, a+1);
> insert into t1 values (1,4,7,0);
> insert into t1 values (2,5,8,0);
> insert into t1 values (0,1,2,0);
> 
> explain select b from t1 --splice-properties index=t1_idx
> where a+1 in (2,3,4) and b+1 in (5,6,7) and c+1 in (8,9,10);
> ERROR XJ001: DERBY SQL error: SQLCODE: -1, SQLSTATE: XJ001, SQLERRMC: java.lang.IllegalStateExceptionASSERT FAILED getClone() not expected to be called for com.splicemachine.db.impl.sql.compile.BinaryArithmeticOperatorNodeXJ001.U
> 
> -- No IndexLookup should be done in the following test case because both a+1 and b (expressions from the query) are in the index:
> drop table t1;
> create table t1 (a int, b int, primary key(a));
> create index t1_idx on t1(a+1, b);
> explain select b from t1 --splice-properties index=t1_idx
> where a+1 > 1;
> Plan
> ----
> Cursor(n=4,rows=7,updateMode=READ_ONLY (1),engine=OLTP (default))
>   ->  ScrollInsensitive(n=4,totalCost=34.479,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
>     ->  ProjectRestrict(n=3,totalCost=30.413,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
>       ->  IndexLookup(n=1,totalCost=30.413,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
>         ->  IndexScan[T1_IDX(1905)](n=0,totalCost=4.013,scannedRows=7,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1,baseTable=T1(1888),preds=[((A[1:1] + 1) > 1)])
> Table statistics are missing or skipped for the following tables:
> MWS3.T1
> 
> -- Same comment for this case, we are selecting a constant here.  No need for IndexLookup:
> create table t1 (a int, b int, primary key(a));
> create index t1_idx on t1(ln(a));
> explain select 1 from t1 --splice-properties index=t1_idx
> where ln(a) > 1;
> Plan
> ----
> Cursor(n=4,rows=7,updateMode=READ_ONLY (1),engine=OLTP (default))
>   ->  ScrollInsensitive(n=4,totalCost=34.476,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
>     ->  ProjectRestrict(n=3,totalCost=30.41,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
>       ->  IndexLookup(n=1,totalCost=30.41,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
>         ->  IndexScan[T1_IDX(2001)](n=0,totalCost=4.01,scannedRows=7,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1,baseTable=T1(1984),preds=[(dataTypeServices: DOUBLE  > 1)])
> Table statistics are missing or skipped for the following tables:
> MWS3.T1
> 

Please let me know if these are fixed in one of the other PRs.

ascend1:
Thanks for the review @msirek ! These are indeed problems not covered in this PR.
> > drop table t1;
> > create table t1 (a int, b int, c int, d int);
> > create index t1_idx on t1(b+1, c+1, a+1);
> > insert into t1 values (1,4,7,0);
> > insert into t1 values (2,5,8,0);
> > insert into t1 values (0,1,2,0);
> > explain select b from t1 --splice-properties index=t1_idx
> > where a+1 in (2,3,4) and b+1 in (5,6,7) and c+1 in (8,9,10);
> > ERROR XJ001: DERBY SQL error: SQLCODE: -1, SQLSTATE: XJ001, SQLERRMC: java.lang.IllegalStateExceptionASSERT FAILED getClone() not expected to be called for com.splicemachine.db.impl.sql.compile.BinaryArithmeticOperatorNodeXJ001.U

This error happens because of the multiple-column in-list predicate built internally. This is DB-10421 and the fix is here: https://github.com/splicemachine/spliceengine/pull/4365. It requires some new utilities and refactoring introduced in DB-10236 and DB-10176, so it's on top of those changes.

> > No IndexLookup should be done in the following test case because both a+1 and b (expressions from the query) are in the index:
> > drop table t1;
> > create table t1 (a int, b int, primary key(a));
> > create index t1_idx on t1(a+1, b);
> > explain select b from t1 --splice-properties index=t1_idx
> > where a+1 > 1;
> > Plan
> > Cursor(n=4,rows=7,updateMode=READ_ONLY (1),engine=OLTP (default))
> > ->  ScrollInsensitive(n=4,totalCost=34.479,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
> > ->  ProjectRestrict(n=3,totalCost=30.413,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
> > ->  IndexLookup(n=1,totalCost=30.413,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1)
> > ->  IndexScan[T1_IDX(1905)](n=0,totalCost=4.013,scannedRows=7,outputRows=7,outputHeapSize=13 B,partitions=1,parallelTasks=1,baseTable=T1(1888),preds=[((A[1:1] + 1) > 1)])
> > Table statistics are missing or skipped for the following tables:
> > MWS3.T1

We have an `IndexLookup` in this case because in this PR (DB-10175), indexes on expressions are always treated as non-covering indexes. Implementation of covering indexes on expressions is in DB-10236: https://github.com/splicemachine/spliceengine/pull/4162. You reviewed that one, too. Running the example on that branch gives no `IndexLookup`. I thought separate the changes in this way could limit the scope of each change and make it easier for review. But I see that it also creates confusion for veterans. :)

> > Same comment for this case, we are selecting a constant here.  No need for IndexLookup:
> > create table t1 (a int, b int, primary key(a));
> > create index t1_idx on t1(ln(a));
> > explain select 1 from t1 --splice-properties index=t1_idx
> > where ln(a) > 1;
> > Plan
> > Cursor(n=4,rows=7,updateMode=READ_ONLY (1),engine=OLTP (default))
> > ->  ScrollInsensitive(n=4,totalCost=34.476,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
> > ->  ProjectRestrict(n=3,totalCost=30.41,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
> > ->  IndexLookup(n=1,totalCost=30.41,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1)
> > ->  IndexScan[T1_IDX(2001)](n=0,totalCost=4.01,scannedRows=7,outputRows=7,outputHeapSize=6 B,partitions=1,parallelTasks=1,baseTable=T1(1984),preds=[(dataTypeServices: DOUBLE  > 1)])
> > Table statistics are missing or skipped for the following tables:
> > MWS3.T1
> 
> Please let me know if these are fixed in one of the other PRs.

Similar to the case above, this one should contain no `IndexLookup` if running on DB-10236 branch.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1838/

msirek (approved):

PR #4115, opened 2020-09-12
DB-10245 set current row in CachedOperation for scenarios where rowco…(3.0)

yxia92:
…unt exceed the cache limit


carolp-503 (approved):
Approved for 3.0

PR #4119, opened 2020-09-14
DB-10196 Create new SPLICE schema for new database

arnaud-splice:


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -569,6 +576,24 @@ public void boot(boolean create,Properties startParams) throws StandardException
         booting=false;
     }
 
+    protected void createOrLoadDictionaryTables() {
+
+    }
```
ascend1, line 581: Do we need this empty method?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -603,9 +604,10 @@ public void boot(boolean create,Properties startParams) throws StandardException
                     databaseVersion.getMinorVersionNumber(),databaseVersion.getPatchVersionNumber(),
                     databaseVersion.getSprintVersionNumber());
         }
-        if(create){
+        if(create || Boolean.TRUE.equals(EmbedConnection.isCreate.get())) {
```
ascend1, line 610: Is it possible to reuse `isCurrentlyCreatingSecondaryDatabase` here?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
arnaud-splice: Removed. Thanks!

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
arnaud-splice: Unfortunately, no, because spliceDbDesc is not yet defined at this point.

ascend1: I see. Thanks!

ascend1 (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -444,14 +446,15 @@ public void boot(boolean create,Properties startParams) throws StandardException
 
             boolean nativeAuthenticationEnabled=PropertyUtil.nativeAuthenticationEnabled(startParams);
 
-            if(create){
-                String userName=IdUtil.getUserNameFromURLProps(startParams);
-                authorizationDatabaseOwner=IdUtil.getUserAuthorizationId(userName);
-                HashSet newlyCreatedRoutines=new HashSet();
+            HashSet newlyCreatedRoutines = null;
```
martinrupp, line 449: why have you changed the scope of this variable?

martinrupp (approved):

carolp-503 (approved):
+1

PR #4112, opened 2020-09-11
DB-10240 Rework test infra to accept DB parameter in connections 

arnaud-splice:

PR #4124, opened 2020-09-15
Revert "DB-10240 Rework test infra to accept DB parameter in connections "

hatyo:
Reverts splicemachine/spliceengine#4112


hatyo:
Change in 3.0 branch was failing, needs splice-ee change which requires CR.

PR #4129, opened 2020-09-15
DB-10110 Fix retrieving aliases of user-defined tables.

hatyo:


hatyo:

spliceci:
TEST SUCCEEDED +1

yxia92 (changes requested):
We may need an upgrade script to update all the rows in sys.sysaliases that are corresponding to  synonyms to NULL for this field.

hatyo:
> We may need an upgrade script to update all the rows in sys.sysaliases that are corresponding to synonyms to NULL for this field.

Thanks for the good observation, I added an upgrade script that checks whether `JAVACLASSNAME` is null and rewrites it to `"NULL"` string literal.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1692,4 +1693,50 @@ public void addMinRetentionPeriodColumn(TransactionController tc) throws Standar
...
+                        ScanController.NA,
+                        null,
+                        null,
+                        ScanController.NA);
+
+        while(sc.fetchNext(rowTemplate)){
+            /* If JAVACLASSNAME was set to null, rewrite it to "NULL" string literal instead. */
+            for (int i=0; i<rowTemplate.length; i++) {
+                if (i+1 == SYSALIASESRowFactory.SYSALIASES_JAVACLASSNAME)
+                    if(replaceRow[i] == null) {
```
yxia92, line 1786: We need to check `rowTemplate[I] == null` here which has the fetched row content.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1692,4 +1693,50 @@ public void addMinRetentionPeriodColumn(TransactionController tc) throws Standar
...
+    public void setJavaClassNameColumnInSysAliases(TransactionController tc) throws StandardException {
+        SchemaDescriptor sd = getSystemSchemaDescriptor();
+        TabInfoImpl ti=getNonCoreTI(SYSALIASES_CATALOG_NUM);
+        faultInTabInfo(ti);
+
+        FormatableBitSet columnToReadSet=new FormatableBitSet(SYSALIASESRowFactory.SYSALIASES_COLUMN_COUNT);
+        FormatableBitSet columnToUpdateSet=new FormatableBitSet(SYSALIASESRowFactory.SYSALIASES_COLUMN_COUNT);
+        for(int i=0;i<SYSALIASESRowFactory.SYSALIASES_COLUMN_COUNT;i++){
+            // we do not want to read the saved serialized plan
+            if (i+1 == SYSALIASESRowFactory.SYSALIASES_JAVACLASSNAME) {
```
yxia92, line 1758: From DB-9388, I learnt that for rows in dictionary table, partial row update does not seem to work well. To be safe, we should read all the columns of the row, and also provide the complete update row for update. That is, set all the bits for both `columnToReadSet` and `columnToUpdateSet`.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1692,4 +1693,50 @@ public void addMinRetentionPeriodColumn(TransactionController tc) throws Standar
...
+                        null,
+                        null,
+                        ScanController.NA);
+
+        while(sc.fetchNext(rowTemplate)){
+            /* If JAVACLASSNAME was set to null, rewrite it to "NULL" string literal instead. */
+            for (int i=0; i<rowTemplate.length; i++) {
+                if (i+1 == SYSALIASESRowFactory.SYSALIASES_JAVACLASSNAME)
+                    if(replaceRow[i] == null) {
+                        replaceRow[i] = new SQLLongvarchar("NULL");
```
yxia92, line 1787: It is better to populate all columns of the replaceRow.

yxia92: Can you run the upgrade test with some alias/synonym created before the upgrade, and see if the rows in sys aliases are properly updated and can be queried?

yxia92 (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: > Can you run the upgrade test with some alias/synonym created before the upgrade, and see if the rows in sys aliases are properly updated and can be queried?

I already did that before pushing the PR, works as expected.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSALIASESRowFactory.java
```diff
@@ -48,346 +48,338 @@
...
-
-		/* now build and return the descriptor */
-		return new AliasDescriptor(dd, aliasUUID, aliasName,
-										schemaUUID, javaClassName, cAliasType,
-										cNameSpace, systemAlias,
-										aliasInfo, specificName);
-	}
+public class SYSALIASESRowFactory extends CatalogRowFactory {
+    public static final String TABLENAME_STRING = "SYSALIASES";
+    public static final int SYSALIASES_ALIASINFO = 8;
```
jyuanca, line 72: Move this down with other column names for better readability

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -75,6 +75,7 @@ else if(v1==v2)
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1970), new UpgradeScriptForAddTablesAndViewsInSYSIBMADM(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1971), new UpgradeScriptToAddCatalogVersion(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1974), new UpgradeScriptToAddMinRetentionPeriodColumnToSYSTABLES(sdd, tc));
+        scripts.put(new Splice_DD_Version(sdd,3,1,0, 1975), new UpgradeScriptToSetJavaClassNameColumnInSYSALIASES(sdd, tc));
```
jyuanca, line 78: change build number to 1976

jyuanca (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSALIASESRowFactory.java
hatyo: done.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: Thanks for the information, I did not know that partial row updates are not working properly, we should investigate this further and fix it.
I will follow your proposed safe measure and mark all columns for read and update.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: Thanks!

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: > It is better to populate all columns of the replaceRow.

done.

dgomezferro (approved):

dgomezferro:
@hatyo there are conflicts

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1715,46 +1719,45 @@ public void addMinRetentionPeriodColumn(TransactionController tc) throws Standar
...
+                ScanController.NA,
+                null,
+                null,
+                ScanController.NA);
+
+        while (sc.fetchNext(rowTemplate)) {
+            for (int i = 0; i < rowTemplate.length; i++) {
+                replaceRow[i] = rowTemplate[i].cloneValue(false);
+                /* If JAVACLASSNAME was set to null, rewrite it to "NULL" string literal instead. */
+                if (i + 1 == SYSALIASESRowFactory.SYSALIASES_JAVACLASSNAME && rowTemplate[i].isNull()) {
```
yxia92, line 1756: Shouldn't we use the check "rowTemplate[I] == null"? I think we only have problem when this field is not populated at all, and thus null. Do we have a problem if rowTemplate[I] has a NULL value?

yxia92 (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: The problem we’re facing here is that we store null in a non-nullable column. The column type is defined as the following:
`SystemColumnImpl.getColumn("JAVACLASSNAME", java.sql.Types.LONGVARCHAR, false /*isNullable?*/, Integer.MAX_VALUE)`
Later on, we create a DVD and store a null into it:
`row.setColumn(SYSALIASES_JAVACLASSNAME, dvf.getLongvarcharDataValue(javaClassName)); // javaClassName is null`
I think this is causing the decoder to be confused leading to corrupted decoded data since we send `null` value in a non-nullable field.
To solve this problem, I made sure that we’re writing `"NULL"` literal leaving no chance for the field to be `null`.
I think the interesting part is we can provide a much simpler solution by simply making the column nullable! i.e. it is a one-liner change from:
`SystemColumnImpl.getColumn("JAVACLASSNAME", java.sql.Types.LONGVARCHAR, false /*isNullable?*/, Integer.MAX_VALUE)`
to
`SystemColumnImpl.getColumn("JAVACLASSNAME", java.sql.Types.LONGVARCHAR, true /*isNullable?*/, Integer.MAX_VALUE)`
This makes the decoder happy and everything works as expected (just tested it). Moreover, I don’t think we need an upgrade script for it.
However it means that we will have to change the `SYSALIASES` schema. Not sure if this is something we want to do, that’s why I lean towards the proposed solution in this PR.

yxia92: Thank you @hatyo for the explanation! I'm good with the current approach taken and will approve the PR.

yxia92 (approved):
+1

jyuanca (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -76,6 +76,7 @@ else if(v1==v2)
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1971), new UpgradeScriptToAddCatalogVersion(sdd,tc));
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1974), new UpgradeScriptToAddMinRetentionPeriodColumnToSYSTABLES(sdd, tc));
         scripts.put(new Splice_DD_Version(sdd,3,1,0, 1977), new UpgradeScriptToAddSysKeyColUseViewInSYSIBM(sdd, tc));
+        scripts.put(new Splice_DD_Version(sdd,3,1,0, 1978), new UpgradeScriptToSetJavaClassNameColumnInSYSALIASES(sdd, tc));
```
martinrupp, line 79: 1979

martinrupp (changes requested):

carolp-503 (approved):
+1

martinrupp:
i fixed conflicts and updated upgrad version to current 1979

martinrupp (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
martinrupp: fixed

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSALIASESRowFactory.java
```diff
@@ -51,343 +51,335 @@
...
+    //
+    //	METHODS
+    //
+    /////////////////////////////////////////////////////////////////////////////
+
+    public SYSALIASESRowFactory(UUIDFactory uuidf, ExecutionFactory ef, DataValueFactory dvf, DataDictionary dd) {
+        super(uuidf, ef, dvf, dd);
+        initInfo(SYSALIASES_COLUMN_COUNT, TABLENAME_STRING, indexColumnPositions, uniqueness, uuids);
+    }
+
```
martinrupp, line 139: this is a constructor, not a method...?

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSALIASESRowFactory.java
```diff
@@ -51,343 +51,335 @@
...
+    //
+    //	CONSTRUCTORS
+    //
+    /////////////////////////////////////////////////////////////////////////////
+    private static int[][] indexColumnPositions =
+            {
+                    {SYSALIASES_SCHEMAID, SYSALIASES_ALIAS, SYSALIASES_NAMESPACE},
+                    {SYSALIASES_ALIASID},
+                    {SYSALIASES_SCHEMAID, SYSALIASES_SPECIFIC_NAME},
+            };
```
martinrupp, line 127: not really a constructor, should be with the other variables.

PR #4131, opened 2020-09-15
DB-10110 Fix retrieving aliases of user-defined tables (2.8)

hatyo:


carolp-503 (approved):
Approved for 2.8

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSALIASESRowFactory.java
```diff
@@ -51,63 +51,68 @@
...
-	// null means all unique.
-    private	static	final	boolean[]	uniqueness = null;
-
-	private static int[][] indexColumnPositions =
-	{
-		{SYSALIASES_SCHEMAID, SYSALIASES_ALIAS, SYSALIASES_NAMESPACE},
-		{SYSALIASES_ALIASID},
-		{SYSALIASES_SCHEMAID, SYSALIASES_SPECIFIC_NAME},
-	};
-
```
martinrupp, line 102: i think you wanted to move this, but it got lost :-/
see DB-10110 on master
https://github.com/splicemachine/spliceengine/pull/4129/files?diff=split&w=1#diff-20f0560c4e3b16f5e0923ce041346429fa9d9084ee732fb35e9d5afe28778dbdR122-R127

hatyo: That's right, sorry about that and thanks for spotting it.

PR #4132, opened 2020-09-16
DB-10049 user-defined statement separator in sqlshell.

hatyo:


martinrupp (approved):

martinrupp:
verified we can do now things like
```
SET TERMINATOR '%';

CREATE TRIGGER mytrig2
AFTER UPDATE OF a,b
ON t1
REFERENCING OLD AS OLD NEW AS NEW
FOR EACH ROW
WHEN (EXISTS (SELECT 1 from t2 where t2.a = OLD.a))
BEGIN ATOMIC
    SELECT * FROM SYS.SYSTABLES;
    SIGNAL SQLSTATE '87101' SET MESSAGE_TEXT = 'a ';
END%
```

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
```diff
@@ -692,206 +667,185 @@ private void checkScrollableCursor(ResultSet rs, String operation)
...
+    }
+
+    public void setOmitHeader(boolean omitHeader) {
+        this.omitHeader = omitHeader;
+    }
+
+    public void setTerminator(String terminator) {
+        if (terminator.length() != 1 || !StatementFinder.isValidTerminator(terminator.charAt(0))) {
+            out.println(langUtil.getTextMessage("IJ_TerminatorInvalid", terminator));
+        }
```
martinrupp, line 847: `return` missing

martinrupp (changes requested):

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
martinrupp: e.g. this fails

```
set terminator ',';
show tables;
```

arnaud-splice (changes requested):
The following scenario doesn't seem to work:
- `set terminator 'a'; create schema foo a` -> Create schema does not see 'a' as a terminating character

Additionally
- `set terminator '~~'` raises an error but still silently sets the terminator to the first character of `~~`
We should either say that in the error message or not set it at all.

dgomezferro (commented):
Looks good to me except for the already raised issues.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/utilMain.java
hatyo: Thanks Martin, fixed.

hatyo:
@arnaud-splice:

> set terminator 'a'; create schema foo a -> Create schema does not see 'a' as a terminating character

Thanks! solving this problem (and other related problems) requires changing the way we pre-parse incoming stream of characters before passing it to respective parsers, for more information please have a look at my comment on respective JIRA items.

I pushed a new commit that disallows users from choosing an alphabet character as a statement separator since I don't think it is a common use case anyway.

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/StatementFinder.java
```diff
@@ -31,549 +31,567 @@
...
+        assert isValidTerminator(terminator);
+        this.terminator = terminator;
+    }
+
+    private static boolean isAsciiCharacter(char c) {
+        return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
+    }
+
+    public static boolean isValidTerminator(char terminator) {
+        return !isAsciiCharacter(terminator) && !DISALLOWED_TERMINATORS.contains(terminator);
```
martinrupp, line 591: (minor comment) also ; is an ASCII character... i think you mean Character.isLetter(terminator)

martinrupp (approved):

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/StatementFinder.java
arnaud-splice: +1

arnaud-splice (approved):
Good change! Please address Martin's comment and merge the latest master to resolve the build issue

carolp-503 (approved):
+1

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #4126, opened 2020-09-15
DB-10232 external CSV tables write correctly PARTITIONED BY

martinrupp:

PR #4133, opened 2020-09-16
DB-9770 Fix External Table Row Counts after cross-join insert

martinrupp:


martinrupp:
this is based on #4126 , actual commits for this PR are only the ones starting with DB-9770

martinrupp:
closing (including the speed improvement in DB-10267)

PR #4148, opened 2020-09-18
DB-9979 reduce the amount of data dictionary calls for row triggers under spark execution (2.8)

yxia92:


carolp-503 (approved):
Approved for 2.8

PR #4150, opened 2020-09-18
DB-9705 reduce the amount of self-inserted rows in DefaultSourceIT (3.0)

hatyo:
- Adding too many rows seems to cause memory pressure in Jenkins environment
  causing builds to fail sporadically.
- We should move this test to a performance benchmark where we have more
  suitable infrastructure.


carolp-503 (approved):

carolp-503 (approved):
Approved for 3.0

PR #4162, opened 2020-09-21
DB-10236 Take expression-based indexes as covering indexes properly

ascend1:


ascend1:

ascend1:

ascend1:

ascend1:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
```diff
@@ -1512,4 +1522,70 @@ public int getOuterJoinLevel() {
     public void setOuterJoinLevel(int level) {
         return;
     }
+
+    public ValueNode replaceIndexExpression(ResultColumnList childRCL) throws StandardException {
+        // by default, try replace this whole subtree
+        if (childRCL != null) {
+            for (ResultColumn childRC : childRCL) {
+                if (this.equals(childRC.getIndexExpression())) {
```
msirek, line 1530: It would be a good idea to use a different method than equals in all of the locations where we compare index expressions, like a wrapper function.  For now, perhaps the wrapper function would call equals, but later, when we want a+1 to be matched equivalent with 1+a, we can simply add the new logic to that method and all the required locations will see the new logic.

msirek (approved):
Approve, but please consider replacing use of the equals method with a new custom method in all expression-based index logic, either through this PR, or a separate PR.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -561,6 +566,48 @@ public boolean isCoveringIndex(ConglomerateDescriptor cd) throws StandardExcepti
         return coveringIndex;
     }
 
+    private boolean areAllReferencingExprsCoveredByIndex(IndexDescriptor id) throws StandardException {
+        if (referencingExpressions == null || referencingExpressions.isEmpty()) {
+            return false;
+        }
+        return getRefExprIndexPositions(id).size() == referencingExpressions.size();
```
msirek, line 573: This check is too simple to cover all combinations.  For example:

> create table t1 (a int, b int, primary key(a));
> create index t1_idx on t1(a+1, b);
> explain select count(*) from t1 --splice-properties index=t1_idx
> where a+1 > 1 and b+b > 1;
>       ->  GroupBy(n=4,totalCost=38.017,outputRows=1,outputHeapSize=0 B,partitions=1)
>         ->  ProjectRestrict(n=3,totalCost=30.413,outputRows=4,outputHeapSize=7 B,partitions=1)
>           ->  ProjectRestrict(n=2,totalCost=30.413,outputRows=4,outputHeapSize=7 B,partitions=1,preds=[((B[1:2] + B[1:2]) > 1)])
>             ->  IndexLookup(n=1,totalCost=30.413,outputRows=7,outputHeapSize=7 B,partitions=1)
>               ->  IndexScan[T1_IDX(7249)](n=0,totalCost=4.013,scannedRows=7,outputRows=7,outputHeapSize=7 B,partitions=1,baseTable=T1(7232),preds=[((A[1:1] + 1) > 1)])
> 

Even though column B is in the index, an expensive IndexLookup operation is done in order to evaluate b+b > 1.  Without expression-based indexes, the reference to b+b does not cause an IndexLookup.

I'm not sure how we should handle this, perhaps by not marking rc.isReferenced for column A in this case, or maybe detecting whether all column references of a given column only occur as matched index expressions.  But having to do IndexLookup is very expensive so we want to always avoid it if possible.

msirek (changes requested):
See previous comments.

yxia92 (changes requested):

yxia92:
In FromBaseTable.changeaccessPath(), for expression index serving as non-covering index for a query, is it true that we only rewrite the expression in predicate to index column, but not other expressions even if they appear in the expression index? if so, we may need a JIRA to track the rewrite for those also.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -495,4 +494,36 @@ public ValueNode genSQLJavaSQLTree() throws StandardException
 
 		return this;
 	}
+
+	@Override
+	public ValueNode replaceIndexExpression(ResultColumnList childRCL) throws StandardException {
+		if (childRCL == null) {
+			return this;
+		}
+		if (receiver != null) {
```
yxia92, line 503: There is no need to check `receiver`. It is either left operand or right operand and used at the code generation phase.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -495,4 +494,36 @@ public ValueNode genSQLJavaSQLTree() throws StandardException
...
+		}
+		return this;
+	}
+
+	@Override
+	public boolean collectExpressions(Map<Integer, List<ValueNode>> exprMap) {
+		boolean result = true;
+		if (receiver != null) {
+			result = receiver.collectExpressions(exprMap);
+		}
```
yxia92, line 520: Same here, no need to look at receiver.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
```diff
@@ -472,4 +473,24 @@ public int getOuterJoinLevel() {
...
+        return this;
+    }
+
+    @Override
+    public boolean collectExpressions(Map<Integer, List<ValueNode>> exprMap) {
+        boolean result = true;
+        for (Object leftItem : leftOperandList) {
+            result = result && ((ValueNode) leftItem).collectExpressions(exprMap);
+        }
+        return result;
```
yxia92, line 494: We need to check the rightOperandList also, as it could also contain expressions:
For example:
```
create index idx_t1 on t1(a1+1);
create table t2(a2 int, b2 int, c2 int);

splice> explain select a1+1 from t1
> , t2 where a2 in (1, a1+1);
ERROR XJ001: DERBY SQL error: SQLCODE: -1, SQLSTATE: XJ001, SQLERRMC: java.lang.IllegalStateExceptionASSERT FAILED sourceResultSetNumber expected to be >= 0 for null.A1XJ001.U
```

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
```diff
@@ -472,4 +473,24 @@ public int getOuterJoinLevel() {
...
+
+    public ValueNode replaceIndexExpression(ResultColumnList childRCL) throws StandardException {
+        ValueNodeList newList = (ValueNodeList) getNodeFactory().getNode(
+                C_NodeTypes.VALUE_NODE_LIST,
+                getContextManager());
+        for (Object leftItem : leftOperandList) {
+            newList.addValueNode(((ValueNode) leftItem).replaceIndexExpression(childRCL));
+        }
+        leftOperandList = newList;
+        return this;
```
yxia92, line 485: We need to check the rightOperandList also.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SubqueryNode.java
```diff
@@ -2702,4 +2699,11 @@ public int getCardinalityCheck() throws StandardException {
             return OnceResultSet.DO_CARDINALITY_CHECK;
         }
     }
+
+    @Override
+    public ValueNode replaceIndexExpression(ResultColumnList childRCL) throws StandardException {
+        assert resultSet instanceof Optimizable;
```
yxia92, line 2705: resultSet of a SubqueryNode could be a SelectNode, but when we get it, it is called from modifyAccessPath phase, so I guess SelectNode has already been replaced by a ProjectRestrictNode, so the assertion here is OK. Am I right? A comment may be helpful here.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -889,7 +891,8 @@ public ResultSetNode preprocess(int numTables,GroupByList gbl,FromList fl) throw
          * ProjectRestrictNode. If it is a FromBaseTable, then we will generate
          * the ProjectRestrictNode above it.
          */
-        fromList.preprocess(numTables,groupByList,whereClause);
+        fromList.preprocess(numTables,groupByList,whereClause,
+                exprMap == null ? collectExpression() : exprMap);
```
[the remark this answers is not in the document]
yxia92: In the presence of derived tables, we are only collecting the top-level expressions, as a result, we cannot qualify the index for the following query as a covering index:
```
create table t1 (a1 int, b1 int, c1 int, d1 int);
create index idx_t1 on t1(a1+1, b1+1);
create table t2(a2 int, b2 int, c2 int);

insert into t1 values (2,2,2,2), (3,3,3,3), (4,4,4,4);
insert into t2 values (1,1,1), (2,2,2), (3,3,3), (4,4,4), (5,5,5);

/* this query does not use covering index, however if we run the derived table by itself, we can qualify covering index */
select X from (select a1+1 as X, b2 from t1 --splice-properties index=idx_t1
, t2 where b1+1=b2) dt, t2
where X=3;
Plan
----
Cursor(n=14,rows=324,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=14,totalCost=2654.904,outputRows=324,outputHeapSize=1.016 KB,partitions=1)
    ->  NestedLoopJoin(n=11,totalCost=2514.624,outputRows=324,outputHeapSize=1.016 KB,partitions=1)
      ->  TableScan[T2(1776)](n=9,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=1.016 KB,partitions=1)
      ->  ProjectRestrict(n=8,totalCost=176.316,outputRows=16,outputHeapSize=52 B,partitions=1,preds=[(X[7:1] = 3)])
        ->  ProjectRestrict(n=7,totalCost=92.286,outputRows=16,outputHeapSize=52 B,partitions=1)
          ->  ProjectRestrict(n=6,totalCost=92.286,outputRows=16,outputHeapSize=52 B,partitions=1)
            ->  BroadcastJoin(n=5,totalCost=92.286,outputRows=16,outputHeapSize=52 B,partitions=1,preds=[(B1[5:3] = B2[5:4])])
              ->  TableScan[T2(1776)](n=3,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=52 B,partitions=1)
              ->  ProjectRestrict(n=2,totalCost=84.03,outputRows=20,outputHeapSize=40 B,partitions=1)
                ->  IndexLookup(n=1,totalCost=84.03,outputRows=20,outputHeapSize=40 B,partitions=1)
                  ->  IndexScan[IDX_T1(1761)](n=0,totalCost=4.03,scannedRows=20,outputRows=20,outputHeapSize=40 B,partitions=1,baseTable=T1(1744))

12 rows selected
ELAPSED TIME = 94 milliseconds
```

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
```diff
@@ -809,6 +809,17 @@ else if(!(childResult instanceof FromBaseTable)){
         }
         accessPathModified=true;
 
+        if (childResult.getResultColumns().isFromExprIndex()) {
+            /* We get a shallow copy of the ResultColumnList and its
+             * ResultColumns.  (Copy maintains ResultColumn.expression for now.)
+             */
+            resultColumns = childResult.getResultColumns().copyListAndObjects();
```
yxia92, line 816: I have a concern with the replacement of the `resultColumns` here. As all the column references that are on top of this ProjectRestrictNode that reference the result columns in `resultColumns` will be broken. I'm trying to use s derived table as a case to demonstrate the problem, but currently in the presence of derived table, index are always qualified as a non-covering index, so this logic is not triggered. We need to check back here once the derived table issue is fixed.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
ascend1: Yes, this is done in the latest changes in DB-10175 PR (https://github.com/splicemachine/spliceengine/pull/4110). Then I merged branch DB-10175 into this one and adapt the code. A few places need to be tuned because `JavaToSQLValueNode` is special in a sense that its sub-tree structure changes during binding. For this reason, we have to bind index expressions (partially, at least), too. It causes a bit trouble.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
ascend1: I agree that the check is simple at the moment. However, doing something similar on expressions is probably more difficult than on column references. In general, we can avoid IndexLookup in the following case:

> create index t1_idx on t1(a+1);
> select * from t1 where 2*a+2 > 3;  -- possible to utilize (a+1) column without IndexLookup

I think I can take care of the case you mentioned in the example, in which a column reference is an index expression. But for more general cases, I haven't got clear ideas on that.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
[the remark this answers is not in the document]
ascend1: Fix of this problem is in commit https://github.com/splicemachine/spliceengine/pull/4162/commits/edd0881cf1a975079c18a51632f64ac0eb77a36b. There, expressions are collected right before calling `resultSet.optimize(...)`. Also, a few test cases on derived tables including the two examples above are added.

ascend1:
> In FromBaseTable.changeaccessPath(), for expression index serving as non-covering index for a query, is it true that we only rewrite the expression in single table predicates to index columns, but not other expressions even if they appear in the expression index? if so, we may need a JIRA to track the rewrite for those also.

I think we have discussed this question in a meeting (Yi, Mark, and Zhen) but I'll put the answer here for everyone looking at this PR.

If an index on expressions is not covering, there is no expression rewritten to pseudo column reference because:
- For predicates that can be used to limit a scan (start/stop key or qualifier), column reference is not needed in code generation
- Correctness is kept for other predicates as well because we have an `IndexLookup` node. Other predicates are evaluated after that.

In the meeting, I learned from Yi that it's possible to use output columns from `IndexScan` (and other varieties) even there is an `IndexLookup` on top. I should revisit the way that `ResultColumn`s of `FromBaseTable` are built. Doing so could avoid `IndexLookup` in more cases.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2853,4 +2911,62 @@ public JBitSet collectInnerTablesFromFlattenedOJ() {
         }
         return collected;
     }
+
+    private Map<Integer, List<ValueNode>> collectExpression() {
+        HashMap<Integer, List<ValueNode>> result = new HashMap<>();
+        boolean proceed = true;
```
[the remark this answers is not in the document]
ascend1: You're right. I was thinking that in case we find an expression that references more than one table, we know immediately that index on expressions defined on these tables cannot be covering. So we can quite the process earlier by not setting the `referencingExprssions` to these base tables. But this logic is buggy (for example, the whole map is cleared if we found one such expression, wrong for multiple tables' case anyway) and error-prone. So I removed it and always collect all expressions.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryLogicalOperatorNode.java
```diff
@@ -228,4 +227,36 @@ DataTypeDescriptor resolveLogicalBinaryOperator(
...
+
+	@Override
+	public ValueNode replaceIndexExpression(ResultColumnList childRCL) throws StandardException {
+		if (childRCL == null) {
+			return this;
+		}
+		if (receiver != null) {
+			receiver = receiver.replaceIndexExpression(childRCL);
+		}
+		if (leftOperand != null) {
```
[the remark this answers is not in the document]
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryLogicalOperatorNode.java
```diff
@@ -228,4 +227,36 @@ DataTypeDescriptor resolveLogicalBinaryOperator(
...
+		}
+		return this;
+	}
+
+	@Override
+	public boolean collectExpressions(Map<Integer, List<ValueNode>> exprMap) {
+		boolean result = true;
+		if (receiver != null) {
+			result = receiver.collectExpressions(exprMap);
+		}
```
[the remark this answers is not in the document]
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
ascend1: Done. The example is added as a test case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -561,6 +566,48 @@ public boolean isCoveringIndex(ConglomerateDescriptor cd) throws StandardExcepti
         return coveringIndex;
     }
 
+    private boolean areAllReferencingExprsCoveredByIndex(IndexDescriptor id) throws StandardException {
+        if (referencingExpressions == null || referencingExpressions.isEmpty()) {
+            return false;
```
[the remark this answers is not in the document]
ascend1: Right, this is part of the buggy `proceed` problem discussed below. I thought when the expression collecting process ended early, we set `referencingExpressions` to null as an indicator. But this is just wrong and opposite to its natural meaning. If no expression references this table, then an index on expressions should be covering, too. I removed the `proceed` flag in the other place and fixed the check here.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: Can you open a Jira to take care of the general case?  I believe regular indexes can handle such expressions without doing an IndexLookup.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2853,4 +2911,62 @@ public JBitSet collectInnerTablesFromFlattenedOJ() {
...
+            }
+        }
+
+        if (havingAggregates != null) {
+            for (AggregateNode havingAggregate : havingAggregates) {
+                havingAggregate.collectExpressions(result);
+            }
+        }
+        // no need to check whereAggregates, should be empty
+
```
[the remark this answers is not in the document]
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -1253,6 +1270,28 @@ public ResultSetNode genProjectRestrict(int origFromListSize) throws StandardExc
...
+                    for (AggregateNode havingAggr : havingAggregates) {
+                        havingAggr.replaceIndexExpression(childResultColumns);
+                    }
+                }
+                if (selectAggregates != null) {
+                    for (AggregateNode selectAggr : selectAggregates) {
+                        selectAggr.replaceIndexExpression(childResultColumns);
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
ascend1: Yes, I'll open a Jira ticket to track this issue.

ascend1: Here it is: https://splicemachine.atlassian.net/browse/DB-10653

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OrderByList.java
```diff
@@ -616,4 +616,13 @@ boolean requiresDescending(ColumnReference cRef,int numOptimizables) throws Stan
...
         return false;
     }
+
+    public void replaceIndexExpressions(ResultSetNode child) throws StandardException {
+        ResultColumnList childRCL = child.getResultColumns();
+        for (int i = 0; i < size(); i++) {
+            OrderByColumn obc = (OrderByColumn) elementAt(i);
+            obc.setColumnExpression(obc.getColumnExpression().replaceIndexExpression(childRCL));
+        }
+        bindOrderByColumns(child);
```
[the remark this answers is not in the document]
ascend1: Hmm, I couldn't remember what was the problem but I'm pretty sure I encountered some failure in composing order by tests, then came up with this solution.

I just commented out the bind call and all tests pass now. I think it should be that expression replacement was buggy but I was on the wrong spot. Lots of fixes came later.

I think you're right that OrderByList is not more special than OrderByList. I remove the bind call.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SubqueryNode.java
ascend1: I think that's right. When we enter this call, it should come from the `replaceIndexExpression` in `genProjectRestrict`. At that point, access paths of subqueries, including derived tables, have been modified. The other place index expressions got replaced is in `FromBaseTable`. But it should never ends up here.

I add a comment as you suggested.

msirek (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
ascend1: After two days debugging with cases including derived tables and outer joins, I believe that breaking column mapping is unfortunately inevitable. Main reason is that if `childResult` comes from an index on expressions, the number of RCs may change. Example could be here we have two result columns for A and B, but the index expression is A + B and the `childResult` has only one result column. Or we could have only column A here, but actually we have multiple index expressions defined on A and referenced like A + 1 and A * 3.

Once we enter this code path, parent nodes must have their `resultColumns` rebuilt, too. Good news is that this propagation stops at top-level select of current query block. There we don't have to make new RC instances, but just replace their expressions.

The outer join failure you discovered is a combined one. Fixing it requires 1) expression collection is fixed, 2) result column rebuilding is fixed, and 3) `indexExpression` of RCs must be set and propagated properly. The third item is a newly found issue.

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ProjectRestrictNode.java
ascend1: Based on our discussion yesterday, I made two changes to the code:

1. not propagating `fromIndexExpr` flag across query block boundary
2. when rebuilding result columns of join node, it rebuilds only those from index expressions

With these two changes, all tests work with no need of `UpdateColumnReferenceVisitor`. Thus, I remove it and revert unnecessary changes in the latest commit.

ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2876,4 +2937,60 @@ public JBitSet collectInnerTablesFromFlattenedOJ() {
...
+            if (fromItem instanceof ResultSetNode) {  // this is probably always true?
+                ((ResultSetNode) fromItem).collectExpressions(result);
+            }
+        }
+
+        for (ResultColumn rc : resultColumns) {
+            rc.getExpression().collectSingleExpression(result);
+        }
+
+        return result;
```
yxia92, line 2992: It seems that for non-flattened correlated subquery, we are not collecting all the expressions. Here is an example:
```
create table t1 (a1 int, b1 int, c1 int, d1 int);
create index idx_t1 on t1 (a1+1, b1+2);
create table t2 (a2 int, b2 int, c2 int, d2 int);

explain select 1 from t1 --splice-properties index=idx_t1
where a1+1 not in (select a2 from t2 where b1+2 =b2);
```
In this case, I don't see the expression a1+1 being collected. This condition is in the PR node underneath SubqueryNode but above the SELECT node, so we may examine that also.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/JoinNode.java
```diff
@@ -378,6 +402,33 @@ public Optimizable modifyAccessPath(JBitSet outerTables) throws StandardExceptio
...
+            getNewResultColumns(resultColumns, newResultColumns, rightResultSet);
+            resultColumns = newResultColumns;
+            assert resultColumns.size() == leftResultSet.getResultColumns().size() + rightResultSet.getResultColumns().size();
+        }
+
+        // replace join predicates pushed down to right
+        if (leftResultSet.getResultColumns().isFromExprIndex()) {
+            rightResultSet.replaceIndexExpressions(leftResultSet.getResultColumns());
+        }
+
```
yxia92, line 431: I don't see a problem with the logic here, but the below case returns a wrong plan, maybe we are missing the replacement for post join conditions?
```
create table t1 (a1 int, b1 int, c1 int, d1 int);
create index idx_t1 on t1 (a1+1, b1+2);

explain select a1+1 from t2 left join t1 --splice-properties index=idx_t1
on a1+1=a2 where a1+1 is null;

Plan
----
Cursor(n=8,rows=20,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=8,totalCost=193.12,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=7,totalCost=184.72,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=6,totalCost=184.72,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1,preds=[is null((A1[5:2] + 1))])
        ->  MergeSortLeftOuterJoin(n=5,totalCost=184.72,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1,preds=[(A1[5:3] = A2[5:1])])
          ->  ProjectRestrict(n=4,totalCost=84.03,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1)
            ->  IndexLookup(n=3,totalCost=84.03,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1)
              ->  IndexScan[IDX_T1(1681)](n=2,totalCost=4.03,scannedRows=20,outputRows=20,outputHeapSize=160 B,partitions=1,parallelTasks=1,baseTable=T1(1664),preds=[(A1[5:3] = A2[5:1])])
          ->  TableScan[T2(1696)](n=0,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)
```
the condition a1=a2 is pushed to the the index scan step(n=2),

yxia92 (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -1283,6 +1322,9 @@ public ResultSetNode genProjectRestrict(int origFromListSize) throws StandardExc
         if(hasWindows()){
             // Now we add a window result set wrapped in a PRN on top of what we currently have.
             for (WindowNode windowDefinition : windowDefinitionList) {
+                if (childResultColumns.isFromExprIndex()) {
+                    windowDefinition.replaceIndexExpression(childResultColumns);
+                }
```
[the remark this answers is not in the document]
ascend1: Hmm, you're right, it seems windows and group by columns would refer to child of GBN, not the latest top result columns. Then I wonder why it has been working so far... I added two test cases explicitly test these cases, but they still work.

I then investigated `OrderByList` a bit further. It turns out that in code generation, its expression is not used at all. What matters is `columnPosition`, based on which the generated code knows the indexes of sort keys. This field is assigned in binding phase.

Then I suddenly remember why I bind `OrderByColumns` after replacing index expressions. It's because binding them would resolve the column reference in newly assigned expressions and update their `columnPosition`. `GroupByColumn` doesn't have this issue.

That also explains why replacing index expression late and having wrong column references also work. Because for `OrderByColumn`s, what matters is column references' column position, not their source or chain. We could have a GBN in between and GBN can change the number of result columns, but order by can only refer to the output columns of GBN and that's checked in binding phase. So as long as the order of result columns doesn't change, column position would be correct, despite that the column mapping is not right.

But anyway, I think it's good to make the column mapping right. So I replace `OrderByList` and `WindowDefinition`s together before constructing the `GroupBy` node. Also, I reverted the `replaceIndexExpression()` method for `OrderByList` to still bind it, and added a comment there.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
ascend1: Yes, I found this today when working on multi-column in-subquery, that `leftOperand` is not collected because it's never part of a `SelectNode`. It happens to all subqueries that have `leftOperand` and cannot be flattened, not necessarily correlated.

Without collecting it we may make wrong decisions about an index being a covering one, but actually have an expression that cannot be replaced. I enhanced this by collecting `SubqueryNode`, too. Then collect its `leftOperand` (if not null), and also `restrictionList` in the PRN directly under `SubqueryNode`. That PRN is generated on top of `SelectNode` to store the newly created predicate. I add a test for this case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/JoinNode.java
ascend1: `IndexLookup` is there and `a1+1` is not replaced, root cause seems to be that the join node is removed after preprocessing, but all column references still refer to the join table number, making all expressions belong to table number 1. For this reason, expressions on table number 0 are also sent to table 1, making it think the index is not covering.

Not sure how to fix this issue at the moment.

ascend1: Right, post join conditions are not replaced correctly. I fixed it in the latest commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1921/

spliceci:
TEST SUCCEEDED +1

yxia92 (approved):

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #4170, opened 2020-09-22
DB-9770/DB-10267 Fix External Table Row Counts after cross-join insert (3.0)

martinrupp:
consisting of changes DB-9770 (fix) and DB-10267 (speed improvement)


carolp-503 (approved):
Approved for 3.0

PR #4172, opened 2020-09-22
DB-10200 Filter SYS tables for each DB in SYSVW

arnaud-splice:


db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
```diff
@@ -265,46 +265,52 @@ public TupleDescriptor buildDescriptor(
                 new ColumnDescriptor("AUTHORIZATIONID", 3, 3,
                         DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.VARCHAR,  false,  128),
                         null, null, view, viewId, 0, 0, 0),
-                new ColumnDescriptor("DATABASEID", 4, 4,
-                        DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.CHAR, false, 36),
-                        null, null, view, viewId, 0, 0, 0)
+                //new ColumnDescriptor("DATABASEID", 4, 4,
+                //        DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.CHAR, false, 36),
+                //        null, null, view, viewId, 0, 0, 0)
```
ascend1, line 270: Why is this new column not returned?

ascend1 (commented):
It seems this PR has commits from DB-10327 (https://github.com/splicemachine/spliceengine/pull/4173), and I took a look only to the two DB-10200 commits.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
arnaud-splice: It's not decided yet, but I think that the schema views should not see the database they're in, since we will not be able to see other databases' schemas with `show schemas`

dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSSCHEMASRowFactory.java
ascend1: OK, that sounds reasonable.

ascend1 (approved):

carolp-503 (approved):
+1

PR #4187, opened 2020-09-24
DB-10339 fix ExternalTableIT for CSV (3.0)

martinrupp:
There seems to be a bug in Spark 2.8 when writing "#," with escape=#,
wrongly writes "#," instead of "##,". Spark 3.0 does it correctly.


carolp-503 (approved):
Approved for 3.0

PR #4195, opened 2020-09-25
DB-10259: fix import for varchar bit column

jyuanca:


jyuanca:

spliceci:
TEST SUCCEEDED +1

yxia92 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLBit.java
```diff
@@ -229,6 +230,30 @@ else if (sourceWidth > desiredWidth)
...
+				if (SanityManager.DEBUG)
+					SanityManager.THROWASSERT(
+							"value is null, stream or blob is not null");
+				return "";
+			}
+		}
+		else
+		{
+			String hex = com.splicemachine.db.iapi.util.StringUtil.toHexString(dataValue, 0, dataValue.length);
+			return String.format("X'%s'",hex);
```
yxia92, line 252: It seems that the only difference between the toString() function in SQLBit and that in SQLBinary is the call to String.format(), why don't we move this implementation up to SQLBinary instead? It seems to make sense to apply String.format() there too.

yxia92 (changes requested):
I have a question posted earlier.

dgomezferro (changes requested):
Marking as changes requested until @yxia92 's comment is resolved

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLBit.java
jyuanca: This impacts all SQLBinary's subclasses, include SQLBlob. X'whatever' works for char/varchar/ long char bit data, not for Blob

yxia92: Ok, thanks for the explanation!

yxia92 (approved):

dgomezferro (approved):

PR #4198, opened 2020-09-29
DB-9330 Add option for batched newTimestamp RPCs (3.0)

dgomezferro:
DB-7371 Port Timestamp Client/Server to Netty 4


carolp-503 (approved):
Approved for 3.0

PR #4204, opened 2020-09-30
DB-8896 Introduce DB2 compatible parameter to csv read "" as empty string 

arnaud-splice:


arnaud-splice:

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
```diff
@@ -775,14 +778,21 @@ public static  ConglomerateDescriptor getIndex(TableDescriptor td, String indexN
...
         // set up csv reader
         CsvPreference preference = createCsvPreference(columnDelimiter, characterDelimiter);
         Reader reader = new StringReader(splitKey);
         List<Integer> valueSizeHints = new ArrayList<>(execRow.nColumns());
         for(DataValueDescriptor dvd : execRow.getRowArray()) {
             valueSizeHints.add(dvd.estimateMemoryUsage());
         }
-        MutableCSVTokenizer tokenizer = new MutableCSVTokenizer(reader,preference, false,
+
+        boolean quotedEmptyIsNull = !PropertyUtil.getCachedDatabaseBoolean(
```
dgomezferro, line 793: I think our current behaviour is wrong (treating quoted empty strings as null). Shouldn't our default behaviour be the "DB2-compatible" one?

dgomezferro (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/storage/SpliceRegionAdmin.java
dgomezferro: @arnaud-splice brought me up to speed regarding yesterday's discussion.

dgomezferro (approved):

martinrupp (approved):

jyuanca (approved):

spliceci:
TEST SUCCEEDED +1

PR #4250, opened 2020-10-08
DB-10431 Fix row decoding in inserting into table with expr index

ascend1:


ascend1:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/actions/index/IndexIT.java
```diff
@@ -964,6 +964,21 @@ public void testCreateIndexAndInsertWithExpressionsNullInput() throws Exception
...
+        String tableName = "TEST_IDX_BATCH_INSERT";
+        methodWatcher.executeUpdate(format("create table %s (c char(4), i int, d double)", tableName));
+        methodWatcher.executeUpdate(format("insert into %s values ('abc', 10, 1.1), ('def', 20, 2.2)", tableName));
+
+        methodWatcher.executeUpdate(format("CREATE INDEX %s_IDX ON %s (upper(c), i + 2)", tableName, tableName));
+
+        // Without the fix, the following insert fails sporadically. It depends on the order of inserting rows. If
+        // (NULL, NULL, NULL) is inserted first, it fails. With the fix, it should never fail.
+        methodWatcher.executeUpdate(format("insert into %s values ('jkl', 30, 2.2), (NULL, NULL, NULL)", tableName));
+        methodWatcher.executeUpdate(format("update %s set c = 'def' where i = 10", tableName));
```
hatyo, line 979: could you also verify the results of these operations?

dgomezferro (approved):

yxia92 (changes requested):
Please address @hatyo 's comment to check the content of the index table after the update. Otherwise, it looks good to me.

msirek (approved):

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/actions/index/IndexIT.java
ascend1: Yes. Since we have only the index expr DDL/DML changes in master, we cannot query using the index yet. Nevertheless, I verify the result by issuing a select without the expr index to make sure insert and update ops modify data as expected.

ascend1:

ascend1:
> Please address @hatyo 's comment to check the content of the index table after the update. Otherwise, it looks good to me.

Yes, I pushed a new commit to add a result verify. Although we cannot query using the expression-based index yet, at least we can see that data are modified correctly.

yxia92 (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #4253, opened 2020-10-08
DB-9009 DB-10435 implement Foreign Key ON DELETE SET NULL

hatyo:


hatyo:
In this PR we add support to foreign key `ON DELETE SET NULL` clause which permits deleting rows from the parent by setting FK columns in referencing child rows to `null`.

Moreover, there was a bug that leads to violating FK constraints if certain situations where a parent table has multiple child tables each with own FK constraint. This bug is fixed, and now, if we attempt to delete a row from the parent, we make sure that no FK constraint is violated in each child table.

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyParentInterceptWriteHandler.java
```diff
@@ -80,135 +79,68 @@ private boolean isForeignKeyInterceptNecessary(KVPair.Type type) {
...
-                case SEEK:
-                    cellCount--; //the cell is filtered
-                    break;
-                case INCLUDE:
-                case INCLUDE_AND_NEXT_COL: //the cell is included
-                default:
-                    break;
+    @Override
+    public void flush(WriteContext ctx) throws IOException {
+        if(failed) {
```
dgomezferro, line 128: Shouldn't we flush the rows that were written before the failure?

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/Action.java
```diff
@@ -0,0 +1,44 @@
...
+        this.backingIndexConglomId = backingIndexConglomId;
+        this.failed = false;
+        this.writeResult = null;
+    }
+
+    public boolean hasFailed() {
+        return failed;
+    }
+
+    public WriteResult getWriteResult() {
```
dgomezferro, line 26: Should it be getFailedWriteResult() ?

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/OnDeleteSetNull.java
```diff
@@ -0,0 +1,73 @@
...
+        return new KVPair(rowId, value, KVPair.Type.UPDATE);
+    }
+
+    @Override
+    protected WriteResult handleExistingRow(byte[] indexRowId, byte[] sourceRowKey) throws Exception {
+        byte[] baseTableRowId = new byte[0];
+        try {
+            baseTableRowId = toChildBaseRowId(indexRowId, constraintInfo);
+        } catch (StandardException e) {
+            e.printStackTrace();
```
dgomezferro, line 63: Please handle properly

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyActionIT.java
```diff
@@ -0,0 +1,350 @@
...
+
+/**
+ * Foreign key tests for referential actions:
+ *
+ * ON DELETE NO ACTION
+ * ON DELETE CASCADE
+ * ON DELETE SET NULL
+ * ON UPDATE NO ACTION
+ */
+public class ForeignKeyActionIT {
```
dgomezferro, line 40: Can you test it with "permissive" imports? If we allow up to N badRecords and some records violate the FK constraint, we should log them but import the rest, does that work?

dgomezferro (changes requested):
Awesome! I have some small comments

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyParentInterceptWriteHandler.java
hatyo: Wouldn't that interfere with transaction atomicity in case of failure? e.g.
Imagine the parent table `p` has FKs defined on three child tables (`c1`, `c2`, and `c3`), the `DELETE` on the parent key comprises three rows `r1`, `r2`, and `r3`:
- for `r1`: accepted by all children => add mutations `r1c1`, `r1c2`, `r1c3` to respective buffers.
- for `r2`: accepted only by `c1` => generate mutation `r2c1` and add it to its buffer.
Now if we flush previous results we will end up with semi-updated child tables (`r1c1` and `r2c1` for `c1`, `r1c2` for `c2`, and `r1c3` for `c3`).
I am not sure about the internals of the write buffer so maybe in this situation the statement is rolled back properly at some other level.

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/Action.java
hatyo: Fixed, thanks!

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/OnDeleteSetNull.java
hatyo: Fixed, thanks!

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyActionIT.java
hatyo: I tried to run the following scenario via a unit test but it failed, I suspected it is not related to this PR so I ran it against master and ran into the same issue:

```
create table PI(col1 int, col2 varchar(2), primary key (col1));
create table C1I(col1 int, col2 int, constraint ci_fk_key foreign key(col1) references PI(col1) on delete no action);
insert into PI values (1, 'a');
insert into PI values (2, 'b');
-- create a file: /tmp/bad.csv containing the following rows:
-- 1,1
-- 1000,1000
-- 1,1
-- 2000,2000
-- 2,2
-- create a directory /tmp/BAD
call SYSCS_UTIL.import_data('SPLICE', 'C1I', null, '/tmp/bad.csv', ',', null, null, null, null, 2, '/tmp/BAD', null, null);
rowsImported        |failedRows          |files      |dataSize            |failedLog
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
3                   |2                   |1          |32                  |/tmp/BAD
select * from C1I;
COL1       |COL2
-----------------------
1          |1
1000       |1000
1          |1
2000       |2000
2          |2
```

The `SYSCS_UTIL.IMPORT_DATA` is not rejecting the rows that violate the FK constraint. I will open a JIRA for it so we can track it separately.

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyParentInterceptWriteHandler.java
dgomezferro: I was thinking of "permissive" writes, where we would want to write some rows and fail others. For a non-permissive write we would rollback the transaction and those written records shouldn't matter. But this is fine for now until we handle permissive writes for FK.

dgomezferro (approved):

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyParentInterceptWriteHandler.java
```diff
@@ -42,35 +65,261 @@
...
+            } else if (typeProvider.isScalar(fkConstraintInfo.getFormatIds(i))) {
+                position += multiFieldDecoder.skipLong();
+            } else {
+                position += multiFieldDecoder.skip();
+            }
+        }
+        int lastKeyIndex = position - 2;
+
+        if (lastKeyIndex == indexRowId.length - 1) {
+            return null; // special case.
```
ascend1: Maybe a bit more description about the special case?

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyActionIT.java
```diff
@@ -0,0 +1,339 @@
...
+            assertEquals(1L,s.executeUpdate("delete from P where a = 2"));
+        }
+    }
+
+    // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+    //
+    // FK - ON DELETE SET NULL
+    //
+    // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
```
ascend1, line 185: What will happen if the foreign key column is not nullable? Do we have such a test case?

`create table SRT (a int primary key, b int not null, constraint fk foreign key (B) references SRT(a) on delete set null);`

Probably should throw already in binding phase since it's semantically conflicting.

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyParentInterceptWriteHandler.java
hatyo: This is an outdated version of this method. Check out the new version of it in `OnDeleteAbstractAction`, we throw if we encounter this special code path.

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyActionIT.java
hatyo: You are correct, we get a semantic check error:
```
create table SRT (a int primary key, b int not null, constraint fk foreign key (B) references SRT(a) on delete set null);
ERROR 42834: SET NULL cannot be specified because FOREIGN KEY 'FK'  cannot contain null values.
```

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/FKWriteFactoryHolder.java
```diff
@@ -115,7 +117,8 @@ public void handleForeignKeyAdd(DDLChange ddlChange, long onConglomerateNumber)
         TentativeFK tentativeFKAdd = ddlChange.getTentativeFK();
         // We are configuring a write context on the PARENT base-table or unique-index.
         if (onConglomerateNumber == tentativeFKAdd.getReferencedConglomerateNumber()) {
-            addParentInterceptWriteFactory(tentativeFKAdd.getReferencedTableName(), ImmutableList.of(tentativeFKAdd.getReferencingConglomerateNumber()),ImmutableList.of(tentativeFKAdd.getFkConstraintInfo()));
+            addParentInterceptWriteFactory(tentativeFKAdd.getReferencedTableName(), ImmutableList.of(tentativeFKAdd.getReferencingConglomerateNumber()),
+                    ImmutableList.of(tentativeFKAdd.getFkConstraintInfo().getTable().getConglomerate()),ImmutableList.of(tentativeFKAdd.getFkConstraintInfo()));
```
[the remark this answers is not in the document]
hatyo: Thanks! I kind of extended the API without paying too much attention to cleaning it up a little, I removed that extraneous parameter.

ascend1 (approved):
+1

carolp-503 (approved):
+1

PR #4255, opened 2020-10-08
DB-10405: make sure to reset compaction state if a compaction fails

jyuanca:


dgomezferro (approved):
This is just making sure we call BackupEndpointObserver.postCompact(), is that right?

ascend1 (approved):

jyuanca:
@dgomezferro Yes, you are right!

carolp-503 (approved):
+1

PR #4258, opened 2020-10-09
DB-10359 allow JDBC driver to ignore unrecognized ClientInfo properties.

hatyo:
After checking `ClientInfo` properties, the driver looks up the environment variable `SPLICEMACHINE_JDBC_IGNORE_UNSUPPORTED_PROPERTIES` and checks whether it is set to `true`. If so, any unrecognized properties in `ClientInfo` will be ignored instead of throwing a `SQLClientInfoException` exception.

This allows JDBC programs that targets other database systems such as DB2, potentially with many properties set in `ClientInfo` that are not supported by SpliceMachine, to run without modification and recompilation against SpliceMachine.

### DIFF
--- a/db-client/src/main/java/com/splicemachine/db/client/net/NetConnection40.java
+++ b/db-client/src/main/java/com/splicemachine/db/client/net/NetConnection40.java
@@ -259,18 +259,19 @@ public class  NetConnection40 extends com.splicemachine.db.client.net.NetConnect
     }
 
     /**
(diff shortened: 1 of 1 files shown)


yxia92 (approved):
+1

arnaud-splice (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #4266, opened 2020-10-12
DB-10350 Special import mode that does not check referential integrity or fire triggers

martinrupp:
Added new import function which behaves like IMPORT_DATA, but is replacing the content of the table without firing triggers and without checking Forein Key constraints (referential integrity).

e.g.

`call SYSCS_UTIL.LOAD_REPLACE('<schemaName>', '<tableName>', null, '<inFilePath>', null, null, null, null, null, -1,      '<badRecordLogDirectory>', true, null);`

For this added special modes for INSERT and DELETE, also ignoring triggers / referential integrity:
`INSERT ... --splice-properties insertMode=LOAD_REPLACE `
`DELETE ... --splice-properties  noTriggerRI=1`


martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1735/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
```diff
@@ -703,12 +730,14 @@ public void generate(ActivationClassBuilder acb,
...
+
             if (colMap != null && colMap.length > 0) {
                 mb.push(acb.addItem(colMap));
             } else {
                 mb.push(-1);
             }
-            argCount += 3;
+            argCount += 4;
         }
         mb.callMethod(VMOpcode.INVOKEINTERFACE, (String) null, resultSetGetter, ClassName.ResultSet, argCount+3);
```
hatyo, line 742: maybe the last parameter should be `argCount+4` as well?

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
```diff
@@ -770,6 +828,9 @@ private static void doImport(String schemaName,
                 rs.open();
                 results[0] = new EmbedResultSet40((EmbedConnection) conn, rs, false, null, true);
             } catch (SQLException | StandardException | IOException e) {
+                if( isLoadReplaceMode ) {
+                    conn.rollback();
```
ipraznik-splice, line 832: makes sense to check, that conn != null, as SpliceAdmin.getDefaultConn can throw SQLException

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperation.java
```diff
@@ -38,6 +38,7 @@
 import com.splicemachine.derby.impl.sql.execute.actions.InsertConstantOperation;
 import com.splicemachine.derby.impl.sql.execute.sequence.SequenceKey;
 import com.splicemachine.derby.impl.sql.execute.sequence.SpliceSequence;
+import com.splicemachine.derby.stream.control.ControlDataSet;
```
ipraznik-splice, line 41: looks like not used

splice_machine/src/main/java/com/splicemachine/derby/stream/output/insert/InsertPipelineWriter.java
```diff
@@ -58,6 +57,7 @@
     protected InsertOperation insertOperation;
     protected boolean isUpsert;
     private Partition table;
+    private boolean load_replace_mode;
```
ipraznik-splice, line 60: field name is not in Java style

splice_machine/src/main/java/com/splicemachine/derby/stream/output/insert/InsertTableWriterBuilder.java
```diff
@@ -243,4 +247,16 @@ public static InsertTableWriterBuilder getInsertTableWriterBuilderFromBase64Stri
     public String getInsertTableWriterBuilderBase64String() throws IOException, StandardException {
         return Base64.encodeBase64String(SerializationUtils.serialize(this));
     }
+
+
+    @Override
+    public DataSetWriterBuilder loadReplaceMode(boolean load_replace_mode) {
+        this.loadReplaceMode = load_replace_mode;
```
ipraznik-splice, line 254: parameter load_replace_mode  name not in Java style

ipraznik-splice (changes requested):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/TableDescriptor.java
```diff
@@ -784,6 +784,18 @@ public int getQualifiedNumberOfIndexes(int minColCount,boolean nonUniqeTrumpsCol
         return matches;
     }
 
+    private static void assertValidStatementType(int statementType)
```
hatyo, line 787: nice refactoring.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
```diff
@@ -112,6 +114,32 @@ public void init(Object targetTableName,
...
+        return properties == null ? false :
+                properties.getProperty(BULK_DELETE_DIRECTORY) != null;
+    }
+
+    static public Properties getDeleteProperties(Properties tableProperties)
+    {
+        if (tableProperties == null) {
+            return null;
+        }
+        String dir = (String)tableProperties.remove(BULK_DELETE_DIRECTORY);
```
hatyo, line 128: Out of curiosity, why do we call `remove` here causing a side-effect? If that's intentional, should be maybe change the method name into something like `spliceDeleteProperties` ( no puns intended! :) )

hbase_pipeline/src/main/java/com/splicemachine/derby/hbase/SpliceIndexEndpoint.java
```diff
@@ -198,15 +199,26 @@ private boolean useToken(BulkWrites bulkWrites) {
...
     }
 
+    // todo: improve this
+    private boolean isLoadReplaceMode(BulkWrites bulkWrites) {
+        boolean loadReplaceMode = false;
+        Iterator<BulkWrite> iterator = bulkWrites.getBulkWrites().iterator();
+        if (iterator.hasNext()) {
+            BulkWrite bw = iterator.next();
+            loadReplaceMode = bw.isLoadReplaceMode();
+        }
```
hatyo, line 209: I guess you want to have something similar to reverse iterator, right? (not that it exists AFAIK, but maybe you can wrap this in a special method to make the intention clear.

pipeline_api/src/main/java/com/splicemachine/pipeline/contextfactory/LocalWriteContextFactory.java
```diff
@@ -231,7 +222,8 @@ private void addWriteHandlerFactories(int expectedWrites, PipelineWriteContext c
             ddlFactories.addFactories(context,true,expectedWrites);
 
             // FK - child intercept (of inserts/updates)
-            fkGroup.addFactories(context,false,expectedWrites);
+            if( foreignKeyChecks )
```
hatyo, line 225: please write a comment here explaining situations where `foreignKeyChecks` can be set to false.

hatyo (changes requested):
Nice work

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
martinrupp: good find! fixed

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/InsertOperation.java
martinrupp: done

splice_machine/src/main/java/com/splicemachine/derby/stream/output/insert/InsertPipelineWriter.java
martinrupp: done (also others in change)

splice_machine/src/main/java/com/splicemachine/derby/stream/output/insert/InsertTableWriterBuilder.java
martinrupp: done

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
martinrupp: no, the `argCount+=4` is specific for the `if ("getDeleteResultSet".equals(resultSetGetter))`.
We push here to mb the 4 parameters printExplainInformationForActivation, bulkDeleteDirectory, isNoTriggerRIMode and colMap.
The arg+3 is for the parameters `(String) null, resultSetGetter, ClassName.ResultSet`.
I mean to be honest, we should change how we do this. There's reflection and stuff that should make this rock solid.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
martinrupp: you are right. i have no idea, i just mimicked the original code

pipeline_api/src/main/java/com/splicemachine/pipeline/contextfactory/LocalWriteContextFactory.java
martinrupp: done

ipraznik-splice (approved):
Thank you!

hatyo (approved):

carolp-503 (approved):
+1

splice_machine/src/main/java/com/splicemachine/derby/stream/output/delete/DeletePipelineWriter.java
```diff
@@ -44,23 +44,27 @@
...
 
-    public void open(TriggerHandler triggerHandler, SpliceOperation operation) throws StandardException {
-        super.open(triggerHandler, operation);
+    public void open(TriggerHandler triggerHandler, SpliceOperation operation, boolean loadReplaceMode) throws StandardException {
+        super.open(triggerHandler, operation, loadReplaceMode);
         try{
             WriteConfiguration writeConfiguration = writeCoordinator.defaultWriteConfiguration();
             if(rollforward)
                 writeConfiguration = new RollforwardWriteConfiguration(writeConfiguration);
+            writeConfiguration.setLoadReplaceMode(loadReplaceMode);
```
dgomezferro, line 67: Isn't it possible you are changing here a shared writeConfiguration?

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
```diff
@@ -770,6 +828,10 @@ private static void doImport(String schemaName,
                 rs.open();
                 results[0] = new EmbedResultSet40((EmbedConnection) conn, rs, false, null, true);
             } catch (SQLException | StandardException | IOException e) {
+                if( conn != null && isLoadReplaceMode ) {
+                    // rolling back the DELETE we've done
+                    conn.rollback();
```
dgomezferro, line 833: If the parent connection has autocommit=true, does this work? In any case it would be better to use savepoints I think, rather than rolling back the whole user transaction.

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
```diff
@@ -670,18 +715,31 @@ private static void doImport(String schemaName,
...
                     + ", samplingOnly=" + samplingOnly + ", outputKeysOnly=" + outputKeysOnly + ", skipSampling=" + skipSampling
                     + (skipConflictDetection ? ", skipConflictDetection=true" : "") + (skipWAL ? ", skipWAL=true" : "")
                     + (indexName !=null ? (", index=" + indexName):"") + "\n" +
                     " SELECT "+ selectList +
                     " from " + vtiTable;
 
+            if(isLoadReplaceMode)
+            {
+                // delete table before inserting
+                try (PreparedStatement ips = conn.prepareStatement("DELETE FROM " + entityName + DeleteNode.NO_TRIGGER_RI_PROPERTY)) {
```
dgomezferro, line 736: Are we deleting all rows? Should we use truncate here instead?

pipeline_api/src/main/java/com/splicemachine/pipeline/contextfactory/LocalWriteContextFactory.java
```diff
@@ -216,11 +204,19 @@ private void isInitialized(TxnView txn) throws IOException, InterruptedException
...
+     *                         this is used e.g. with INSERT ... insertMode=LOAD_REPLACE or DELETE ... noTriggerRI=1
+     */
+    private void addWriteHandlerFactories(int expectedWrites, PipelineWriteContext context,
+                                          boolean foreignKeyChecks) throws IOException, InterruptedException {
         isInitialized(context.getTxn());
         //only add constraints and indices when we are in a RUNNING state
         if (state.get() == State.RUNNING) {
             //add Constraint checks before anything else
+
+            // ?
```
dgomezferro, line 219: missing comment?

dgomezferro (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
jyuanca: My understanding is that not all rows are to be deleted from the table. Should join the table with input file using PK as join key, and delete all rows from the table whose PK is in the join result.

jyuanca (changes requested):

martinrupp:
thanks @jyuanca , will revisit the change. i guess then it's more like a UPSERT.

also @dgomezferro hinted me i need to check the FK indices from child to parent tables.

setting to WIP for now.

martinrupp:
> thanks @jyuanca , will revisit the change. i guess then it's more like a UPSERT.
> 
> also @dgomezferro hinted me i need to check the FK indices from child to parent tables.
> 
> setting to WIP for now.

I talked with Gene about this, and he thinks the DELETE+INSERT approach is good for the customer.
I checked/debugged the code about the FK indices. They are used when values of the parent table are deleted, so i added special test scenarios for this. A "wrong" implementation would e.g. prevent to delete table A (parent) if the referencing table B has been deleted with noTriggerRI=1.
I also checked if we can, instead of `DELETE from <NAME> --noTriggerRI=1` use `TRUNCATE TABLE <NAME> --noTriggerRI=1` . I added code for this, but the problem is that TRUNCATE is not supposed to work when there's triggers or Foreign Keys - it will prompt the user this. When we disable this warning (since we want noTriggerRI=1) it works, however then the FK indices are not correctly updated, so that we than are not able to delete the parent table A. We can check this later, since this would be a performance improvement. Follow up jira is DB-10642 [TRUNCATE with ignoring RI/Triggers (TRUNCATE --noTriggerRI=1)] .

db-engine/src/main/java/com/splicemachine/db/iapi/sql/execute/ResultSetFactory.java
```diff
@@ -185,12 +185,13 @@ ResultSet getDeleteVTIResultSet(NoPutResultSet source, double optimizerEstimated
...
      *               contain one column which provides RowLocations that are
      *               valid in the target table.
+     * @param noTriggerRI if set to 1, DELETE will not fire triggers or check foreign key constraints
      * @return the delete operation as a result set.
      * @throws StandardException thrown when unable to perform the delete
      */
     ResultSet getDeleteResultSet(NoPutResultSet source, double optimizerEstimatedRowCount,
                                  double optimizerEstimatedCost, String tableVersion,
-                                 String explainPlan, String bulkDeleteDirectory, int colMapRefItem)
+                                 String explainPlan, String bulkDeleteDirectory, String noTriggerRI, int colMapRefItem)
```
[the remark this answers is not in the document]
martinrupp: done

splice_machine/src/main/java/com/splicemachine/derby/stream/output/delete/DeletePipelineWriter.java
martinrupp: you're right, i replaced with newDefaultWriteConfiguration, which returns a new object

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
martinrupp: ok, using savepoints now and added a test

splice_machine/src/main/java/com/splicemachine/derby/impl/load/HdfsImport.java
martinrupp: discussed with Gene, he thinks the DELETE+INSERT is the thing the customer wants. We can add an UPSERT mode later if needed.
TRUNCATE would be possible, but is more difficult to implement. As it is a performance improvement over this, i created DB-10642 .

dgomezferro (approved):

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1840/

jyuanca (approved):

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1843/

carolp-503 (approved):
+1

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1846/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1847/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1854/

PR #4267, opened 2020-10-12
DB-9009 DB-10435 implement Foreign Key ON DELETE SET NULL (3.0)

hatyo:


carolp-503 (approved):
Approved for 3.0 - thanks for the more-extensive unit tests!

PR #4274, opened 2020-10-13
DB-10460 fix default value for bit data type in show_create_table sta…

yxia92:
…tement

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
@@ -2387,6 +2387,10 @@ public class SpliceAdmin extends BaseAdminProcedures{
                         ) {
                     if ((defaultText = defaultText.toUpperCase()).startsWith("'"))
                         defaultText = "'" + defaultText + "'";
+                    if (columnDescriptor.getType().getTypeId().isBitTypeId() &&
+                            defaultText.startsWith("X'")) {
+                        defaultText = "X'" + defaultText.substring(1) + "'";
+                    }
                 }
             }
             colDef.append(defaultText);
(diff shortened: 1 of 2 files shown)


yxia92:

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

dgomezferro (approved):

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2387,6 +2387,10 @@ private static String createColumn(ColumnDescriptor columnDescriptor) throws SQL
                         ) {
                     if ((defaultText = defaultText.toUpperCase()).startsWith("'"))
                         defaultText = "'" + defaultText + "'";
+                    if (columnDescriptor.getType().getTypeId().isBitTypeId() &&
+                            defaultText.startsWith("X'")) {
+                        defaultText = "X'" + defaultText.substring(1) + "'";
+                    }
```
martinrupp, line 2393: this is necessary since we are putting that generated DDL string into this

```
StringBuilder sb = new StringBuilder("SELECT * FROM (VALUES '");
sb.append(DDL);
String extStr = extTblString.toString();
if (extStr.length() > 0)
	sb.append(extStr);
sb.append(";') FOO (DDL)");
resultSet[0] = executeStatement(sb);
```
and therefore all quotes like ' have to be escaped, resulting in double quotes ‘'. For X’000' this means X''000''. 

This scheme of creating a `”SELECT * FROM (VALUES ‘“ + result + “;) FOO(col1, col2)’ is used in all of SpliceAdmin.java, might be up for some more surprises.

Actually maybe simple would have been to solve escaping generally here by using sth like

```
String escapeAsValue(String val) {
  return val.replaceAll("'", "''"); 
}
```

and then use everywhere sth  like `”SELECT * FROM (VALUES ‘“ + doEscaping(result) + “;) FOO(col1, col2)”;`

martinrupp (approved):

carolp-503 (approved):
+1

jyuanca (approved):

PR #4281, opened 2020-10-13
DB-9010 implement FK ON DELETE CASCADE

hatyo:
In this PR we provide support to foreign key `ON DELETE CASCADE`. The write pipeline is extended in a way that allows executing recursive cascading deletes, self-referencing foreign keys are also covered and support recursive cascading deletes.


splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/OnDeleteAbstractAction.java
```diff
@@ -0,0 +1,229 @@
...
+            } else if (fkConstraintInfo.getFormatIds(i) == StoredFormatIds.SQL_REAL_ID) {
+                position += multiFieldDecoder.skipFloat();
+            } else if (typeProvider.isScalar(fkConstraintInfo.getFormatIds(i))) {
+                position += multiFieldDecoder.skipLong();
+            } else {
+                position += multiFieldDecoder.skip();
+            }
+        }
+        int lastKeyIndex = position - 2;
+
```
jyuanca, line 170: Can you comment the magic number 2?

jyuanca (changes requested):

pipeline_api/src/main/java/com/splicemachine/pipeline/config/SharedWriteConfiguration.java
```diff
@@ -105,7 +105,7 @@ public WriteResponse partialFailure(BulkWriteResult result, BulkWrite request) t
                         context.failed(main, mutationResult);
                     }
                 }
-                return WriteResponse.IGNORE;
+                return WriteResponse.THROW_ERROR;
             }
```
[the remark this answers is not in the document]
hatyo: Changing the `WriteResponse` to `THROW_ERROR` will cause the caller is throw an exception, in `BuldWriteAction`:
```
WriteResponse writeResponse=writeConfiguration.partialFailure(bulkWriteResult,currentBulkWrite);
                        switch(writeResponse){
                            case THROW_ERROR:
                                partialThrownErrorRows.add(currentBulkWrite.getSize());
                                thrown=true;
                                throw parseIntoException(bulkWriteResult);
```
The thrown exception informs the caller of the `BulkWriteAction` (which in our case, could be yet another `WriteHandler`) that something went wrong so we also cancel persisting the mutations during the call to `flush`.

Without changing this, I was running into the following problem that is best illustrated with an example; we have three tables `parent`, `child`, and `grandChild` related to each other with FK relationships (notice that `child`'s FK has `on delete cascade` while `grandChild`'s FK has `on delete no action`):
```
create table parent(col1 int, col2 varchar(2), col3 int, col4 int, primary key (col2, col4));
insert into parent values (1, 'a', 1, 1);
insert into parent values (2, 'b', 2, 2);
insert into parent values (3, 'c', 3, 3);
create table child(col1 int primary key, col2 varchar(2), col3 int, col4 timestamp, constraint fkey_1 foreign key(col2, col3) references parent(col2, col4) on delete cascade);
insert into child values (1, 'b', 2, '2020-09-09 10:10:10.123456');
insert into child values (400, 'a', 1, '2019-09-09 10:10:10.123456');
insert into child values (2, 'b', 2, '2018-09-09 10:10:10.123456');
create table grandChild(col1 int references child(col1) on delete no action, col2 varchar(2));
insert into grandChild values (1, 'z');
insert into grandChild values (1, 'z');
insert into grandChild values (1, 'z');
insert into grandChild values (400, 'y');
```
If we attempt to delete from parent:
```
delete from parent where col1 = 2;
```
the deletion will remove the row `(2, 'b', 2, 2)` from `parent` and also cause these rows to be deleted from `child`:
```
(1, 'b', 2, '2020-09-09 10:10:10.123456')
(2, 'b', 2, '2018-09-09 10:10:10.123456')
```
However the deletion of `(1, 'b', 2, '2020-09-09 10:10:10.123456')` will violate the `grandChild`'s FK since it is referenced by the following rows:
```
(1, 'z')
(1, 'z')
(1, 'z')
```
With `WriteResponse.IGNORE;`, the failure triggered by the child foreign key interceptor is not send to the parent foreign key interceptor, this means the deletion on the `parent` works but fails on the `child`. As a result, we get the following inconsistent results:
```
splice> select * From parent;
COL1       |C&|COL3       |COL4
--------------------------------------
1          |a |1          |1
3          |c |3          |3
2 rows selected
ELAPSED TIME = 29 milliseconds
splice> select * from child;
COL1       |COL2|COL3       |COL4
----------------------------------------------------------
1          |b   |2          |2020-09-09 10:10:10.123456
2          |b   |2          |2018-09-09 10:10:10.123456
400        |a   |1          |2019-09-09 10:10:10.123456
3 rows selected
ELAPSED TIME = 14 milliseconds
splice> select * from grandchild;
COL1       |COL2
----------------
1          |z
1          |z
1          |z
400        |y
```
With `WriteResponse.THROW_ERROR` the parent foreign key interceptor gets the notification and also cancels the deletion causing the overall operation to rollback maintaining referential integrity of the database.

I honestly still do not know the *full* scope of changing the `WriteRespose` like that, but it looks to be more correct than the previous behavior (to simply ignore partial failures), I think also @dgomezferro agrees on this. I am hoping we have enough test coverage that can unravel any potential issue introduced with this.

pipeline_api/src/main/java/com/splicemachine/pipeline/context/PipelineWriteContext.java
```diff
@@ -124,14 +124,21 @@ public void result(KVPair put, WriteResult result) {
...
             if (Arrays.equals(currentRowKey, resultRowKey)) {
                 resultsMap.put(kvPair, result);
-                return;
+                added = true;
+                break;
+            }
+        }
+        if (!added) {
+            for (KVPair kvPair : resultsMap.keySet()) {
+                byte[] currentRowKey = kvPair.getRowKey();
```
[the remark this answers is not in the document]
hatyo: Fixed it, thanks!

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/actions/OnDeleteAbstractAction.java
hatyo: I simplified it a bit and added more documentation.

pipeline_api/src/main/java/com/splicemachine/pipeline/context/PipelineWriteContext.java
```diff
@@ -124,14 +124,21 @@ public void result(KVPair put, WriteResult result) {
 
     @Override
     public void result(byte[] resultRowKey, WriteResult result) {
```
dgomezferro, line 126: I don't understand this change, if the rowKey is found it's updated with the result, otherwise all rows are updated with the result? What's the rationale behind it?

hatyo: I think previously it wasn't expected to reach this code path with a `resultRowKey` that is not part of the `result`.
Now it is possible when a failure from another inner pipeline is propagated, I thought the best way to handle it is to write the same result (in failing cascade it would be failure) to all rows rolling it back.
It kind of makes sense for the particular case of `on delete cascade` that I demonstrated to Jun in another comment, but I am not this does not break other code paths, therefore I would appreciate more feedback on this if possible.

jyuanca (approved):

carolp-503 (approved):
Approved for master once Daniel's accepted the response to his comment.

pipeline_api/src/main/java/com/splicemachine/pipeline/context/PipelineWriteContext.java
dgomezferro: I think we need to map the lower level failure to the specific rows that failed at this level, and keep the same code we had here.

dgomezferro (changes requested):

pipeline_api/src/main/java/com/splicemachine/pipeline/context/PipelineWriteContext.java
hatyo: Thanks Daniel, will do.

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1719/

hatyo:
A couple of issues were solves in the last few commits:
1. usage of incorrect parent conglomerate ID which triggered failures in some primary key tests.
2. proper error propagation, only the rows which indirectly cause failures in other (remote) `PipelineWriteContext` are marked as failed. Important for permissive bulk inserts later.
3. fixed another bug in error propagation which caused non-deterministic failures due to storing multiple values in the pipeline result set sharing the same row id but belonging to different tables.

dgomezferro (approved):
Great job!

PR #4282, opened 2020-10-13
DB-9010 implement FK ON DELETE CASCADE (3.0)

hatyo:
In this PR we provide support to foreign key `ON DELETE CASCADE`. The write pipeline is extended in a way that allows executing recursive cascading deletes, self-referencing foreign keys are also covered and support recursive cascading deletes.


carolp-503 (approved):
Approved for 3.0 once the concerns @dgomezferro raised in the Master branch are resolved.

PR #4287, opened 2020-10-14
DB-10385 Add JAVA_HOME check to standalone sqlshell

bklo94:
Runs into an error with spark queries if JAVA_HOME is not set. We should check if it is correctly set before sqlshell starts in standalone.


hatyo (approved):

arnaud-splice (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #4290, opened 2020-10-14
DB-10487 VacuumIT needs to wait for possibly still open Transactions

martinrupp:


martinrupp:

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -105,46 +100,77 @@
             .around(spliceTableIWatcher)
             .around(spliceTableJWatcher);
 
+    @Before
+    public void waitForStaleTransactions() throws Exception
```
dgomezferro, line 104: Can you create a JIRA to remove this and reference it here?

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -105,46 +100,77 @@
...
+    {
+        // wait at max 60s for other transactions to finish
+        // we have defined SerialTest.class, so no other test should run in parallel
+        for(int i=0; i<60; i++) {
+            try (ResultSet rs = methodWatcher.executeQuery(
+                    "call SYSCS_UTIL.SYSCS_GET_ACTIVE_TRANSACTION_IDS()")) {
+                String actual = TestUtils.FormattedResult.ResultFactory.toString(rs);
+                if( actual.equals("") ) return;
+                Thread.sleep(1000);
+                System.out.println("VacuumIT: Waiting for active transactions:\n" + actual);
```
dgomezferro, line 114: Should this be a LOG instead of println?

dgomezferro (changes requested):

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
martinrupp: log would be only visible when logging is enabled for this class, i would rather use println here to have this prominently visible in the output.

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
martinrupp: https://splicemachine.atlassian.net/browse/DB-10490

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
dgomezferro: Logging at INFO level should be visible by default, but if it's too much of a hassle println is fine since this is only a test

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
dgomezferro: I meant referencing the JIRA in a comment as a TODO for instance

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -105,46 +100,77 @@
...
+    @Before
+    public void waitForStaleTransactions() throws Exception
+    {
+        // wait at max 60s for other transactions to finish
+        // we have defined SerialTest.class, so no other test should run in parallel
+        for(int i=0; i<60; i++) {
+            try (ResultSet rs = methodWatcher.executeQuery(
+                    "call SYSCS_UTIL.SYSCS_GET_ACTIVE_TRANSACTION_IDS()")) {
+                String actual = TestUtils.FormattedResult.ResultFactory.toString(rs);
+                if( actual.equals("") ) return;
```
hatyo, line 112: To check if the result is empty:
```
if(!rs.next()) {
  return;
}
```

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
```diff
@@ -105,46 +100,77 @@
...
+                Thread.sleep(1000);
+                System.out.println("VacuumIT: Waiting for active transactions:\n" + actual);
+            }
+        }
+        Assert.fail("VacuumIT can't run with active transactions still happening. This is a bug in other " +
```
hatyo, line 117: maybe we should just ignore the test instead of failing since `VacuumIT` itself didn't even start.

hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
martinrupp: done

PR #4302, opened 2020-10-15
DB-10176 Support expression-based indexes in cost estimation

ascend1:


ascend1:

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/MergeJoinSelectivityIT.java
```diff
@@ -51,6 +51,16 @@ public void innerJoin() throws Exception {
...
+
+        try(Statement s = methodWatcher.getOrCreateConnection().createStatement()){
+            rowContainsQuery(s,
+                    new int[]{1,3},
+                    "explain select upper(ts_10_spk.c2), upper(ts_5_spk.c2) from --splice-properties joinOrder=fixed\n " +
+                            "ts_10_spk --splice-properties index=ts_10_spk_expr_idx\n, " +
+                            "ts_5_spk --splice-properties index=ts_5_spk_expr_idx, joinStrategy=MERGE\n " +
+                            "where upper(ts_10_spk.c2) = upper(ts_5_spk.c2)",
+                    "rows=10","MergeJoin");
+        }
```
martinrupp, line 63: afaik you can just do SpliceUnitTest.rowContainsQuery(levels, query, "rows=10","MergeJoin"); which does the try/createStatement for you

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
```diff
@@ -277,29 +283,37 @@ private boolean mergeable(RowOrdering outerRowOrdering,
...
+                        innerColumnNumber = innerColumn.getColumnNumber();
+                    }
+
+                    Pair<Integer, Integer> outerColumnInfo = getOuterColumnInfo(bron, innerTable, innerColumn);
+                    if (outerColumnInfo != null) {
+                        outerTableNum = outerColumnInfo.getFirst();
+                        outerColNum = outerColumnInfo.getSecond();
+                    } else {
+                        continue;
+                    }
```
martinrupp, line 307: i would do a slightly reordering here:
```
if (outerColumnInfo == null) continuel

int outerTableNum = outerColumnInfo.getFirst();
int outerColNum = outerColumnInfo.getSecond();
```

that way, there's no doubt outerTableNum and outerColNum are initialized.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
```diff
@@ -218,33 +219,38 @@ private boolean mergeable(RowOrdering outerRowOrdering,
...
                      */
-                    int outerTableNum = outerColumn.getTableNumber();
-                    int outerColNum = outerColumn.getColumnNumber();
+                    Pair<Integer, Integer> outerColumnInfo = getOuterColumnInfo(bron, innerTable, null);
+                    if (outerColumnInfo != null) {
+                        outerTableNum = outerColumnInfo.getFirst();
+                        outerColNum = outerColumnInfo.getSecond();
+                    } else {
+                        continue;
+                    }
```
martinrupp, line 253: reorder
```
if (outerColumnInfo == null) continue;
int outerTableNum = outerColumnInfo.getFirst();
int outerColNum = outerColumnInfo.getSecond();
```
(now clear everything is initialized)

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
```diff
@@ -915,4 +915,10 @@ public boolean isConstantOrParameterTreeNode() {
         // for count(*), operand is null
         return operand != null && operand.isConstantOrParameterTreeNode();
     }
+
+    public double getBaseOperationCost() throws StandardException { return getOperandCost(); }
+
+    protected double getOperandCost() throws StandardException {
+        return operand == null ? 0.0f : operand.getBaseOperationCost();
```
martinrupp, line 922: (nitpic) 0.0f is float, you mean 0.0 (double)

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryDateTimestampOperatorNode.java
```diff
@@ -181,4 +183,12 @@ public void generateExpression(ExpressionClassBuilder acb,
         mb.cast(ClassName.DataValueDescriptor);
         mb.callMethod(VMOpcode.INVOKEINTERFACE,(String)null,methodName,getTypeCompiler().interfaceName(),1);
     } // end of generateExpression
+
+    @Override
+    public double getBaseOperationCost() throws StandardException {
+        double lowerCost = getOperandCost();
+        double localCost = SIMPLE_OP_COST * (operand == null ? 1.0 : 2.0);
```
martinrupp, line 190: you have consts for everything but not for 1.0 and 2.0, and i'm not sure what they represent. Why is the cost twice if there's an operand?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TimestampOperatorNode.java
```diff
@@ -162,4 +162,10 @@ public ValueNode getClone() throws StandardException {
         return newTS;
     }
 
+	@Override
+	public double getBaseOperationCost() throws StandardException {
+		double localCost = SIMPLE_OP_COST * 2.0;
```
martinrupp, line 167: why 2

martinrupp (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/HalfMergeSortJoinStrategy.java
```diff
@@ -144,29 +144,39 @@ public int maxCapacity(int userSpecifiedCapacity, int maxMemoryPerTable, double
     private boolean mergeable(IndexRowGenerator innerRowGenerator,
```
dgomezferro, line 144: This class is not used, it can be removed

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/HalfMergeSortSparkIT.java
```diff
@@ -156,6 +151,7 @@ public void testHalfSortMergeAfterSplit() throws Throwable {
             "(c1 int, c2 int, c3 int, c4 int)");
 
     protected static SpliceIndexWatcher aIndex = new SpliceIndexWatcher(A, CLASS_NAME, A_IDX,CLASS_NAME,"(c1 desc, c2 asc, c3 desc)");
```
dgomezferro, line 153: This test is ignored, and the class is unused, it can be removed.

dgomezferro:
Change looks good to me, but I'd rather have a review from @yxia92 or @msirek (or both!) since they know the optimizer much better

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DB2LengthOperatorNode.java
```diff
@@ -194,5 +194,17 @@ private int getConstantNodeLength() throws StandardException
...
         return -1;
-    }        
+    }
+
+    @Override
+    public double getBaseOperationCost() throws StandardException {
+        double cost = 0.0;
+        if (operand != null) {
+            if (getConstantLength() == -1) {
+                cost += SIMPLE_OP_COST * operand.getTypeServices().getNull().getLength();
```
msirek, line 204: Is the cost really multiplied by the length of the data type?  What if you have a really long string, is it really that expensive?

msirek (approved):

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/MergeJoinSelectivityIT.java
ascend1: That's right. Currently, `xJoinSelectivityIT` tests are derived from `BaseJoinSelectivityIT` but not from `SpliceUnitTest`. So we don't have the `rowContainsQuery()` here. The one we see in these tests are actually a method in `BaseJoinSelectivityIT` that does the same...

All these tests probably should be adapted to derive from `SpliceUnitTest`. But I guess I'll do this in the very end since they are just tests.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
ascend1: Done.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeJoinStrategy.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryDateTimestampOperatorNode.java
ascend1: It just indicates that if the function has an operand, probably it executes more instructions to finish. I didn't look into the code of `date` and `timestamp`. But the costing is also like a heuristic. No need to make the cost absolutely accurate because in the end, it's about comparing options, always relative.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TimestampOperatorNode.java
ascend1: Similar to the other place, no particular reason for having 2.0, but just to indicate that it should be at least more expensive than a simple addition, which has cost of `SIMPLE_OP_COST`. The operator may be much more expensive than two addition, but I'm worried that expression evaluation cost accumulates fast, don't want this part to dominate the cost.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/HalfMergeSortJoinStrategy.java
ascend1: Done.

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/HalfMergeSortSparkIT.java
ascend1: Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DB2LengthOperatorNode.java
ascend1: I have the same concern here actually...I took a look at `SQLChar.getLength()` and it seems to have chance to calculate it in O(n). But then the expression evaluation cost could accumulate fast, soon dominating the plan cost. In the latest commit, I add an upper bound on this, hopefully better than linear.

carolp-503 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/StoreCostController.java
```diff
@@ -120,9 +123,10 @@
      * @throws StandardException Standard exception policy.
      * @see RowUtil
      */
-    void getFetchFromFullKeyCost(BitSet validColumns,
+    void getFetchFromFullKeyCost(boolean forExprIndex,
```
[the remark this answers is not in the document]
ascend1: That's a good idea. I did it in the current way because when an index on expressions is not covering, we need both base table statistics and index on expressions statistics at the same time. That's why we have them together in `StoreCostController` and have this flag in methods. It's a quick step.

Ideally we should be able to have multiple `StoreCostController`s representing statistics on different conglomerates. A bit more refactoring is needed to achieve it, though. I think when we want to improve index statistics handling, this is a necessary prior step.

ascend1:

dgomezferro (approved):

dgomezferro:
Awesome work @ascend1 !

martinrupp (approved):

spliceci:
TEST SUCCEEDED +1

PR #4311, opened 2020-10-16
DB-10496 Fix ignored deletes after restore (3.0)

dgomezferro:


jyuanca (commented):
approve for branch 3.0

PR #4320, opened 2020-10-19
DB-10526: fix CachedOperation initialization

jyuanca:


jyuanca:

msirek (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/CachedOperation.java
```diff
@@ -110,6 +110,11 @@ public String prettyPrint(int indentLevel) {
         if (!isOpen)
             throw new IllegalStateException("Operation is not open");
         populateCache();
+        try {
+            source.init(SpliceOperationContext.newContext(activation));
```
yxia92, line 114: @jyuanca shouldn't we have called source.init() during the initialization of the CachedOperation? Why do we still need to init the source again? Thanks!

jyuanca: populateCache() closes source, so it needs to be re-initalized and open again.

yxia92 (approved):
+1

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/CachedOperation.java
yxia92: Thanks @jyuanca !

carolp-503 (approved):
+1

PR #4326, opened 2020-10-20
DB-10544 fix missing serialization items in LastIndexKeyOperation / ScanOperation (3.1)

martinrupp:
fixes a problems caused by DB-10328 -> DB-10253 -> REVERT DB-10328


carolp-503 (approved):
Approved for 3.1

PR #4350, opened 2020-10-22
DB-10562 Refactor SpliceTransactionResourceImpl.getMarshallTransaction 

arnaud-splice:


martinrupp (approved):

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
```diff
@@ -376,36 +367,45 @@ public static void preAlterStats(DDLMessage.DDLChange change, DataDictionary dd,
...
+        DDLMessage.DropSchema dropSchema = change.getDropSchema();
+        UUID schemaId = ProtoUtil.getDerbyUUID(dropSchema.getSchemaUUID());
+        try {
+            try (SpliceTransactionResourceImpl transactionResource = new SpliceTransactionResourceImpl()) {
+                TxnView txn = DDLUtils.getLazyTransaction(change.getTxnId());
+                transactionResource.marshallTransaction(txn);
+                SchemaDescriptor sd = dd.getSchemaDescriptor(schemaId, transactionResource.getLcc().getTransactionExecute());
+                if(sd==null) // Schema Descriptor transaction never committed
+                    return;
+                dm.invalidateFor(sd,DependencyManager.DROP_SCHEMA,transactionResource.getLcc());
```
hatyo, line 379: Looks like an additional logic besides refactoring right? Could you explain why this is necessary?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
```diff
@@ -707,18 +707,12 @@ public int getMaxBaseColumnPosition() {
     }
 
     private LanguageConnectionContext getLcc(DDLMessage.TentativeIndex tentativeIndex) throws StandardException {
-        boolean prepared = false;
-        SpliceTransactionResourceImpl transactionResource = null;
-        try {
-            TxnView txn = DDLUtils.getLazyTransaction(tentativeIndex.getTxnId());
-            transactionResource = new SpliceTransactionResourceImpl();
-            prepared = transactionResource.marshallTransaction(txn);
+        TxnView txn = DDLUtils.getLazyTransaction(tentativeIndex.getTxnId());
```
arnaud-splice, line 710: Returning the LCC here seems wrong. Rework that class a bit to not have to return a LCC whose contextmanager got closed

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
```diff
@@ -79,17 +80,17 @@ public SpliceTransactionResourceImpl(String url,Properties info) throws SQLExcep
...
-        return this.marshallTransaction(txn, propertyCache, null, null, null);
+    public void marshallTransaction(TxnView txn, ManagedCache<String, Optional<String>> propertyCache) throws StandardException, SQLException {
+        this.marshallTransaction(txn, propertyCache, null, null, null);
     }
 
-    public boolean marshallTransaction(TxnView txn, ManagedCache<String, Optional<String>> propertyCache,
+    public void marshallTransaction(TxnView txn, ManagedCache<String, Optional<String>> propertyCache,
                                        TransactionController reuseTC, String localUserName, Integer sessionNumber) throws StandardException, SQLException{
-        boolean updated = false;
+        assert !updated;
```
hatyo, line 93: the method sets the `updated` to true and is public, so it is possible that it is called twice from the outside (which was permissible in the previous version)

I think it is better to return immediately and maybe issue a log.WARN("unexpected") or even throw an exception.

I would also add more documentation about the intended usage of this method, and mark the class `SpliceTransactionResourceImpl` as thread-unsafe.

hatyo (changes requested):
It is really nice that you replaced `try ... catch` with `try-with-resource` where applicable. But I am afraid this PR is doing a bit more besides refactoring. Please have a look at my comments.
Excellent work!

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
arnaud-splice: Done

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
arnaud-splice: This is true. We found out with Daniel that the code in `preDropSchema` did not invalidate potential dependencies to that schema. In a pre multidb world, it probably does not matter, but we should do in `preDropSchema` what we do in `preDropTable` anyway

hatyo: OK, thanks for the clarification.

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
hatyo: Great

hatyo (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
jyuanca: lcc is only used in getExecutableIndexExpression(). You can refactor to marshall transactions there only when it is being used

jyuanca (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
arnaud-splice: There is a dedicated jira for this specific issue: DB-10485. Zhen is currently working on it, so I would prefer not modify this further.

jyuanca (approved):

PR #4353, opened 2020-10-22
DB-10534 fix VacuumIT / unclosed Transactions in Tests (2.8)

martinrupp:
This fixes the issues because of un-closed Transactions which are preventing VacuumIT from running correctly.
It also introduces a @AfterClass function for each tests that logs messages of the following form:
`WARNING: Test failed to close all transactions. This might be due to multiple tests running in parallel.`
in case it finds left open transactions.
However, this message might be also caused by multiple tests running in parallel (that's why it's not Assert.fail).

To check if these Warnings are actually issues (especially if you see multiple of these messages), you can turn `failOnError=true` in SpliceUnitTest.waitForStaleTransactions() and check the tests that reported issues one by one (forcing Serial mode).

Also, SYSCS_GET_OLDEST_ACTIVE_TRANSACTION will now return same oldest Transaction as VACUUM, that is, the oldest active transaction over all region servers

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/derby/utils/VacuumIT.java
@@ -112,18 +112,16 @@ public class VacuumIT extends SpliceUnitTest{
             .around(spliceTableIWatcher)
             .around(spliceTableJWatcher);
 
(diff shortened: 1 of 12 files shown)


carolp-503 (approved):
Approved for 2.8

PR #4363, opened 2020-10-23
DB-10566 fix backward compatibility GenericColumnDescriptor

martinrupp:


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericColumnDescriptor.java
```diff
@@ -230,9 +230,19 @@ public static String readNullableUTF(ObjectInput in) throws IOException {
      * @exception IOException thrown on error
      */
     public void writeExternal(ObjectOutput out) throws IOException {
-        writeNullableUTF(out, name);
-        writeNullableUTF(out, tableName);
-        writeNullableUTF(out, schemaName);
```
dgomezferro, line 235: You can keep the `writeNullableUTF` calls for tableName and schemaName, right?

dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericColumnDescriptor.java
martinrupp: as it's hard to reproduce i'd rather change as little as possible

jyuanca: +1 if you can test and validate

jyuanca:
+1 if you can test and validate

martinrupp:
> +1 if you can test and validate

i could reproduce the issue and validate that this fix fixes it.

jyuanca (approved):

carolp-503 (approved):
+1

jyuanca (approved):

PR #4163, opened 2020-09-21
DB-10084 HBase flush and compactions are aware of min retention period 

arnaud-splice:

PR #4364, opened 2020-10-23
DB-10558 Improve getTxnAt performance

hatyo:
This PR addresses TPC-C performance regressions reported recently, the root cause was [DB-10084](https://github.com/splicemachine/spliceengine/pull/4163) PR, in that PR we made our purge-during-flush logic play nicely with time travel's minimum retention period (MRP) such that if we want to flush we check whether the involved transaction is within MRP, if so, we skip the purging preserving as a result the validity of past transactions within MRP.

The new check leverages the `getTxnAt` RPC which we introduced a while ago with time travel. This method performs HBase scans in a very inefficient way potentially causing a full scan of Splice txn table. That is maybe OK if the system is not under heavy load.

The TPC-C benchmark is intentionally putting the system under a heavy transnational workload to [benchmark](http://www.tpc.org/tpcc/detail5.asp) its OLTP performance, as a result the Splice txn table could grow to a much larger size causing HBase scans on it to take considerable amount of time and exposing as a result the inefficiency in `getTxnAt` RPC implementation since now flushes are taking a considerable amount of time causing the overall performance to drop heavily.

The proposed fix includesm aking HBase scans needed by `getTxnAt` algorithm much more efficient by setting proper start/stop rows for the scan and return a single row instead of (unnecessary) many rows.

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
+++ b/hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
@@ -5,13 +5,17 @@ import com.splicemachine.si.impl.TxnUtils;
 import com.splicemachine.utils.Pair;
 import com.splicemachine.utils.Source;
 import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.filter.FilterBase;
+import org.apache.hadoop.hbase.filter.FilterList;
 import org.apache.hadoop.hbase.filter.PrefixFilter;
 import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.RegionScanner;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.NoSuchElementException;
 
@@ -40,11 +44,10 @@ public class HBaseTxnFinder implements TxnFinder {
                     return stateKv.getTimestamp();
                 }
             }
-            return -1L; // in some rare scenarios, we could have a series of ROLLBACKs.
+            return -1L; // should never happen.
         }
 
         protected Pair<Long, Long> decode(List<Cell> cells) {
-            if (cells.size() <= 0) return null;
             long ts = 0;
                 ts = getTimestampOfActiveState(cells);
                 if(ts == -1) {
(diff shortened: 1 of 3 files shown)


hatyo:

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
```diff
@@ -89,15 +92,81 @@ public void close() throws IOException {
         }
     }
 
+    private static class TxWithTimestampFilter extends FilterBase {
+
+        @Override
+        public boolean filterRowKey(Cell cell) {
```
dgomezferro, line 98: `filterRowKey` should only filter based on the rowkey, here you are looking at specific values. This is most likely called only once per row.

I'm guessing the current implementation wouldn't work on SPLICE_TXN, since we'd inspect the latest written cell (COMMITTED or ROLLEDBACK), see that it's not the ACTIVE state and filter it out, which would filter out the whole row.

dgomezferro (changes requested):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1761/

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
hatyo: Thanks Daniel, I fixed it, now we simply return `false` in `filterRowKey()` method and I provided an implementation for `filterCell` instead, which returns `INCLUDE` for `ACTIVE` state.

hatyo:
Reverted the commit which dispatches `getTxnAt` to the high-priority queue since it is not working as expected, Oleg tried it already and it didn't have the desired effect. Once we figure out why this is happening we will do it again, reverting for now since it is not the lead cause of the reported issue.

hatyo:

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/region/HBaseTxnFinder.java
dgomezferro: Looks good, I think you don't need to implement it at all, the implementation on the super class should be good too, but it's OK with me.

dgomezferro (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4378, opened 2020-10-25
DB-10527 support for the comparison between bit data type and regular…(3.0)

yxia92:
… string type


carolp-503 (approved):
Approved for 3.0

PR #4379, opened 2020-10-26
DB-10360 support operation on current_timestamp

yxia92:


yxia92:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1774/

yxia92:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1775/

yxia92:

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

ipraznik-splice (approved):

arnaud-splice (approved):

hatyo (approved):
It is amazing that you managed to provide this functionality without having to implement the `INTERVAL` data type, I think we might be find it also useful for defining the minimum retention period.
Great work!!

ascend1 (commented):
1. The requirement definition is blurry because it says only "DB2-like operations on current_timestamp". Not surprisingly, DB2 supports interval values here:
```
values current timestamp + interval('1 day 3 hours');
1                         
------------------------- 
2020-10-28 12:27:14.42119 
```
Since we don't have interval syntax, it's not easy for us to have a complete "DB2-like" support for this. Maybe we should check if the fix is sufficient for the customer. I left this comment to the Jira ticket, too.

2. Now we have HOURS, MINUTES, and SECONDS, the following statement becomes valid:
```
splice> select timestamp('2020-01-01 00:00:00') + 2 minutes;
ERROR 2202F: Invalid time span operation.
```
At the moment it throws invalid time span operation, which could be confusing.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -7040,7 +7078,8 @@ convertAddTimeIntervalToFunction(ValueNode base, TimeSpanNode timeSpanNode, bool
                 function = "ADD_YEARS";
```
arnaud-splice, line 7078: To echo Zhen's comment: We should probably support hours/minutes/seconds in this code path as well.
Alternatively, why not merge the two code paths? They seem redundant.

carolp-503 (approved):
+1

yxia92:
Thanks @hatyo @ascend1 and @arnaud-splice for the comment. In this JIRA, I only extended the support for current_timestamp, but not any operation on  general timestamp value/columns. So for the example that @ascend1 listed:
```
splice> select timestamp('2020-01-01 00:00:00') + 2 minutes;
ERROR 2202F: Invalid time span operation.
```
it is not supported.
This is because at the time we parse the expression, syntaxer does not know about the type of the expression. So our syntax rules only expect 3 categories:
1. regular +/- operation for any data type without the time span units; or
2. Current_timestamp node +/- value node followed by time span units, this will be converted to timestampadd() functions; or otherwise,
3. value/column +/- value node followed by time span units except for #2, the value/column will be assumed as a DateType. and for date type, the time span unit can only support day/month/year, but not minute/hour/second.
If needed, we can have a separate JIRA to define the scope of general support.

PR #4383, opened 2020-10-26
DB-10252 Query hint to avoid returning error when AS OF is past min retention period

hatyo:
In this PR we did two things:

1. Time travel queries fail if they target past data beyond the relation(s) minimum retention period.
2. A new query hint `unboundedTimeTravel` is added, it allows disabling minimum retention period (MRP) checks for time travel queries, using this hint could lead to returning incorrect time travel (since history of rows could be altered during flush and compaction), but in certain situations it could still be useful to have the possibility to bypass MRP checks.


jyuanca (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1797/

PR #4384, opened 2020-10-26
DB-10252 Query hint to avoid returning error when AS OF is past min retention period (3.0)

hatyo:


carolp-503 (approved):
Approved for 3.0

PR #4394, opened 2020-10-27
DB-10312 Improve index on expressions handling

ascend1:


ascend1:
Commits in this PR have been included in PR of DB-10175, DB-10236, and DB-10176. No need to have this PR anymore. Specifically, 7eb8bb886bf2da4105e12844ea365755893600d6 is not a correct solution. Need to think about alternatives.

PR #4396, opened 2020-10-27
DB-10514-2.8 Reject adding foreign key with delete actions via ALTER statement

hatyo:
We do not allow tables to be created with a foreign key that has a delete action (`on delete cascade` and `on delete set null`) since they these actions are not supported. However we still incorrectly allow the user to add such foreign keys via the `alter` statement.

In this PR we fix this behavior disallowing the user to add a foreign key with delete action via `alter` statement.


yxia92 (approved):
+1

arnaud-splice (approved):

carolp-503 (approved):
Approved for 2.8

PR #4410, opened 2020-10-28
DB-10587 Resolve timed out transactions properly

dgomezferro:


dgomezferro:

jyuanca (approved):

hbase_storage/src/main/java/com/splicemachine/si/impl/region/V2TxnDecoder.java
```diff
@@ -197,9 +200,15 @@ protected long toLong(Cell data){
         if (taskKv != null) {
             taskId = decodeTaskId(taskKv);
         }
-        return composeValue(destinationTables,level,txnId,beginTs,parentTxnId,hasAdditive,
-                isAdditive,commitTs,globalTs,state,kaTime,rollbackSubIds,taskId);
+        TxnMessage.Txn txn = composeValue(destinationTables, level, txnId, beginTs, parentTxnId, hasAdditive,
+                isAdditive, commitTs, globalTs, state, kaTime, rollbackSubIds, taskId);
+
+        boolean committedWithoutGlobalTS = txn.getInfo().getParentTxnid() > 0 && txn.getGlobalCommitTs() < 0;
```
msirek, line 206: This is labelled "committedWithoutGlobalTS", but we're not checking if the transaction state is COMMITTED.  Should we add this state check so we don't resolve active transactions?

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1801/

dgomezferro:

hbase_storage/src/main/java/com/splicemachine/si/impl/region/V2TxnDecoder.java
dgomezferro: Good catch @msirek ! It's even worse than that, because we could resolve RolledBack subtransactions and end up in a similar situation to what this was trying to solve. I also added a test. Thanks!

arnaud-splice (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

msirek (approved):

PR #4418, opened 2020-10-29
DB-9656 better error output for nonmatching parameters in CALL

martinrupp:
# Description
Adding better error output for when users try to call functions with wrong parameters.

# Motivation

Users can make various mistakes when calling a function, and the error message should help solving those.
Examples of current behavior:

## parameter count wrong (old behavior)
```
splice> call SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY( 'a', 'b', 'c');
ERROR 42Y03: 'SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY' is not recognized as a function or procedure.
```

This is misleading, as there IS a function, however the parameter count is wrong.

## wrong parameter type (old behavior)

```
splice> call SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY( 123, 123 );
ERROR 42846: Cannot convert types 'INTEGER' to 'VARCHAR'.
```

This is better than the behavior when having wrong parameter count. However, when calling a function with multiple parameters, it might not be clear which one of the integers is wrong.

In both cases, error output should list the parameters of the function that was found.

# How to test

New error output looks like this:

## parameter count wrong (NEW behavior)
```
splice> call SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY( 'a', 'b', 'c');
ERROR 42Y03: Could not resolve function or procedure call 'SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY':
- Parameter count is wrong (provided 3, but needs 2)
  for SYSCS_SET_GLOBAL_DATABASE_PROPERTY(IN "KEY" VARCHAR(128), IN "VALUE" VARCHAR(32672))
```

## wrong parameter type (NEW behavior)
```
splice> call SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY( 123, 123 );
ERROR 42Y03: Could not resolve function or procedure call 'SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY':
- ERROR for parameter 0: Cannot convert types 'INTEGER' to 'VARCHAR'.:
  for SYSCS_SET_GLOBAL_DATABASE_PROPERTY(IN "KEY" VARCHAR(128), IN "VALUE" VARCHAR(32672))
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/types/RoutineAliasInfo.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/types/RoutineAliasInfo.java
@@ -41,12 +41,14 @@ import java.io.ObjectOutput;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.util.IdUtil;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  * Describe a routine (procedure or function) alias.
  *
  * @see com.splicemachine.db.catalog.AliasInfo
  */
+@SuppressFBWarnings("EI_EXPOSE_REP")
 public class RoutineAliasInfo extends MethodAliasInfo
 {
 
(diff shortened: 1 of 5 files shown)


jyuanca (approved):

arnaud-splice (approved):

dgomezferro (approved):
This is great @martinrupp ! My only concern is that the parameter messages are hardcoded and wouldn't be translatable like other errors, is there any way to work around that?

martinrupp:
> This is great @martinrupp ! My only concern is that the parameter messages are hardcoded and wouldn't be translatable like other errors, is there any way to work around that?

yes, i moved them to messages.xml and used `MessageService.getCompleteMessage` to get the strings.

carolp-503 (approved):
+1

PR #4419, opened 2020-10-29
DB-10615 Support known keywords as tokens instead of string in EXPORT 

arnaud-splice:


spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportOperationIT.java
```diff
@@ -536,9 +544,9 @@ private String buildExportSQL(String selectQuery, String exportPath, String comp
         if (useNativeSyntax) {
             StringBuilder sql = new StringBuilder();
             sql.append("EXPORT TO '").append(exportPath).append("'");
-            sql.append(" AS 'csv' ");
```
hatyo, line 539: but wouldn't this remove the tests for exporter checking the syntax internally (not the parser)?

martinrupp (approved):
nice

hatyo (commented):
Nice work! It would be nice to have a couple of tests checking that the new syntax works (`EXPORT TO '/tmp/' AS CSV COMPRESSION SNAPPY`) as well as the old syntax (`EXPORT TO '/tmp/' AS 'CSV' COMPRESSION 'SNAPPY'`)

hatyo (approved):
great!

carolp-503 (approved):
+1

PR #4426, opened 2020-10-30
DB-10620 fix sporadically-failing SelectTimeTravelIT

hatyo:
The test is written incorrectly, it was using a lower-precision `TIMESTAMP` (from `SpliceDateTimeFormatter.defaultTimestampFormatString = "yyyy-MM-dd HH:mm:ss"`) causing time travel algorithm to return incorrect results when `INSERT` transactions share the same sub-second time.

To solve this problem, we now use the server time (`values current_timestamp`) with a high enough precision to probe the time travel results.


ascend1 (approved):
Awesome!

dgomezferro (approved):

bklo94 (approved):

jyuanca (approved):

PR #4433, opened 2020-11-02
DB-10570 support currentSchema in jdbc connecting string

yxia92:


yxia92:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
```diff
@@ -173,7 +172,9 @@
         drdaID = info.getProperty(Attribute.DRDAID_ATTR, null);
         rdbIntTkn = info.getProperty(Attribute.RDBINTTKN_ATTR, null);
         ipAddress = info.getProperty(Property.IP_ADDRESS, null);
-        defaultSchema = info.getProperty(Property.CONNECTION_SCHEMA, null);
+        defaultSchema = info.getProperty(Property.CONNECTION_CURRENT_SCHEMA, null);
+        if (defaultSchema == null)
+            defaultSchema = info.getProperty(Property.CONNECTION_SCHEMA, null);
```
ascend1, line 177: It seems we support both `schema` and `currentSchema` then and `currentSchema` has priority in a sense that if it's set, then `schema` is ignored. Is that intended?

dgomezferro (approved):

arnaud-splice (approved):

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
hatyo: e.g. what's the expected behavior of `schema=A;currentSchema=B`.

hatyo (changes requested):
LGTM, but I think we should address @ascend1's comment.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
yxia92: Yes, @ascend1 and @hatyo , the code above is intended, so for the scenario @hatyo raised, we have a deterministic behavior. Basically, currentSchema will take precedence over schema.

carolp-503 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
hatyo: Thanks!

hatyo (approved):

ascend1 (approved):

PR #4437, opened 2020-11-03
DB-10187 add base_name, base_schema to sysibm.systables

yxia92:


yxia92:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/catalog/ViewsInSysIbmIT.java
```diff
@@ -189,10 +189,11 @@ public void testColumnDefault() throws Exception {
...
         methodWatcher.executeUpdate(format("create synonym %s.S5 for t5", CLASS_NAME));
-        String sqlText = format("select name, creator, type, colcount, keycolumns, keyunique, codepage from sysibm.systables where CREATOR='%s' and name='S5'", CLASS_NAME);
-        String expected = "NAME |    CREATOR     |TYPE |COLCOUNT |KEYCOLUMNS | KEYUNIQUE |CODEPAGE |\n" +
-                "-------------------------------------------------------------------------\n" +
-                " S5  |VIEWSINSYSIBMIT |  A  |    0    |     0     |     0     |    0    |";
+        String sqlText = format("select * from sysibm.systables where CREATOR='%s' and name in ('S5', 'T5')", CLASS_NAME);
+        String expected = "NAME |    CREATOR     |TYPE |COLCOUNT |KEYCOLUMNS | KEYUNIQUE |CODEPAGE | BASE_NAME |  BASE_SCHEMA   |\n" +
+                "------------------------------------------------------------------------------------------------------\n" +
+                " S5  |VIEWSINSYSIBMIT |  A  |    0    |     0     |     0     |    0    |    T5     |VIEWSINSYSIBMIT |\n" +
+                " T5  |VIEWSINSYSIBMIT |  T  |   13    |     3     |     0     |  1208   |   NULL    |     NULL       |";
```
ascend1, line 196: Is it possible in splice for now that `BASE_NAME` and `BASE_SCHEMA` are not null?

ascend1 (approved):

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/catalog/ViewsInSysIbmIT.java
yxia92: Base_name and base_schema are only meaningful for alias/synonym that points to a table/view/alias, which refer to the base table/view/alias and the schema the referred object resides in. Splice currently does not have these two columns directly, but the relevant info are in the sys.sysaliases table in the column aliasinfo. For other database objects, these two fields are not applicable, so they should stay as NULL.

msirek (approved):

PR #4449, opened 2020-11-03
DB-9556 DB-9586 NSDS v1 Column Names

jpanko1:


jpanko1:

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkUtils.java
```diff
@@ -224,19 +220,37 @@ public ExecRow call(ExecRow row) throws Exception {
...
-            // Generate the schema based on the ResultColumnDescriptors
-            List<StructField> fields = new ArrayList<>();
+        
+        final ResultColumnDescriptor[] columns = serverSideResultSet.getResultDescription().getColumnInfo();
+        // Generate the schema based on the ResultColumnDescriptors
+        List<StructField> fields = new ArrayList<>();
+        for (ResultColumnDescriptor column : columns) {
+            fields.add(column.getStructField());
+        }
+        if( fields.stream().map( f -> f.name().toUpperCase() ).distinct().count() != fields.size() ) {  // has duplicate names
```
yxia92, line 230: Do we need toUpperCase() here? It does not seem consistent with the subsequent HashSet logic, which honors case sensitivity.

spliceci:
TEST SUCCEEDED +1

yxia92 (changes requested):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkUtils.java
jpanko1: I think I could leave that and change the HashSet uses to `used.contains(name.toUpperCase())` and `used.add(name.toUpperCase())`.  I would also leave the case of the name passed to StructField unchanged.

yxia92: Thanks @jpanko1 ! We could have case sensitive column names. For example:
```
splice> create table t1 ("a1" int, "A1" int); 
ELAPSED TIME = 19 milliseconds
splice> select * from t1;
a1         |A1         
-----------------------

0 rows selected
ELAPSED TIME = 28 milliseconds
```
After the binding phase, all the regular column names should have already been in the UPPER case except for the case sensitive column names. With that consideration, should we remove the toUpperCase() call?

jpanko1: I see, thanks Yi!  I've removed `toUpperCase()`.
I also checked about case sensitivity in Spark and found that it can be set either way with a param: `spark.sqlContext.setConf("spark.sql.caseSensitive", "true")`.

jpanko1:

yxia92 (approved):
+1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/SparkUtils.java
yxia92: Thanks @jpanko1 for checking!

jyuanca (approved):

jyuanca:
Code change looks good to me. One question - why is it possible to get duplicate column names?

jpanko1:
> Code change looks good to me. One question - why is it possible to get duplicate column names?

In the event a query joins tables that happen to have the same column name, especially in a `select *` the user may not realize the duplicate column names will be returned.

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #4458, opened 2020-11-04
DB-9843 External Tables: Own directory partition parsing (3.0)

martinrupp:
fixing issues:
- DB-9843: Don't copy files for schema infer, but also don't check all files for schema -> overall faster
- DB-10568 Partitioned by using VARCHAR but value is 1 will be infered as INT and can't be read


carolp-503 (approved):
Approved for 3.0

PR #4473, opened 2020-11-04
redo-DB-10350 Special import mode that does not check referential integrity or fire triggers (3.0)

martinrupp:
Added new import function which behaves like IMPORT_DATA, but is replacing the content of the table without firing triggers and without checking Forein Key constraints (referential integrity).

e.g.

`call SYSCS_UTIL.LOAD_REPLACE('<schemaName>', '<tableName>', null, '<inFilePath>', null, null, null, null, null, -1,      '<badRecordLogDirectory>', true, null);`

For this added special modes for INSERT and DELETE, also ignoring triggers / referential integrity:
`INSERT ... --splice-properties insertMode=LOAD_REPLACE `
`DELETE ... --splice-properties  noTriggerRI=1`


carolp-503 (approved):
PR was previously reviewed before being merged and revoked due to a blocking WriteToken change.  That PR has been addressed, and the previous reviews count for this PR, so ready to merge.

PR #4476, opened 2020-11-04
DB-10484 Support new DECFLOAT data type 

arnaud-splice:


arnaud-splice:

mem_sql/pom.xml
```diff
@@ -259,12 +259,10 @@
                             <key>com.splicemachine.enableLegacyAsserts</key>
                             <value>true</value>
                         </systemProperty>
```
msirek, line 261: This should only be uncommented while debugging in your local branch.  We shouldn't have this uncommented for everyone.  It will fill up the mem_sql folder with class files and depending on the queries run this could take a lot of space.

msirek (changes requested):
Looks good.  I just have one comment that needs to be fixed up.

mem_sql/pom.xml
```diff
@@ -259,12 +259,10 @@
...
-                        <!-- Uncomment the following to dump activation class files for mem platform to the mem_sql directory.
```
msirek, line 262: The other comment pointed to the wrong code.  dumpClassFile is the property we don't want to leave uncommented:

arnaud-splice: Very good point, thanks for catching that!

msirek (approved):

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecfloat.java
```diff
@@ -0,0 +1,995 @@
...
+            setToNull();
+        else {
+            isNull = false;
+            value = row.getDecimal(ordinal).round(MathContext.DECIMAL128);
+        }
+    }
+
+    @Override
+    public StructField getStructField(String columnName) {
+        return DataTypes.createStructField(columnName, DataTypes.BinaryType, true);
```
yxia92, line 979: It may be not right to use DataTypes.BinaryType here. For native spark execution path, the use of defloat type will fail for the following case:
```
create table t1 (a1 decfloat, b1 int);
create table t2 (a2 decfloat, b2 int);
insert into t1 values (3,3);
insert into t2 values (3,3);

select * from t1, t2 --splice-properties useSpark=true, joinStrategy=broadcast
> where a1=a2;
ERROR XJ001: Java exception: 'java.lang.RuntimeException: java.math.BigDecimal is not a valid external type for schema of binary: java.io.IOException'.
ERROR XJ001: Java exception: 'java.math.BigDecimal is not a valid external type for schema of binary: java.lang.RuntimeException'.
```

yxia92 (changes requested):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1858/

db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLDecfloat.java
arnaud-splice: Good point,
SQLDecfloat.getSparkObject / setSparkObject now work with Strings. 
Until we define a proper native spark sql type for decfloat, ordering will not be possible

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1867/

arnaud-splice:

yxia92 (approved):
+1

carolp-503 (approved):
+1

spliceci:
TEST SUCCEEDED +1

PR #4505, opened 2020-11-06
DB-10678 Modify accepted params for [VAR]CHAR function 

arnaud-splice:


arnaud-splice:

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
```diff
@@ -725,5 +726,46 @@ public void testDateTimeToCharVarcharConstants() throws Exception {
...
+        try (ResultSet ignored = methodWatcher.executeQuery("select char(current date, 20)")) {
+            Assert.fail("expect failure since we cannot specify a length with something which is not a char/varchar");
+        } catch (SQLException e) {
+            assertEquals("42846", e.getSQLState());
+            Assert.assertThat(e.getMessage(), containsString("Cannot set a length"));
+        }
+        try (ResultSet ignored = methodWatcher.executeQuery("select char(42, 20)")) {
+            Assert.fail("expect failure since we cannot specify a length with something which is not a char/varchar");
+        } catch (SQLException e) {
+            assertEquals("42846", e.getSQLState());
```
martinrupp, line 741: why not use SQLState.LANG_INVALID_CAST

arnaud-splice: I'm using a variation of it. Same prefix number, but more comprehensive error message

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
```diff
@@ -9023,7 +9023,17 @@ ValueNode
...
+     |
+     charType = charOrVarchar()
+     <LEFT_PAREN>
+     operand = additiveExpression(null,0)
+     [
+        <COMMA> (
+            format = datetimeStringFormat()
+        |
+            length = length()
+        )
```
ascend1, line 9035: Ah, now I see why the regression is there. Because they both start with `<COMMA>`, and the length one is shadowed. Thanks for the fix!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
```diff
@@ -320,6 +324,20 @@ else if (this.targetCharType == Types.VARCHAR)
...
+                throw StandardException.newException(
+                        SQLState.LANG_INVALID_CAST_TO_CHAR_WITH_LENGTH_NOT_FROM_CHAR,
+                        sourceCTI.getSQLTypeName(),
+                        getTypeId().getSQLTypeName());
+            }
+            if (dateToStringFormat != -1 && !sourceCTI.isDateTimeTimeStampTypeID()) {
+                throw StandardException.newException(
+                        SQLState.LANG_INVALID_CAST_TO_CHAR_WITH_FORMAT_NOT_FROM_DATE
+                );
+            }
```
ascend1, line 338: I'm not sure if this is always fine. I wanted to do check here, but then there are cases where `sourceCTI` is null.

ascend1 (approved):
With two minor comments.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
arnaud-splice: Do you know when that can be null?
In the meantime, I added a nullcheck

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/sqlgrammar.jj
arnaud-splice: 👍

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
ascend1: I moved line 367 and 368 out, then added a check like `if (sourceJDBCTypeId == Types.TIMESTAMP && stringFormat >= 0)` then throw (because DB2 doesn’t accept datetime format for timestamp), and run tests, then there will be NPE.

arnaud-splice:

dgomezferro (approved):

carolp-503 (approved):
+1

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1882/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1885/

PR #4513, opened 2020-11-06
DB-10638 DB-10643 DB-10671 Implement new foreign key checker.

hatyo:


db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/CreateConstraintConstantAction.java
```diff
@@ -302,7 +302,7 @@ public void	executeConstantAction( Activation activation )
 			case DataDictionary.FOREIGNKEY_CONSTRAINT:
 				ReferencedKeyConstraintDescriptor referencedConstraint = DDUtils.locateReferencedConstraint
 					( dd, td, constraintName, columnNames, otherConstraintInfo );
-				DDUtils.validateReferentialActions(dd, td, constraintName, otherConstraintInfo,columnNames);
+				// DDUtils.validateReferentialActions(dd, td, constraintName, otherConstraintInfo,columnNames);
```
hatyo, line 305: Is `com.splicemachine.db.impl.sql.execute.CreateConstraintConstantAction#executeConstantAction` reachable? if so, how? if not, I suggest we remove it.

It looks almost identical to `com.splicemachine.derby.impl.sql.execute.actions.CreateConstraintConstantOperation#executeConstantAction`

hatyo: ... to be removed, follow up in [DB-10689](https://splicemachine.atlassian.net/browse/DB-10689).

hatyo:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,151 @@
...
+
+    EdgeClassification edgeClassification(int x, int y) {
+        if(dfsParent[y] == x) return EdgeClassification.TREE;
+        if(discovered[y] && !processed[y]) return EdgeClassification.BACK;
+        if(processed[y] && (entryTime[y] > entryTime[x])) return EdgeClassification.FORWARD;
+        if(processed[y] && (entryTime[y] < entryTime[x])) return EdgeClassification.CROSS;
+        throw new IllegalArgumentException("unknown edge class between " + x + " and " + y);
+    }
+
+    public void run(int v) throws StandardException {
```
martinrupp, line 74: add a comment what `run` is doing

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,151 @@
...
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.utils.Pair;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Stack;
+
```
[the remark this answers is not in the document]
martinrupp: please add unit tests for this class

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+
+import java.util.*;
+
+public class Graph {
```
martinrupp, line 21: please add unit tests for Graph

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
```diff
@@ -0,0 +1,36 @@
...
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+public class EdgeNode {
+    public EdgeNode(int y, Type type) {
+        this.y = y;
+        this.type = type;
+        this.next = null;
+    }
+
+    enum Type {C, R, SN, NA};
```
martinrupp, line 23: C, R, SN, NA?

C -> CASCADE;
NA -> NOACTION;
R -> RESTRICT;
SN -> SETNULL;

are those other values even used? Maybe a
`bool cascade` would be sufficient and somewhat more "general"

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -0,0 +1,181 @@
...
+        } else {
+            EdgeNode previous = edgeNode;
+            edgeNode = edgeNode.next;
+            while(true) {
+                if(edgeNode == null) {
+                    throw new IllegalArgumentException("no edge between " + getName(from) + " and " + getName(to));
+                }
+                if(edgeNode.y == to) {
+                    previous.next = edgeNode.next;
+                    edgeNode.next = null; // bye, GC.
```
martinrupp, line 165: `edgeNode.next = null; // bye, GC.`
this line is superflous. the thing you want to clean up is edgeNode, not edgeNode.next. edgeNode.next will survive in previous.next. at this point, no one should be pointing to edgeNode anymore, so no need for this line.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,151 @@
...
+    public Dfs(Graph g, String newConstraintName) {
+        this.graph = g;
+        dfsParent = new int[g.getVertexCount()];
+        processed = new boolean[g.getVertexCount()];
+        discovered = new boolean[g.getVertexCount()];
+        edgeTypes = new EdgeClassification[g.getVertexCount()];
+        entryTime = new int[g.getVertexCount()];
+        exitTime = new int[g.getVertexCount()];
+        init();
+        stack = new Stack<Integer>();
```
dgomezferro, line 51: A Deque is the preferred Java Stack type (Stack is based on Vector, which is synchronized by default and the API is not as consistent as Deque)

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    public void removeEdge(int from, int to) {
+        EdgeNode edgeNode = getEdge(from);
+        if(edgeNode == null) {
+            throw new IllegalArgumentException("no edge between " + getName(from) + " and " + getName(to));
+        }
+        if(edgeNode.y == to) {
+            edgeNodes.set(from, null);
```
martinrupp, line 155: if it's the first edge, but there are other edges, this code would delete all edges, but i guess it should only delete the first one ->
`edgeNodes.set(from, edgeNodes.next);`

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,151 @@
...
+    boolean[] processed;
+    boolean[] discovered;
+    private boolean trackParents;
+
+    enum EdgeClassification {TREE, BACK, FORWARD, CROSS};
+    EdgeClassification[] edgeTypes;
+    Stack<Integer> stack;
+    int[] entryTime;
+    int[] exitTime;
+    int time;
```
dgomezferro, line 37: is Step a better name than Time?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -0,0 +1,181 @@
...
+        String name = getName(toIdx) + "__SURROGATE__" + surrogateCounter++;
+        surrogates.put(surrogateIdx, toIdx);
+        vertexIndex.put(name, surrogateIdx);
+        EdgeNode edgeNode = new EdgeNode(surrogateIdx, type);
+        edgeNode.next = edgeNodes.get(fromIdx);
+        edgeNodes.set(fromIdx, edgeNode);
+        edgeNodes.add(null); // for the surrogate
+    }
+
+    void addEdgeInternal(int fromIdx, int toIdx, EdgeNode.Type type) throws StandardException {
```
martinrupp, line 59: // adds am edge and checks for cycles

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -0,0 +1,181 @@
...
+                    // add the new edge again, but check again for cycles!
+                    addEdgeInternal(fromIdx, toIdx, type);
+                    return;
+                }
+            }
+        }
+        // cycle is unbreakable, bail out
+        StringBuilder sb = new StringBuilder();
+        sb.append("adding the constraint between ").append(
+                getName(p.get(0))).append(" and ").append(getName(p.get(p.size()-1))).append(" would cause the following illegal delete action cascade cycle");
```
martinrupp, line 92: i think normal + should be performant enough for this part of the code

sb.append("adding the constraint between " + getName(p.get(0)) + " and " + getName(p.get(p.size()-1)) + " would cause the following illegal delete action cascade cycle");

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/GraphAnnotator.java
```diff
@@ -0,0 +1,159 @@
...
+    static class Annotation {
+        List<Path> paths;
+        public Annotation() {
+            paths = new ArrayList<>();
+        }
+    }
+
+    Annotation[] annotations;
+
+    public GraphAnnotator(String newConstraintName, Graph graph) {
```
martinrupp, line 77: this also looks easily unit-testable

martinrupp (changes requested):
very nice change youssef! some remarks

dgomezferro (dismissed):
Great change @hatyo , I really like the isolated Graph classes! I haven't looked deeply at the core graph algorithm yet, but I'd love to see some unit tests for that part, which should be easy to achieve given how well you've decoupled them.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -29,12 +36,66 @@ public Graph(Set<String> vertices) {
             vertexIndex.put(vertex, i++);
         }
         parents = new int[vertices.size()][];
+        surrogateCounter = 0;
+        this.newConstraintName = newConstraintName;
+        this.surrogates = new HashMap<>();
+    }
+
+    void addEdge(String from, String to, EdgeNode.Type type) throws StandardException {
```
ascend1, line 44: Maybe add some comments describing the graph structure. Having an example with three nodes would be very helpful in understanding how it works. Currently, interface doesn't match call site (from <-> to reversed) and it's a bit confusing.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DDUtils.java
```diff
@@ -261,7 +261,7 @@ public static void validateReferentialActions(DataDictionary dd,
                                                               deleteConnectionsMap,
                                                               newForeignKeyConstraintName);
             }
-        }
+        } // would be nice to have a logger so we can log that we are not going to perform foreign key checks.
```
arnaud-splice, line 264: you can define `    private static final Logger LOG = Logger.getLogger(DDUtils.class);` at the top of the class.

arnaud-splice (changes requested):

arnaud-splice:
What a great change! I love the new Graph class and I love the clean ITs for for foreign keys valid / invalid cases!

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
```diff
@@ -71,4 +132,50 @@ public int getVertexCount() {
...
+            }
+            edgeNode = edgeNode.next;
+        }
+    }
+
+    public void removeEdge(int from, int to) {
+        EdgeNode edgeNode = getEdge(from);
+        if(edgeNode == null) {
+            throw new IllegalArgumentException("no edge between " + getName(from) + " and " + getName(to));
+        }
```
ascend1, line 153: Is this an error that we expected? I mean, talking about just an abstract graph, if we want to delete an edge that doesn't exist, might just be a no-op and OK to be safely ignored. Or if we don't expect this ever happen, probably assert instead of throw.

ascend1 (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropConstraintConstantOperation.java
```diff
@@ -239,6 +239,8 @@ public void executeConstantAction(Activation activation ) throws StandardExcepti
 			dm.invalidateFor(conDesc, DependencyManager.DROP_CONSTRAINT, lcc);
 			dm.clearDependencies(lcc, conDesc);
 		}
+
+		dd.getDataDictionaryCache().clearConstraintDescriptorListCache(); // not sure if this is the correct place to do this.
```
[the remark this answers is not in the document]
ascend1: +1 for this one.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,151 @@
+/*
```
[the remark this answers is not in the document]
hatyo: Good point, I refactored accordingly.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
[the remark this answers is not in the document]
hatyo: > a small comment explaining why this class exists and what it is doing would be nice

done.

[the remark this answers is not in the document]
hatyo: > please add unit tests for this class

will do in a [separate JIRA ticket](https://splicemachine.atlassian.net/browse/DB-10719).

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: will do in a [separate JIRA ticket](https://splicemachine.atlassian.net/browse/DB-10719).

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
hatyo: You're right, but I would rather keep the distinction as it helps stringifying error messages and potentially address other future use cases.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: Good point, but it seems it could help generational GC as explained [here](https://stackoverflow.com/a/6936214/337194) and [here](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/LinkedList.java#l450).
What do you think?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: Great point, thanks for spotting this, fixed it.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/GraphAnnotator.java
hatyo: True, planning for extensive testing of the graph and its algorithms in [this JIRA](https://splicemachine.atlassian.net/browse/DB-10719), that is actually why I did the separation of the graph checker from the rest of the data dictionary so I can unit test it.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: I always try to avoid String concatenation and use a `StringBuilder` instead, I see your point, the code is ugly, I will make it more readable by restructuring it.

It is not only for performance, but it is kind of a habit I think :)

Let me know what you think about the new version once I push it.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropConstraintConstantOperation.java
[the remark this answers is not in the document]
hatyo: We store in this cache the following map: `Map(<constraint UUID>, <list of all foreign key constraint descriptors referencing that constraint>`, I think we can be more intelligent in cleaning up this cache such that:

1. If we find the UUID of the  constraint we're about to drop in the cache we delete.
2. we iterate over the result of the entries, and for each entry-value, we iterate over all the foreign keys, if we find a foreign key descriptor with UUID matching the one we're about to drop, we remove the entire entry from the map.

I think step(1) removes the children and step(2) removes the parents, both steps are necessary to remove all traces of this constraint from the cache, do you agree?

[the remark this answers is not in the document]
hatyo: ... I just checked the interface of the `ManagedCache`, I couldn't find a way to get all the present keys from the cache which is necessary for step 2.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,142 @@
...
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+import javafx.util.Pair;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Stack;
+
+public class Dfs {
```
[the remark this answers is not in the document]
hatyo: > Can you document this class?

done

> And add unit tests for it?

will do in a [separate JIRA ticket](https://splicemachine.atlassian.net/browse/DB-10719).

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+public class EdgeNode {
+    public EdgeNode(int y, Type type) {
+        this.y = y;
```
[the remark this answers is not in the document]
hatyo: It is the "y" part of an edge between (x,y) in the graph.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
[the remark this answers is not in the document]
hatyo: I will write comments about each one's meaning, they are just used internally and they make quite easy to quickly check a complex graphviz diagram.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -144,6 +145,8 @@ public DataDictionaryCache(Properties startParams,DataDictionary dd) throws Stan
                 Property.LANG_PROPERTY_CACHE_SIZE_DEFAULT);
         int catalogVersionCacheSize = getCacheSize(startParams, Property.LANG_PROPERTY_CACHE_SIZE,
                 Property.LANG_PROPERTY_CACHE_SIZE_DEFAULT);
+        int constraintDescriptorListCacheSize = 10000;//getCacheSize(startParams, Property.LANG_PROPERTY_CACHE_SIZE,
```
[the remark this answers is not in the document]
hatyo: Correct, this is done in a more recent commit.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -728,6 +734,32 @@ public void conglomerateDescriptorCacheRemove(Long conglomId) throws StandardExc
...
     }
 
+    public void constraintDescriptorListCacheAdd(UUID id, ConstraintDescriptorList item) throws StandardException {
+        if (LOG.isDebugEnabled())
+            LOG.debug("constraintDescriptorListCacheAdd " + id);
+        constraintDescriptorListCache.put(id,item);
+    }
+
+    public ConstraintDescriptorList constraintDescriptorListCacheFind(UUID id) throws StandardException {
+        //if (!dd.canReadCache(null))
```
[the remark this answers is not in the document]
hatyo: Correct, it is already restored in more recent commits.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DropConstraintConstantOperation.java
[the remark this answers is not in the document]
dgomezferro: @hatyo you could add an `asMap` method to the ManagedCache that delegates it to the guava cache. But this is not too important, and iterating over all elements might be more expensive than repopulating the cache when needed. Up to you.

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyGraphIT.java
```diff
@@ -0,0 +1,195 @@
...
+            return String.format("ALTER TABLE %s ADD CONSTRAINT \"%s\" FOREIGN KEY (C2) REFERENCES %s(C1) ON DELETE %s",
+                                 tables.get(child), constraintName, tables.get(parent), toString(action));
+        }
+
+        public DDLGenerator link(int child, int parent, Action action) {
+            ddls.add(generateConstraintDdl(child, parent, action));
+            return this;
+        }
+
+        public DDLGenerator materialize() throws Exception {
```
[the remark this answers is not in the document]
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Dfs.java
```diff
@@ -0,0 +1,142 @@
...
+ * See the GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+import javafx.util.Pair;
```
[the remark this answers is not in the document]
hatyo: Correct, it has been already addressed in recent commits.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/DictionaryGraphBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+
+        TableDescriptor referencedTableDescriptor = newConstraintInfo.getReferencedTableDescriptor(dd);
+        descriptors.add(referencedTableDescriptor);
+
+        Set<UUID> visited = new HashSet<>();
+
+        while(!descriptors.isEmpty()) {
+            TableDescriptor descriptor = descriptors.remove();
+            visited.add(descriptor.getUUID());
+            String tableName = descriptor.getSchemaName() + "." + descriptor.getName();
```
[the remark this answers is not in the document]
hatyo: Thanks! fixed in multiple places.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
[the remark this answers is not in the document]
hatyo: OK, I give up :) (read the comment I wrote to Arnaud), two opinions beat one, I will fix accordingly! :)

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/GraphAnnotator.java
```diff
@@ -0,0 +1,145 @@
...
+ * See the GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.splicemachine.db.iapi.sql.dictionary.fk;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
+import javafx.util.Pair;
```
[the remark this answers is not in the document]
hatyo: already fixed in more recent commits.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: You're right, it should never happen, but still I would rather keep an exception to get more context if something goes wrong.

hatyo (commented):
Thanks a lot for your feedback.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
hatyo: totally agree, I fixed it, and added more documentation.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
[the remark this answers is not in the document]
hatyo: I refactored it and renamed it to `to`.

ascend1 (approved):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
martinrupp: woah hm yeah i didn't think about that! a GC hack, interesting. 👍

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/Graph.java
martinrupp: 👍

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/fk/EdgeNode.java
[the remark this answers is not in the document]
martinrupp: wait, you can create a graphviz diagram out of this? how?

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
```diff
@@ -309,6 +309,14 @@ public static BitSet getIndexedCols(int[] indexColsToMainColMap) {
         }
     }
 
+    public static void preDropForeignKey(DDLMessage.DDLChange change, DataDictionary dd) throws StandardException {
```
dgomezferro, line 312: Nitpick, name is a bit confusing since it's being used for both Add/Drop, preAlter.. might be better.

martinrupp (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
hatyo: Thanks Daniel, you're right, I will address it in another PR where I improve the FK checker.

PR #4517, opened 2020-11-09
DB-10688 refactor sqlgrammar.jj

martinrupp:


arnaud-splice (approved):

hatyo (commented):
I don't fully understand why it is better this way (besides reducing the size of the JJ file), but I appreciate your effort doing this. I am ok with this PR if other colleagues approve it.

martinrupp:
> I don't fully understand why it is better this way (besides reducing the size of the JJ file), but I appreciate your effort doing this. I am ok with this PR if other colleagues approve it.

jj files are not supported by all editors, and e.g. doing refactorings is not automatically possible. i have a jj plugin for IDEA, but it is way worse than working on normal jar files, consuming a lot of CPU etc.. Working in sqlgrammar.jj is not pleasent, as there the file is very long, and refactorings are hard. In my opinion, sqlgrammar.jj should focus on GRAMMAR things, and not have code that could be one elsewhere like for example that `getLength(String s)` function. So it's about separating responsibilities and putting less code in hard to maintain/work with jj files.

martinrupp:
> I don't fully understand why it is better this way (besides reducing the size of the JJ file), but I appreciate your effort doing this. I am ok with this PR if other colleagues approve it.

other reasons why we shouldn't have too much code in there:
- can't set a breakpoint in sqlgrammar.jj , as it is not the code, but something that is code generated from
- for that reason, also can't use Java's HotSwap feature, making a full recompile necessary every time
- i guess we also can't unit test parts in there
- not auto complete / type or syntax error help when modifying sqlgrammar.jj

hatyo:
Thanks for the explanation, do you think it makes sense to separate the huge grammar file into multiple functional units? e.g. one file for: data types, expressions, derived tables, window functions, transactions, aggregation, ... etc?

martinrupp:
> Thanks for the explanation, do you think it makes sense to separate the huge grammar file into multiple functional units? e.g. one file for: data types, expressions, derived tables, window functions, transactions, aggregation, ... etc?

in general that would be cool i guess. however, modularization is not supported by JavaCC:
https://stackoverflow.com/questions/26698266/how-to-modularize-a-javacc-grammar-file-jj

> There is no way built in to JavaCC to modularize .jj files. The best thing to do is often to use JJT, as this allows you to move all actions out of the grammar file. If you don't want to use JJT, the next best thing may be to use the builder pattern.

I think this PR is already a step towards builder pattern.

ascend1 (commented):
The jj plugin eats all CPU cycles...Is it better with this change?

ascend1 (approved):

martinrupp:
> The jj plugin eats all CPU cycles...Is it better with this change?

not really. goal is to move more and more code into GrammarImpl.java so we need to go into sqlgrammar.jj only if we really change the *grammar*

martinrupp:
> The jj plugin eats all CPU cycles...Is it better with this change?

line count went down in sqlgrammar.jj from 18'744 to 17'937... so it's a start...

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1974/

martinrupp:
Hi, i had to redo this because of some merge conflicts and found a better solution. Somehow i thought the generated class SQLParser can't inherit from another class, but it is possible. So we don't have to use the `GrammarImpl g = new GrammarImpl()` thing and the `g.` everywhere, which made things a bit easier. I also moved all member variables to GrammarImpl. So now, only "real" grammar and helper functions related to getToken are still in sqlgrammar.jj, rest moved out.

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1991/

carolp-503 (approved):
+1

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2005/

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2007/

PR #4528, opened 2020-11-10
DB-10685 Fix SYSCS_INVALIDATE_STORED_STATEMENTS

msirek:
The SYSCS_INVALIDATE_STORED_STATEMENTS procedure should clear stored statements from the dictionary cache so that they will be recompiled the next time they are executed.  Currently it does not, it only invalidates stored statements on disk.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
@@ -48,17 +48,22 @@ import com.splicemachine.db.iapi.util.StringUtil;
 import com.splicemachine.db.impl.jdbc.EmbedDatabaseMetaData;
 import com.splicemachine.db.impl.jdbc.Util;
 import com.splicemachine.db.impl.load.Import;
+import com.splicemachine.db.impl.sql.catalog.DataDictionaryCache;
 import com.splicemachine.db.impl.sql.execute.JarUtil;
 import com.splicemachine.db.jdbc.InternalDriver;
 import com.splicemachine.db.shared.common.reference.AuditEventType;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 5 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeScriptToCreateNewInvalidateStoredStatementsProcedure.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    protected void upgradeSystemTables() throws StandardException {
+        sdd.replaceInvalidateStoredStatementsAndDependentProcedures(tc);
+
+        SpliceLogUtils.info(LOG, "Catalog upgraded: created new version of SYSCS_UTIL.SYSCS_INVALIDATE_STORED_STATEMENTS.");
+
+    }
+}
```
jyuanca, line 36: This is not needed. If software version is higher than data dictionary catalog version, createOrUpdateAllSystemProcedures() recreates all system procedures.

jyuanca (changes requested):

yxia92 (approved):
+1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeScriptToCreateNewInvalidateStoredStatementsProcedure.java
msirek: Thanks @jyuanca, I removed the upgrade script.

jyuanca (approved):

PR #4544, opened 2020-11-11
DB-10483 Support DEC and DECIMAL functions

hatyo:
Implementation of DB2 DECIMAL and DEC functions is provided in this PR. These functions allow the user to create `decimal` object from almost any other object such as `time`, `timestamp`, `varchar`, `integer` and so on.

To get more details about these functions visit DB2's [documentation](https://www.ibm.com/support/producthub/db2/docs/content/SSEPGG_11.5.0/com.ibm.db2.luw.sql.ref.doc/doc/r0000791.html) and take a look at the tests in this PR.


hatyo:

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
```diff
@@ -0,0 +1,130 @@
...
+import java.math.BigInteger;
+import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+public interface DecimalUtil {
+
+    /**
```
arnaud-splice, line 34: I think you can remove that reference. We will have our own documentation.

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
```diff
@@ -0,0 +1,130 @@
...
+                bigDecimal = SQLDecimal.getBigDecimal(dvd);
+                if (bigDecimal == null) return null;
+            }
+            break;
+            case DECFLOAT: {
+                bigDecimal = SQLDecfloat.getBigDecimal(dvd);
+                if (bigDecimal == null) return null;
+            }
+            break;
+            case CHAR: // fallthrough
```
arnaud-splice, line 68: Why not construct a SQLDecimal directly from the string?

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
```diff
@@ -0,0 +1,130 @@
...
+        Calendar cal = null;
+        switch (DataValueFactoryImpl.Format.formatFor(dvd)) {
+            case TINYINT: // fallthrough
+            case SMALLINT: // fallthrough
+            case INTEGER: // fallthrough
+            case LONGINT: {
+                if (String.valueOf(Math.abs(dvd.getLong())).length() > (precision - scale)) {
+                    throw StandardException.newException(SQLState.LANG_INVALID_DECIMAL_CONVERSION, dvd.toString(), precision, scale); // DB2 SQLSTATE 22003
+                }
+                bigDecimal = new BigDecimal(new BigInteger(Bytes.toBytes(dvd.getLong())), 0);
```
dgomezferro, line 55: Use `BigInteger.valueOf(long val)` instead

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
```diff
@@ -0,0 +1,130 @@
...
+            case DECFLOAT: {
+                bigDecimal = SQLDecfloat.getBigDecimal(dvd);
+                if (bigDecimal == null) return null;
+            }
+            break;
+            case CHAR: // fallthrough
+            case VARCHAR: {
+                String input = dvd.getString();
+                if (input == null) return null;
+                input = input.trim().replace(decimalCharacter, "");
```
dgomezferro, line 72: Isn't this removing the decimalCharacter? How do we use it later?

mem_sql/pom.xml
```diff
@@ -259,12 +259,10 @@
                             <key>com.splicemachine.enableLegacyAsserts</key>
                             <value>true</value>
                         </systemProperty>
-                        <!-- Uncomment the following to dump activation class files for mem platform to the mem_sql directory.
```
dgomezferro, line 262: Please restore this, otherwise we are going to generate lots of files on mem builds

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
dgomezferro: After seeing the ITs I think you are using as "digit group separator" but I'm pretty sure it's supposed to be the "fractional separator". From the docs:
>  it can appear at most once in string-expression

and

> Digits are truncated from the end of the decimal number if the number of digits to the right of the decimal separator character is greater than the scale scale. An error is returned if the number of significant digits to the left of the decimal character (the whole part of the number) in string-expression is greater than precision - scale

dgomezferro (changes requested):

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
hatyo: I would rather keep it for reference (even our documentation will be mirroring that some way of the other I think).

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
hatyo: The string could contain invalid characters, it could also contain a comma for example and decimal separator(s), that's why it is not possible.

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
hatyo: You're right, I glimpsed over that, thanks for pointing it out.

mem_sql/pom.xml
hatyo: Sorry, checked in by mistake, reverted.

arnaud-splice (approved):

arnaud-splice:
Awesome change!

dgomezferro (approved):

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1929/

carolp-503 (approved):
+1

PR #4551, opened 2020-11-11
DB-10555 Support multi-column (not-)in-subquery

ascend1:


ascend1:

arnaud-splice (changes requested):
Great change! One small request: can you add a test with a table that does not have matching columns? For example, instead of (1, 10) (2,20) (3,30)… have something like (1, 10) (1, 20) (2,20)… 
That would catch potential bugs where we'd compare only the first column of the subquery.

msirek (approved):
Looks good.  There can be issues with NOT IN subqueries with nulls.  It would be good to add some test cases using a nullable column in one of the NOT IN columns and a non-nullable column in the other, and then test that the answer is the same when the order of the columns is swapped.  Use nulls in the data too.

ascend1:
> Great change! One small request: can you add a test with a table that does not have matching columns? For example, instead of (1, 10) (2,20) (3,30)… have something like (1, 10) (1, 20) (2,20)…
> That would catch potential bugs where we'd compare only the first column of the subquery.

Sure, I add one such test in the latest commit.

ascend1:
> Looks good. There can be issues with NOT IN subqueries with nulls. It would be good to add some test cases using a nullable column in one of the NOT IN columns and a non-nullable column in the other, and then test that the answer is the same when the order of the columns is swapped. Use nulls in the data too.

Yes, I add one test to cover this case in the latest commit.

ascend1:

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1916/

spliceci:
TEST SUCCEEDED +1

PR #4552, opened 2020-11-11
DB-10320 execute transaction RPCs in the priority thread pool

OlegMazurov:
Set transaction RPC priority to HIGH to dispatch such RPCs to the priority thread pool.

### DIFF
--- a/hbase_storage/src/main/java/com/splicemachine/si/impl/SkeletonTxnNetworkLayer.java
+++ b/hbase_storage/src/main/java/com/splicemachine/si/impl/SkeletonTxnNetworkLayer.java
@@ -39,7 +39,8 @@ public abstract class SkeletonTxnNetworkLayer implements TxnNetworkLayer{
     @Override
     public void beginTransaction(byte[] rowKey,TxnMessage.TxnInfo txnInfo) throws IOException{
         TxnMessage.TxnLifecycleService service=getLifecycleService(rowKey);
-        ServerRpcController controller=new ServerRpcController();
+        SpliceRpcController controller = new SpliceRpcController();
+        controller.setPriority(HBaseTableDescriptor.HIGH_TABLE_PRIORITY);
         service.beginTransaction(controller,txnInfo,new BlockingRpcCallback<>());
         dealWithError(controller);
     }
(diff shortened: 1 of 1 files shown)


arnaud-splice (approved):

hbase_storage/src/main/java/com/splicemachine/si/impl/SkeletonTxnNetworkLayer.java
```diff
@@ -39,7 +39,8 @@
     @Override
     public void beginTransaction(byte[] rowKey,TxnMessage.TxnInfo txnInfo) throws IOException{
         TxnMessage.TxnLifecycleService service=getLifecycleService(rowKey);
-        ServerRpcController controller=new ServerRpcController();
+        SpliceRpcController controller = new SpliceRpcController();
+        controller.setPriority(HBaseTableDescriptor.HIGH_TABLE_PRIORITY);
```
dgomezferro, line 43: Any reason you used `HBaseTableDescriptor.HIGH_TABLE_PRIORITY` instead of `HConstants.ADMIN_QOS` or `HConstants.HIGH_QOS` ?

dgomezferro (approved):
Looks good, I wonder about the chosen priority (different from what we use for getTxn()) but we treat everything over 10 the same as far as I can see.

hbase_storage/src/main/java/com/splicemachine/si/impl/SkeletonTxnNetworkLayer.java
OlegMazurov: `getTxn()` originally used `HConstants.HIGH_QOS`. DB-10230 introduced `HBaseTableDescriptor.HIGH_TABLE_PRIORITY/NORMAL_TABLE_PRIORITY` which are mapped to `HConstants.ADMIN_QOS/NORMAL_QOS`. I reused `HBaseTableDescriptor.HIGH_TABLE_PRIORITY` for other transaction RPCs for consistency.

OlegMazurov:
@dgomezferro 
`HBaseTableDescriptor.HIGH_TABLE_PRIORITY` is consistently used for all RPCs including `getTxn()`.

PR #4567, opened 2020-11-12
DB-10718 fix HEX function for non-ASCII Strings

martinrupp:


splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -377,4 +378,13 @@ public long getConglomId(String tableName, String schemaName) throws Exception {
         }
         return -1l;
     }
+
+    public String executeGetString(String sql, int index) throws SQLException {
+        try( Statement s = getOrCreateConnection().createStatement();
+             ResultSet rs = s.executeQuery(sql))
+        {
+            Assert.assertTrue(rs.next());
+            return rs.getString(index);
```
hatyo, line 387: please make this [`null`](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#wasNull()) friendly:

```
String result = rs.getString(index);
rs.wasNull() ? null : result;
```

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceStringFunctionsIT.java
```diff
@@ -702,40 +702,37 @@ public void testCastToFixedLengthCharType() throws Exception {
...
+                "'B'",              "42",
+                "'000020'",         "303030303230",
+                "'000020'" ,        "303030303230",
+                "'ß'",              "C39F",
+                "'\u00E4'",         "C3A4",     // ä  https://www.compart.com/de/unicode/U+00E4
+                "'\uD83D\uDE02'",   "F09F9882", // 😂 https://www.compart.com/de/unicode/U+1F602
+                "'Hello, World!'",    "48656C6C6F2C20576F726C6421",
+
+                "''",               "",
+                "null",             null,
```
hatyo, line 718: see above, in order to test null properly.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceStringFunctions.java
```diff
@@ -134,17 +135,15 @@ else if (i >= 256)
...
-    public static String HEX(String s)
-    {
+    public static String HEX(String s) throws UnsupportedEncodingException {
         if (s == null)
             return null;
-        else
-            return StringUtil.toHexString(s.getBytes(),0,s.length()).toUpperCase();
-
+        else {
+            byte[] b = s.getBytes("UTF-8");
```
hatyo, line 142: I would rather use a constant such as `com.splicemachine.primitives.Bytes.UTF8_CHARSET`

hatyo (changes requested):

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
martinrupp: this is already covered by getString, see docu https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getString(int)

however you're right if we would add sth like `Integer executeGetInteger` we would need to check wasNull since getInt only returns the primitive type `int`.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceStringFunctions.java
martinrupp: thanks, good suggestion!

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
hatyo: Thanks! I didn't know that it is specified in JDBC interface that reference types must return null if they are null.

hatyo (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4574, opened 2020-11-12
DB-10705 fix transaction time travel search algorithm behavior

hatyo:
The time travel search algorithm is a simple binary search algorithm whose search domain is splice transaction table. When implementing this algorithm I made the wrong assumption that transactions have uniquely identified active timestamps. This false assumption went unnoticed under normal workloads, however testing with a TPC-C which puts the system under heavy transactional load exposed it because many transactions could actually share the same time causing the algorithm to loop forever.

A unit test is added that simulates such workload and verifies the fix.


dgomezferro (approved):
Looks good and nice simple test!

hbase_storage/src/main/java/com/splicemachine/si/impl/region/TxnTimeTraveler.java
```diff
@@ -52,14 +52,14 @@ public TxnTimeTraveler(TxnFinder txnFinder) {
...
                 if(middleTS == rightTS)
                 {
                     reversed = true;
                     changed = true;
                 }
                 else if(middleTS < rightTS)
                 {
-                    right = toRowKeyWithBucket(middleTx.getFirst(), bucket);
                     rightTS = middleTS;
                     changed = true;
```
OlegMazurov, line 64: `changed = true;` is common in both conditional branches, can be moved outside `if`

OlegMazurov (approved):

hbase_storage/src/main/java/com/splicemachine/si/impl/region/TxnTimeTraveler.java
hatyo: Thanks! I actually saw it after I pushed PRs to 3.0 and 3.1, will fix it later if that's OK.

arnaud-splice (approved):

PR #4577, opened 2020-11-12
DB-10690 fix duplicate FirstWriteToken

martinrupp:


martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1933/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1934/

dgomezferro (changes requested):
Very nice find @martinrupp !

I'd say the problem is only constrained to ForeignKey indices, because they have uniqueness to them but they don't have a constraint checker, so we run into the problem you described (always adding a first write token).

The proposed solution adds a lot of overhead for non-pk inserts, so I'd rather special case the FK indices by creating a dummy constraintChecker for them so that we always do conflict detection for them.

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -284,6 +284,17 @@ private void releaseLocksForKvBatch(Pair<KVPair, Lock>[] locks){
...
             KVPair.Type writeType=kvPair.getType();
+
+            DataResult possibleConflicts = null;
+            if(!skipConflictDetection)
+            {
+                // we need the possibleConflicts for
+                // - has constraintChecker && writeType == UPSERT/UPDATE/DELETE: checking for constraint violations
+                // - writeType == INSERT/UPSERT/DELETE: evaluating addFirstOccurrenceToken
+                if( bloomInMemoryCheck == null || bloomInMemoryCheck.get(i) ) {
+                    possibleConflicts = table.getLatest(kvPair.getRowKey(), possibleConflicts);
```
martinrupp, line 295: note for myself: are these conflicts collected? why does getLatest have parameter possibleConfilcts?

martinrupp:
> Very nice find @martinrupp !
> 
> I'd say the problem is only constrained to ForeignKey indices, because they have uniqueness to them but they don't have a constraint checker, so we run into the problem you described (always adding a first write token).
> 
> The proposed solution adds a lot of overhead for non-pk inserts, so I'd rather special case the FK indices by creating a dummy constraintChecker for them so that we always do conflict detection for them.

i also had the feeling this adds overhead. The problem however is, that currently if we for some cases do not execute the line  `possibleConflicts = table.getLatest(kvPair.getRowKey(), possibleConflicts);`, then possibleConflicts is null, which is interpreted as "we need to write FIRST_WRITE_TOKEN or DELETE_RIGHT_AFTER_FIRST_WRITE_TOKEN", which will cause again duplicated FIRST_WRITE_TOKENS

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1937/

hbase_storage/src/test/java/com/splicemachine/si/impl/server/SICompactionStateMutateTest.java
```diff
@@ -412,6 +441,8 @@ public void mutatePurgeLatestTombstoneInMinorCompaction() throws IOException {
                 SITestUtils.getMockFirstWriteCell(100)
         ));
 
+        System.out.println(toString(inputCells, false));
```
dgomezferro, line 444: Remove

hbase_storage/src/test/java/com/splicemachine/si/impl/server/SICompactionStateMutateTest.java
```diff
@@ -422,8 +453,9 @@ public void mutatePurgeLatestTombstoneInMinorCompaction() throws IOException {
                 transaction2,
                 transaction1
         ));
-        cutPurgeDuringMinorCompaction.mutate(inputCells, transactions, outputCells);
+        SICompactionStateMutate.mutate(cutPurgeDuringMinorCompaction, context, inputCells, transactions, outputCells);
         assertThat(outputCells, is(empty()));
+        System.out.println(toString(outputCells, false));
```
dgomezferro, line 458: Remove

dgomezferro (approved):
Looks good! Please remove System.out calls, there are a few more than what I highlighted

msirek (approved):

jyuanca (approved):

PR #4583, opened 2020-11-12
DB-10236 Take expression-based indexes as covering indexes properly (3.1)

ascend1:
…#4162)

* DB-10175 Expression-based index scan with relational operator preds

* DB-10175 Expression-based index scan with in-list predicates

* DB-10175 Fix SpotBugs

* DB-10175 Address comments

* DB-10236 Rewrite query for a covering expression-based index

In this change, rewriting happens in:

- where clause
- group by clause
- order by clause
- having clause, including aggregates
- select list, including aggregates

* DB-10236 Make joins work correctly for covering expr index

* DB-10236 Fix SpotBugs

* DB-10236 Fix TernaryOperator handling in index expression rewriting

* DB-10175 Address comments

* DB-10175 Address comments / add semanticallyEquals()

DB-10312 Cache index expression ASTs is partially applied here so that
the code binding index expressions is not duplicated everywhere.

* DB-10175 Address comments / fix potential concurrent issue

Index expression ASTs should not be cached in IndexDescriptor because
it is in turns cached in data dictionary cache, which is global to all
queries. For now, always return a copy of ASTs.

* DB-10236 Adapt to the latest changes made in DB-10175

Note: IndexIT.testCoveringExpressionBasedIndexSubquery fails in
current state because we cannot bind expressions after resultColumns
are changed to index columns after changeAccessPath(). As a result,
index expressions containing JavaToSQLValueNode do not work.

* DB-10236 Fix IndexIT.testCoveringExpressionBasedIndexSubquery

The fix is done by allowing binding base table columns even after
table's access path has been changed to index columns. This ability is
limited to index on expressions only.

* DB-10312 Add hashCode() to ValueNode class hierarchy

* DB-10236 Address comments / fix expr collecting and derived tables

* DB-10236 Address comments / fix index expr in RHS of in-list

* WIP: DB-10236 Address comments / fix count(*) and index expr in window

* Resolved missed conflicts

* DB-10236 Column mapping manipulation for index on expressions

* DB-10236 Fix outer join without correcting column references

* DB-10236 Address comments / fix collecting expressions

* DB-10236 Address comment / fix post join condition replacing

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
@@ -391,7 +391,7 @@ public class IndexDescriptorImpl implements IndexDescriptor, Formatable {
                 indexColumnTypes[i].readExternal(in);
             }
         }
-        synchronized(this) {
+        synchronized (this) {
             executableExprs = new BaseExecutableIndexExpression[numIndexExpr];
         }
     }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
@@ -31,11 +31,12 @@
 
 package com.splicemachine.db.iapi.sql.compile;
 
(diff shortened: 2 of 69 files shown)


carolp-503 (approved):
Approved for 3.1

PR #4585, opened 2020-11-13
DB-10732 NSDS IT Made schema call compatible with both cdh and hdp.

msirek:


msirek:
Fixes errors on the HDP platform in ColumnCaseIT.

PR #4592, opened 2020-11-13
DB-10723 Support cast to [VAR]CHAR as SBCS DATA 

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1944/

arnaud-splice:

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
```diff
@@ -782,5 +782,36 @@ public void testLengthOfCastToVarchar() throws Exception {
...
+        }
+    }
+
+    @Test
+    public void testCastToCharForSbcsData() throws Exception
+    {
+        methodWatcher.executeUpdate("drop table sbcs if exists");
+        methodWatcher.executeUpdate("create table sbcs(a varchar(30) for bit data, b char(30) for bit data)");
+        methodWatcher.executeUpdate("insert into sbcs values (cast('a' as varchar(30) for bit data), cast('b' as char(30) for bit data))");
+        methodWatcher.executeUpdate("insert into sbcs values (case when 1 = 2 then 'dummy' end, case when 1 = 2 then 'dummy' end)");
```
dgomezferro, line 802: You can do `insert into sbcs values (NULL, NULL)` the case trick is only needed when you want a NULL in the select portion of a query I think.

db-engine/src/main/java/com/splicemachine/db/catalog/types/TypeDescriptorImpl.java
```diff
@@ -65,6 +65,8 @@
     private int                        scale;
     private boolean                    isNullable;
     private int                        maximumWidth;
+
+    transient private boolean          forSbcsData;
```
dgomezferro, line 69: @arnaud-splice have you tested this with triggers? 
@jyuanca @yxia92 do you think this could cause issues if we don't serialize it?

dgomezferro (approved):

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FunctionIT.java
arnaud-splice: Ah, good to know, thanks!

db-engine/src/main/java/com/splicemachine/db/catalog/types/TypeDescriptorImpl.java
arnaud-splice: I have, everything works well. We only use this variable in `CastNode.generateExpression`. I think it'd be cleaner to add this boolean to the cast node itself, but that means reworking the grammar quite a bit, I'll see what I can do there.

arnaud-splice: Refactoring done.

martinrupp (approved):

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):
Nice refactoring!

carolp-503 (approved):
+1

PR #4614, opened 2020-11-16
DB-10503 fix foreign key child interceptor's checks for parent row existence

hatyo:


splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyChildInterceptWriteHandler.java
```diff
@@ -159,24 +159,20 @@ public void close(WriteContext ctx) throws IOException {
...
-                    case INCLUDE:
+                    case INCLUDE: // fallthrough
                     case INCLUDE_AND_NEXT_COL: //the cell is included, so we have some data
-                    default: //do nothing
+                        return true;
+                    case SKIP: // fallthrough
+                    case NEXT_COL: // fallthrough
+                    case SEEK: // fallthrough
+                    case NEXT_ROW: // fallthrough
+                    default: //do nothing, keep trying
```
ipraznik-splice, line 172: does it make sense still to use switch here, as most of the cases empty? Maybe "if" will look simpler

dgomezferro (approved):

ascend1 (approved):

splice_machine/src/main/java/com/splicemachine/pipeline/foreignkey/ForeignKeyChildInterceptWriteHandler.java
hatyo: If it is OK, I would rather keep it this way mainly to illustrate the exact purpose of the algorithm for each `DataFilter.ReturnCode` return type. But I don't have strong opinion about it and can still change it if you think this doesn't make too much sense.

ipraznik-splice (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4628, opened 2020-11-17
DB-10746 Fix UPDATE statement index corruption.

msirek:
The logic in IndexTransformer and elsewhere to avoid updating an index on an UPDATE statement which doesn't modify any index columns is broken.  If an index shares any columns with some other index that is getting updated, then it will also be updated.  UpdateNode.getXAffectedIndexes() detects which indexes shouldn't get updates, and skips adding their columns to the update, meaning they will be inaccessible.  Any attempt to rebuild an index row with missing columns will corrupt the index (produce columns with bad data, possibly a base rowid which cannot be properly read).   Attempting to read the bad index row will return incorrect results or throw an error.  Deleted the corresponding base table row may fail to delete the index row.

The fix is to add all indexes to the update for now, so that all possible index columns are accessible during the update.  A follow-on Jira will fix the logic to properly detect which indexes to skip.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLModStatementNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLModStatementNode.java
@@ -1519,11 +1519,17 @@ abstract class DMLModStatementNode extends DMLStatementNode
             ** If this index doesn't contain any updated
             ** columns, then we can skip it.
             */
-            if ((updatedColumns != null) &&
-                    (!updatedColumns.updateOverlaps(
-                            cd.getIndexDescriptor().baseColumnPositions()))) {
-                continue;
-            }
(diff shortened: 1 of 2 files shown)


msirek:

ascend1 (approved):
Awesome finding!

dgomezferro (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1972/

arnaud-splice (approved):

carolp-503 (approved):
+1

jyuanca (approved):

ascend1 (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1978/

PR #4633, opened 2020-11-17
DB-10708 SpliceCK interactive mode and other improvements (3.1)

martinrupp:
- `./spliceck.sh interactive` starts spliceck in "interactive shell" `spliceck>` mode. Commands stay the same, however since -we're caching the hbase connection, it's much faster. exit with `exit`
- rget can now also get all rows (use `all` instead of rowid)
- since this can be quite a lot, `--limit X` limits the number of rows to X. Default is 100.
- rget can limit the version it prints with e.g. `-V 1` (like default for hbase shell)
- rget can print additional cells like hbase (`--hbase`)
- tlist can accept filters e.g. `tlist *DUMMY` list all schema.table that end with DUMMY
- txlist also limits output by default to 100 (change with --limit)
- both rget and txlist notify when they cut off results due to --limit.

### DIFF
--- a/splice_ck/src/main/java/com/splicemachine/ck/ConnectionCache.java
+++ b/splice_ck/src/main/java/com/splicemachine/ck/ConnectionCache.java
@@ -1,0 +1,26 @@
+package com.splicemachine.ck;
+
+import com.splicemachine.ck.hwrap.ConnectionWrapper;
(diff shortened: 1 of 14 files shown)


carolp-503 (approved):
Approved for 3.1

PR #4635, opened 2020-11-17
DB-10734 expose entire row history to foreign key child interceptor

hatyo:
When adding a row in the child table, the child-interceptor logic was making decisions about the existence of the foreign key in the parent table using latest rows' latest versions from HBase which could lead to true negatives. This behavior is corrected by letting the child-interceptor examine the entire row history of the parent keys.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
@@ -22,6 +22,7 @@ import com.splicemachine.test_tools.TableCreator;
 import org.junit.*;
 import org.junit.experimental.categories.Category;
 
+import java.sql.Connection;
 import java.sql.Statement;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
```diff
@@ -706,6 +707,36 @@ public void multipleTablesReferencingSameTable() throws Exception {
...
+                    .create();
+            conn.commit();
+
+        try(Connection c = newNoAutoCommitConnection()) {
+            c.setAutoCommit(false);
+
+            try (Statement statement = c.createStatement()) {
+                statement.executeUpdate("update P set col1 = 42");
+            }
+            Thread.sleep(30 * 1000);
```
dgomezferro, line 728: I don't like this long pause here. Without the pause I think most of the times the next insert will happen before we have a chance to remove the rollback (specially if we have everything set up, the connection, prepared statement, etc), so I'd rather do that than introduce 30 more seconds of runtime on ITs

dgomezferro (changes requested):
Nice clean fix @hatyo ! I'd rather remove the sleep() on the IT if possible though

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
hatyo: done.

dgomezferro (approved):

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
```diff
@@ -706,6 +707,35 @@ public void multipleTablesReferencingSameTable() throws Exception {
...
+            }
+            c.rollback();
+        }
+
+        try(Connection conn = newNoAutoCommitConnection()) {
+            try(Statement s = conn.createStatement()) {
+                s.executeUpdate("insert into C values (400, 'a', 1, 42)"); // should not fail
+            }
+            conn.commit();
+        }
```
dgomezferro, line 736:
```suggestion
        try(Connection c = newNoAutoCommitConnection();
        Connection c2 = newNoAutoCommitConnection();
        PreparedStatement ps = c2.preparedQuery("insert into C values (400, 'a', 1, 42)")) {
            c.setAutoCommit(false);

            try (Statement statement = c.createStatement()) {
                statement.executeUpdate("update P set col1 = 42");
            }
            c.rollback();
            
            ps.executeUpdate();
            c2.commit();
        }
```

dgomezferro: I think something like this would be better to reduce the window of time between the rollback and the update, but as it is it looks fine to me.

splice_machine/src/test/java/com/splicemachine/foreignkeys/ForeignKeyCheckIT.java
```diff
@@ -706,6 +707,35 @@ public void multipleTablesReferencingSameTable() throws Exception {
...
+                    .withInsert("insert into P values(?,?,?,?)")
+                    .withRows(rows(row(1, "a", 1, 1)))
+                    .create();
+            new TableCreator(conn)
+                    .withCreate("create table C (col1 int primary key, col2 varchar(2), col3 int, col4 int, constraint CHILD_FKEY foreign key(col2, col3) references P(col2, col4) )")
+                    .create();
+            conn.commit();
+
+        try(Connection c = newNoAutoCommitConnection()) {
+            c.setAutoCommit(false);
```
arnaud-splice, line 723: setAutoCommit(false) is redundant

arnaud-splice (approved):

carolp-503 (approved):
+1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/1977/

PR #4659, opened 2020-11-19
DB-10719 add unit tests for foreign key dependency checker.

hatyo:
In this PR we add few hundreds of positive and negative unit tests to the foreign key dependency graph. the sheer amount of unit tests comes from constructing each graph in the test cases using all possible edge permutations. This brute-force test uncovered one bug in graph construction mechanism causing the checker to construct an incorrect graph and ultimately causing it to accept an incorrect foreign key dependency graph. The generation of the permutations is done using [Heap's algorithm](https://en.wikipedia.org/wiki/Heap%27s_algorithm).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/Graph.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/Graph.java
@@ -208,6 +208,7 @@ public class Graph {
                     edgeNode.next = null; // bye, GC.
                     return;
                 }
+                previous = edgeNode;
                 edgeNode = edgeNode.next;
             }
         }
--- a/db-engine/src/test/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/GraphTest.java
+++ b/db-engine/src/test/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/GraphTest.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright (c) 2012 - 2020 Splice Machine, Inc.
+ * This file is part of Splice Machine.
+ * Splice Machine is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Affero General Public License as published by the Free Software Foundation, either
+ * version 3, or (at your option) any later version.
+ * Splice Machine is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License along with Splice Machine.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.splicemachine.db.iapi.sql.dictionary.foreignkey;
+
+import com.splicemachine.db.iapi.error.StandardException;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import splice.com.google.common.collect.Lists;
+
+import java.util.*;
+import java.util.stream.IntStream;
+
+@RunWith(Enclosed.class)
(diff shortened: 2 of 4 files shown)


db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/Graph.java
```diff
@@ -208,6 +208,7 @@ public void removeEdge(int from, int to) {
                     edgeNode.next = null; // bye, GC.
                     return;
                 }
+                previous = edgeNode;
```
dgomezferro, line 211: Is this a legitimate issue on the current version? What kind of failures would it cause if so?

hatyo: Yes it is, it causes the graph to miss some edges (which leads to incorrect conclusions about the validity of the graph). Since we're not updating `previous` correctly, we end up removing the wrong edge (line 207). Here is an example that exposes this bug:

```
digraph {
  C -> A[label="Restrict"];
  A -> B[label="Cascade"];
  C -> A[label="SetNull"];
  B -> C[label="Cascade"];
}
```

In this graph, adding the final edge `B -> C` causes two cycles to appear:
1. `A -Cascade-> B -Cascade-> C -SetNull-> A`
2. `A -Cascade-> B -Cascade-> C -Restrict-> A`

We remove the first cycle by rerouting `C-SetNull->A` to surrogate node `C-SetNull->S_A_0`. At this point, the edge list for node `C` looks something like this:
`'C' -> 'S_A_0 (setnull)' -> 'A (restrict)'`.

Now, we try check again if we have a cycle (line 124), and we find that we have another one: `A -Cascade-> B -Cascade-> C -Restrict-> A`. So now we reroute `C -Restrict-> A` to a new surrogate node `C -Restrict->S_A_1`. (at this point, the edge list for node `C` will be `'C' -> 'S_A_1 (restrict)' -> 'S_A_0 (setnull)' -> 'A (restrict)'`) And we attempt to remove the edge between `C` and `A` (line 122).

Here is where we hit the problem, when `removeEdge` is called we set the `previous` to edge `'S_A_1 (restrict)'`, and we don't update it anymore, so when we find the edge to remove, we end removing this edge instead of the edge `'A (restrict)'`, therefore we end up building a different graph which happens to be OK.

hatyo: Failures it causes is accepting invalid graphs, I think this is unlikely to happen with customers who are migrating (hopefully) valid schemas from other DBMS to SpliceMachine.

db-engine/src/test/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/GraphTest.java
```diff
@@ -0,0 +1,239 @@
...
+                                 "A -> B[label=\"Cascade\"];\n" +
+                                 "B -> C[label=\"Cascade\"];\n" +
+                                 "C -> A[label=\"Cascade\"];\n" +
+                                 "}").generateGraph("something");
+                Assert.fail("expected graph creation to fail with error message: adding the constraint between A and C would cause the following " +
+                                    "illegal delete action cascade cycle A B C");
+            } catch (Exception e) {
+                Assert.assertTrue(e instanceof StandardException);
+                StandardException sqlException = (StandardException) e;
+                Assert.assertEquals("42915", sqlException.getSQLState());
```
msirek, line 234: Maybe use LANG_DELETE_RULE_VIOLATION instead of 42915.

msirek (approved):

dgomezferro (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

db-engine/src/test/java/com/splicemachine/db/iapi/sql/dictionary/foreignkey/GraphTest.java
hatyo: Thanks! will keep that in mind when adding more fixes to FKs.

PR #4662, opened 2020-11-19
DB-10790 Fix missing parameter in DROP TABLE error.

msirek:
Regression of DB-10525.

> drop table cjpsplat;
> ERROR 42Y55: 'CJPSPLAT' cannot be performed on '{1}' because it does not exist.
> 
> ... should be:
> 
> drop table cjpsplat;
> ERROR 42Y55: 'DROP TABLE' cannot be performed on 'CJPSPLAT' because it does not exist.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
@@ -512,7 +512,7 @@ public class SpliceUnitTest {
         }
     }
 
-    public static void testFail(String sqlText,
+    protected void testFail(String sqlText,
                             List<String> expectedErrors,
                             SpliceWatcher methodWatcher) throws Exception {
         try (ResultSet rs = methodWatcher.executeQuery(sqlText)) {
@@ -526,7 +526,7 @@ public class SpliceUnitTest {
         }
     }
 
-    public static void testUpdateFail(String sqlText,
+    protected void testUpdateFail(String sqlText,
                                   List<String> expectedErrors,
                                   SpliceWatcher methodWatcher) throws AssertionError {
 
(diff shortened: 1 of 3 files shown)


msirek:

bklo94 (approved):

ascend1 (approved):

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
```diff
@@ -526,7 +526,7 @@ protected void testFail(String sqlText,
         }
     }
 
-    protected void testUpdateFail(String sqlText,
+    public static void testUpdateFail(String sqlText,
```
OlegMazurov, line 529: This change is not warranted. Make DropTableIT extend SpliceUnitTest and call testUpdateFail().
Use DecimalIT as an example.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
```diff
@@ -512,7 +512,7 @@ protected void testQueryUnsorted(String sqlText, String expected, SpliceWatcher
         }
     }
 
-    protected void testFail(String sqlText,
+    public static void testFail(String sqlText,
```
OlegMazurov, line 515: See below

OlegMazurov (changes requested):

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
msirek: Done.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceUnitTest.java
msirek: Reverted.

carolp-503 (approved):
+1

PR #4668, opened 2020-11-20
DB-10786 Remote ITs are broken (2.8)

OlegMazurov:
No remote_it in 2.8.
Fixed SpliceUnitTest.getOldestActiveTransaction() to return long.


carolp-503 (approved):
Approved for 2.8

PR #4672, opened 2020-11-23
DB-10854 Fix txn leak in FROM FINAL TABLE.

msirek:
- With autocommit off, the transaction stack would get pushed for every FROM FINAL TABLE statement, but never popped.
  This is due to a missing needsSavepoint flag in the outermost prepared statement plus code from SPLICE-1625 that's no longer needed (trigger logic has added an extra transaction wrapping the statement since then) which caused the transaction to be elevated when the operation was closed (elevating it upon close causes it never to get popped).
- The cleanup code in TriggerHandler was too convoluted and not kicking in, so it's been simplified.
- I also moved the ThreadLocals added by DB-10525 from DataDictionaryCache to TriggerReferencingStruct per [DB-10738](https://splicemachine.atlassian.net/browse/DB-10738).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
@@ -46,14 +46,13 @@ import com.splicemachine.db.iapi.sql.depend.Dependency;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.sql.dictionary.SchemaDescriptor;
 import com.splicemachine.db.iapi.sql.execute.ExecutionContext;
-import com.splicemachine.db.iapi.types.TypeId;
 import com.splicemachine.db.iapi.util.ByteArray;
 import com.splicemachine.db.iapi.util.InterruptStatus;
 import com.splicemachine.db.impl.ast.JsonTreeBuilderVisitor;
 import com.splicemachine.db.impl.sql.compile.CharTypeCompiler;
-import com.splicemachine.db.impl.sql.catalog.DataDictionaryCache;
 import com.splicemachine.db.impl.sql.compile.ExplainNode;
 import com.splicemachine.db.impl.sql.compile.StatementNode;
+import com.splicemachine.db.impl.sql.compile.TriggerReferencingStruct;
 import com.splicemachine.db.impl.sql.conn.GenericLanguageConnectionContext;
 import com.splicemachine.db.impl.sql.misc.CommentStripper;
 import com.splicemachine.system.SimpleSparkVersion;
@@ -65,7 +64,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.Timestamp;
-import java.sql.Types;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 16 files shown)


msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -687,8 +685,14 @@ else if (newMergeJoinString.equals("forced"))
                 preparedStmt.compilingStatement=false;
                 preparedStmt.notifyAll();
             }
-            DataDictionaryCache.fromTableTriggerDescriptor.remove();
-            DataDictionaryCache.fromTableTriggerSPSDescriptor.remove();
+            TriggerReferencingStruct.fromTableTriggerDescriptor.remove();
+            TriggerReferencingStruct.fromTableTriggerSPSDescriptor.remove();
+            // Communicate to the immediate parent statement if its child
+            // contains a FROM TABLE clause.
+            if (BoundAndOptimizedStatement != null)
```
dgomezferro, line 692: Can you rename `BoundAndOptimizedStatement` in all places to firstLowerCase?

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -249,6 +249,7 @@ public ResultSet executeQuery(String sql, String userName, String password) thro
         while (rs.next()) {
             resultList.add((T) rs.getObject(1));
         }
+        rs.close();
```
dgomezferro, line 252: I think a try-with-resources would be better, in case we raise an exception for instance.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -265,6 +266,7 @@ public ResultSet executeQuery(String sql, String userName, String password) thro
             }
             resultList.add(row);
         }
+        rs.close();
```
dgomezferro, line 269: Same

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
```diff
@@ -122,9 +127,16 @@ public static void createData(Connection conn, String schemaName) throws Excepti
 
     @Before
     public void createDataSet() throws Exception {
+        runningOperations = getNumberOfRunningOperations();
```
dgomezferro, line 130: Shouldn't this always be 0, since the test is Serial?

dgomezferro (changes requested):
Change looks good, great turn around @msirek ! I only have some style comments

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
msirek: Theoretically.  But the user may be doing something else on the machine.  If they have another running operation in a user session I don't want it to raise a false alarm in this test.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
msirek: Made the change.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
msirek: Made the change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
msirek: Done.

msirek:
> Change looks good, great turn around @msirek ! I only have some style comments

Thanks.

dgomezferro (approved):

arnaud-splice (approved):

ascend1 (approved):

carolp-503 (approved):
+1

PR #4675, opened 2020-11-23
DB-10483 DEC[IMAL] functions round DECFLOAT instead of truncating it.

hatyo:


dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/iapi/util/DecimalUtil.java
```diff
@@ -64,6 +65,7 @@ static DataValueDescriptor getDecimalDb2(DataValueDescriptor dvd,
             break;
             case DECFLOAT: {
                 bigDecimal = SQLDecfloat.getBigDecimal(dvd);
+                roundingMode = BigDecimal.ROUND_HALF_EVEN; // should be configurable (DB-10672)
```
msirek, line 68: How was this rounding mode picked?  Normal rounding (as done in math class, or in spark) uses ROUND_HALF_UP I believe.

msirek: Found out in the standup ROUND_HALF_EVEN is the DB2 default.

msirek (approved):

carolp-503 (approved):
+1

PR #4678, opened 2020-11-23
DB-10542 Support returning long result DDL from SHOW_CREATE_TABLE

ascend1:


ascend1:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
```diff
@@ -618,4 +618,2118 @@ private static void  checkEqualIgnoreConstraintOrder(String expected, String act
         String expectedResult = Stream.of(expected.split("CONSTRAINT ")).map(str -> str.substring(0, str.length() - 2)).sorted().collect(Collectors.joining(", "));
         Assert.assertEquals(expectedResult, result);
     }
+
+    @Test
+    public void testTable2100Columns() throws Exception {
+        String ddl = "CREATE TABLE \"" + SCHEMA + "\".\"TEST_2100\" (\n" +
```
hatyo, line 624: I would write a DDL generator instead and use it in multiple tests.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -2276,9 +2274,29 @@ public static void INVALIDATE_GLOBAL_DICTIONARY_CACHE() throws Exception {
     public static void SHOW_CREATE_TABLE(String schemaName, String tableName, ResultSet[] resultSet) throws SQLException
     {
         List<String> ddls = SHOW_CREATE_TABLE_CORE(schemaName, tableName, false, true);
-        StringBuilder sb = new StringBuilder("SELECT * FROM (VALUES '");
-        sb.append(ddls.get(0));
-        sb.append(";') FOO (DDL)");
+        StringBuilder sb = new StringBuilder("SELECT * FROM (VALUES ");
```
hatyo, line 2277: You should (over)estimate the capacity of the `StringBuilder` otherwise you leave it grow exponentially which could stress the GC e.g. if you have hundreds of such queries running on parallel.

i.e. do something like this:
```
long size = 22 /* size of 'SELECT * FROM (VALUES ' */ + 11 /* size of ') FOO (DDL)' */ + ddl.length + 15 /* size of 'cast('' as clob)'*/*(ddl.length / MAX_STR_LITERAL_LENGTH) + 7 /* size of '|| ';''*/
StringBuilder sb = new StringBuildre(size);
```

hatyo (changes requested):

splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
```diff
@@ -618,4 +618,2118 @@ private static void  checkEqualIgnoreConstraintOrder(String expected, String act
...
+                "\"COL_0\" VARCHAR(30)\n" +
+                ",\"COL_1\" CHAR(10)\n" +
+                ",\"COL_2\" BIGINT\n" +
+                ",\"COL_3\" DATE\n" +
```
martinrupp, line 628: can you write a for-loop generating this? i guess you already have, or did you create 2100 columns by hand?

martinrupp (changes requested):

splice_machine/src/test/java/com/splicemachine/derby/utils/ShowCreateTableIT.java
ascend1: I do have a program generating it, but it randomize the column types...Also, we have to generate the DDL as the way how `show create table` formats it, so that it can be used for both creating and comparing. I'll tune my current code to do it.

ascend1:
The test case added this this PR is proved to work with DB-10720 from @ipraznik-splice . We will move the test there, implement the comments of this PR, and close this one.

PR #4679, opened 2020-11-23
DB-10542 Support returning long result DDL from SHOW_CREATE_TABLE (3.1)

ascend1:


ascend1:

PR #4688, opened 2020-11-24
DB-10878 more server-side logging.

hatyo:
In this PR we add more DRDA logging on the server. The purpose of this is to log as many exceptions as possible for easier postmortems.


db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -97,6 +97,8 @@
 
     private static final Logger AUDITLOG =Logger.getLogger("splice-audit");
 
+    private static final Logger log = Logger.getLogger(DRDAConnThread.class);
```
OlegMazurov, line 100: _log_ -> _LOG_ for consistency

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -306,13 +309,22 @@ private void parsePRDDTA() throws DRDAProtocolException
         initialize();
     }
 
+    private String getSessionName() {
+        return session == null ? "" : session.toString();
+    }
+
+    private void debug(String message) {
+        if(log.isDebugEnabled()) {
```
OlegMazurov, line 317: The reason we redundantly check if Debug/Trace levels are enabled outside the logging methods is to avoid argument evaluation which is done unconditionally and may have quite noticeable overhead.
Introducing an intermediate convenience method apparently defeats the purpose. 
The existing _trace()_ method doesn't conform to that principle and should be reworked (separately from this PR; also to use Logger). It should not be considered as acceptable still.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -380,6 +393,8 @@ public void run() {
                 // TODO: Could make use of Throwable.addSuppressed here when
                 //       compiled as Java 7 (or newer).
                 try {
+                    debug("encountered error " + error.getMessage());
```
OlegMazurov, line 396: Use `LOG.debug(message, error)`. It will also print the `error.getMessage()` and the stack trace.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -1188,6 +1200,9 @@ private void processCommands() throws DRDAProtocolException
                     server.consoleExceptionPrint(sqle);
                     SanityManager.THROWASSERT("Unexpected exception after " +
                             "codePoint "+cpStr, sqle);
+                } catch (Exception e) {
+                    server.consoleExceptionPrint(e);
```
OlegMazurov, line 1204: Use `LOG`. Ideally we'll have just one logging system and it won't be the Derby one.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -8697,14 +8714,24 @@ private void closeSession()
...
             // Print an error to the console and close this
             //thread. (6013)
+            debug("encountered exception while closing session " + getSessionName());
             sendUnexpectedException(se);
             close();
         }
+        catch (Exception e)
+        {
+            // log the exception and rethrow.
+            debug("encountered exception while closing session "+ getSessionName());
```
OlegMazurov, line 8731: `LOG.debug(message, error)`

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
```diff
@@ -8725,6 +8752,7 @@ private void handleException(Exception e)
             if (e instanceof DRDAProtocolException) {
                 // protocol error - write error message
                 sendProtocolException((DRDAProtocolException) e);
+                server.consoleExceptionPrintTrace(e);
```
OlegMazurov, line 8755: `LOG.debug(message, error)`

db-drda/src/main/java/com/splicemachine/db/impl/drda/NetworkServerControlImpl.java
```diff
@@ -3973,6 +3972,10 @@ void addSession(Socket clientSocket) throws Exception {
                 if ((maxThreads == 0) || (threadList.size() < maxThreads)) {
                     thread = new DRDAConnThread(session, this, getTimeSlice(),
                                                 getLogConnections());
+                    thread.setUncaughtExceptionHandler((t, e) -> {
+                        consoleMessage("tracing unhandled exception originating from thread: '" + t.getName() + "'", true);
```
OlegMazurov, line 3976: `LOG.debug(message, error)`

OlegMazurov (changes requested):

jpanko1 (commented):
I approve contingent on Oleg's comments being resolved.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: done.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: Yes, I also think we should change `trace()` to use our `Logger` but I would do it separately since it is orthogonal to this change.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: done.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: Yes, that's definitely something we should do, I didn't want to introduce that here because I thought it is out of scope. Will create a JIRA for this at least.

For now, I would prefer to keep it like this, because the error will appear anyway (although in a different log file).

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: done.

db-drda/src/main/java/com/splicemachine/db/impl/drda/DRDAConnThread.java
hatyo: I would make it part of this [JIRA](https://splicemachine.atlassian.net/browse/DB-10893).

db-drda/src/main/java/com/splicemachine/db/impl/drda/NetworkServerControlImpl.java
hatyo: I would make it part of this [JIRA](https://splicemachine.atlassian.net/browse/DB-10893).

hatyo:
@OlegMazurov thanks for the feedback, I addressed most of your comments, I would address the rest of the comments related to using LOG instead of Derby's own logging mechanisms such as `trace`, `consoleMessage`, and `consoleExceptionPrintTrace` in a separate [JIRA](https://splicemachine.atlassian.net/browse/DB-10893), since it is orthogonal to the purpose of the this ticket.

OlegMazurov (approved):

jpanko1 (approved):

carolp-503 (approved):
+1

PR #4690, opened 2020-11-24
DB-10654 Hit an assertion error during restore

ipraznik-splice:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1925,4 +1904,18 @@ public long getSystablesMinRetentionPeriod() {
...
+        SchemaDescriptor sd = getSystemSchemaDescriptor();
+        tc.elevate("dictionary");
+
+        TableDescriptor td = getTableDescriptor("SYSCONSTRAINTS", sd, tc);
+        if (td != null) {
+            dropAllColumnDescriptors(td.getUUID(), tc);
+            dropTableDescriptor(td, sd, tc);
+        }
+
+        createNaturalNumbersTable(tc);
```
jyuanca, line 1918: This drops SYSCONSTRAINTS table. However, you only want to drop a column from it.

If this is an extra column that's to be ignored, it is OK not to drop it.

jyuanca (approved):

ipraznik-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2015/

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
msirek: I'm interested to see the final code here before reviewing this.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCONSTRAINTSRowFactory.java
```diff
@@ -198,9 +195,6 @@ public ExecRow makeRow(boolean latestVersion, TupleDescriptor td,TupleDescriptor
 		/* 6th column is STATE (char(1)) */
         row.setColumn(SYSCONSTRAINTS_STATE,new SQLChar(isEnabled?"E":"D"));
 
-		/* 7th column is REFERENCED */
-        row.setColumn(SYSCONSTRAINTS_REFERENCECOUNT,new SQLInteger(referenceCount));
-
```
dgomezferro, line 203: Should we keep writing something here to prevent issues if we rollback to a previous version?
Say we upgrade to this version, create a constraint, we write only 6 values. There's an issue with the upgrade, so we rollback to the previous version. When deserializing the descriptor for the new constraint we try to read the 7th column but fail.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ReferencedKeyConstraintDescriptor.java
```diff
@@ -270,9 +263,8 @@ else if (type == DISABLED)
 	 *
 	 * @return false
 	 */
-	public boolean isReferenced()
-	{
-		return referenceCount != 0;
+	public boolean isReferenced() throws StandardException {
+		return getReferenceCount() > 0;
```
dgomezferro, line 267: The optimization in `getForeignKeyConstraints()` no longer applies, because for the "fast" path we are getting the list of foreign keys from the data dictionary. It actually hurts because if we have the list cached we will always request it from the DD first. Please remove the optimization there and coordinate with @hatyo to make sure this doesn't make the creation of large FK hierarchies too slow.

dgomezferro: Thinking a bit more about it, we could use the same cache field `fkConstraintList` in `getReferenceCount()`, if it's already populated we have the list and don't need to retrieve it from the DD. If we haven't, we retrieve the list and update the field.

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCONSTRAINTSRowFactory.java
jyuanca: I recommend not to drop the column even it is not used

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConstraintDescriptor.java
```diff
@@ -308,7 +306,7 @@ public int getReferenceCount()
 	 *
 	 * @return true/false
 	 */
-	public abstract boolean needsToFire(int stmtType, int[] modifiedCols);
+	public abstract boolean needsToFire(int stmtType, int[] modifiedCols) throws StandardException;
```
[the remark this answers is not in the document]
ipraznik-splice: the methods isReferenced and getReferenceCount are changed to use fkConstraintList directly or to load from DataDictionary via getForeignKeys. The method getForeignKeys throws StandardException. I've added throws StandardException to this chain of methods. I can of course wrap with try/catch in getOrLoadFKConstraintList, but with throws is cleaner from my point of view.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ReferencedKeyConstraintDescriptor.java
```diff
@@ -318,8 +284,7 @@ public int decrementReferenceCount()
 	 *
 	 * @return true/false
 	 */
-	public boolean needsToFire(int stmtType, int[] modifiedCols)
-	{
+	public boolean needsToFire(int stmtType, int[] modifiedCols) throws StandardException {
```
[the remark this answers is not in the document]
ipraznik-splice: please see below

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
ipraznik-splice: I've rolled back my changes. The field will not be dropped.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCONSTRAINTSRowFactory.java
ipraznik-splice: I've rolled back my changes. The field will not be dropped.

ipraznik-splice:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

carolp-503 (approved):
+1

PR #4693, opened 2020-11-25
DB-10885 Fix FINAL TABLE txn leak in all paths.

msirek:
FROM FINAL Table must elevate the transaction to writeable via a call to getCurrentTransaction() in VTIOperation, but not all flavors of SpliceOperation were calling getCurrentTransaction on its source result set, so we ended up deferring the elevation of the transaction, which leaks it.

All operations that could have a VTIOperation as its source are now modified with a version of getCurrentTransaction that checks the source operation.  A new error is returned if we try to use FROM TABLE in a SetOp query.

Also, compiling the temporary trigger prepared statement caused a new transaction to be created in preparation for a write to the data dictionary, which caused in our case transactions to go to the transaction table instead of using subtransactions.  FROM TABLE does not write the temporary trigger to the data dictionary, so this new transaction is avoided in this fix.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
@@ -1488,4 +1488,8 @@ public interface LanguageConnectionContext extends Context {
 
     void resetDB2VarcharCompatibilityMode();
 
+    void setCompilingFromTableTempTrigger(boolean newVal);
+
+    boolean isCompilingFromTableTempTrigger();
+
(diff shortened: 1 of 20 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2016/

msirek:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4723, opened 2020-11-27
DB-10911 Fix Activation leak in FINAL TABLE query.

msirek:
The Activation that executes the INSERT/UPDATE/DELETE of a FROM TABLE query (VTIOperation.executeFromTableDML -> fromTableDML_SPS.executeSubStatement) is never saved off anywhere so that it can be closed.  So, all of the resources that it references are leaked until the LanguageConnectionContext is cleared, which may be until the sqlshell session ends.

The solution is to register the Activation as a closeable item of the SpliceOperation that's executing, so that when it closes, the Activation will automatically be closed.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/Activation.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/Activation.java
@@ -84,7 +84,7 @@ import java.util.Hashtable;
  *
  */
 
-public interface Activation extends Dependent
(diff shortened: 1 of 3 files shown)


msirek:

dgomezferro (approved):

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4732, opened 2020-11-30
DB-10871 more efficient API for getting active Txns

hatyo:


hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2028/

dgomezferro (commented):
Change looks good @hatyo ! I'm not sure if we are always registering the 'destination' tables though, because it looks like we are only registering them when we create the transaction but a transaction could start writing to a different table later.
Btw, I'm not sure we are correctly registering written to tables for transactions that write through triggers, for instance. Could you double check that?

hatyo:
> Change looks good @hatyo ! I'm not sure if we are always registering the 'destination' tables though, because it looks like we are only registering them when we create the transaction but a transaction could start writing to a different table later.
> Btw, I'm not sure we are correctly registering written to tables for transactions that write through triggers, for instance. Could you double check that?

Thanks @dgomezferro, I will check that.

PR #4735, opened 2020-11-30
DB-10914 + DB-10926 Long varchar to timestamp in prepared statements 

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2030/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -55,154 +55,101 @@
...
+        TypeId rightTypeId = rightOperand.getTypeId();
+
+        if (! leftTypeId.isStringTypeId() && rightTypeId.isStringTypeId())
+        {
+            rightOperand = addCastNodeForStringToNonStringComparison(leftOperand, rightOperand);
+            rightTypeId = rightOperand.getTypeId();
+        }
+        else if (! rightTypeId.isStringTypeId() && leftTypeId.isStringTypeId())
+        {
+            leftOperand.addCastNodeForStringToNonStringComparison(rightOperand, leftOperand);
```
dgomezferro, line 150: Should this be like this instead?
```suggestion
            leftOperand = addCastNodeForStringToNonStringComparison(rightOperand, leftOperand);
``` 
Is it possible we are not testing this branch? Can we add a test?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
```diff
@@ -240,6 +240,20 @@ public ValueNode bindExpression(FromList fromList,
         return this;
     }
 
+    private void addCastOnRightOperandForStringToNonStringComparison() throws StandardException {
+        if (singleLeftOperand) {
```
dgomezferro, line 244: We already checked this before calling this function, do you want it here too to emphasize it?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
```diff
@@ -252,8 +266,13 @@ public void bindComparisonOperator() throws StandardException{
 
         /* Can the types be compared to each other? */
         /* Multicolumn IN list cannot currently be constructed before bind time. */
-        if (singleLeftOperand)
-            rightOperandList.comparable(getLeftOperand());
+        if (singleLeftOperand) {
+            if (!rightOperandList.comparable(getLeftOperand(), false)) {
```
dgomezferro, line 270: `ValueNodeList.comparable()` is only used here, I'd rather make it always return a boolean and explicitly raise an exception here (instead of adding the extra boolean parameter), to me it's surprising that comparable() would raise an exception instead of returning true/false

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryListOperatorNode.java
arnaud-splice: Yes. Would you rather I make that an assertion?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
arnaud-splice: Good point, I will add a test and fix that.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
```diff
@@ -1622,6 +1624,46 @@ public boolean collectExpressions(Map<Integer, Set<ValueNode>> exprMap) {
...
+     * an index on a string column. This is because the string types
+     * are all of low precedence, so the comparison rules of the non-string
+     * value are used, so it may not find values in a string index because
+     * it will be in the wrong order. So, cast the string value to its
+     * own type. This is easier than casting it to the non-string type,
+     * because we would have to figure out the right length to cast it to.
+     * @return the new rightOperand
+     */
+    protected ValueNode addCastNodeForStringToNonStringComparison(
+            ValueNode leftOperand, ValueNode rightOperand) throws StandardException {
```
ascend1, line 1641: Does it make sense to put this function in `BinaryOperatorNode`? It seems not used in any other context.

ascend1 (approved):

ascend1 (commented):
Maybe also have a test case on like predicate, something like `'2000-01-01' like ... escape ...`. Not sure if it's a valid case.

hatyo (approved):

hatyo:
Maybe you want to change the PR title to include also DB-10926

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
arnaud-splice: It is used in both `BinaryComparisonOperatorNode` and `BinaryListOperatorNode` whose closest common ancestor is `ValueNode`

dgomezferro (approved):

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #4740, opened 2020-11-30
DB-10898 DB-10897 configurable timestamp precision.

hatyo:
In this PR we support configurable `TIMESTAMP` precision (between 0 and 9), this alters the behavior of `CHAR(TIMESTAMP...))` making the resulting textual representation of the timestamp appear with different precision.

Changing the precision can be done by running `call SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY( 'splice.function.timestampPrecision', '9' );` where `9` in this example is the desired `TIMESTAMP` precision.


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -603,6 +604,10 @@ else if (newMergeJoinString.equals("forced"))
                 }
                 cc.setDisablePerParallelTaskJoinCosting(disablePerParallelTaskJoinCosting);
 
+                /**
```
arnaud-splice, line 607: I guess we can't quite deprecate that one until we support an alternative that can be used for a transition though, but I agree we need to deprecate this in favour of a single parameter for timestamp precision, and a new way to specify the precision at function level: e.g. `CURRENT TIMESTAMP(9)`

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
```diff
@@ -1101,6 +1123,7 @@ private void genDataValueConversion(ExpressionClassBuilder acb,
              * of VSDV.
              */
 
+            // not sure if this is important
```
arnaud-splice, line 1126: @martinrupp are you familiar with that? I seem to remember that you worked with precisions before.

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
hatyo: Yes, maybe I shouldn't write that until we have an alternative. will remove it.

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2064/

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -1283,6 +1283,9 @@ public static void SYSCS_SET_GLOBAL_DATABASE_PROPERTY(final String key,final Str
             PropertyInfo.setDatabaseProperty(key, value);
             DDLMessage.DDLChange ddlChange = ProtoUtil.createSetDatabaseProperty(tc.getActiveStateTxn().getTxnId(), key);
             tc.prepareDataDictionaryChange(DDLUtils.notifyMetadataChange(ddlChange));
+            // we need to invalidate the statement caches since we could set parameters that affect query plans.
+            SYSCS_INVALIDATE_STORED_STATEMENTS();
+            SYSCS_EMPTY_GLOBAL_STATEMENT_CACHE();
```
msirek, line 1288: I like the idea of automatically forcing recompile for convenience, but are you sure we want to do this?  It may have a performance impact.  Every time a flag is changed we have to recompile all SQLs and triggers.  There is no option for the user to set a database property without calling these cache-clearing functions if they wish to avoid it.  Something we should think about, perhaps, before making this change.

hatyo:

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -619,6 +620,22 @@ else if (newMergeJoinString.equals("forced"))
...
+                int timestampPrecision = CompilerContext.DEFAULT_TIMESTAMP_PRECISION;
+                try {
+                    if (timestampPrecisionString != null)
+                        timestampPrecision = Integer.parseInt(timestampPrecisionString);
+                    if (timestampPrecision < Limits.MIN_TIMESTAMP_PRECISION)
+                        timestampPrecision = Limits.MIN_TIMESTAMP_PRECISION;
+                    if (timestampPrecision > Limits.MAX_TIMESTAMP_PRECISION)
+                        timestampPrecision = Limits.MAX_TIMESTAMP_PRECISION;
+                } catch (Exception e) {
+                    // If the property value failed to convert to a boolean, don't throw an error,
```
jyuanca, line 634: Better to throw the error so that the user can correct the property values

jyuanca (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
msirek: @jyuanca I believe this is not the code to handle the statement itself, but to parse the value once it's already been set.  So, if we throw the Exception here, that would cause all SQL requests to fail after a user has set an out-of-range value for this property.  So, the current code may be OK.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
msirek: We can decide this point later, so the build can proceed.  Opened the following Jira:  https://splicemachine.atlassian.net/browse/DB-10948

msirek (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2070/

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
hatyo: Thanks @msirek for your observation, maybe it is OK to keep this implementation as setting a global database property doesn't happen very often but let's follow up in the JIRA you opened to discuss improvements (one improvement that @dgomezferro mentioned was e.g. to only invalidated affected statements using the dependency manager).

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
hatyo: @jyuanca does @msirek's address your comment?

hatyo:
I adapted the failing test in EE in [this](https://github.com/splicemachine/spliceengine-ee/pull/572/files) PR.

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2077/

dgomezferro (approved):

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
jyuanca: This is not needed for all database properties. You can do this if the property is timestamp precision

jyuanca (approved):

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
hatyo: Thanks @jyuanca! I will address it in DB-10948 then.

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2079/

PR #4744, opened 2020-12-01
DB-10895 Complete DB2 varchar compatibility for MultiProbeScan.

msirek:
This completes the work to support DB2 varchar compatibility mode for queries using MultiProbeScan.
Trailing spaces are ignored when comparing a varchar with any other string.  Normally in Splice, comparisons of varchars do not ignore trailing spaces.  The original work was done by DB-10597, which utilizes a new SQLVarcharDB2Compatible DataValueDescriptor in place of SQLVarchar, where required.  This follow-on work uses the same approach.

> call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', true);
> create table t (v varchar(14), c char(2));
> create index ti on t (v,c);
> insert into t values ('ABC','A ');
> /* Should return one row. */
> select * from t --splice-properties index=ti
>                      WHERE v in ( ' ', 'ABC      ');

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
@@ -50,7 +50,6 @@ import com.splicemachine.db.iapi.sql.execute.ExecutionFactory;
 import com.splicemachine.db.iapi.store.access.ScanController;
 import com.splicemachine.db.iapi.types.*;
 import com.splicemachine.db.iapi.util.JBitSet;
-import com.splicemachine.db.impl.db.BasicDatabase;
 
 import java.lang.reflect.Modifier;
 import java.util.*;
@@ -848,10 +847,6 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
         newPred.setOriginalInListPredList(origList);
     }
 
-    private boolean isMemPlatform() {
-        return "true".equals(System.getProperty("com.splicemachine.isMemPlatform"));
-    }
-
     private void orderUsefulPredicates(Optimizable optTable,
                                        ConglomerateDescriptor cd,
                                        boolean pushPreds,
@@ -867,6 +862,9 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
         int usefulCount=0;
         Predicate predicate;
         boolean rowIdScan;
+        final boolean varcharDB2CompatibilityMode =
+                      getCompilerContext().getVarcharDB2CompatibilityMode();
+        int varcharRangeKeyPos = Integer.MAX_VALUE;
 
         if(cd!=null && !cd.isIndex() && !cd.isConstraint()){
             List<ConglomerateDescriptor> cdl=optTable.getTableDescriptor().getConglomerateDescriptorList();
@@ -1014,6 +1012,11 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
 
             Integer position=isIndexUseful(pred,optTable,pushPreds,skipProbePreds,cd);
             if(!pred.isFullJoinPredicate() && position!=null){
+                if (varcharDB2CompatibilityMode &&
+                    pred.isVarcharBinaryRelationalOperator()) {
+                    if (position < varcharRangeKeyPos)
+                        varcharRangeKeyPos = position;
+                }
                 if (pred.isInListProbePredicate()) {
                     inlistPreds.put(position, pred);
                     if (position >= 0 && multiColumnMultiProbeEnabled)
@@ -1080,7 +1083,7 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
             }
             TreeMap<Integer, Predicate> inlistPredsCopy = (TreeMap < Integer, Predicate>)inlistPreds.clone();
             for (Map.Entry<Integer, Predicate> p : inlistPredsCopy.entrySet()) {
-                if (isEquality[p.getKey()]) {
+                if (p.getKey() <= varcharRangeKeyPos && isEquality[p.getKey()]) {
                     p.getValue().setIndexPosition(p.getKey());
                     inlistQualified = true;
                     /* Remember the useful predicate */
@@ -1116,34 +1119,20 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
         ArrayList<Predicate> predsForNewInList = null;
     
         if (inlistQualified) {
-          // In DB2 mode, probe keys can be overlapping with eachother.
-          // On HBase the overlapping keys are merged together, but on
-          // mem they aren't causing duplicate rows.  So here we avoid
-          // constructing the overlapping keys on mem.
-          boolean varcharDB2CompatibilityMode = getCompilerContext().getVarcharDB2CompatibilityMode();
-          if (varcharDB2CompatibilityMode && isMemPlatform()) {
-              BinaryRelationalOperatorNode bron =
-                  (BinaryRelationalOperatorNode) usefulPredicates[0].getRelop();
-              boolean isVarChar =
-                  bron.getLeftOperand().getTypeServices().
-                         getTypeName().equals(TypeId.VARCHAR_NAME);
-              if (isVarChar)
-                  usefulCount = 1;
-          }
-          if (inlistPreds.size() > 1 && usefulCount > 1) {
+          if (inlistPreds.size() > 1) {
             predsForNewInList = new ArrayList<>();
             int firstPred = -1, lastPred = -1, lastIndexPos = -1, firstInListPred = -1;
 
             boolean foundPred[] = new boolean[usefulCount];
             int numConstants = 1;
             int maxMulticolumnProbeValues = getCompilerContext().getMaxMulticolumnProbeValues();
-
             for (int i = 0; i < usefulCount; i++) {
                 final Predicate pred = usefulPredicates[i];
 
                 if (pred.getIndexPosition() == lastIndexPos + 1) {
                     BinaryRelationalOperatorNode bron =
                         (BinaryRelationalOperatorNode) pred.getRelop();
+
                     boolean inList = pred.isInListProbePredicate();
                     if (inList ||
                         (bron != null && bron.getOperator() == RelationalOperator.EQUALS_RELOP)) {
@@ -1166,14 +1155,6 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
                         predsForNewInList.add(usefulPredicates[i]);
                         lastIndexPos = pred.getIndexPosition();
                         foundPred[i] = true;
-
-                        if (varcharDB2CompatibilityMode && isMemPlatform()) {
-                            boolean isVarChar =
-                            bron.getLeftOperand().getTypeServices().
-                                   getTypeName().equals(TypeId.VARCHAR_NAME);
-                            if (isVarChar)
-                                break;
-                        }
                     }
                 }
                 // Can't have any gaps in index position.
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
@@ -50,6 +50,7 @@ import com.splicemachine.db.iapi.sql.execute.ExecutionFactory;
 import com.splicemachine.db.iapi.store.access.ScanController;
 import com.splicemachine.db.iapi.types.*;
 import com.splicemachine.db.iapi.util.JBitSet;
+import com.splicemachine.db.impl.db.BasicDatabase;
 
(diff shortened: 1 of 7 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MultiProbeDerbyScanInformation.java
```diff
@@ -293,39 +301,64 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
+							 // unless we're in DB2 varchar compatibility mode.
+							 if (isVarCharType) {
+							 	if (targetIsSQLVarcharDB2Compatible) {
+							 		if (dvd1.getClass().equals(SQLVarchar.class))
+							 			listData.setDVD(position,
+										                new SQLVarcharDB2Compatible(dvd1.getString()));
+								}
+							 	else if (!(dvd1 instanceof SQLVarchar))
+								     listData.setDVD(position, new SQLVarchar(dvd1.getString()));
+							 }
```
martinrupp, line 336: same code as 347-361. i think we should combine those in a function

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DB2VarcharCompatibilityIT.java
```diff
@@ -325,4 +340,140 @@ public void testInvalidateStoredStatements() throws Exception {
         // Restore the flag setting to the value when this test started.
         methodWatcher.executeUpdate("call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', true)");
     }
+
+    @Test
+    public void testMultiProbeScan() throws Exception {
+        String sqlTemplate = "select * from t a --splice-properties useSpark=" + useSpark.toString() +
+        ", index=%s\n" +
+        "where v in ('SBVGCCC    ', 'G')\n" +
+        "and c in ( ' ', CAST('A' AS CHAR(2)))";
```
martinrupp, line 349: this strings are weirdly intended. i think we should intend multi-line commands like e.g.
```
String sqlTemplate = "select * from t a --splice-properties useSpark=" + useSpark.toString() +
        ", index=%s\n" +
        "where v in ('SBVGCCC    ', 'G')\n" +
        "and c in ( ' ', CAST('A' AS CHAR(2)))";
```

martinrupp (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -1111,15 +1116,28 @@ private void orderUsefulPredicates(Optimizable optTable,
...
+          // constructing the overlapping keys on mem.
+          boolean varcharDB2CompatibilityMode = getCompilerContext().getVarcharDB2CompatibilityMode();
+          if (varcharDB2CompatibilityMode && isMemPlatform()) {
+              BinaryRelationalOperatorNode bron =
+                  (BinaryRelationalOperatorNode) usefulPredicates[0].getRelop();
+              boolean isVarChar =
+                  bron.getLeftOperand().getTypeServices().
+                         getTypeName().equals(TypeId.VARCHAR_NAME);
+              if (isVarChar)
+                  usefulCount = 1;
```
ascend1, line 1131: It feels dangerous to change `usefulCount` to 1 here because `usefulCount` is used not only for multi-probe scan construction but also later to determine start/stop keys. Suppose we have several other predicates together with one in-list on a varchar column so we enter this code path. Setting `usefulCount` to 1 would prevent all other predicates to be scan start/stop keys. Or am I missing anything?

If we don't want to build a `MultiProbe...Scan` for this in-list, is it sufficient to remove it from `inlistPreds` and clean up its states in other data structures (e.g., usefulPredicates[], usefulCount--, isEquality[], etc.) ?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MultiProbeDerbyScanInformation.java
ascend1: +1 for this

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -847,6 +848,10 @@ private void saveOriginalInListPreds(Predicate newPred, List<Predicate> predsFor
         newPred.setOriginalInListPredList(origList);
     }
 
+    private boolean isMemPlatform() {
```
dgomezferro, line 851: Is it possible to avoid this?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -1111,15 +1116,28 @@ private void orderUsefulPredicates(Optimizable optTable,
         ArrayList<Predicate> predsForNewInList = null;
     
         if (inlistQualified) {
-          // Only combine multiple IN lists if not using Spark.
-          // Adding extra RDDs and unioning them together hinders performance.
-          if (inlistPreds.size() > 1) {
+          // In DB2 mode, probe keys can be overlapping with eachother.
+          // On HBase the overlapping keys are merged together, but on
+          // mem they aren't causing duplicate rows.  So here we avoid
```
dgomezferro, line 1121: Can we fix this on Mem by merging those keys? I dislike introducing knowledge of the platform in the engine itself.

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: This is a good point you bring up.  The intent of this code is to disallow any IN list predicates to be used for multiprobing after the first usefulPredicate which is or type varchar, whether it be in an equality predicate or an in list predicate.  However, usefulCount is not the right position to use.

Just for clarity, let's look at a test case.  Suppose we have this test case:
> create table t (v varchar(14), c char(2));
> insert into t values ('SBVGCCC','  ');
> insert into t values ('SBVGCCC c ','  ');
> insert into t values ('SBVGCCC','A ');
> create index ti on t (v,c);
> select * from t a --splice-properties index=ti
>         where v in ('SBVGCCC    ', 'G')
>         and c in ( ' ', CAST('A' AS CHAR(2)));
We will build 4 probe keys for this.  Two of these are:
Probe Number         Start Key                                    Stop Key
--------------------        -------------                                    ------------
          1                     'SBVGCCC', '  '                           'SBVGCCC       ', '  '
          2                     'SBVGCCC', 'A '                           'SBVGCCC       ', 'A  '

Since the first probe value is a varchar, its start and stop keys must cover a range of values, the start key having zero trailing spaces, and the stop key having the maximum number of trailing spaces to fit the field width.  That means that for all other matching values in between (e.g., 'SBVGCCC ','SBVGCCC  ', 'SBVGCCC   ' ...), for those we also match all values for column c.  The same is true whether we use 'v =' or 'v in'.  However, we could have a third column, i, and have an index:  create index ti2 on t (v,c,i);  with predicates v = 'SBVGCCC    ' and c = 'A ' and i in (1,2,3).  In this case we don't want to disallow use of the column c predicate in the IndexScan because equality predicates are marked as qualifiers, so even if more values than c = 'A ' could initially be scanned, applying the qualifier filters those values out, and this predicate could still be useful as a start key.  Multiprobe IN list predicates don't currently use qualifiers.  I will fix up the code to make the correct usefulCount adjustment and see what other adjustments are appropriate.  Keep in mind, this only affects the mem platform, so isn't something that should affect customers.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: It's possible, but may take a bit of effort for not much performance benefit.  It's not actually merging that could be done on mem, but expansion, since each probe value is applied independently.  For example, if we have t1 (a varchar(101), b int);  create index ix1 on t1 (a, b);  select * from t1 where a in ('a', 'b', 'c') and b in (1,2,3);  Each probe value for column a would have to be chopped up into 100 probe values to cover each possible number of trailing spaces.  For this case that would be 900 total probe values.  It would likely be more efficient to just use keys from column a resulting in 3 start/stop keys, and apply the column b predicate as a filter.  There really is a difference in how mem does probing vs. HBase.  HBase does a straight scan through rows in a single pass, while mem does actual probing where each probe is done independently.  If we don't want the optimizer to be aware of the platform, we would either have to limit which predicates we use on HBase to match mem, or we could change mem to do a straight scan through in a single thread.  Keep in mind that this is only for DB2 compatibility mode and only mem platform that customers never see, which we are talking about optimizing.  I would argue that the implementation effort compared to possible benefit isn't worth it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: See my comment below.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
dgomezferro: After a discussion with @msirek we thought it better to open a JIRA to address this at a later time and keep this optimization here for now.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MultiProbeDerbyScanInformation.java
msirek: OK, I've refactored this into a function.

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DB2VarcharCompatibilityIT.java
msirek: OK, I've fixed the indentation.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: @ascend1 I've reworked the logic to simply avoid adding the problematic IN list predicates to usefulPredicates, so the adjustment on usefulCount is removed.  I found problem cases on HBase too, so I've removed the distinction between mem platform and non-mem.

msirek:
@dgomezferro I've found that HBase has problematic cases as well as mem, they just didn't manifest the same way.  My understanding of the issue was incorrect.  The varchar key at the first level, expanded to a range condition, will also return unwanted values for a predicate on the 2nd index column, etc.  So I cannot keep the optimization in place, and there is no longer a difference between mem and HBase.  I will open a Jira for optimizations though.

msirek:

msirek:
@dgomezferro I opened [DB-10933](https://splicemachine.atlassian.net/browse/DB-10933) for the index enhancement.

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2048/

msirek:

spliceci:
TEST SUCCEEDED +1

mem_sql/src/main/java/com/splicemachine/derby/lifecycle/MemDatabase.java
```diff
@@ -59,6 +59,7 @@ public static void main(String...args) throws Exception{
             System.setProperty("com.splicemachine.enableLegacyAsserts",Boolean.TRUE.toString());
             SanityManager.DEBUG_SET("DumpOptimizedTree");
         }
+        System.setProperty("com.splicemachine.isMemPlatform",Boolean.TRUE.toString());
```
dgomezferro, line 62: Remove this since we don't rely on it anymore.

dgomezferro (changes requested):
Change looks good to me , thanks @msirek ! I just have a small request to remove an unneeded line.

mem_sql/src/main/java/com/splicemachine/derby/lifecycle/MemDatabase.java
msirek: Thanks @dgomezferro , I've removed the line.

dgomezferro (approved):

carolp-503 (approved):
+1

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2066/

jpanko1:

spliceci:
TEST SUCCEEDED +1

PR #4747, opened 2020-12-01
DB-10926 Support timestamp(LONG VARCHAR) 

arnaud-splice:


arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -55,154 +55,101 @@
...
+        TypeId rightTypeId = rightOperand.getTypeId();
+
+        if (! leftTypeId.isStringTypeId() && rightTypeId.isStringTypeId())
+        {
+            rightOperand = addCastNodeForStringToNonStringComparison(leftOperand, rightOperand);
+            rightTypeId = rightOperand.getTypeId();
+        }
+        else if (! rightTypeId.isStringTypeId() && leftTypeId.isStringTypeId())
+        {
+            leftOperand.addCastNodeForStringToNonStringComparison(rightOperand, leftOperand);
```
martinrupp, line 150: why is this line not the equivalent of 145? i would have expected sth like
`leftOperand = addCastNodeForStringToNonStringComparison(rightOperand, leftOperand);`

nice pulling out of a function!

hatyo: Also the result is ignored. Probably a bug?

spliceci:
TEST SUCCEEDED +1

PR #4751, opened 2020-12-01
DB-10919 fix DEC conversion null in Spark for getExecRowDefinition (3.1)

martinrupp:
DecimalUtil.getDecimalDb2 needs to return a SQL null, not a Java null.
The projection (and therefore also the getDecimalDb2 function) will be called once from Olap Server with
a"null" row from getExecRowDefinition to get the types in the result rows.
Then later, the Spark Executor will get the correct results and insert the correct values.


carolp-503 (approved):
Approved for 3.1

PR #4776, opened 2020-12-03
DB-10896 Keep track of compacted files in HFiles for crash recovery

dgomezferro:


dgomezferro:

dgomezferro:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2081/

dgomezferro:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -81,7 +83,8 @@
  *
  */
 public class SpliceDefaultCompactor extends DefaultCompactor {
-    private static final boolean allowSpark = true;
+    public static boolean allowSpark = true;
```
arnaud-splice, line 86: Could you make this a construction parameter of SpliceDefaultCompactor instead? 
This non final `allowSpark` only for the sake of tests is a bit scary in my opinion.

hbase_sql/src/test/java/org/apache/hadoop/hbase/regionserver/TestCleanupCompactedFileAfterFailover.java
```diff
@@ -0,0 +1,281 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class TestCleanupCompactedFileAfterFailover {
```
arnaud-splice, line 53: Can you rename this to `CleanupCompactedFileAfterFailoverTest` instead to adhere to our naming conventions?

arnaud-splice (changes requested):
Awesome change! Two small comments to address (one of them linked to the spotbug error) before we can merge that.

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
```diff
@@ -92,6 +97,45 @@
     protected boolean authTokenEnabled;
     protected Optional<RegionObserver> optionalRegionObserver = Optional.empty();
 
+    /**
+     * We need to keep track of compacted files in HFiles in order to avoid forgetting about any of them when
+     * recovering from a crash, since the compaction marker might have been lost due to a WAL roll or another crash
+     * See DB-10896 and HBASE-20724 for more details
+     */
+    @Override
+    public void postOpen(ObserverContext<RegionCoprocessorEnvironment> c) {
```
msirek, line 106: Awesome fix!   So glad this issue is finally understood.  Do you think we need to do something similar in paths that don't use snapshot isolation, for example, table and schema backup, so that the backup doesn't include the compacted store files?

jyuanca (approved):

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
dgomezferro: We can't make it a constructor parameter because the constructor is called directly by HBase, but I've made it a private non-static final field that gets initialized in the Constructor from the Configuration object.

hbase_sql/src/test/java/org/apache/hadoop/hbase/regionserver/TestCleanupCompactedFileAfterFailover.java
dgomezferro: Good point, made the change.

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
dgomezferro: Good point @msirek , we are already doing it for Full backups (see https://github.com/splicemachine/spliceengine-ee/blob/master/splice_backup/src/main/java/com/splicemachine/backup/HBaseBackupUtils.java#L384 ) but maybe we are missing some cases.
@jyuanca do we need the same thing for Incremental backups and maybe table/schema backups?

dgomezferro:

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
arnaud-splice: Neat!

arnaud-splice (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
Love the unit test - thank you!

+1

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/SIObserver.java
jyuanca: @dgomezferro Database/table/schema backup excludes all compacted HFiles. If we can make sure compacted HFiles do not come alive after an HBase restart and recover, backup would not copy these files.

PR #4777, opened 2020-12-04
DB-10949 Support parameterized FROM FINAL TABLE queries.

msirek:
Example of query this PR adds support for:
> prepare ps as 'select * from old table (update t1 set a = ?) where a >= ?';
> execute ps using 'values (2, 1)';

The parameter list is normally one per activation, but FROM TABLE queries use two activations, so this change makes sure both activations can access the same parameter list.

Two new system procedures are also added: SYSCS_GET_TABLE_COUNT and SYSCS_IS_MEM_PLATFORM

**SYSCS_GET_TABLE_COUNT** 
Returns the number of HBase user tables, or on mem, the number of tables allocated in MPartitionFactory.Admin.partitionMap.

**SYSCS_IS_MEM_PLATFORM**
Returns true if splice is running on the mem platform, otherwise false.

> splice> CALL SYSCS_UTIL.SYSCS_GET_TABLE_COUNT();
> NumTables  
> //-----------
> 111        
> 
> 1 row selected
> ELAPSED TIME = 337 milliseconds
> splice> call syscs_util.syscs_is_mem_platform();
> IsMe&
> //-----
> false

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
@@ -49,6 +49,7 @@ public class FromFinalTableIT extends SpliceUnitTest {
     
     private Boolean useSpark;
     private static int numTables = 0;
+    private static boolean isMemPlatform = false;
     
     @Parameterized.Parameters
     public static Collection<Object[]> data() {
@@ -134,6 +135,7 @@ public class FromFinalTableIT extends SpliceUnitTest {
 
     @BeforeClass
     public static void recordNumTables() throws Exception {
+        isMemPlatform = isMemPlatform();
         dropObjects();
         vacuum();
         numTables = getNumTables();
@@ -144,8 +146,11 @@ public class FromFinalTableIT extends SpliceUnitTest {
         dropObjects();
         vacuum();
         int newNumTables = getNumTables();
-        assertEquals("\nStarted with " + numTables + " tables and ended with " + newNumTables,
-                     numTables, newNumTables);
+        // Mem platform doesn't physically remove dropped tables, so
+        // this is an HBase-only check.
+        if (!isMemPlatform)
+            assertEquals("\nStarted with " + numTables + " tables and ended with " + newNumTables,
+                         numTables, newNumTables);
     }
 
     @Before
@@ -162,13 +167,9 @@ public class FromFinalTableIT extends SpliceUnitTest {
 
     public static void
     vacuum() throws Exception{
-        try {
+        // Mem doesn't have vacuum.
+        if (!isMemPlatform)
             spliceClassWatcher.executeUpdate("CALL SYSCS_UTIL.VACUUM()");
-        }
-        catch (SQLException e) {
-            // Don't fail on the mem platform.
-        }
-        return;
     }
 
     public static int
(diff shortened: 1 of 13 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2082/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2083/

dgomezferro (approved):

dgomezferro:
@msirek is it possible the IT is leaking transactions or conglomerates? I think that's what's breaking `com.splicemachine.derby.impl.HTablePrioritiesIT#testTablesPriority()` (it expects 0 user tables, Vacuum should have cleaned up all older ones)

msirek:
> @msirek is it possible the IT is leaking transactions or conglomerates? I think that's what's breaking `com.splicemachine.derby.impl.HTablePrioritiesIT#testTablesPriority()` (it expects 0 user tables, Vacuum should have cleaned up all older ones)

@dgomezferro Thanks for the tip.  I will look into the issue from this perspective.

msirek:
@dgomezferro I added a new system procedure to count the number of HBase tables and check in the IT whether the count is the same before and after the test.  There was a conglomerate leak, fixed by a one-line change.  I also fixed problems with executing a prepared FROM FINAL TABLE statement multiple times.

dgomezferro:

dgomezferro (approved):

dgomezferro:
That's great @msirek , thanks!

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2091/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2092/

msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
```diff
@@ -44,6 +48,8 @@
 public class FromFinalTableIT extends SpliceUnitTest {
```
martinrupp, line 48: i'm not exactly sure why you don't have to specify `@Category(value = {SerialTest.class})` here. but looks like this already runs only in serial. otherwise the assert in line 153 with checkNumTables() would cause sporadics. Maybe you want to add the SerialTest.class as a precaution still?

arnaud-splice (approved):

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
msirek: Thanks.  Maybe it's not showing for you, but line 46 has this:
> @Category(value = {SerialTest.class})

carolp-503 (approved):
+1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/FromFinalTableIT.java
martinrupp: ah thanks, i missed that it was retracted by github

PR #4781, opened 2020-12-04
DB-10817 Fix plan structure for non-flattenable nested joins

ascend1:


ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2087/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TableOperatorNode.java
```diff
@@ -529,6 +529,19 @@ protected FromTable getFromTableByName(String name,String schemaName,boolean exa
...
+         * This has nothing to do with flattening, but to build a PRN on top of each
+         * nested joins. This is necessary because JoinConditionVisitor may add extra
+         * hash key columns into children's result column lists. If any child is a
+         * JoinNode, not PRN, then it breaks the equation that
+         * resultColumns.size() = leftResultSet.resultColumns.size() + rightResultSet.resultColumns.size(),
+         * which is the assumption in generated code.
+         */
+        if (!isFlattenableJoinNode()) {
+            leftResultSet.notFlattenableJoin();
+            rightResultSet.notFlattenableJoin();
```
msirek, line 543: It might be a good idea to introduce a new flag, such as 'needsProjectRestrict' so we don't overload flattenableJoin.  Is my understanding correct?  We can still flatten these child JoinNodes, but we just need to mark them so a ProjectRestrict is built?

ascend1: Hmm, I'm not sure if we will flatten the child JoinNodes in this case. Here is my understanding from the source code:

`FromSubquery` and `JoinNode`s flattening happens in `SelectNode`'s `preprocess()`, when calling `fromList.flattenFromTables(...)`. So at this time, join order is the one specified in query, no reordered.

Looking into `flattenFromTables`, we see that it iterates over each `FromTable` in `fromList`. If a `FromTable` is a `JoinNode` that can be flattened, its children will be put into `fromList` and `flattened` is set to true. The outer `while` loop continues, meaning we will try to flatten the nested join.

However, if the `JoinNode` is not flattenable, then `flattened` will not be set, and the outer loop terminates. That means all nested loops do not have a chance to be considered. Basically, that means if a `JoinNode` is not flattenable, the whole sub-tree is not flattenable. Another clue about this is the implementation of `notFlattenableJoin`. This method is actually recursive. Calling `leftResultSet.notFlattenableJoin()` sets all `JoinNode`s in the sub-tree to be non-flattenable.

To me, the question is why we need this flag. If a join cannot be flattened, then all nested joins in the sub-tree doesn't participate join ordering? But I see that we maintain dependencies between joins, too. If that's the case, shouldn't all joins participate join ordering and dependencies prevent some illegal order? Do you have a clue on this Mark?

ascend1:

spliceci:
TEST SUCCEEDED +1

martinrupp (approved):

hatyo (approved):
LGTM

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/TableOperatorNode.java
msirek: @ascend1 I'm not sure, unless a JoinNode can show up in multiple FromLists, or maybe the code author just wanted the nodes to all be consistent (indicate that nothing was flattened).  In any case, the code you added, I see the same code in JoinNode.java, so this is the current convention for marking the tree, so is consistent.  Also, I cannot find a test case where the recursive marking of notFlattenableJoin makes a difference.

msirek (approved):

carolp-503 (approved):
+1

PR #4789, opened 2020-12-04
DB-10963 Add timeouts to connection.close() and call abort() on timeout

dgomezferro:


dgomezferro:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -160,21 +173,61 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
         return ps;
     }
 
+    /**
+     * Try closing connections gracefully from different threads, if that takes long it could be some queries are stuck,
+     * so abort the connections forcefully instead
+     */
     private void closeConnections() {
+        int numThreads = connections.size() + 2; // two extra for the .abort() call
+        ExecutorService executor = Executors.newFixedThreadPool(numThreads,
```
hatyo, line 182: The thread pool is shared right? according to the doc:
> public static ExecutorService newFixedThreadPool(int nThreads)
> ... At any point, at most nThreads threads will be active processing tasks. If additional tasks are submitted when all threads are active, they will wait in the queue until a thread is available.

If we run Test suites in parallel which use `SpliceWatcher`, could it lead to competition between different suites calling `closeConnection` on parallel and causing some of them to time out (because of line 198)?

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -160,21 +173,61 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
...
-                if (connection != null && !connection.isClosed()) {
-                    try {
-                        connection.close();
-                    } catch (SQLException e) {
-                        connection.rollback();
+            connections.stream().forEach(connection -> completionService.submit(() -> {
+                try {
+                    if (connection != null && !connection.isClosed())
                         connection.close();
+                } catch (SQLException e) {
```
hatyo, line 191: Does it make sense to at least log this exception?

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceAdmin_OperationsIT.java
```diff
@@ -239,11 +246,9 @@ public void testKillLongRunningQuery(boolean useSpark) throws Exception {
...
             @Override
             public void run() {
-                PreparedStatement ps = null;
-                try (TestConnection connection = methodWatcher.createConnection()) {
-                    ps = connection.prepareStatement(sql);
-                    ResultSet rs = ps.executeQuery();
-                    assertTrue(rs.next());
+                try (TestConnection connection = methodWatcher.createConnection();
+                     PreparedStatement ps = connection.prepareStatement(sql);
+                     ResultSet rs = ps.executeQuery()) {
```
hatyo, line 251: we were verifying previously that the result set contained at least one row, this is removed now, is this intentional?

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -160,21 +173,61 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
...
+                }
+                return null;
+            }));
+
+            int futures = 0;
+            while(futures++ < connections.size()) {
+                Future<Void> future = completionService.poll(10, TimeUnit.SECONDS);
```
martinrupp, line 198: (minor comment) this would wait 10 second not for each, but 10 second every time you call it, so the first one is for 10s, the second 20s, third 30s etc.. Maybe sth like
`Future<Void> future = completionService.poll(futures  == 0 ? 10 : 1, TimeUnit.SECONDS);`

I'm not entirely sure you need a CompletionService here at all, maybe a simple
`ExecutorService.invokeAll(Callable, timeout, TimeUnit)` would work.

msirek (approved):

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
dgomezferro: It wasn't shared, but I think it's better if it is. I made the Executor a cachedThreadPool without a thread limit, so resources should be enough for concurrent ITs.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
dgomezferro: Good point, made the change.

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceAdmin_OperationsIT.java
dgomezferro: Yes, this query was never completing in the first place (it's a long running query on purpose) and it was running in a separate thread, so even if the assertion failed it wouldn't have failed the test.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
dgomezferro: I like the invokeAll approach, we need to check the futures afterwards to see if they are done or not, but I think it's cleaner.

dgomezferro:
@hatyo @martinrupp Thank you for your review! I've addressed your comments, please take another look.

splice_machine/src/test/java/com/splicemachine/derby/test/framework/SpliceWatcher.java
```diff
@@ -51,6 +54,8 @@
 public class SpliceWatcher extends TestWatcher implements AutoCloseable {
 
     private static final Logger LOG = Logger.getLogger(SpliceWatcher.class);
+    private static ExecutorService executor = Executors.newCachedThreadPool(
+            new ThreadFactoryBuilder().setDaemon(true).setNameFormat("connection-close-%d").build());
```
martinrupp, line 58: wouldn't this cause problems when running tests in parallel?

dgomezferro: I don't think so, each task would get it's own thread and we are not limiting the amount of threads, so there should be no conflict.

martinrupp: right, ok!

martinrupp (approved):

dgomezferro:

spliceci:
TEST SUCCEEDED +1

PR #4807, opened 2020-12-08
DB-10579 Use truly nested transactions for triggers.

msirek:
This fixes trigger rollback cases where data is not fully rolled back upon an error, or when we exceed the 1 million row OLTP trigger limit, so roll back the work and start over from scratch on Spark.

Previously, nested triggers (one nested inside the other) shared the same parent transaction, that of the top-level DML statement.
Now, each nested trigger will use the transaction of the containing trigger as the parent transaction of the child transaction it creates and uses for its own execution.
This allows us to wait until the very end of each operation to commit it, since child transactions can see rows written by their active parent transaction.  Previously we had to commit before firing triggers, which makes rollback problematic.


msirek:

msirek:

msirek:

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2108/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2110/

msirek:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):
Change looks good to me, thanks @msirek ! 
I believe the change is somewhat risky though, I'd be more comfortable if we only merge it to master and run regression tests on it (general and focused on problematic areas like triggers)

msirek:
> Change looks good to me, thanks @msirek !
> I believe the change is somewhat risky though, I'd be more comfortable if we only merge it to master and run regression tests on it (general and focused on problematic areas like triggers)

@dgomezferro Thanks for the review.  I have the same feeling, this is not a small change and could use a lot more testing.  I will close the PRs for 3.0 and 3.1.

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/TransactionController.java
```diff
@@ -804,6 +804,52 @@ TransactionController startNestedUserTransaction(
 			boolean flush_log_on_xact_end)
         throws StandardException;
 
+    /**
+     * Get a nested internal transaction.
+     * <p>
+     * A nested internal transaction is used in place of a user transaction
+	 * when a DML statement is executing its operation tree and a new writable
+	 * child transaction needs to be created with the possibility that within
```
ascend1, line 812: Indentation is weird in this newly added block. Comment lines start to shift to the right and method parameters are not aligned.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/DMLWriteOperation.java
```diff
@@ -179,15 +182,44 @@ protected TxnView getTransactionForWrite(DataSetProcessor dsp) throws StandardEx
...
+                    Txn childTxn = (Txn) ((BaseSpliceTransaction) rawStoreXact).getActiveStateTxn();
+                    nestedTxn = childTxn;
+                    txn = (TxnView) childTxn;
+                }
+                else {
+                    nestedTxn =
                     SIDriver.driver().lifecycleManager().beginChildTransaction(
-                            parent,
-                            parent.getIsolationLevel(),
-                            parent.isAdditive(),
```
ascend1, line 186: I don't quite understand this branch. It seems right in case of `!needsfullyNestedTransaction`. But if it's `needsfullyNestedTransaction && lcc.hasNestedTransaction()`, I assume that's because some other write operation enters the branch above earlier? In that case, what does the code mean in this branch? Does it start a child transaction in a different way and push into a different stack? But what happens if a third write operation comes?

Sorry this might be a stupid question. I haven't worked on transactions in Splice yet.

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/SpliceTransactionFactory.java
```diff
@@ -119,16 +119,27 @@ public Transaction startTransaction(HBaseStore hbaseStore,ContextManager context
      * @param hbaseStore the hbase store relevant to the transaction
      * @param contextMgr the context manager
      * @param parentTxn  the parent transaction
+	 * @param destinationTable The byte representation of the conglomerate
+	 *                         number of the target table as a String.
+	 * @param inMemoryTxn If true, attempt to begin the child transaction
+	 *                    as an in-memory transaction.  This is only
+	 *                    applicable to non-Spark queries.
```
ascend1, line 126: Also weird indentation here.

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/TransactionController.java
msirek: Thanks, I've fixed the indentation in the whole file.

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/SpliceTransactionFactory.java
msirek: I've removed all tabs in this file.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/DMLWriteOperation.java
msirek: @ascend1 Not a stupid question at all.  The first time around, lcc.pushNestedTransaction needs to be called to create a SpliceInternalTransactionManager instead of the default transaction manager.  This special transaction manager has the ability to do full nesting of transactions.  The first time around, we don't need to create a nested transaction, we are just using the main transaction that was elevated to writable on line 85 of QueryJob.java:
    TxnView parent = root.getCurrentTransaction();

Every time after this that we create a new child transaction (because we have a new nested operation) and this transacction is pushed onto the transaction manager's transaction stack at line 220:

internalTransaction.pushInternalTransaction((Txn) txn); 

So, we can go to the else branch once the transaction manager that supports fully nested transactions is set up, and we only create the child transaction on the 2nd, 3rd, 4th, etc. nested operations.

Maybe the method name hasNestedTransaction is causing the confusion.  Do you think I should change the name to something like hasSpliceInternalTransactionManager() ?

splice_machine/src/main/java/com/splicemachine/derby/stream/control/ControlDataSetWriter.java
```diff
@@ -64,26 +76,59 @@ public ControlDataSetWriter(ControlDataSet<ExecRow> dataSet, AbstractPipelineWri
     @Override
     public DataSet<ExecRow> write() throws StandardException{
         SpliceOperation operation=operationContext.getOperation();
-
+        LanguageConnectionContext lcc =
+            operationContext.getActivation().getLanguageConnectionContext();
         try{
-            boolean inMemoryTxn = !operation.isOlapServer();
+            boolean inMemoryTxn = !operation.isOlapServer() && !SpliceClient.isClient();
```
arnaud-splice, line 82: Why is `!SpliceClient.isClient` needed here? (And why wasn't it needed before this PR?)

jyuanca (approved):

carolp-503 (approved):
+1

arnaud-splice (approved):

PR #4810, opened 2020-12-08
DB-10819 Fix inconsistent results with substr between spark and native spark (3.1)

arnaud-splice:


carolp-503 (approved):
Approved for 3.1

PR #4811, opened 2020-12-08
DB-10579 Use truly nested transactions for triggers (3.1)

msirek:


msirek:
Merging this master only at this time so there can be more testing.

PR #4812, opened 2020-12-08
DB-10579 Use truly nested transactions for triggers (3.0)

msirek:


carolp-503:
Not required for 3.0.  Please don't merge without further discussion.

msirek:
Merging this to master only at this time so there can be more testing.

PR #4817, opened 2020-12-09
DB-10976 Add SYSCAT.INDEXCOLUSE system view

ascend1:


ascend1:

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -351,6 +351,25 @@ public void createKeyColumnUseViewInSysIBM(TransactionController tc) throws Stan
...
+        // drop it if it exists
+        if (td != null) {
+            ViewDescriptor vd = getViewDescriptor(td);
+
+            // drop the view deifnition
+            dropAllColumnDescriptors(td.getUUID(), tc);
+            dropViewDescriptor(vd, tc);
+            dropTableDescriptor(td, sysCatSchemaDesc, tc);
+        }
+
```
jyuanca, line 366: If it already exists, skip dropping and recreating the view.

jyuanca (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
jyuanca: I accidentally approved it. Changing to "request cahnges"

jyuanca (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
ascend1: Hi Jun, what do you mean by "recreating the view"? I mean, if the existing one is not dropped, we cannot create the view with the same name (`createOneSystemView` fails). Or do I miss anything here?

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
ascend1: Hi Jun, as we talked previously, I tested doing an upgrade from 3.0.0.1989 to 3.2.0.1989 locally. Here is what I did:

1. compile DB-10976-3.0 branch and start a cdh6.3.0 cluster
2. check SYSCAT.INDEXCOLUSE is there and query on it works (108 rows in fresh start state)
3. copy platform_it/target to another place
4. compile DB-10976 branch and start a cdh6.3.0 cluster, verify its version and the existence of SYSCAT_INDEXCOLUSE
5. shutdown the cluster and replace platform_it/target with the previously copied 3.0.0.1989 version
6. start cdh6.3.0 cluster again using -b on the 3.0 persistent files
7. cluster started normally and verify the existence of SYSCAT_INDEXCOLUSE , query on it should return 108 rows

ascend1:

jyuanca (approved):

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -259,15 +259,15 @@ public String indexType()
 	/** @see IndexDescriptor#isAscending */
 	public boolean isAscending(Integer keyColumnPosition) {
 		int i = keyColumnPosition - 1;
-		if (i < 0 || i >= baseColumnPositions.length)
+		if (i < 0 || i >= isAscending.length)
 			return false;
```
msirek, line 263: The code looks good.  I'm just wondering in the base code why both isAscending and isDescending return false for an out-of-range column.  I would assume that since ascending is the default, an out-of-range column should return true (unless we want to catch the out-of-range condition and throw an exception.

msirek (approved):

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
ascend1: Indeed, that's a good point. I'll take a look at the call sites of these two functions and file a Jira if there is no special logic relying on the false return values. Probably we should throw than returning a default to prevent subtle bugs.

carolp-503 (approved):
+1

PR #4822, opened 2020-12-10
DB-10728 Fix maxExecutorCores calculation on clusters.

msirek:
SPLICE-2289 reads spark properties from the wrong location on cluster configurations, causing the estimated number of parallel tasks to always be 1, which can negatively affect costing of join plans for OLAP queries.

explain select * from lineitem: 

> Cursor(n=5,rows=1,updateMode=,engine=OLAP (cost))
>   ->  ScrollInsensitive(n=5,totalCost=60721.573,outputRows=1,outputHeapSize=0 B,partitions=1,**parallelTasks=1**)
>     ->  ProjectRestrict(n=4,totalCost=705.423,outputRows=1,outputHeapSize=0 B,partitions=1,**parallelTasks=1**)
>       ->  GroupBy(n=3,totalCost=705.423,outputRows=1,outputHeapSize=0 B,partitions=1,**parallelTasks=1**)
>         ->  ProjectRestrict(n=2,totalCost=11286.284,outputRows=6001215,outputHeapSize=0 B,partitions=2,**parallelTasks=1**)
>           ->  TableScan[LINEITEM(129456)](n=0,totalCost=11286.284,scannedRows=6001215,outputRows=6001215,outputHeapSize=0 B,partitions=2,**parallelTasks=1**)

This change picks up the spark and yarn properties from the proper structures.
After the fix:

> Cursor(n=5,rows=1,updateMode=,engine=OLAP (cost))
>   ->  ScrollInsensitive(n=5,totalCost=60721.573,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=16)
>     ->  ProjectRestrict(n=4,totalCost=705.423,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=16)
>       ->  GroupBy(n=3,totalCost=705.423,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=16)
>         ->  ProjectRestrict(n=2,totalCost=11286.284,outputRows=6001215,outputHeapSize=0 B,partitions=2,parallelTasks=16)
>           ->  TableScan[LINEITEM(129456)](n=0,totalCost=11286.284,scannedRows=6001215,outputRows=6001215,outputHeapSize=0 B,partitions=2,parallelTasks=16)


msirek:

spliceci:
TEST SUCCEEDED +1

hbase_storage/pom.xml
```diff
@@ -136,6 +136,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>com.cedarsoftware</groupId>
+          <artifactId>java-util</artifactId>
+          <version>1.61.0</version>
+        </dependency>
```
martinrupp, line 143: i'm not an expert here, should we add dependencies like this?

martinrupp (approved):

hbase_storage/pom.xml
dgomezferro, line 143: I'd rather not add this dependency, since we are only using .deepEquals() it should be easy to remove.

dgomezferro (changes requested):
I don't think this approach would work well for the cloud, since we don't use Yarn. If we have to parse the Yarn configuration I would do it on the "server side" (OlapServerMaster) I think it fits better there. We aren't taking into account Application Server Queues either.

hbase_storage/pom.xml
msirek: @dgomezferro OK, is it because this is not a well-known library that it's not a good idea to include?  Just trying to know the reason in case I want to include other libraries in the future.  They use the Apache license, so I'm assuming it's not a license issue.

Also, if we don't use YARN in the cloud, can you point me to the code that determines the memory available to start spark executors?  I only have one YARN config setting I'm using that's related to how much memory we can use, which determines the number of spark executors we use.

msirek: @dgomezferro OK, I've removed the usage and reference to DeepEquals.

msirek:
> I don't think this approach would work well for the cloud, since we don't use Yarn. If we have to parse the Yarn configuration I would do it on the "server side" (OlapServerMaster) I think it fits better there. We aren't taking into account Application Server Queues either.
@dgomezferro  OK,  I've opened the following Jira to handle changes for the cloud and any updates needed for Olap Server queues: https://splicemachine.atlassian.net/browse/DB-11052.  

 As for "server side" (OlapServerMaster) parsing of Yarn config, that's what this Jira is doing.  All configs are parsed in OlapServerMaster, and the values are stored in Zookeeper via a new SparkYarnConfiguration class.

msirek:
@dgomezferro Please let me know if you have any further comments for this PR and if you're OK with continuing the remaining work through [DB-11052](https://splicemachine.atlassian.net/browse/DB-11052).  Thanks!

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
```diff
@@ -68,6 +71,8 @@
     // numNodes is written to zookeeper by OlapServerMaster, so we have
     // to wait until the Olap Server comes up before getting numNodes from zookeeper.
     private static int MAX_EXECUTOR_CORES = -1;
+    // A value to use until the Olap Server comes up.
+    private static int DEFAULT_MAX_EXECUTOR_CORES = 8;
     private static int numSparkNodes = -1;
```
dgomezferro, line 76: I think this is not used, remove if so

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
```diff
@@ -265,21 +270,38 @@ public int getMaxExecutorCores() {
...
-                                        getProperty("splice.spark.executor.instances"),
-                                        getProperty("splice.spark.executor.cores"),
-                                        getProperty("splice.spark.executor.memory"),
-                                        getProperty("splice.spark.dynamicAllocation.maxExecutors"),
-                                        getProperty("splice.spark.executor.memoryOverhead"),
-                                        getProperty("splice.spark.yarn.executor.memoryOverhead"),
-                                        sparkNodes > 0 ? sparkNodes : 1);
-            if (sparkNodes > 0) {
-                numSparkNodes = sparkNodes;
+              calculateMaxExecutorCores(conf.getYarnNodemanagerResourceMemoryMB(),
```
dgomezferro, line 290: I'd rather have this function in OlapServerMaster or a class logically close to the OlapServer, I think this class shouldn't have that much knowledge about YARN. Ideally we should just read the number of executor cores from Zookeeper.

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
```diff
@@ -78,22 +83,22 @@
...
 
-    // Find the number of nodes on which Spark executors can be run.
-    private static int getNumSparkNodes() {
-        int numNodes = -1;
+    // Find the Spark and YARN configuration from Zookeeper.
+    private static byte [] getSparkYarnConfigBytes() {
+        byte [] returnData = null;
         try {
-            byte [] data = ZkUtils.getData(sparkNumNodesZkPath);
+            byte [] data = ZkUtils.getData(sparkYarnConfigZkPath);
```
dgomezferro, line 94: If it's not a lot of work make this a Protobuf message for future extendability.

hbase_storage/src/main/java/com/splicemachine/access/SparkYarnConfiguration.java
```diff
@@ -0,0 +1,213 @@
...
+
+package com.splicemachine.access;
+
+import java.io.*;
+
+/**
+ * A serializable class for storing YARN and Spark properties in Zookeeper.
+ *
+ */
+public class SparkYarnConfiguration implements Externalizable {
```
dgomezferro, line 23: If we compute the number of cores in OlapServerMaster (or close by) I don't think we need this class anymore

dgomezferro (changes requested):

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
msirek: This is needed.  I ran into problems if I made getMaxExecutorCores non-retryable.  If we call it now before ZK data is there we can use the default.

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
msirek: OK, I moved the calculation to OlapServerMaster.

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
msirek: This is an ephemeral ZK field, so we don't need to be backwards compatible.  It goes away every time zookeeper reboots.  Would it matter to use protobuf?

hbase_storage/src/main/java/com/splicemachine/access/SparkYarnConfiguration.java
msirek: Yes, it is removed.

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
dgomezferro: I meant the `numSparkNodes` field, I believe it's only written to but never used. Sorry for not being clear, I can see how my message was confusing!

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
dgomezferro: I created https://splicemachine.atlassian.net/browse/DB-11068 to change all messages in OlapServerMaster. It could matter for online upgrade.

dgomezferro (approved):

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
msirek: I see.  It might actually help for debugging to have a static field that tells us how many spark nodes there are (or how many we think there are), so if you don't mind I'd like to leave it in for now.

dgomezferro: Ok

msirek:
The Jenkins build hooks are failing for this PR for some unknown reason, but the branch does compile on my laptop.  The last IT run I did passed.

hbase_sql/src/main/java/com/splicemachine/derby/lifecycle/HEngineSqlEnv.java
msirek: OK, we can handle it through the Jira.

carolp-503 (approved):
+1

PR #4830, opened 2020-12-10
DB-10420 Push down disjunction of index expressions when possible

ascend1:


ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
```diff
@@ -521,34 +522,64 @@ public final boolean isStoreQualifier(){
...
-                        return (false);
+                    if (isOnExpression) {
+                        tempAnd.setLeftOperand(or_node.getLeftOperand());
+                        tempAnd.postBindFixup();
+                        Predicate tempPred = (Predicate) getNodeFactory().getNode(
+                                C_NodeTypes.PREDICATE,
+                                tempAnd,
+                                or_node.getLeftOperand().getTablesReferenced(),
+                                getContextManager());
+                        boolean skipProbePreds = pushPreds && optTable.getTrulyTheBestAccessPath().getJoinStrategy().isHashJoin();
```
msirek, line 561: pushPreds seems to be true when join planning is complete, and false in the middle of join planning.  If we use skipProbePreds as false while planning a hash join, then we might think an IN list predicate is useful, which may affect the cost of the join and lead us to pick it.  But then at the end of planning we would enter this code again, and see that we can't actually push the predicate, so the join is not as cheap as we thought.  Perhaps the code should be modified to detect during join planning whether the predicate can be pushed, something like:
> 
> JoinStrategy joinStrategy = pushPreds ? optTable.getTrulyTheBestAccessPath().getJoinStrategy() :
>                                                            optTable.getCurrentAccessPath().getJoinStrategy();
> boolean skipProbePreds = joinStrategy.isHashJoin();

msirek (changes requested):

dgomezferro (approved):
@ascend1 Regarding the comment in JIRA about IndexScan vs TableScan, is it easy for us to return "PKTableScan" / "RangeTableScan" or something similar when scanning a base table with start/stop keys to differentiate it from a Full table scan?

ascend1:
> @ascend1 Regarding the comment in JIRA about IndexScan vs TableScan, is it easy for us to return "PKTableScan" / "RangeTableScan" or something similar when scanning a base table with start/stop keys to differentiate it from a Full table scan?

Do you mean in the output of explain statement? I think it's doable since each predicate pushed down to a scan has flags indicating if it's a start/stop key or qualifier, or the scan has no predicate at all.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
ascend1: That makes total sense! I didn't really check this logic before but simply copy-pasted. I improve it as you suggested with a bit modification mainly because when join planning is complete, this method can also get called with `pushPreds = false`. This happens in some helper functions, e.g. `transferPredicates()`.

ascend1:

msirek (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #4833, opened 2020-12-10
DB-10999 fix MAXIMUMDISPLAYWIDTH / refactor more ij.jj

martinrupp:


martinrupp:

spliceci:
TEST SUCCEEDED +1

sqlshell.sh
```diff
@@ -9,14 +9,25 @@ else
...
+    fi
+    export MAVEN_OPTS="${MAVEN_OPTS} -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=$port"
+fi
+
+
 if [ -z "${CLIENT_SSL_KEYSTORE}" ]; then
   cd splice_machine ; ${RLWRAP} mvn exec:java ; cd ${DIR}
 else
-  cd splice_machine ; ${RLWRAP} mvn  exec:java \
+  cd splice_machine ; ${RLWRAP} MAVEN_OPTS="${MY_MAVEN_OPTS}" mvn exec:java \
```
msirek, line 30: Is there a reason to only support the new debug command-line option when we are using a CLIENT_SSL_KEYSTORE?

dgomezferro: I think this is an error, `MY_MAVEN_OPTS` is not defined

dgomezferro (changes requested):
Nice cleanup! I think there's an error using `MY_MAVEN_OPTS`, looks good otherwise

martinrupp:
> Nice cleanup! I think there's an error using `MY_MAVEN_OPTS`, looks good otherwise

yes, you're correct, i fixed it. also now works for both SSL and non-SSL

dgomezferro (approved):

msirek (approved):
+1

carolp-503 (approved):
+1

PR #4842, opened 2020-12-11
DB-11012 Truncate when casting timestamp to [var]char (3.1)

arnaud-splice:


carolp-503 (approved):
Approved for 3.1

liprais:
This should be documented and users should be altered that the timestamp cast to varchar can not be cast back if the length is not correct.

PR #4844, opened 2020-12-11
DB-10952 use reflection to ensure created system procedures are correct (3.0)

martinrupp:
removes the following functions since they are not existent:

- SYSCS_GET_AUTO_INCREMENT_ROW_LOCATIONS
- PARTITION_EXISTS
- CLOBGETPROSITIONFROMSTRING

re-adding GET_REPLICATION_PROGRESS as it was removed accidentally in DB-8990


carolp-503 (approved):
Approved for 3.0

PR #4850, opened 2020-12-11
DB-10472 Restrict sys visibility to only splicedb.splice - Other db owners can see everything in their db

arnaud-splice:


mem_sql/pom.xml
```diff
@@ -259,12 +259,10 @@
                             <key>com.splicemachine.enableLegacyAsserts</key>
                             <value>true</value>
                         </systemProperty>
-                        <!-- Uncomment the following to dump activation class files for mem platform to the mem_sql directory.
                            <systemProperty>
```
dgomezferro, line 262: Comment out again

dgomezferro (approved):

mem_sql/pom.xml
arnaud-splice: Thanks for catching that! Done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -1054,8 +1083,8 @@ public SchemaDescriptor getSchemaDescriptor(UUID dbId, String schemaName,
                 break;
             case SchemaDescriptor.IBM_SYSTEM_ADM_SCHEMA_NAME:
                 return sysIBMADMSchemaDesc;
-            case SchemaDescriptor.STD_SYSTEM_UTIL_SCHEMA_NAME:
-                return getSystemUtilSchemaDescriptor();
+            //case SchemaDescriptor.STD_SYSTEM_UTIL_SCHEMA_NAME:
+            //    return getSystemUtilSchemaDescriptor();
```
msirek, line 1087: Should the commented code be removed?

msirek (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
arnaud-splice: done

PR #4851, opened 2020-12-11
DB-10961 remove node factory part 1

martinrupp:
# Description

This is a first part of starting to remove the NodeFactory (spliceengine/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/NodeFactory.java) that is currently used by our SQL Parser to create the nodes from the Syntax Tree (sqlfgrammar.jj).

# Motivation

We want to remove NodeFactory and replace with simple, strongly typed constructors, because:

1. Node Factory is quite complicated code

2. we don't have safety when constructing the nodes. Potential errors that will ONLY be detected at run time include

    1. calling getNode with wrong number of parameters. will result (at run time) in a throw in SanityManager.DEBUG, otherwise “nothing”, so Node will not be initialized)

    2. calling getNode with wrong TYPE of parameters: will result (at run time) in java.lang.ClassCastException.

    3. calling getNode, but forgot to add type to the getNode switch: will result (at run time) in a Not Implemented exception

    4. typo in C_NodeNames.XXX_NAME : will result (at run time) in ClassNotFound Exceptions

3. refactoring is very hard, as e.g. Intellij can’t help you add a parameter to all invocations. If you rename a class, you might forget rename the name in C_NodeNames.XXX_NAME.

4. adding nodes is complicated and error-prone.

For all those reasons, Derby also decided to remove the node factory a long time ago, see https://issues.apache.org/jira/browse/DERBY-673 .

This is part 1 of the removal of the Node Factory. We decided to remove it in multiple steps to make this manageable and e.g. avoid one huge change that might cause a lot of merge conflicts. I suppose we have 2-4 of these changes.

# How to test

The new code is only a refactoring and doesn’t add or remove any features or fix any bugs.

# What the old code looks like

Currently, we create nodes like this:

```
        StatementNode retval =
                (StatementNode) nodeFactory.getNode(
                        C_NodeTypes.UPDATE_NODE,
                        tableName, /* target table for update */
                        resultSet, /* SelectNode just created */
                        getContextManager());
```

which then calls this code:

`Node retval = getNode(nodeType, cm);`

Where getNode will call basically a large switch statement mapping C_NodeTypes.UPDATE_NODE to C_NodeNames.UPDATE_NODE_NAME ("com.splicemachine.db.impl.sql.compile.UpdateNode”). This alone is already a bit dangerous, as a typo at creating or if someone adds another node will not be detected at compile time, but ONLY when running integrations tests and ONLY if the tests run exactly through that code (which they of course should, but maybe not).

Next, we getNode will pass the parameters to the node like this

`retval.init(arg1, arg2);`

To be able to do this, UpdateNode extends QueryTreeNode which implements Node, which has interfaces for all counts of parameters, e.g.

```
   @Override
    public void init(Object arg1, Object arg2) throws StandardException{
        if(SanityManager.DEBUG){
            SanityManager.THROWASSERT("Two-argument init() not implemented for "+getClass().getName());
        }
    }
```

# What the new code looks like

```
        UpdateNode node = new UpdateNode(
                tableName, /* target table for update */
                resultSet, /* SelectNode just created */
                getContextManager());
```

When we finally removed the NodeFactory, we will remove C_NodeNames.java (354 lines), SpliceNodeFactoryImpl.java (663 lines), NodeFactory (605 lines) and more code in other classes, AND get type safety at run compile time.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
@@ -38,6 +38,7 @@ import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.services.classfile.VMOpcode;
 import com.splicemachine.db.iapi.services.compiler.LocalField;
 import com.splicemachine.db.iapi.services.compiler.MethodBuilder;
+import com.splicemachine.db.iapi.services.context.ContextManager;
 import com.splicemachine.db.iapi.services.io.FormatableBitSet;
 import com.splicemachine.db.iapi.services.io.FormatableProperties;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
@@ -101,6 +102,20 @@ public class DeleteNode extends DMLModStatementNode
     private int[] colMap;
     private Boolean cursorDelete;
 
+    public DeleteNode() {}
+
+    public DeleteNode(ContextManager cm, TableName targetTableName,
+               SelectNode queryExpression, Boolean cursorDelete,
+               Properties targetProperties)
+    {
+        super.init(queryExpression);
+        setContextManager(cm);
+        setNodeType(C_NodeTypes.DELETE_NODE);
+        this.targetTableName = targetTableName;
+        this.targetProperties = targetProperties;
+        this.cursorDelete = cursorDelete;
+    }
+
     /**
      * Initializer for a DeleteNode.
      *
@@ -108,7 +123,6 @@ public class DeleteNode extends DMLModStatementNode
      * @param queryExpression    The query expression that will generate
      *                the rows to delete from the given table
      */
-
     public void init(Object targetTableName,
                      Object queryExpression,
                      Object cursorDelete,
@@ -902,9 +916,7 @@ public class DeleteNode extends DMLModStatementNode
         ((FromBaseTable) fromTable).setTableProperties(targetProperties);
 
         fromList.addFromTable(fromTable);
-        SelectNode resultSet = (SelectNode) nodeFactory.getNode(
-                                                     C_NodeTypes.SELECT_NODE,
-                                                     null,
+        SelectNode resultSet = new SelectNode( null,
                                                      null,   /* AGGREGATE list */
                                                      fromList, /* FROM list */
                                                      whereClause, /* WHERE clause */
@@ -953,9 +965,7 @@ public class DeleteNode extends DMLModStatementNode
 
         fromList.addFromTable(fromTable);
 
-        SelectNode resultSet = (SelectNode) nodeFactory.getNode(
-                                                     C_NodeTypes.SELECT_NODE,
-                                                     getSetClause(tableName, cdl),
+        SelectNode resultSet = new SelectNode(getSetClause(tableName, cdl),
                                                      null,   /* AGGREGATE list */
                                                      fromList, /* FROM list */
                                                      whereClause, /* WHERE clause */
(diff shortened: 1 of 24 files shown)


martinrupp:

martinrupp:

spliceci:
TEST SUCCEEDED +1

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2209/

spliceci:
TEST SUCCEEDED +1

msirek:
@martinrupp  What is the motivation for removing the node factory?  This abstraction layer can yield some benefits.  For example, if there are constant nodes which are created too frequently, memory could be saved by returning the same node multiple times (e.g. for constant 'true' or 'false').  Or, if we wanted to add some sort of logging or profiling of node allocations, it would be easier to do so with a node factory.

martinrupp:
> @martinrupp What is the motivation for removing the node factory?

1. It's quite complicated code, and not immediately obvious why. You don't really have safety when constructing the nodes, you can write code that only fails when you run the code, and maybe that part of the code is not hit by the test cases.
2. you neither have safety of amounts of parameters not the types.
3. you can't do any refactoring
4. adding nodes is complicated and error-prone.
For all those reasons, Derby also decided to remove the node factory a long time ago.

> For example, if there are constant nodes which are created too frequently, memory could be saved by returning the same node multiple times (e.g. for constant 'true' or 'false'). 

These nodes are not small nodes like true or false, so i don't think we would ever do that

> Or, if we wanted to add some sort of logging or profiling of node allocations, it would be easier to do so with a node factory.

We could do that in the base class of the nodes.

martinrupp:
Derby removed the NodeFactory code 7 years ago:
 https://issues.apache.org/jira/browse/DERBY-673

arnaud-splice (approved):

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2264/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DeleteNode.java
```diff
@@ -964,12 +974,7 @@ private StatementNode getEmptyUpdateNode(String schemaName,
                                                      null, /* windows */
                                                      getContextManager());
 
-        return (StatementNode) nodeFactory.getNode(
-                                                    C_NodeTypes.UPDATE_NODE,
-                                                    tableName,
-                                                    resultSet,
-                                                    getContextManager());
-
+        return new UpdateNode( tableName, resultSet, cursorDelete, getContextManager());
```
martinrupp, line 977: this is actually fixing a bug that was introduced because NodeFactory does no type checking (UpdateNode's init requires `tableName, resultSet, cursorDelete`, but would have only been given `tableName, resultSet` )

hatyo (approved):
LGTM, I see for many `Node` types you didn't remove yet the `init` method, but I think you will remove it later on as it becomes dead code anyway, right?

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2265/

martinrupp:
> LGTM, I see for many `Node` types you didn't remove yet the `init` method, but I think you will remove it later on as it becomes dead code anyway, right?

yes, this is the plan

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/108/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/107/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2268/

carolp-503 (approved):
+1

jyuanca (approved):

spliceci:
TEST SUCCEEDED +1

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/153/

PR #4853, opened 2020-12-14
DB-11019 Use the selectivity of only one join predicate per column set.

msirek:
Fixes join cardinality underestimation errors, which could lead to large poorly-performing nested loop joins in large multi-table queries.

 Simplified cardinality underestimation example:
> create table t1 ( a int, b int, c int, d int, e int );
> insert into t1 values (1,1,1,1,1),(2,2,2,2,2),(3,3,3,3,3),(4,4,4,4,4),(5,5,5,5,5),(6,6,6,6,6),(7,7,7,7,7),(8,8,8,8,8),(9,9,9,9,9),(10,10,10,10,10);
> analyze table t1;
> /* The following should have an outputRows estimate of 10, not 1. */
> explain select * from t1 a, t1 b, t1 c where a.a = b.a and b.a = c.a;
> Cursor(n=11,rows=1,updateMode=READ_ONLY (1),engine=OLTP (default))
>   ->  ScrollInsensitive(n=11,totalCost=32.552,outputRows=1,outputHeapSize=60 B,partitions=1,parallelTasks=1)
>     ->  BroadcastJoin(n=8,totalCost=20.249,outputRows=1,outputHeapSize=60 B,partitions=1,parallelTasks=1,preds=[(A.A[8:1] = C.A[8:11])])
>       ->  TableScan[T1(1680)](n=6,totalCost=4.012,scannedRows=10,outputRows=10,outputHeapSize=60 B,partitions=1,parallelTasks=1)
>       ->  BroadcastJoin(n=4,totalCost=12.135,outputRows=10,outputHeapSize=400 B,partitions=1,parallelTasks=1,preds=[(A.A[4:1] = B.A[4:6])])
>         ->  TableScan[T1(1680)](n=2,totalCost=4.012,scannedRows=10,outputRows=10,outputHeapSize=400 B,partitions=1,parallelTasks=1)
>         ->  TableScan[T1(1680)](n=0,totalCost=4.012,scannedRows=10,outputRows=10,outputHeapSize=200 B,partitions=1,parallelTasks=1)

1. When there is more than one join predicate involving the same column on a table, either explicitly, or derived through transitive closure, the selectivity of the redundant term is applied.  The old formula for selectivity used only the join predicate with the lowest selectivity as the selectivity of the entire join.  The fix is to apply a similar rule.  Among a set of columns with overlapping column references, only use the predicate with lowest selectivit.  For example:  t1.a + t1.b > t2.a and t1.a = t2.a both involve column t1.a, so only one of these predicates contributes to the join selectivity.  Non-overlapping predicate selectivities will still be combined.

2.  The formula for combining join predicate selectivities is a little too aggresive (pred1.selectivity * pred2.selectivity * pred3.selectivity...) and implies absolute independence of join column values.  This change was made by SPLICE-2289, as we only counted a single predicates selectivity before, which caused row count overestimation.  The formula is now updated to match the default scan selectivity formula, which used function computeSqrtLevel to take the square root of the combined selectivity successively, the same number of times as the number of join predicates.

3.  Sometimes join predicates are applied to the scan selectivity.  This is done for nested loop join, because it applies the join predicates as scan predicates.  But this has the effect of applying a predicate's selectivity more than once when other types of joins are picked.  The solution is to only apply join predicates to the scan selectivity if the current access path uses nested loop join.

To handle part 1, new class ReferencedColumnsMap is added to map table number to column numbers, tracking the ColumnReferences present in each predicate.  Method categorize, which collects metadata about a predicate, is updated with ReferencedColumnsMap as a parameter and builds this map for each Predicate.

For part 2, a new SelectivityHolder, named JoinPredicateSelectivity, is added to facilitate the sorting of selectivities as is done for scan predicates.

For part 3, a new method is added to the SelectivityHolder API called shouldApplySelectivity() which checks the number of tables referenced in the predicate and current join strategy.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
@@ -33,17 +33,17 @@ package com.splicemachine.db.impl.sql.compile;
 
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.compile.Optimizable;
+import com.splicemachine.db.impl.ast.RSUtils;
 
 /**
  * Predicate Selectivity Computation
  *
  */
 public class DefaultPredicateSelectivity extends AbstractSelectivityHolder {
-    private final Predicate p;
     private final Optimizable baseTable;
     private final double selectivityFactor;
     public DefaultPredicateSelectivity(Predicate p, Optimizable baseTable, QualifierPhase phase, double selectivityFactor){
-        super(false,0,phase);
+        super(false,0,phase,p);
         this.p = p;
         this.baseTable = baseTable;
         this.selectivityFactor = selectivityFactor;
@@ -57,4 +57,14 @@ public class DefaultPredicateSelectivity extends AbstractSelectivityHolder {
         }
         return selectivity;
     }
+
+    @Override
+    public boolean shouldApplySelectivity() {
+        return
+            getNumReferencedTables() < 2             ||
+            baseTable.getCurrentAccessPath() == null ||
+            RSUtils.isNLJ(baseTable.getCurrentAccessPath());
+    }
+
+    public Predicate getPredicate() { return p; }
 }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
@@ -38,9 +38,8 @@ package com.splicemachine.db.impl.sql.compile;
  * BASE - > Qualifier is going against a key value in a way where splice can perform partitioning.
  * FILTER_BASE -> Qualifier for base table but does not restrict I/O from base scan.  This is evaluated prior to a possible lookup.
  * FILTER_PROJECTION -> Qualifer applied after possible lookup in the projection node on top of the base conglomerate scan.
- * JOIN -> Join predicate used to adjust the join selectivity.
  *
  */
 public enum QualifierPhase {
-    BASE,FILTER_BASE,FILTER_PROJECTION,JOIN
+    BASE,FILTER_BASE,FILTER_PROJECTION
 }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
@@ -38,8 +38,9 @@ package com.splicemachine.db.impl.sql.compile;
  * BASE - > Qualifier is going against a key value in a way where splice can perform partitioning.
  * FILTER_BASE -> Qualifier for base table but does not restrict I/O from base scan.  This is evaluated prior to a possible lookup.
  * FILTER_PROJECTION -> Qualifer applied after possible lookup in the projection node on top of the base conglomerate scan.
+ * JOIN -> Join predicate used to adjust the join selectivity.
  *
  */
 public enum QualifierPhase {
-    BASE,FILTER_BASE,FILTER_PROJECTION
+    BASE,FILTER_BASE,FILTER_PROJECTION,JOIN
 }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
@@ -168,11 +168,12 @@ public class SelectivityUtil {
                 ReferencedColumnsMap referencedColumnsMap = p.getReferencedColumns();
 
                 Integer mapIndex = null;
-                for (Integer I:columnSet) {
-                    mapIndex = selectivityIndexMap.get(I);
-                    if (mapIndex != null)
-                        break;
-                }
+                if (columnSet != null)
+                    for (Integer I:columnSet) {
+                        mapIndex = selectivityIndexMap.get(I);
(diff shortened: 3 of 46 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2144/

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
```diff
@@ -130,14 +132,73 @@ public static double estimateJoinSelectivity(Optimizable innerTable, Conglomerat
...
+                Set<Integer> columnSet = p.getReferencedColumns().get(tableNumber);
+                ReferencedColumnsMap referencedColumnsMap = p.getReferencedColumns();
+
+                Integer mapIndex = null;
+                if (columnSet != null)
+                    for (Integer I:columnSet) {
+                        mapIndex = selectivityIndexMap.get(I);
+                        if (mapIndex != null)
+                            break;
+                    }
```
ascend1, line 176: Maybe a pair of curly braces for the if block, since it contains multiple lines.

db-engine/src/test/java/com/splicemachine/db/impl/sql/compile/ReferencedColumnsMapTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ *
+ * Test Class for ReferencedColumnsMap
+ *
+ */
+public class ReferencedColumnsMapTest {
```
martinrupp, line 29: very nice

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
```diff
@@ -38,8 +38,9 @@
  * BASE - > Qualifier is going against a key value in a way where splice can perform partitioning.
  * FILTER_BASE -> Qualifier for base table but does not restrict I/O from base scan.  This is evaluated prior to a possible lookup.
  * FILTER_PROJECTION -> Qualifer applied after possible lookup in the projection node on top of the base conglomerate scan.
+ * JOIN -> Join predicate used to adjust the join selectivity.
  *
  */
 public enum QualifierPhase {
-    BASE,FILTER_BASE,FILTER_PROJECTION
+    BASE,FILTER_BASE,FILTER_PROJECTION,JOIN
```
ascend1, line 45: One thing that could lead to some confusion later is that the JOIN phase newly added here is not used in `addPredicate()` in `ScanCostFunction`, but only used in `estimateJoinSelectivity()` in `SelectivityUtil`. For this reason, I wonder if it's necessary to have this phase. In `estimateJoinSelectivity()`, there is only one phase involved, so practically, phase doesn't really matter.

ascend1 (approved):
Great change!

arnaud-splice (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QualifierPhase.java
msirek: Good point.  I will remove it and just pass FILTER_BASE as the parameter value.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
msirek: Added braces.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
```diff
@@ -57,4 +57,14 @@ public double getSelectivity() throws StandardException {
         }
         return selectivity;
     }
+
+    @Override
+    public boolean shouldApplySelectivity() {
+        return
+            getNumReferencedTables() < 2             ||
+            baseTable.getCurrentAccessPath() == null ||
```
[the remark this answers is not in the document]
msirek: I don't know, this is more like defensive programming so that no errors occur in any state of FromTable.  For example resetAccessPaths() can set currentAccessPath to null.  If currentAccessPath could never be set to null then this condition is not needed.  But if the method assumes currentAccessPath must be non-null, it places a precondition upon calling this method, which should be documented.  Adding the check just makes life easier because there is no precondition for calling this function.

db-engine/src/test/java/com/splicemachine/db/impl/sql/compile/ReferencedColumnsMapTest.java
msirek: Thanks

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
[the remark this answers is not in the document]
ascend1: That makes sense. I debugged it a little bit and it seems `getCurrentAccessPath()` can indeed return null. Not sure if it can ever happen during cost estimation, but it seems when the best conglomerate is chosen, current is then set to null. Anyway, I agree that it safer. All good.

ascend1 (approved):

carolp-503 (approved):
+1

ipraznik-splice:
jenkins please test branch @mem,skipTestsLongerThan2Minutes

PR #4855, opened 2020-12-14
DB-11019 Use the selectivity of only one join predicate per column set (3.0)

msirek:


carolp-503 (approved):
Approved for 3.0

carolp-503 (approved):
Approved for 3.0

PR #4856, opened 2020-12-14
DB-10139 Migrate Jenkins Env (3.1)

bklo94:
Branch-3.1 specific changes so the correct can be pulled from SCM. Otherwise similar to the master branch.


carolp-503 (approved):
Approved for 3.1

PR #4862, opened 2020-12-14
DB-11036 Move platform specific code to its own submodule

dgomezferro:


OlegMazurov:
_mvn -B -Pcore,cdh6.3.0,ee -Denv=cdh6.3.0 clean install_
does not work any more:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:copy-dependencies (copy) on project platform_it: Artifact has not been packaged yet. When used on reactor artifact, copy should be executed after packaging: see MDEP-187. -> [Help 1]

dgomezferro:
@OlegMazurov Thanks for checking it out! That should be fixed, can you have another look?

martinrupp (approved):
tested and works much better in Intellij now. Switching profiles works now (a thing i always recommended against) and also the dbaas3.0 profile works, which always had some wrongly interpreted dependencies. This is really really great, thank you!

bklo94 (approved):
Looks good to me for the assembly and pom.xml changes

carolp-503 (approved):
+1

dgomezferro:
Converting to draft to prevent merging, there are some IT failures in the long IT run

PR #4863, opened 2020-12-15
DB-10773 Honor joinStrategy hints on the target table of an UPDATE.

msirek:
This allows joinStragegy query hints to be applied to the target table of an UPDATE, for example, if there is a join predicate on the primary key and you'd like to use nested loop join to avoid scanning the entire table:
```
explain update t4 --splice-properties joinStrategy=nestedloop
                             set (a4,b4) = (select 1,1 from t1 where c1 = a4);
```
Sometimes such hints may be inside a trigger, but often the optimizer cost estimates are so low on triggers that the parsing time limit is set to less than a millisecond, and the query planner gives up before even considering the hinted join.  In general prepared statements such as triggers are recompiled infrequently, so it is OK if parsing uses a little extra time, especially if it means that will allow more optimal plans and plans including all query hints to get built.

A fix is added to give extra allowed parsing time to prepared statements.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
@@ -84,6 +84,7 @@ public class OptimizerImpl implements Optimizer{
     private static final int JUMPING=2;
     private static final int WALK_HIGH=3;
     private static final int WALK_LOW=4;
+    private static final int PREPARED_STATEMENT_TIME_LIMIT_LOWER_BOUND = 2000;
     protected DataDictionary dDictionary;
     /* The number of tables in the query as a whole.  (Size of bit maps.) */
     protected final int numTablesInQuery;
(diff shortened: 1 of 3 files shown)


msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2178,6 +2196,17 @@ private void rememberBestCost(CostEstimate currentCost,int planType) throws Stan
...
+
+            // Triggers and other prepared statements may have a very low
+            // cost estimate because the size of the involved trigger rows
+            // is not known up-front.  We don't want to quit query compilation
+            // early, resulting in a bad query plan or a plan which fails to
+            // pick up query hints, especially since the
+            // prepared statement does not need to be compiled on every
+            // invocation, just once in a while.  So, it is time well
+            // spent if it means we can compile a more optimal plan.
+            if (timeLimit < 2000 && isPreparedStatement())
```
dgomezferro, line 2208: Nitpick, can you make it a static constant instead of a hardcoded number?

dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: Done.

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4874, opened 2020-12-16
DB-11008 Introduce EXPORT and CAST parameter to set floating point notation (3.0)

arnaud-splice:


carolp-503 (approved):
Approved for 3.0

PR #4890, opened 2020-12-17
DB-11063 Left outer broadcast join for more cases.

msirek:
Currently the Splice optimizer only allows nested loop join to perform left outer join with join predicates that don't reference left table columns, for example:
> select * from t1 left outer join t2  
> on t2.b = 1;

This Jira adds support for left outer join to use broadcast join when each join predicate only references columns from a single table (or perhaps from neither table).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
@@ -167,7 +167,8 @@ public abstract class HashableJoinStrategy extends BaseJoinStrategy {
         if (hashKeyColumns == null && skipKeyCheck) {
             // For full outer join, broadcast join is the default join strategy that can be applied
             // to any kinds of join predicate, so make the eligibility check more general
-            if (outerCost.getJoinType() == JoinNode.FULLOUTERJOIN &&
+            if ((outerCost.getJoinType() == JoinNode.FULLOUTERJOIN ||
+                 outerCost.getJoinType() == JoinNode.LEFTOUTERJOIN) &&
                 (innerTable.isMaterializable() ||
                  innerTable.supportsMultipleInstantiations())) {
                 ap.setMissingHashKeyOK(true);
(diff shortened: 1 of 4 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
```diff
@@ -167,7 +167,8 @@ public boolean feasible(Optimizable innerTable,
         if (hashKeyColumns == null && skipKeyCheck) {
             // For full outer join, broadcast join is the default join strategy that can be applied
             // to any kinds of join predicate, so make the eligibility check more general
-            if (outerCost.getJoinType() == JoinNode.FULLOUTERJOIN &&
+            if ((outerCost.getJoinType() == JoinNode.FULLOUTERJOIN ||
+                 outerCost.getJoinType() == JoinNode.LEFTOUTERJOIN) &&
```
dgomezferro, line 171: @msirek should we also allow RIGHTOUTERJOIN here, or at this point we only have LEFT OUTER ?

dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
msirek: @dgomezferro For join algorithms we currently only have left outer join.  If we specify a right outer join in SQL it is executed using a left outer join algorithm (but with left and right tables swapped).  There is a test for it in BroadCastJoinIT.

dgomezferro: @msirek got it, thanks!

jyuanca (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #4896, opened 2020-12-18
DB-11067/DB-11111 SpliceCatalogUpgradeScripts UpgradeScriptForTablePriorities fix, cleanup, tests

martinrupp:
- SpliceCatalogUpgradeScripts can run multiple scripts for one version 
- fix UpgradeScriptForTablePriorities not running
- added some sort of tests for SpliceCatalogUpgradeScripts
- aligned master/3.0/3.1 code
- cleaned up upgrade scripts (see DB-11111)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/Splice_DD_Version.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/Splice_DD_Version.java
@@ -117,4 +117,9 @@ public class Splice_DD_Version extends DD_Version {
         return majorVersionNumber * 10000000000l + minorVersionNumber * 10000000l + patchVersionNumber * 10000l + sprintVersionNumber;
     }
 
+    static public int compare(Splice_DD_Version version1,Splice_DD_Version version2){
+        long v1=version1.toLong();
+        long v2=version2.toLong();
+        return Long.compare(v1, v2);
+    }
 }
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
@@ -50,11 +50,21 @@ public class SpliceCatalogUpgradeScripts{
                 version.getMinorVersionNumber(), version.getPatchVersionNumber(), sprint), script));
     }
 
-    // master / branch-3.0 / branch-3.1
-    // baseVersion1 = 2.8.0 / 2.8.0 / 2.8.0
-    // baseVersion2 = 3.1.0 / 3.1.0 / 3.0.0 // 3. fork
-    // baseVersion3 = 3.1.0 / 3.1.0 / 3.0.1 // hickup on branch 3.0
-    // baseVersion4 = 3.2.0 / 3.1.0 / 3.0.1 // 3.2 fork
+    // Upgrade scripts should be mostly the same on master/3.0/3.1.
+    // However the base versions (i.e. major.minor.patch without sprint) are different.
+    // e.g. UpgradeScriptToAddSysNaturalNumbersTable is for sprint 1985, which is [baseVersion4, 1985].
+    // that means on master = 3.2.0.1985, branch-3.0 = 3.1.0.1985, branch-3.0 = 3.0.1.1985.
+
+    // following table gives an overview over different base versions in the branches
+    //                 master | branch-3.0 | branch-3.1 | remarks
+    // baseVersion1 =  2.8.0  |   2.8.0    |   2.8.0    |
+    // baseVersion2 =  3.1.0  |   3.1.0    |   3.0.0    | 2.8 -> 3.x fork
+    // baseVersion3 =  3.1.0  |   3.1.0    |   3.0.1    | hickup on branch 3.0
+    // baseVersion4 =  3.2.0  |   3.1.0    |   3.0.1    | 3.2 fork
+    // ...
+    // baseVersionX = ...                               | add here new versions
+    //
+    // also check SpliceCatalogUpgradeScriptsTest for some unit tests for this.
 
     static final public Splice_DD_Version baseVersion1 = new Splice_DD_Version(null, 2, 8, 0);
     static final public Splice_DD_Version baseVersion2 = new Splice_DD_Version(null, 3, 1, 0);
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
@@ -21,10 +21,7 @@ import com.splicemachine.derby.impl.sql.catalog.SpliceDataDictionary;
 import com.splicemachine.derby.impl.sql.catalog.Splice_DD_Version;
 import com.splicemachine.si.impl.driver.SIDriver;
 import org.apache.log4j.Logger;
-
-import java.util.Comparator;
-import java.util.NavigableSet;
-import java.util.TreeMap;
+import java.util.*;
 
 /**
  * Created by jyuan on 10/14/14.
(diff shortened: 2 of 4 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/Splice_DD_Version.java
```diff
@@ -117,4 +117,15 @@ public long toLong() {
...
+    static public int compare(Splice_DD_Version version1,Splice_DD_Version version2){
+        long v1=version1.toLong();
+        long v2=version2.toLong();
+
+        if(v1<v2)
+            return -1;
+        else if(v1==v2)
+            return 0;
+        else
+            return 1;
```
dgomezferro, line 129: Nitpick, you can use Long.compare(v1, v2);

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -35,52 +32,53 @@
...
+        addUpgradeScript(3,1,0, 1948, new UpgradeScriptForAddDefaultToColumnViewInSYSIBM(sdd,tc));
+        addUpgradeScript(3,1,0, 1953, new UpgradeScriptForRemoveUnusedIndexInSYSFILESTable(sdd,tc));
+        addUpgradeScript(3,1,0, 1959, new UpgradeScriptForTriggerMultipleStatements(sdd,tc));
+        addUpgradeScript(3,1,0, 1962, new UpgradeScriptForAddDefaultToColumnViewInSYSVW(sdd,tc));
+        addUpgradeScript(3,1,0, 1964, new UpgradeScriptForAliasToTableView(sdd,tc));
+        addUpgradeScript(3,1,0, 1970, new UpgradeScriptForAddTablesAndViewsInSYSIBMADM(sdd,tc));
+        addUpgradeScript(3,1,0, 1971, new UpgradeScriptToAddCatalogVersion(sdd,tc));
+        addUpgradeScript(3,1,0, 1974, new UpgradeScriptToAddMinRetentionPeriodColumnToSYSTABLES(sdd, tc));
+        addUpgradeScript(3,1,0, 1977, new UpgradeScriptToAddSysKeyColUseViewInSYSIBM(sdd, tc));
+        addUpgradeScript(3,1,0, 1979, new UpgradeScriptForTablePriorities(sdd, tc));
```
dgomezferro, line 77: Move this one to the current version so that it gets executed with the new fixed logic. I think it already handles tables that have the expected priority, right?

dgomezferro (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -33,56 +30,81 @@
...
+        addUpgradeScript(3,1,0, 1974, new UpgradeScriptToAddMinRetentionPeriodColumnToSYSTABLES(sdd, tc));
+        addUpgradeScript(3,1,0, 1977, new UpgradeScriptToAddSysKeyColUseViewInSYSIBM(sdd, tc));
+        addUpgradeScript(3,1,0, 1979, new UpgradeScriptToSetJavaClassNameColumnInSYSALIASES(sdd, tc));
+        addUpgradeScript(3,2,0, 1983, new UpgradeScriptToAddBaseTableSchemaColumnsToSysTablesInSYSIBM(sdd,tc));
+        addUpgradeScript(3,2,0, 1985, new UpgradeScriptToAddSysNaturalNumbersTable(sdd, tc));
+        addUpgradeScript(3,2,0, 1989, new UpgradeScriptToAddIndexColUseViewInSYSCAT(sdd, tc));
+        addUpgradeScript(3,2,0, 1990, new UpgradeScriptForTablePriorities(sdd, tc));
+    }
+
+    public static List<VersionAndUpgrade> GetScriptsToUpgrade(List<VersionAndUpgrade> scripts,
```
dgomezferro, line 82: Start method names with lowerCase

splice_machine/src/test/java/com/splicemachine/derby/utils/SpliceCatalogUpgradeScriptsTest.java
```diff
@@ -0,0 +1,110 @@
...
+            "3.1.0.1979: com.splicemachine.derby.impl.sql.catalog.upgrade.UpgradeScriptToSetJavaClassNameColumnInSYSALIASES\n" +
+            "3.2.0.1983: com.splicemachine.derby.impl.sql.catalog.upgrade.UpgradeScriptToAddBaseTableSchemaColumnsToSysTablesInSYSIBM\n" +
+            "3.2.0.1985: com.splicemachine.derby.impl.sql.catalog.upgrade.UpgradeScriptToAddSysNaturalNumbersTable\n";
+
+    String s2 = "3.2.0.1989: com.splicemachine.derby.impl.sql.catalog.upgrade.UpgradeScriptToAddIndexColUseViewInSYSCAT\n" +
+            "3.2.0.1990: com.splicemachine.derby.impl.sql.catalog.upgrade.UpgradeScriptForTablePriorities\n";
+    // add more scripts here
+
+    @Test
+    public void t1()
```
dgomezferro, line 35: It'd be great to have more descriptive test names

dgomezferro (changes requested):
Just a couple of minor issues

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -33,56 +30,81 @@
...
 
     SpliceDataDictionary sdd;
-    Splice_DD_Version catalogVersion;
     TransactionController tc;
-    TreeMap<Splice_DD_Version, UpgradeScript> scripts;
-    Comparator<Splice_DD_Version> ddComparator;
-    
-    public SpliceCatalogUpgradeScripts(SpliceDataDictionary sdd,Splice_DD_Version catalogVersion,TransactionController tc){
+
+    List<VersionAndUpgrade> scripts;
```
hatyo, line 35: Instead of using a `List` and perform a linear search each time we want to perform an upgrade, I would keep the tree structure used before and replace it with something like Guava's [`MultiMap`](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html) instead.

martinrupp: i made this especially easy since the whole upgrade path is really hard to test, people modify it, and if it's broken it's a high cost, while optimal performance is lower priority here. last version was using a sophisticated map, then someone changed it to still iterate over all element.
this upgrade is called ONCE when database is started, the number of items in the list is currently 22. Even for 100 items i would doubt that creating a multimap + searching is faster than just looping over a list. even if - I would really like to keep this rock-solid-bullet-proof code, and not trying to squeeze out a second or two.

dgomezferro: I agree with @martinrupp , I think it's ok to have a list here since this code is not performance sensitive (at least for now)

hatyo: > I would really like to keep this rock-solid-bullet-proof code, and not trying to squeeze out a second or two.

I think these are mutually exclusive.

Anyway, it is a fair point that the number of items is too small that the overhead is negligible.

martinrupp: > > I would really like to keep this rock-solid-bullet-proof code, and not trying to squeeze out a second or two.

> I think these are not mutually exclusive

yes, i agree.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/Splice_DD_Version.java
martinrupp: done

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
martinrupp: done

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
martinrupp: done

martinrupp:
aligning code of master/3.0/3.1, and make it so we don't have to adjust the upgrade code line for every branch.
it was easier git-wise to squash all this into one commit.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -33,79 +30,121 @@
...
+    {
+        scripts.add(new VersionAndUpgrade(new Splice_DD_Version(sdd, version.getMajorVersionNumber(),
+                version.getMinorVersionNumber(), version.getPatchVersionNumber(), sprint), script));
+    }
+
+    // master / branch-3.0 / branch-3.1
+    // baseVersion1 = 2.8.0 / 2.8.0 / 2.8.0
+    // baseVersion2 = 3.1.0 / 3.1.0 / 3.0.0 // 3. fork
+    // baseVersion3 = 3.1.0 / 3.1.0 / 3.0.1 // hickup on branch 3.0
+    // baseVersion4 = 3.2.0 / 3.1.0 / 3.0.1 // 3.2 fork
```
dgomezferro, line 57: Can you add two lines here explaining what the table means? It took me a while to understand

dgomezferro (approved):

hatyo (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
martinrupp: yes, good point

carolp-503 (approved):
+1

PR #4900, opened 2020-12-19
DB-10759 Tell Spark where to find jars for user VTIs

msirek:
This allows user-defined VTIs to be written as FlatMapFunctions to process rows in the splice DataSet operation tree.
SparkContext.addJar is called on the VTI jar file that splice internally stores in HDFS.

Note that if the VTI definition is changed, the Olap Server must be restarted to pick up the change because there is no way to remove a jar from the spark context.

In the future we'll try to do away with this limitation, perhaps by switching internally to a new OlapServer every time a VTI is compiled.


msirek:

spliceci:
TEST SUCCEEDED +1

ipraznik-splice (approved):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
```diff
@@ -1491,4 +1492,16 @@ void setDB2VarcharCompatibilityModeNeedsReset(boolean newValue,
 
     boolean isCompilingFromTableTempTrigger();
 
+    void addUserJarsToSparkContext();
+
+    boolean isSparkJob();
+
+    void setSparkContext(SparkContext sparkContext);
+
+    SparkContext getSparkContext();
```
dgomezferro, line 1501: We should try to reduce direct dependencies on Spark in "core" modules such as db-engine.

arnaud-splice (approved):

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
msirek: @dgomezferro OK, I converted this to class Object and moved Spark class and API references to the spark_sql module.

msirek:

carolp-503 (approved):
+1

spliceci:
TEST SUCCEEDED +1

spark_sql/src/main/spark2.1/src/main/java/com/splicemachine/sparksql/SparkSQLUtilsImpl.java
```diff
@@ -0,0 +1,44 @@
...
+package com.splicemachine.sparksql;
+
+
+import com.splicemachine.utils.SparkSQLUtils;
+import org.apache.spark.SparkContext;
+
+/* Purpose:  Provide interfaces to access Spark objects or APIs
+ *           which may differ depending on the Spark version.
+ */
+public class SparkSQLUtilsImpl implements SparkSQLUtils {
```
dgomezferro, line 24: The implementation for all platforms is the same, it should be in `spark_sql/src/main/java` only once instead of copied in multiple platforms.

dgomezferro (changes requested):

spark_sql/src/main/spark2.1/src/main/java/com/splicemachine/sparksql/SparkSQLUtilsImpl.java
msirek: Good point.  I've made the change.  If we need to have different method implementations in the future based on spark version, we could create separate source files for it at that time.

dgomezferro (approved):

PR #4906, opened 2020-12-21
DB-11075 Make standalone durable (3.1)

dgomezferro:
Reduce HMaster timeout for faster procedure execution


carolp-503 (approved):
Approved for 3.1

PR #4919, opened 2020-12-22
DB-11047 add option splice.function.preserveLineEndings (3.1)

martinrupp:
if splice.function.preserveLineEndings = true, multiline strings imported into the database will contain the `\r`, `\r\n`, `\n` exactly like in file. Otherwise (current default) all these will be replaced by `\n`.


carolp-503 (approved):
Approved for 3.1

carolp-503 (approved):
Approved for 3.1

PR #4923, opened 2020-12-24
DB-10743 resolve conflicts at commit time

hatyo:
In this PR we support deferring write-write conflict resolution up until the commit time. When a conflict is detected, the transaction registers the conflicting transaction and proceeds with the write, at commit time, the transaction tries to rollback all conflicting transactions, if some of these transactions were already committed it aborts.

Note that this behavior is configurable with the parameter `splice.txn.conflictResolutionStrategy`, it has two values:
1. `IMMEDIATE` preserves the old behavior, when a conflict is detected the write is immediately rejected.
2. `DEFERRED` the new behavior, when a conflict is detected, it is remembered and checked at commit time.

Implementation-wise, we introduce a new modifier in `SPLICE_TXN` table which keeps track of all conflicting transaction(s) a transaction could come across during its life-time. This column is checked when a transaction is about to commit on the server-side, this is necessary because we want to perform roll backs under the transaction's write lock. The following picture illustrates how this feature works:

image

To avoid potential deadlocks with transactions trying to rollback each other at commit time, we implement a new `rollback` method which takes as parameter both the transaction we want to rollback plus the originating transaction which issues the rollback. If the rollback fails to obtain a lock on the transaction we want to rollback we double check whether that transaction is attempting to commit itself (hence we can't obtain the lock), and then either cancel the rollback or try again, cancelling the rollback is done by checking the "smallest" of the transaction Ids, we can maybe take a more intelligent decision on which transaction to rollback in the future.

To know if a transaction is about to commit, we introduced a new in-memory "state" `commit-pending` which is active during the commit logic.

When we rollback a transaction, we check first if it exists in the region itself so we roll it back immediately, otherwise, we make an RPC call to invoke the rollback on the respective remote region.

Testing this new feature is a bit tricky since it requires testing static configuration of the cluster, i.e. the tests should only run if `splice.txn.conflictResolutionStrategy` is set correctly. Luckily there is a system procedure which lists all system configuration `SYSCS_GET_REGION_SERVER_CONFIG_INFO`. We use it in the test suite to sense the resolution strategy and run (or ignore) the tests, unfortunately doing this for all other tests is very time consuming, so we decided to adapt all affected tests in /master to run with the new `DEFERRED` strategy.

A simple test DSL is written to describe on somewhat a high level a specific scenario of concurrent transactions interactions and what should happen. It can be useful to easily test other aspects of the system in the future, some tests are moved from `SITransactorTest` to it to properly test deletion logic.

Previously we had some logic that ignores WW conflicts coming from transactions trying to write to `SYSSTATEMENTS`, the reasoning behind it was that many concurrent transactions could be writing the same context (activation, ...) to that table and from usability perspective, it doesn't make sense to make them conflict with each other, so we simply ignore all conflicts pertaining writes to that table and move on. It turned out that this logic is not easy to preserve considering we don't conflict at all from now on when writing, only at commit time we detect react to conflicts as described above.

To fix this issue, we examine the table affected by the transaction right before we collect the list of conflicting transactions and check whether its descriptor matches the `SYSSTATEMENTS`, if so we avoid conflict detection all together preserving the old behavior. This solution is a bit hacky but any other solution (e.g. making the transaction itself carry a conflict-detection strategy state) requires substantial amount of work for little gain. Maybe instead of working around the problem we should fix the root cause of it, but again, that's a separate work stream.

Here is an example on how this works:
```
time      session1                   session2
 1     create table t1(col1 int);
 2     insert into t1 values 42;
 3     autocommit off;
 4                                   autocommit off;
 5     update t1 set col1 = 100;
 6                                   update t1 set col1 = 200; -- succeeds
 7     commit;
 8                                   commit; -- fails with an error
```
Before this PR, we would've failed at step 6, now we fail at commit time (step 8).

#### SpliceCK Adaptations

SpliceCK was adapted to display the new column in `SPLICE_TXN` which could be useful to examine the internal behavior of this feature, we added a new column to the `txlist` command called `conflicting transaction` which, as the name implies, read the respective conflicting transactions modifier from `SPLICE_TXN` table and displays it, here an example similar to the above but transaction IDs:

```
time      session1                                        session2
 1     create table t1(col1 int);
 2     insert into t1 values 42;
 3     autocommit off;
 4     CALL SYSCS_UTIL.SYSCS_GET_CURRENT_TRANSACTION();
         -- returns 7936
 5                                                     autocommit off;
 6                                                     CALL SYSCS_UTIL.SYSCS_GET_CURRENT_TRANSACTION();
                                                       -- returns 8192
 7     update t1 set col1 = 100;
 8                                                     update t1 set col1 = 200; -- succeeds
```
If we examine now the transaction table with SpliceCK (extra columns removed for better readability):
```
spliceck> txlist

transaction id     ...     target tables     conflicting transactions
===================...====================================================
512                ...     [boot]            []
1024               ...     [boot]            []
1536               ...     [boot]            []
2048               ...     [dictionary]      []
3072               ...     [dictionary]      []
6912               ...     [dictionary]      []
7424               ...     [1648]            []
7936               ...     [1648]            []
8192               ...     [1648]            [7936]
```

we can see from the table above in column "conflicting transactions" that transaction 8192 is conflicting with transaction 7963.


splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
```diff
@@ -180,6 +181,10 @@
     public static final String DURABILITY = "splice.txn.durability";
     public static final String DEFAULT_DURABILITY = "SYNC";
 
+    // write-write conflict resolution strategy
+    public static final String CONFLICT_RESOLUTION_STRATEGY = "splice.txn.conflictResolutionStrategy";
+    public static final String DEFAULT_CONFLICT_RESOLUTION_STRATEGY = ConflictResolutionStrategy.DEFERRED.toString();
```
hatyo, line 186: as discussed, on master we activate the new behavior, but when cherry-picking this PR on 3.0 and 3.1 we keep the old one (`IMMEDIATE`)

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2225/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2226/

hatyo:
> TEST FAILED
> http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2226

We should adapt tests to the new behavior, DB-11143

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2243/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2250/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2253/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2255/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2267/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/111/

hatyo:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/114/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2271/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2273/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/116/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2274/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/117/

hatyo:

martinrupp (commented):
this is an awesome change Youssef!
I only have some minor thingies:
- can you please add an example how to observe the behavior in sqlshell both with `splice.txn.conflictResolutionStrategy` set to `IMMEDIATE` and `DEFERRED`, so that the QA team can check later that this feature is still working.
- you mentioned in the description that SpliceCK can be used to check the behavior, can you maybe add a small (shortened) example as well?

martinrupp (changes requested):
see above

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/TransactionController.java
```diff
@@ -247,66 +247,95 @@ gaining access to resources (conglomerates, scans, etc.) in the transaction
...
-	byte IS_TEMPORARY	=	(byte) 0x01; // conglom is temporary
-	byte IS_KEPT		=	(byte) 0x02; // no auto remove
+    byte IS_DEFAULT	=	(byte) 0x00; // initialize the flag
+    byte IS_TEMPORARY	=	(byte) 0x01; // conglom is temporary
+    byte IS_KEPT		=	(byte) 0x02; // no auto remove
 
+    /**
+     * Conflict resolution strategies.
+     */
+    enum ConflictResolutionStrategy {
```
martinrupp, line 269: we also have a ConflictResolutionStrategy.java, do we need both?

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -8690,42 +8673,36 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
-</text>
+                <text>------------ END ERROR MESSAGE -------------</text>
             </msg>
 
             <msg>
                 <name>L003</name>
                 <text>
-------------  BEGIN SHUTDOWN ERROR STACK -------------
-</text>
+                    ------------ BEGIN SHUTDOWN ERROR STACK -------------</text>
```
martinrupp, line 8687: i guess it is intended you added 20 spaces in front of the message? (just checking)

martinrupp (changes requested):
maybe we could add some tracing in some spots - i'm not very much into this code i don't know how much that would help later debugging issues on a cluster, might also be a new jira

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/119/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2276/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2280/

hatyo:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/123/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2288/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/131/

hatyo:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/165/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2323/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/165/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2343/

martinrupp:
- sql example is done
- i would have a small suggestion on the words in the description about how splice_ck displays this, e.g. "SpliceCK was adapted to display the new column `conflicting transactions` when doing `txlist`. This proved to be useful to examine the internal behavior when running some tests during development."

- please check comments i made here
https://github.com/splicemachine/spliceengine/pull/4923#pullrequestreview-571135465

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2558/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2559/

hatyo:

spliceci:
TEST SUCCEEDED +1

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
hatyo: Actually not, this was part of IntelliJ's reformatting the file, I restored it

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/TransactionController.java
hatyo: Thanks, this is a left over from a previous implementation approach that is refactored. I just removed it.

hatyo:
> maybe we could add some tracing in some spots - i'm not very much into this code i don't know how much that would help later debugging issues on a cluster, might also be a new jira

Since your review I added a number of trace messages and I will add few more in another commit.

hatyo:
> you mentioned in the description that SpliceCK can be used to check the behavior, can you maybe add a small (shortened) example as well?

Sure, I will add an example of that in the PR message, thanks!

martinrupp (approved):

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -574,72 +588,142 @@ private ConflictResults checkCommitTimestampForConflict(TxnView updateTransactio
...
+            LOG.warn("transaction " + txn + " conflicts with " + activeTx.toString() + " on table " + table.getTableName() + " on row " + cell);
+        }
+        return Longs.toArray(activeTx);
+    }
+
+    private boolean shouldIgnoreConflicts(Partition table) throws IOException {
+        return "SYSSTATEMENTS".equals(table.getDescriptor().getValue("tableDisplayName"))
+                && (table.getDescriptor().getValue("schemaDisplayName") == null
+                || table.getDescriptor().getValue("schemaDisplayName").equals("SYS"));
+    }
```
dgomezferro, line 717: Did you change the SYSSTATEMENTS conflict avoidance in the end? The PR description still describes the old way.

dgomezferro (changes requested):
Some minor comments

db-tools-i18n/src/main/resources/com/splicemachine/db/loc/messages.xml
```diff
@@ -8941,17 +8920,13 @@ Shutting down instance {0} on database directory {1} with class loader {2} </tex
...
             <title>Undocumented Message Service Messages from MessageId.java</title>
 
             <msg>
                 <name>M001</name>
                 <text>
-# ********************************************************************
-# ***                Please do NOT edit this file.                 ***
-# *** CHANGING THE CONTENT OF THIS FILE MAY CAUSE DATA CORRUPTION. ***
-# ********************************************************************</text>
+                    # ******************************************************************** # *** Please do NOT edit this file. *** # *** CHANGING THE CONTENT OF THIS FILE MAY CAUSE DATA CORRUPTION. *** # ********************************************************************</text>
```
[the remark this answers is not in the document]
hatyo: fixed.

hbase_storage/src/main/java/com/splicemachine/si/impl/CoprocessorTxnStore.java
```diff
@@ -61,7 +57,8 @@
     private volatile long oldTransactions;
```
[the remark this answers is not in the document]
hatyo: done.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/V2TxnDecoder.java
```diff
@@ -57,172 +55,363 @@
...
-    protected long toLong(Cell data){
-        return Encoding.decodeLong(data.getValueArray(),data.getValueOffset(),false);
+    /**
+     * Encodes transaction objects using the new, packed Encoding format
+     *
+     * The new way is a (more) compact representation which uses the Values CF (V) and compact qualifiers (using
+     * the Encoding.encodeX() methods) as follows:
+     *
+     * "d" -- packed tuple of (beginTimestamp,parentTxnId,isDependent,additive,isolationLevel) // Youssef: looks like isDependent is not encoded?
+     * "c" -- counter (using a packed integer representation)
```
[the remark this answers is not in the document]
hatyo: done.

hbase_storage/src/main/java/com/splicemachine/si/impl/region/V2TxnDecoder.java
```diff
@@ -57,172 +55,363 @@
...
+     * "w" -- conflictingTxnIds
+     *
+     * The additional columns are kept separate so that they may be updated(and read) independently without
+     * reading and decoding the entire transaction.
+     *
+     * In the new format, if a transaction has been written to the table, then it automatically allows writes
+     *
+     * order: c,d,e,g,k,s,t,v
+     * order: counter,data,destinationTable,globalCommitTimestamp,keepAlive,state,commitTimestamp,taskId
+     */
```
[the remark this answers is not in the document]
hatyo: done.

mem_storage/src/main/java/com/splicemachine/si/impl/MemTxnStore.java
```diff
@@ -176,6 +179,48 @@ public void rollback(long txnId) throws IOException{
         }
     }
 
+    @Override
+    public void rollback(long txnId, long originatorTxnId) throws IOException {
+        long beginTS = txnId & SIConstants.TRANSANCTION_ID_MASK;
+        Lock lock = lockStriper.get(beginTS).writeLock();
+        // make sure that the region doesn't close while we are working on it
```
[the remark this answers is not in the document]
hatyo: removed, that was copy/paste mistake.

mem_storage/src/main/java/com/splicemachine/si/impl/MemTxnStore.java
```diff
@@ -176,6 +179,48 @@ public void rollback(long txnId) throws IOException{
...
+        boolean lockAcquired = false;
+        while (!lockAcquired) {
+            try {
+                lockAcquired = lock.tryLock(200, TimeUnit.MILLISECONDS);
+                if (!lockAcquired && commitPendingTxns.contains(txnId)) {
+                    if (originatorTxnId < txnId) { // simplest comparison that leads that
```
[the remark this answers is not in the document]
hatyo: Yes, fixed it (also in `StripedTxnLifecycleStore`), thanks!

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
hatyo: updated the PR description, thanks for checking it.

hbase_storage/src/main/java/com/splicemachine/si/data/hbase/coprocessor/TxnLifecycleEndpoint.java
```diff
@@ -73,30 +75,35 @@ public void shutdown(TransactionResolver instance){
...
+                                                                          resolver,
+                                                                          txnKeepAliveTimeout,
+                                                                          new SystemClock(),
+                                                                          durability);
+                TimestampSource timestampSource = driver.getTimestampSource();
                 int txnLockStrips = configuration.getTransactionLockStripes();
-                lifecycleStore = new StripedTxnLifecycleStore(txnLockStrips,regionStore,
-                        new RegionServerControl(region, (RegionServerServices)rce.getOnlineRegions()),timestampSource);
-                isTxnTable=true;
+                ClientTxnLifecycleManager lifecycleManager = new ClientTxnLifecycleManager(timestampSource, siEnv.exceptionFactory());
```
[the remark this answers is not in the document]
hatyo: I think I missed that `SIDriver` has an instance of `ClientTxnLifecycleManager`, I fixed it, thanks a lot

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2564/

hatyo:

hatyo:
Investigating the failure on mem, the test failed because a transaction which is cleaning up the schema after finishing has conflicted with another transaction on syscols table, not sure exactly why, but not having a wait-for-stale-tx in mem profile could provide a hint, will continue looking into this on Monday, retriggered Jenkins job because I am interested in knowing how often this fails.

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2567/

hatyo:

spliceci:
TEST SUCCEEDED +1

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -574,72 +588,156 @@ private ConflictResults checkCommitTimestampForConflict(TxnView updateTransactio
...
+            }
+            // get the parent
+            TxnView txnView = txnSupplier.getTransaction(dataTxnId);
+            long id = txnView.getTxnId();
+            while(true) {
+                TxnView parent = txnView.getParentTxnView();
+                if(parent == Txn.ROOT_TRANSACTION) break;
+                txnView = parent;
+            }
+            Txn.State state = txnView.getState();
```
msirek, line 701: Should we add the '((txnView.getTxnId() & SIConstants.TRANSANCTION_ID_MASK) == txn.getTxnId())' check here and continue, or is it OK to add this transaction's txnId to the HashSet?

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -574,72 +588,156 @@ private ConflictResults checkCommitTimestampForConflict(TxnView updateTransactio
...
+            } else if(state == Txn.State.COMMITTED && txnView.getCommitTimestamp() > txn.getBeginTimestamp()) { // bail out
+                throwWriteWriteConflict(txn, cell, dataTxnId);
+            }
+        }
+
+        return Longs.toArray(activeTx);
+    }
+
+    private static boolean shouldIgnoreConflictsInternal(Partition table) throws IOException {
+        String[] ignoredTables = new String[]{"SYSSTATEMENTS", "SYSCOLUMNS", "SYSDEPENDS"};
```
msirek, line 722: What's the reason for ignoring conflicts on these tables?  Do we do this elsewhere in the code?  Maybe that's why we keep getting corruption on these tables.  Maybe add a comment about why this is needed.

msirek (changes requested):

hbase_storage/src/main/java/com/splicemachine/si/impl/region/RegionTxnStore.java
```diff
@@ -579,4 +592,29 @@ public void close() throws IOException{
...
+        if (result == null || result == Result.EMPTY_RESULT)
+            throw new HReadOnlyModificationException("Transaction " + txnId + " is read-only, and was not properly elevated.");
+        Cell kv = result.getColumnLatestCell(FAMILY, identifier);
+        byte[] newBytes;
+        if (kv == null || kv.getValueLength() <= 0) {
+            newBytes = Encoding.encodeBytesUnsorted(payload);
+        } else {
+            newBytes = new byte[payload.length + kv.getValueLength() + 1];
+            System.arraycopy(payload, 0, newBytes, 0, payload.length);
+            System.arraycopy(kv.getValueArray(), kv.getValueOffset(), newBytes, payload.length + 1, kv.getValueLength());
```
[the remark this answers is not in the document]
hatyo: This logic is just moved [as-is](https://github.com/splicemachine/spliceengine/blob/0e5019bfc437088642e8691ec4b31ad85831ac7c/hbase_storage/src/main/java/com/splicemachine/si/impl/region/RegionTxnStore.java#L164) into a separate function, as to why we have this extra byte, I honestly can't say much, maybe @dgomezferro knows?

splice_si_api/src/main/java/com/splicemachine/si/api/txn/lifecycle/CannotRollbackException.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2012 - 2020 Splice Machine, Inc.
```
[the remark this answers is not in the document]
hatyo: done (also for other newly added files), thanks!

splice_si_api/src/main/java/com/splicemachine/si/impl/data/StripedTxnLifecycleStore.java
```diff
@@ -139,6 +170,63 @@ public void rollbackTransaction(long txnId) throws IOException{
...
+                     * acquired).
+                     */
+                    serverControl.ensureNetworkOpen();
+                } catch (IOException ioe) {
+                    if (lockAcquired) { // the lock was acquired, so it needs to be unlocked
+                        unlock(lock);
+                    }
+                    throw ioe;
+                }
+            } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
hatyo: Nice catch! thank you, I added `finally` for unlocking to this method and also fixed it similarly in the `acquireLock` method as well.

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
hatyo: Currently we ignore WW conflicts if they happen in certain scenarios:
- calling stored procedures causes writes in `SYSCOLUMNS` and `SYSDEPENDS` (for more information please check Jira DB-3386).
- calling an SPS could cause a write in `SYSSTATEMENTS`
these writes could conflicts, and the conflicts are ignored in these cases. With immediate conflict detection it is easily possible to localize this behavior by surrounding the code path with try-catch-ErrorState-WWConflict-and-ignore. However, with deferred conflict detection it is not possible to follow the same logic since we gather conflicts and check them against at commit-time, so I added this method to do exactly that. I am not claiming this is the right approach, but that's the easiest approach I could think of, I am open to suggestions on how to improve it.

I documented the method accordingly.

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
hatyo: I cleaned it a bit, the resulting `txnView` should always have its last byte set to zero, added an assertion for that.

splice_access_api/src/main/java/com/splicemachine/access/configuration/SIConfigurations.java
```diff
@@ -215,8 +220,16 @@ public void setDefaults(ConfigurationBuilder builder, ConfigurationSource config
...
+            LOG.error(String.format("Couldn't parse %s option: '%s'", DURABILITY, durability));
             builder.durability = Durability.SYNC;
         }
+
+        String conflictResolutionStrategyString = configurationSource.getString(CONFLICT_RESOLUTION_STRATEGY, DEFAULT_CONFLICT_RESOLUTION_STRATEGY);
+        try {
+            builder.conflictResolutionStrategy = ConflictResolutionStrategy.valueOf(conflictResolutionStrategyString);
+        } catch (IllegalArgumentException ex) {
+            LOG.error(String.format("Couldn't parse %s option: '%s'", CONFLICT_RESOLUTION_STRATEGY, conflictResolutionStrategyString));
+            builder.conflictResolutionStrategy = ConflictResolutionStrategy.IMMEDIATE;
```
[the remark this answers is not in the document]
hatyo: Thanks, that makes sense, please check whether the new version matches your suggestion.

msirek (approved):

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2625/

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
hatyo: **update** I was mistaken, I think if we could reach this code path by a in-memory sub-transaction which has its last byte set to its child-order (I think) to the parent. In that case we should reset the last byte in order to get the parent transaction (since we want to rollback the parent transaction if we conflict with a child sub-transaction).

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #4925, opened 2020-12-25
DB-10773+DB-11084 Honor joinStrategy hints on the target table of an UPDATE (3.1)

msirek:
This is to re-merge DB-10773, including DB-11084 which fixes a performance regression in query plans with hinted joins.


carolp-503 (approved):
Approved for 3.1

PR #4939, opened 2021-01-04
DB-11089 Fix NPE in SQLChar.setWidth if warning occurs without resultset 

arnaud-splice:


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):
Change looks good to avoid the NPE. Arnaud is going to create a JIRA to fix the places where we are missing the activation or the ResultSet so that we don't skip attaching the Warning to it.

ipraznik-splice (approved):

arnaud-splice:
> Change looks good to avoid the NPE. Arnaud is going to create a JIRA to fix the places where we are missing the activation or the ResultSet so that we don't skip attaching the Warning to it.

Tracked by DB-11106

carolp-503 (approved):
+1

PR #4954, opened 2021-01-06
DB-11069 Fix HDP restart script for Hbase

bklo94:
Fix logic for Hbase resart. Also the omission of `hbase.SpliceDriver:` from the grep makes it consistent to the flatten.sh script.


jyuanca (approved):

carolp-503 (approved):
+1

PR #4960, opened 2021-01-07
DB-11113 fix LCC NPE when using getPreserveLineEndings in SpliceFileVTI

martinrupp:


martinrupp:

jyuanca:

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2216/

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -7545,8 +7545,7 @@ public TransactionController getTransactionCompile() {
             assert !booting:"booting is expected to be false";
 
             LanguageConnectionContext lcc=getLCC();
-            return lcc.getTransactionCompile();
-
+            return lcc == null ? null : lcc.getTransactionCompile();
```
dgomezferro, line 7548: Is this necessary to fix the problem or is it just preventing a similar NPE in the future?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1124,7 +1124,7 @@ public boolean canWriteCache(TransactionController xactMgr) throws StandardExcep
         DDLWatcher ddlWatcher=driver.ddlWatcher();
         if(xactMgr==null)
             xactMgr = getTransactionCompile();
-        return ddlWatcher.canWriteCache((TransactionManager)xactMgr);
+        return xactMgr != null && ddlWatcher.canWriteCache((TransactionManager)xactMgr);
```
dgomezferro, line 1127: Is this necessary to fix the problem or is it just preventing a similar NPE in the future?

dgomezferro (changes requested):
If that code tries to prevent a similar NPE I think we should remove it, since it would hide errors.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
martinrupp: good point, will check

martinrupp: actually it is needed, `pass in language connection context` is not sufficient. if we remove this, `HdfsImportIT.testPropertiesCacheNPE` fails. However i agree that this is not completely understood. I'm creating a new jira.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
martinrupp: see above

dgomezferro (approved):
Thanks for the explanation @martinrupp , I'm ok with merging the current fix if we make https://splicemachine.atlassian.net/browse/DB-11117 a priority for next sprint

martinrupp:
DB-11117 will look at this more thoroughly

carolp-503 (approved):
+1

martinrupp:
@OlegMazurov found a better solution in https://splicemachine.atlassian.net/browse/DB-11117

PR #4964, opened 2021-01-07
DB-11134 testing upgrades in spliceengine

martinrupp:


check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
martinrupp, line 1: this assumes current branch has been compiled

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
...
+tar -xzvf platform_it_${VERSION}.tar.gz
+rm platform_it_${VERSION}.tar.gz
+
+# restart cluster
+./start-splice-cluster -l
+
+# test
+if mvn -B -e surefire:test -Pcore,cdh6.3.0 -Dtest='UpgradeTestIT#*' -DskipServerStart -DfailIfNoTests=false; then
+	echo "\n\n\nUPGRADE SUCCEEDED"
+	return 0
```
martinrupp, line 24: return: can only `return' from a function or sourced script

martinrupp: cat platform_it/splice.log | grep 'upgrade scripts'
	cat platform_it/splice.log | grep 'Running upgrade script'

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# creates a file platform_it_${VERSION}.tar.gz
+function create_upgrade_targz
+{
+	git checkout tags/${VERSION}
+	cd platform_it
+	git clean -dfx
+	cd ..
+
+	./start-splice-cluster
```
hatyo, line 47: Maybe you should shut down the cluster `start-splice-cluster -k` before starting to copy `hbase` and `zookeeper` folders, because as the cluster is running, the content of these directories keeps changing which could lead to data race conditions.

check_upgrade.sh
hatyo: Maybe you could add some checks for that, e.g. that the `target` folder exists.

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+VERSION=3.1.0.1971
+
+# stop current cluster
+./start-splice-cluster -k
+
+# clean up platform_it
+cd platform_it
```
hatyo, line 9: (optional) you also assume you're in the base directory of the project which is not necessarily true, e.g. I could this script from `/tmp`.

Maybe you could add some sanity checks that you're in a git repository and then go to the base folder using `git rev-parse --show-cdup`

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+VERSION=3.1.0.1971
```
hatyo, line 3: should this be a parameter?

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
...
+# download the previous standalone data
+aws s3 cp s3://splice-snapshots/upgrade_tests/platform_it_${VERSION}.tar.gz .
+tar -xzvf platform_it_${VERSION}.tar.gz
+rm platform_it_${VERSION}.tar.gz
+
+# restart cluster
+./start-splice-cluster -l
+
+# test
+if mvn -B -e surefire:test -Pcore,cdh6.3.0 -Dtest='UpgradeTestIT#*' -DskipServerStart -DfailIfNoTests=false; then
```
hatyo, line 22: Is `mvn` return code reliable here?

hatyo (approved):
Couple of minor comments, otherwise, it looks good.

check_upgrade.sh
```diff
@@ -0,0 +1,57 @@
...
+git clean -dfx
+cd ..
+
+# download the previous standalone data
+aws s3 cp s3://splice-snapshots/upgrade_tests/platform_it_${VERSION}.tar.gz .
+tar -xzvf platform_it_${VERSION}.tar.gz
+rm platform_it_${VERSION}.tar.gz
+
+# restart cluster
+./start-splice-cluster -l
```
dgomezferro, line 19: I would use `-b` here since we haven't changed any code

check_upgrade.sh
dgomezferro, line 1: Maybe this script is better in `platform_it/src/test/bin/` similar to `start-splice-its` for instance.

Can you try integrating it into the verification maven goal too? After running the NSDS tests we would shut splice down, download the tarball, unpack it, restart (without building) and run the UpgradeIT test.

dgomezferro (changes requested):

dgomezferro:
This is great @martinrupp , if we can integrate it into the verify maven phase it would be amazing!

check_upgrade.sh
martinrupp: done

check_upgrade.sh
martinrupp: yes, returning non-null in case of failed test

check_upgrade.sh
martinrupp: i did a `git clean -dfx` in platform_it, so which removes jars from there, so i need to rebuild those.

check_upgrade.sh
martinrupp: yes, working on it

martinrupp:
good bye :-)

PR #4970, opened 2021-01-08
DB-11103 Fix Jenkins pipeline for scm pull (3.1)

bklo94:
Similar to the one master branch, but also removes the missed duplicate settings.xml


carolp-503 (approved):
Approved for 3.1

PR #4973, opened 2021-01-08
DB-11117 NPE in getPreserveLineEndings/SpliceFileVTI (3.0)

martinrupp:
Create a delegating ContextManager. No need to push LCC using initCurrentLCC() (will be available from the parent ContextManager). Cleanup must be done on the same thread. Newly created ContextManager must be destroyed.
I'm also fixing memory leak warnings in ContextService.


jyuanca (approved):
Approved for 3.0

PR #4983, opened 2021-01-11
DB-11136 Use SYNC_WAL in standalone (3.0)

dgomezferro:
ASYNC_WAL doesn't eventually force a SYNC as I thought, so we need to
use SYNC_WAL if we want durability for standalone. See
https://issues.apache.org/jira/browse/HBASE-16689 for details


carolp-503 (approved):
Approved for 3.0

PR #4986, opened 2021-01-11
DB-6509 Support UPDATE table SET xx=yy FROM ...

ascend1:


ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SQLGrammarImpl.java
```diff
@@ -724,6 +725,55 @@ StatementNode getUpdateNodeWithSub(FromTable fromTable, /* table to be updated *
...
+                innerRCL,  /* SELECT list */
+                null, /* AGGREGATE list */
+                fromList,  /* FROM list */
+                null, /* WHERE clause */
+                null, /* GROUP BY list */
+                null, /* having clause */
+                null, /* window list */
+                getContextManager());
+
+        return (SubqueryNode) nodeFactory.getNode(
```
hatyo, line 765: Does it make sense to also have aggregations supported in the select subquery?

hatyo (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SQLGrammarImpl.java
ascend1: Aggregation works. You can write a subquery that selects aggregations and put it in the from list. The `SelectNode` built here is merely a wrapper used in query processing. Suppose the user subquery has `select avg(x), sum(y) from ...`, the select we build here is something like
`select a, b from (select avg(x) as a, sum(y) as b from ...) ...`
That's why we don't have aggregation list or any other clauses in this select.

Of course, something like `update set x = sum(y) from a, b where ...` is beyond the scope, I think. It might be possible, though.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/UpdateFromSubqueryIT.java
```diff
@@ -491,6 +494,91 @@ public void testNestedFromSubquery() throws Exception {
...
+                " from (select a2, b2 from t2 --splice-properties joinStrategy=%s,useSpark=%s\n" +
+                " cross join (select a3 from t3)) where a1=a2", this.joinStrategy, this.useSparkString));
+
+        try (ResultSet rs = spliceClassWatcher.executeQuery("select * from t1")) {
+            assertEquals(expected, TestUtils.FormattedResult.ResultFactory.toString(rs));
+        }
+
+        // UPDATE ... FROM, from list
+        // where clause is part of the update statement
+        spliceClassWatcher.executeUpdate(format("update t1 set b1 = b2" +
```
dgomezferro, line 510: Do we support updating multiple columns with this syntax? Can you add a test if so?

dgomezferro (changes requested):

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/UpdateFromSubqueryIT.java
ascend1: Yes, updating multiple columns using this syntax is supported. I added a new test for this case. Basically, it's an existing test but the query is rewritten using the new syntax.

ascend1:

spliceci:
TEST SUCCEEDED +1

msirek (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #5009, opened 2021-01-14
DB-11176 Remove cdh/hdp repository from pom.xml

bklo94:
Currently we have a proxy for these repositories in our nexus repository (splicemachine). After removing my local maven cache and the repositories, my instance is able to successfully build. Therefore this PR removes the redundant endpoints since if maven is not able to find the jars in our repository, then it wouldn't find it in the other repositories either.

Also took out the mapr repo since we also mirror that repository

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,26 +29,10 @@
             <id>splicemachine</id>
             <url>http://nexus.splicemachine.com/nexus/content/groups/developers</url>
         </repository>
-        <repository>
-            <id>cloudera</id>
-            <url>https://repository.cloudera.com/artifactory/cloudera-repos/</url>
-        </repository>
-        <repository>
-            <id>mapr</id>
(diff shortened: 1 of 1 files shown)


dgomezferro (approved):
That's great @bklo94 !

OlegMazurov:
Can we or external developers still build _spliceengine_ with only _splicemachine-public_ access?

bklo94:
Yes it builds currently, but the builds will break after Jan. So after Jan, core can be built without issue, but not any of the cdh/hdp platforms. Since there are references to platform specific artifacts such as `org.apache.spark:spark-sql_2.11:jar:2.4.0-cdh6.3.0` that will need to solved.

hatyo (approved):

carolp-503 (approved):
+1

PR #5013, opened 2021-01-15
SPLICE-2399 Index access with no predicate on leading column.

msirek:
## Description

This feature allows the use of a primary key or index to avoid scanning all rows when there is no predicate on the leading index column.  Statistics must be collected via ANALYZE TABLE for this access path to be chosen by the optimizer.  The explain text will indicate "IndexPrefixIteratorMode" followed by the estimated number of values in the first column of the index or primary key.

## Example
```
create table t1  (a1 int,
                  b1 int,
                  c1 int,
                  d1 int,
                  e1 int,
                  f1 int,
                  g1 int,
                  h1 timestamp, primary key(a1, b1, c1));
analyze table t1;
explain select * from t1
             where b1 = 1;
```

Plan
```
Cursor(n=3,rows=2,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=3,totalCost=8.024,outputRows=2,outputHeapSize=80 B,partitions=1,parallelTasks=1)
    ->  TableScan[T1(1808),**IndexPrefixIteratorMode(3 values)**](n=0,totalCost=4.004,scannedRows=2,outputRows=2,outputHeapSize=80 B,partitions=1,parallelTasks=1,preds=[(B1[0:2] = 1)])
```

See [SPLICE-2399](https://splice.atlassian.net/browse/SPLICE-2399?focusedCommentId=27973) for more details.
[Performance Test Results](https://docs.google.com/spreadsheets/d/19y0iX3d3k8h5_XHKlAQXrYno2foahvCzZsaU4POXU9U/edit?usp=sharing)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2254/

msirek:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/98/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2257/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/100/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2258/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/101/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2259/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/102/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2263/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/106/

ascend1:
It would take some time to review this change. I'll try to finish it this week.

dgomezferro (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -3658,12 +3822,15 @@ public String printExplainInformation(String attrDelim) throws StandardException
         return sb.toString();
     }
 
-    private String getClassName(String niceIndexName) throws StandardException {
+    private String getClassName(String niceIndexName, String attrDelim) throws StandardException {
         String cName = "";
+        String indexPrefixIteratorString =
```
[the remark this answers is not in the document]
msirek: Yes, this is for the explain.  I had the same idea, but forgot to open a Jira, thanks for reminding me.  We probably want to put some time into figuring out the verbiage and how to list out the index key predicates.  I opened a Jira for it: https://splicemachine.atlassian.net/browse/DB-11239

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
```diff
@@ -35,6 +35,7 @@
 import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
```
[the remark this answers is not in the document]
msirek: Removed.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
```diff
@@ -79,6 +80,7 @@
 	private final UUID	uuid;
 	private final UUID	tableID;
 	private final UUID	schemaID;
+	private FirstColumnOfIndexStats firstColumnStats;
```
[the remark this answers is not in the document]
msirek: Yes, firstColumnStats is an attribute of the conglomerate.  The AccessPath is updated with the ConglomerateDescriptor of the conglomerate it is currently considering, so we can get it via currentAccessPath->conglomerateDescriptor->firstColumnStats.  Putting it directly into AccessPath would duplicate information and require more code to keep the access path updated with the correct information.  All of the information about the conglomerate currently being considered as an access path is typically collected together in the ConglomerateDescriptor.

I have moved setting of the ConglomerateDescriptor firstColumnStats into StoreCostControllerImpl and synchronized it for threading.  It makes more sense to initialize it here.

hbase_storage/src/main/java/com/splicemachine/storage/HScan.java
```diff
@@ -14,18 +14,24 @@
...
+import org.apache.hadoop.hbase.filter.PageFilter;
 import org.apache.hadoop.hbase.util.Bytes;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import static com.splicemachine.db.shared.common.reference.SQLState.LANG_INTERNAL_ERROR;
+
```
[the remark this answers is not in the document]
msirek: Removed.

splice_access_api/src/main/java/com/splicemachine/storage/DataScan.java
```diff
@@ -14,6 +14,7 @@
 
 package com.splicemachine.storage;
 
+import com.splicemachine.db.iapi.error.StandardException;
```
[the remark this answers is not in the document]
msirek: Removed.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2322/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/164/

mem_storage/src/main/java/com/splicemachine/storage/MScan.java
```diff
@@ -14,12 +14,15 @@
...
 
+import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.utils.Pair;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
 import java.util.*;
 
+import static com.splicemachine.db.shared.common.reference.SQLState.LANG_INTERNAL_ERROR;
+
```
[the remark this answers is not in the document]
msirek: Removed.

msirek:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/166/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2324/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2327/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2330/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2331/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2332/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -123,6 +124,12 @@
     long defaultRowCount;
     double defaultSelectivityFactor = -1d;
 
+    private boolean disableIndexPrefixIteration = false;
+
+    // Statistics about the first column in the conglomerate currently
+    // being considered as the access path of this table:
+    private FirstColumnOfIndexStats currentIndexFirstColumnStats = new FirstColumnOfIndexStats();
```
ascend1, line 131: Since the code later test on `currentIndexFirstColumnStats == null`, maybe we don't need to default initialize it to an object here?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1730,6 +1791,28 @@ protected ResultSetNode genProjectRestrict(int numTables)
                 getContextManager());
     }
 
+    private void markFirstColumnReferencedForIndexIteratorMode() throws StandardException {
+        if (numUnusedLeadingIndexFields > 0) {
+            int firstIndexCol = getTrulyTheBestAccessPath().getConglomerateDescriptor().
+                                getIndexDescriptor().baseColumnPositions()[0];
```
ascend1, line 1797: This probably doesn't work for indexes on expressions since `baseColumnPositions()[0]` may not be the actual first index column. It looks like we want to mark the base table column that is defined as the first column in the index? This might be difficult for an index on expression which, for example, has the first column as `a + b + c`.

I'm not sure in which step this function is called. But in modifying the `FromBaseTable`'s access path to an index on expression, `resultColumns` are also rebuilt to reflect the index columns. If this function is used in cost estimation step, `resultColumns` are not changed yet and it's difficult to associate the two sets of columns.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1813,13 +1896,32 @@ else if(isOneRowResultSet()){
...
+                }
+            }
+            if (pred.isStartKey() || pred.isStopKey()){
+                if (numUnusedLeadingIndexFields == 0) {
+                    numUnusedLeadingIndexFields = ap.getNumUnusedLeadingIndexFields();
+                    if (numUnusedLeadingIndexFields >= 1) {
+                        currentIndexFirstColumnStats = trulyTheBestConglomerateDescriptor.getFirstColumnStats();
+                        if (numUnusedLeadingIndexFields > 1)
+                            throw StandardException.newException(LANG_INTERNAL_ERROR,
+                               "IndexPrefixIteratorMode currently allows at most one leading index column to be unspecified in predicates.");
```
ascend1, line 1912: This is the first question I have in mind after reading @msirek 's comment to Jira. Does it make sense to extend the support to multiple leading columns? I guess it's a question about theory vs. reality. Probably it's not common that the number of distinct values in all leading columns are small?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
```diff
@@ -1140,11 +1176,22 @@ private void orderUsefulPredicates(Optimizable optTable,
...
+            boolean indexPrefixMissing = accessPath.getNumUnusedLeadingIndexFields() > 0;
+            int firstColumnIteratedValues = !indexPrefixMissing ? 1 :
+                          (int) accessPath.getConglomerateDescriptor().
+                                           getFirstColumnStats().
+                                           getFirstIndexColumnCardinality();
+            int numConstants = firstColumnIteratedValues;
+            // Allow up to 30x the maximum number of probe values in IndexPrefixIteratorMode
+            // because the first column's values don't need to be compiled in the activation,
+            // so won't contribute to parsing or serialization overhead.
+            int multiplier = indexPrefixMissing ? Math.min(30, numConstants) : 1;
```
ascend1, line 1191: How do we get the value of 30 here? Also, maybe consider define it as a constant.

ascend1 (changes requested):
I reviewed only half of the files and stopped at `DerbyScanInformation.java`. I'll continue from there soon.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: We need it because other code that reads currentIndexFirstColumnStats does not check for null, for example, line 3805 of FromBaseTable.  The goal is to always have it be non-null.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: This code is just trying to mark a column as referenced so that it is built into the row during the IndexScan.  It looks like we don't need to do anything here for index on expressions because all of the columns in the index are marked as referenced here:

>         if (irg.isOnExpression()) {
>             assert !oldColumns.isEmpty();
>             DataTypeDescriptor[] indexColumnTypes = irg.getIndexColumnTypes();
>             ValueNode[] exprAsts = irg.getParsedIndexExpressions(getLanguageConnectionContext(), this);
> 
>             for (int i = 0; i < indexColumnTypes.length; i++) {
>                 ResultColumn rc = (ResultColumn) getNodeFactory().getNode(
>                         C_NodeTypes.RESULT_COLUMN,
>                         indexColumnTypes[i],
>                         null,
>                         getContextManager());
>                 rc.setIndexExpression(exprAsts[i]);
>                 rc.setReferenced();

I will update the code to not even enter this path for index on expressions:

>         IndexRowGenerator indexDescriptor = 
>             getTrulyTheBestAccessPath().getConglomerateDescriptor().getIndexDescriptor();
>         if (numUnusedLeadingIndexFields > 0 && !indexDescriptor.isOnExpression()) {
>             int firstIndexCol = getTrulyTheBestAccessPath().getConglomerateDescriptor().
>                                 getIndexDescriptor().baseColumnPositions()[0];

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: Yes, we'll probably want to wait and see if there is a customer need for it.  It possibly could occur if the first column has a very small number of values, like a region code {west, south, east, north}, or a 2-value field like history rows vs. current rows.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: The idea is that, we are trying to limit both the size of the serialized operation and also the total number of values in the MultiRowRangeFilter.  The index prefix values won't contribute to the serialization size, because they are found out during execution time, but they will make the number of items in the MultiRowRangeFilter greater.  The current default limit in number of combined values is 5000.  If we had 1000 values in the first column, with no change in the code, we may end of building a MultiRowRangeFilter with 5 million items, which seems a bit high.  So the first change is to start our count at 1000 values instead of 1:

>  int numConstants = firstColumnIteratedValues;

But this means our new limit in terms of number of combined values from predicates is reduced to 5, which seems a little small.  To allow for slightly more combined values we use the multiplier 30 which would bring the limit on combined values from predicates up to 150, which results in a theoretical maximum MultiRowRangeFilter size of 150000.  Now, it is possible that filters with millions of ranges could possibly run fine on Spark.  I really haven't had the time to dive into fine tuning this and pushing the limit to find the cutoff point where large MultiRowRangeFilters start getting too expensive.  I've noticed some degradation after a couple hundred thousands ranges, and would need to try a lot more cases to feel comfortable with allowing much larger filters.  So, the value 30 comes from trying to cap the MultiRowRangeFilter at 150000 or so.  I have created a new constant for this:

> private static final int INDEX_PREFIX_MODE_MULTI_PROBE_LIMIT_MULTIPLIER = 30;

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2338/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2340/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2341/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2344/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2345/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2346/

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
ascend1: OK. That makes sense.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
ascend1: I see. I think adding your explanation here (or the number calculations) as a comment here would be very helpful.

splice_access_api/src/main/java/com/splicemachine/storage/DataScan.java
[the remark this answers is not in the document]
dgomezferro: I think this one got left behind.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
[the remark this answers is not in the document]
dgomezferro: There was a discussion during the EU code review time between Youssef, Zhen and me regarding this change, some concerns were raised which I'm trying to summarize here for a follow up discussion.
1. This doesn't feel like the right place for FirstColumnOfIndexStats because it doesn't belong semantically in the ConglomerateDescriptor, which should be an immutable description of a conglomerate rather than be updated at runtime with changing statistics. This is mixing different layers and somewhat increasing complexity.
2. Since `ConglomerateDescriptor`s are cached and shared the synchronization for threading could mean some queries get blocked waiting for one another if they get executed concurrently 
3. Are we writing and reading `firstColumnStats` from inside the same `synchronized` block? Otherwise a query could read the stats meant for a different plan.
4. Does `firstColumnStats` always refer to the whole conglomerate? Is it possible it only refers to a part of it if we have a range predicate on the first column?

Since `FirstColumnOfIndexStats` are computed on the fly they might not belong in `StoreCostControllerImpl` either, Zhen suggested introducing a new type of Controller for these runtime computations, but I think a short meeting for a quick discussion and brainstorm would be ideal.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1813,13 +1896,32 @@ else if(isOneRowResultSet()){
...
+                }
+            }
+            if (pred.isStartKey() || pred.isStopKey()){
+                if (numUnusedLeadingIndexFields == 0) {
+                    numUnusedLeadingIndexFields = ap.getNumUnusedLeadingIndexFields();
+                    if (numUnusedLeadingIndexFields >= 1) {
+                        currentIndexFirstColumnStats.setFrom(trulyTheBestConglomerateDescriptor.getFirstColumnStats());
+                        if (numUnusedLeadingIndexFields > 1)
+                            throw StandardException.newException(LANG_INTERNAL_ERROR,
+                               "IndexPrefixIteratorMode currently allows at most one leading index column to be unspecified in predicates.");
```
hatyo, line 1912: that's also what I thought when I saw some assumptions above in the code about considering only the first column, it is a good start to consider only the first column, nevertheless we should probably consider extending it to cover the first N-no-predicate-index-columns to unlock its full potential.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -3871,8 +4014,33 @@ private PredicateList translateBetweenHelper(PredicateList predList) throws Stan
...
+
+        if (getCompilerContext().getDisablePrefixIteratorMode())
+            return false;
+
+        // We must have statistics collected on the base table to pick IndexPrefixIteratorMode
+        // because the operation may get expensive if the number of values is underestimated.
+        // Also, building of the MultiRowRangeFilter to handle the iteration may get expensive
+        // for large numbers of values, so limit it to 20000 on control and 1000000 on Spark
+        // for now.
+        long maxPrefixIteratorValues = optimizer.isForSpark() ? 1000000 : 20000;
```
hatyo, line 4041: Slightly off topic, but where do these magic numbers come from? does it make sense to have them correlating with some environment parameters? e.g. for Spark we take the number of cores into account `c` and available memory `m` to come up with something like: `1000000*m*c`

hatyo (commented):
I haven't reviewed all the files yet, will continue with the review some time later, wrote a few minor comments.

Reading about your implementation proposal, did you have a look at the [FuzzyRowFilter](https://issues.apache.org/jira/browse/HBASE-6509)? If I understand it correctly, it looks like it can solve the problem even more efficiently as it pushes down the fuzzy filter down to HBase.

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
> I haven't reviewed all the files yet, will continue with the review some time later, wrote a few minor comments.
> 
> Reading about your implementation proposal, did you have a look at the [FuzzyRowFilter](https://issues.apache.org/jira/browse/HBASE-6509)? If I understand it correctly, it looks like it can solve the problem even more efficiently as it pushes down the fuzzy filter down to HBase.

Hi @hatyo, Thanks for the input.  Yes, I looked at that, and was hoping it could work, but I don't think it will due to how we encode column data.  Take this example from https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/filter/FuzzyRowFilter.html :

> Let's assume that we need to fetch all users that performed certain action (encoded as "99") in Jan of any year. Then the pair (row key, fuzzy info) would be the following: row key = "????_99_????_01" (one can use any value instead of "?") fuzzy info = "\x01\x01\x01\x01\x00\x00\x00\x00\x01\x01\x01\x01\x00\x00\x00" I.e. fuzzy info tells the matching mask is "????_99_????_01", where at ? can be any value.

So, if our encoding of a column value always takes up x number of bytes, then we could fill in the first x positions with '?' and then specify the next value.  However, our encodings are not fixed length.  A date of '2020-01-01' may take up 3 bytes, and a date of '2022-01-01'  may take up 4 bytes.  Also there are variable-length data types which could take up any number of bytes.  So, FuzzyRowFilter only seems useful to me if we change our encoding schema for fixed-length data types to always have a fixed-length encoding.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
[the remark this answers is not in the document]
msirek: @dgomezferro OK, so it seems ConglomerateDescriptor isn't the best place for firstColumnStats.  The alternative pattern is that attributes of a conglomerate are computed in a StoreCostController, and then copied over into a CostEstimate object associated with the access path in ScanCostFunction.generateCost.  At the time this information is copied, the AccessPath is not accessed.  Attributes of the AccessPath seem to reflect more the decisions made by the optimizer in terms of how to access the rows.  Following this pattern, I will move firstColumnStats from ConglomerateDescriptor into StoreCostController and copy it into the CostEstimate.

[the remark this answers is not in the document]
msirek: @dgomezferro @ascend1  Regarding the following:

> Since FirstColumnOfIndexStats are computed on the fly they might not belong in StoreCostControllerImpl either, Zhen suggested introducing a new type of Controller for these runtime computations, but I think a short meeting for a quick discussion and brainstorm would be ideal.

I'm thinking that we can add a new dictionary cache to store FirstColumnOfIndexStats associated with each conglomerate, which stays in sync with the partitionStatisticsCache.  This way it doesn't have to be computed on-the-fly each time.  I'm not sure if we want to split it into a different Controller since it is a value derived from tablePartitionStatistics which is also associated with StoreCostController.  What do you think?

splice_access_api/src/main/java/com/splicemachine/storage/DataScan.java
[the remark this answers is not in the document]
msirek: Removed it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -881,10 +931,14 @@ public CostEstimate estimateCost(OptimizablePredicateList predList,
         baseTableRestrictionList.removeAllElements();
         baseTableRestrictionList.countScanFlags();
 
-        currentJoinStrategy.getBasePredicates(predList,baseTableRestrictionList,this);
         /* RESOLVE: Need to figure out how to cache the StoreCostController */
         StoreCostController scc=getStoreCostController(tableDescriptor,cd);
         useRealTableStats=scc.useRealTableStatistics();
+        if (!currentIndexFirstColumnStats.setFrom(cd.getFirstColumnStats())) {
```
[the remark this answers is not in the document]
msirek: I would like to guarantee that firstColumnStats is never null in the storeCostController, so better not to check it.  Also, checking it should have no effect because we always do the first pass with disableIndexPrefixIteration == false before setting  disableIndexPrefixIteration to true and re-costing.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1813,13 +1896,32 @@ else if(isOneRowResultSet()){
...
+                    multiProbing = true;
+                }
+            }
+            if (pred.isStartKey() || pred.isStopKey()){
+                if (numUnusedLeadingIndexFields == 0) {
+                    numUnusedLeadingIndexFields = ap.getNumUnusedLeadingIndexFields();
+                    if (numUnusedLeadingIndexFields >= 1) {
+                        currentIndexFirstColumnStats.setFrom(trulyTheBestConglomerateDescriptor.getFirstColumnStats());
+                        if (numUnusedLeadingIndexFields > 1)
+                            throw StandardException.newException(LANG_INTERNAL_ERROR,
```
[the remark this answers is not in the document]
msirek: PredicateList is coded to only ever set numUnusedLeadingIndexFields to at most 1, so if we ever set it higher, that is a logic bug so I'd like to catch it and fix it.  A customer might not report a warning, but they definitely will report an error.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: Right, though the benefit will decrease the more index prefix values we have, and might at some crossover point be more expensive than doing a full table scan.  So, it is safer to start with 1 column, and then when we have more time to validate the costing and benefit of the multicolumn case, add it in later.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: This is a very good point.  The more parallel Spark tasks we have, then more the work will be spread out to different tasks.  For 8 tasks, each task may only process 1/8 of the values.
I have added:
`    // The maximum number of estimated first index column values
    // per parallel task to support for IndexPrefixIteratorMode.
    private final long MAX_PER_PARALLEL_TASK_INDEX_PREFIX_VALUES = 200000;

    // The absolute maximum number of index values for IndexPrefixIteratorMode.
    // Iterating through these values will be divided by n tasks.
    private final long MAX_ABSOLUTE_INDEX_PREFIX_VALUES = 50000000;`
and
`        long sparkMaxPrefixIteratorValues = Long.min(MAX_ABSOLUTE_INDEX_PREFIX_VALUES,
                                                     parallelism * MAX_PER_PARALLEL_TASK_INDEX_PREFIX_VALUES);
        long maxPrefixIteratorValues = optimizer.isForSpark() ? sparkMaxPrefixIteratorValues :
                                       sparkRowThreshold;`

The value 20000 comes from calling getDetermineSparkRowThreshold().  I added this in instead of using a constant since this value is configurable:
`long sparkRowThreshold = lcc.getOptimizerFactory().getDetermineSparkRowThreshold();`

For the value 1000000, this is just a conservative value since I don't want to use too high of a value and make the optimization too aggressive in the first pass.  I've adjusted this to 200000 per parallel task, but maybe we should make it 150000.  I will try some tests.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
msirek: OK, added this comment:

>             // Allow up to 30x the maximum number of probe values in IndexPrefixIteratorMode
>             // because the first column's values don't need to be compiled in the activation,
>             // so won't contribute to parsing or serialization overhead.
>             // The idea is that, we are trying to limit both the size of the serialized operation
>             // and also the total number of values in the MultiRowRangeFilter. The index prefix values
>             // won't contribute to the serialization size, because they are found out during execution time,
>             // but they will make the number of items in the MultiRowRangeFilter greater.
>             // The current default limit in number of combined values is 5000.
>             // If we had 1000 values in the first column, with no change in the code,
>             // we may end of building a MultiRowRangeFilter with 5 million items, which testing has shown
>             // to not perform well.  
>             // Performance starts to degrade with a MultiRowRange filter of around 150000 items,
>             // so, the value 30 for INDEX_PREFIX_MODE_MULTI_PROBE_LIMIT_MULTIPLIER comes from trying to
>             // cap the MultiRowRangeFilter at 150000 (maxMulticolumnProbeValues (5000) * 30).

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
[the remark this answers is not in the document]
msirek: @dgomezferro Regarding:

> Does firstColumnStats always refer to the whole conglomerate? Is it possible it only refers to a part of it if we have a range predicate on the first column?

It refers to the whole conglomerate.  Currently having a range predicate plus IndexPrefixIteratorMode is not supported.  This would be a useful extension which I thought about implementing.  I don't think it's trivial to do.  We might want to do it as a feature add-on through a separate Jira.

msirek:
@dgomezferro @hatyo @ascend1 I tried to address all of your review comments and questions.  Please take a look when you get the chance and let me know if you'd like some more discussion.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2543/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2544/

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
```diff
@@ -45,6 +45,7 @@
 
 import com.splicemachine.db.iapi.services.uuid.UUIDFactory;
 import com.splicemachine.db.iapi.services.monitor.Monitor;
+import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
```
dgomezferro, line 48: nitpick: this can be removed

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -336,7 +336,7 @@ private PreparedStatement prepMinion(LanguageConnectionContext lcc,
         }
 
         // if anyone else also has this prepared statement,
-        // we don't want them trying to compile with it while
+        // we don't want them trying to compile wLook for stored statements that are in a system sith it while
```
dgomezferro, line 339: Looks like a wrongly pasted comment here

dgomezferro (approved):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/ConglomerateDescriptor.java
msirek: Removed it

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
msirek: Fixed it

ascend1 (approved):

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLUtils.java
```diff
@@ -82,7 +82,13 @@ public static void finishMetadataChange(String changeId) throws StandardExceptio
...
             SpliceLogUtils.trace(LOG,"finishMetadataChange changeId=%s",changeId);
         DDLDriver.driver().ddlController().finishMetadataChange(changeId);
-        DDLDriver.driver().ddlWatcher().clearFinishedChange(changeId);
+        /** TODO-msirek: Try to enable this code.  It would allow more TableDescriptors
+         *               to get cached because it would clear out the DDL change
+         *               as soon as it's committed, which is needed because the cacheIsValid()
+         *               check requires that currChangeCount be zero.
+         *               Currently, enabling this code causes a performance regression.
+         */
+        /* DDLDriver.driver().ddlWatcher().clearFinishedChange(changeId);   */
```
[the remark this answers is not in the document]
msirek: No.  It applies to mem platform only so may not be so important.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2569/

martinrupp:

spliceci:
TEST SUCCEEDED +1

martinrupp (changes requested):
hold, @msirek has seen some IT problems

msirek:
@martinrupp False alarm.  ProjectRestrictSparkExpressionIT is failing on master too.

carolp-503 (approved):
+1

msirek:

spliceci:
TEST SUCCEEDED +1

hatyo (approved):

martinrupp (approved):

PR #5014, opened 2021-01-15
DB-11091 fix null expr. evluation with NOT.

hatyo:
In this PR we fix an issue with evaluating `null` expressions with `NOT` operator, during simplification, we simplify expressions like `NOT <NullExpr>` to `<NullExpr>` which will be simplified further later on as in e.g. join-condition expressions where we simply replace `<NullExpr>` with `false`.

Please have a look `NullPredicateIT.java` for example cases.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/NullPredicateIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/NullPredicateIT.java
@@ -191,9 +191,79 @@ public class NullPredicateIT extends SpliceUnitTest {
     }
 
     @Test
+    public void testNotWithNullInInList_1() throws Exception {
+        String query = format("select count(*) from T --SPLICE-PROPERTIES useSpark=%s\n" +
+                                      "where a not in (cast(null as int), 5)", useSpark);
+
+        String expected = "1 |\n" +
+                "----\n" +
+                " 1 |";
+
+        try(ResultSet rs = methodWatcher.executeQuery(query)) {
+            Assert.assertEquals(expected, TestUtils.FormattedResult.ResultFactory.toStringUnsorted(rs));
+        }
+    }
+
+    @Test
+    public void testNotWithNullInInList_2() throws Exception {
+        String query = format("select count(*) from T --SPLICE-PROPERTIES useSpark=%s\n" +
+                                      "where cast(null as int) not in (3, 5)", useSpark);
+
+        String expected = "1 |\n" +
+                "----\n" +
+                " 0 |";
+
+        try(ResultSet rs = methodWatcher.executeQuery(query)) {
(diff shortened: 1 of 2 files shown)


ascend1 (approved):

hatyo:

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/113/

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/NullPredicateIT.java
```diff
@@ -190,6 +190,34 @@ public void testNullInInList_2() throws Exception {
...
     }
 
+    @Test
+    public void testNotWithNullInInList_1() throws Exception {
+        String query = format("select count(*) from T --SPLICE-PROPERTIES useSpark=%s\n" +
+                                      "where a not in (cast(null as int), 5)", useSpark);
+
+        String expected = "1 |\n" +
+                "----\n" +
+                " 1 |";
```
msirek, line 200: 1 not in (null) is unknown meaning should not match.  I think the count should be zero here.

msirek (changes requested):

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/NullPredicateIT.java
msirek: Our representation of predicates as (true and pred) in the AST could be wrong.  Maybe we need to change our representation or avoid pruning some of the predicates.

hatyo: nice catch @msirek, you're right, the problem is that we're pruning this NULL expression too early from the in-list elements resulting with a semantically different NOT expression, i.e.
`not in(null, x)` evaluates to `not in(x)` instead of `unknown` which evaluates to `false` in join predicate.

as discussed with @ascend1, we special-case this by checking the parent of the in-list, if it is a NOT predicate and it contains a NULL element we simply return the NULL element in order to indirectly trigger the transformation of `not null` to `null`.

msirek (approved):

carolp-503 (approved):
+1

ascend1:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/140/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/140/

spliceci:
TEST SUCCEEDED +1

PR #5015, opened 2021-01-15
DB-10702 improve FK DB2 errors compatibility.

hatyo:
The holy grail of this PR is to assign `LANG_INVALID_FK_COL_FOR_SETNULL` severity [-629](https://www.ibm.com/support/knowledgecenter/SSEPEK_11.0.0/codes/src/tpc/n629.html). The rest is just cleanup and dead code removal.


db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
```diff
@@ -567,6 +567,9 @@ else if (sqlState.equals(SQLState.LANG_CHECK_CONSTRAINT_VIOLATED)) {
         else if (sqlState.equals(SQLState.LANG_FK_VIOLATION)) {
             se.setSeverity(-530);
         }
+        else if (sqlState.equals(SQLState.LANG_INVALID_FK_COL_FOR_SETNULL)) {
```
ipraznik-splice, line 570: Is it possible to add IT for this error code? For SQLState.LANG_FK_VIOLATION there are at least some tests in LoadReplaceModeIT.

ipraznik-splice (changes requested):

ascend1 (approved):

dgomezferro (approved):
I'm OK with the change once an IT gets added

jyuanca (approved):

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
carolp-503: I suggest that we Jira this as a separate item if it's going to take very long at all to construct the IT.

carolp-503 (approved):
+1.  

Check with @hatyo about adding a new IT per @ipraznik-splice 's comment.

db-engine/src/main/java/com/splicemachine/db/impl/jdbc/TransactionResourceImpl.java
hatyo: Thanks, I added a test for `SQLState.LANG_FK_VIOLATION` in `ForeignKeyActionIT`.

ipraznik-splice (approved):

PR #5047, opened 2021-01-20
DB-11141: return a warning if no rows are affected by a DML(3.0)

jyuanca:


carolp-503 (approved):
Approved for 3.0

carolp-503 (approved):
Changes look good

PR #5049, opened 2021-01-20
DB-11161 Reset JUMP_MODE to READY_TO_JUMP in prepForNextRound()

ascend1:


ascend1:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/135/

ascend1:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2152,13 +2173,23 @@ private void pullOptimizableFromJoinOrder() throws StandardException{
...
     private boolean isPreparedStatement() {
         LanguageConnectionContext lcc =
             (LanguageConnectionContext) ContextService.
                 getContextOrNull(LanguageConnectionContext.CONTEXT_ID);
         if (lcc == null || lcc.getActivationCount() < 1)
             return false;
-        return lcc.getLastActivation().getPreparedStatement() != null;
+        Activation ac = lcc.getLastActivation();
+        ResultDescription rd = ac.getResultDescription();
+        return ac.getPreparedStatement() != null && rd != null &&
```
msirek, line 2191: Nit: you could save a couple method calls by checking ac.getPreparedStatement() == null first, and returning right away if that is true.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/JoinOrderJumpModeIT.java
```diff
@@ -0,0 +1,262 @@
...
+    @ClassRule
+    public static TestRule chain = RuleChain.outerRule(spliceClassWatcher)
+            .around(spliceSchemaWatcher);
+    @Rule
+    public SpliceWatcher methodWatcher = new SpliceWatcher(CLASS_NAME);
+
+    public static void createData(Connection conn, String schemaName) throws Exception {
+
+        new TableCreator(conn)
+                .withCreate("CREATE TABLE \"TGSCHFAKT\" (\n" +
```
msirek, line 33: We should change the column names or table names if this DDL is coming from a customer.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2152,13 +2173,23 @@ private void pullOptimizableFromJoinOrder() throws StandardException{
...
+     */
     private boolean isPreparedStatement() {
         LanguageConnectionContext lcc =
             (LanguageConnectionContext) ContextService.
                 getContextOrNull(LanguageConnectionContext.CONTEXT_ID);
         if (lcc == null || lcc.getActivationCount() < 1)
             return false;
-        return lcc.getLastActivation().getPreparedStatement() != null;
+        Activation ac = lcc.getLastActivation();
+        ResultDescription rd = ac.getResultDescription();
```
msirek, line 2190: I don't see the assertion in BaseActivation.getResultDescription removed.  Was this change not required after all?

msirek (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
ascend1: It is removed. Please check `BaseActivation.java` in this change.

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2296/

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/138/

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/JoinOrderJumpModeIT.java
ascend1: Yes, you're right. I renamed them in the latest commit.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
ascend1: I think I get the point of this comment and the other about assertion is not removed. There are two `ResultDescription`s that we can get:
```
ac.getResultDescription();
```
and
```
ac.getPreparedStatement().getResultDescription();
```
At a quick glance, they return different objects. One is a member in `BaseActivation` and the other is a member in `GenericPreparedStatement`. Not sure if they eventually point to the same thing, though.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2152,13 +2173,23 @@ private void pullOptimizableFromJoinOrder() throws StandardException{
...
+        return ac.getPreparedStatement() != null && rd != null &&
+                !"Explain".equals(rd.getStatementType());
```
ipraznik-splice, line 2192: I see also many other places with if statements to compare type against string values, like "explain", "sparkexplain", "EXPLAIN", etc. Maybe it makes sense to move all these values as constants to the common class to have one place with all possible values.

ipraznik-splice (approved):

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2334/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: OK, thanks.

ascend1:
The failing test indicates that `explain` on `update` statement with hint on the target table will be different than the actual plan. That means`isPreparedStatement()` is still risky. @msirek Why do we need to give the update statement 2 seconds to let it take the hint?

carolp-503 (approved):
+1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2351/

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #5053, opened 2021-01-21
DB-11125 Allow column definitions in with clause (3.1)

ascend1:


carolp-503 (commented):
Please close - this change will go into master and be released as part of the 3.2 branch

PR #5068, opened 2021-01-22
DB-11162 support CurrentFunctionPath JDBC parameter.

hatyo:
In this PR we support the `CurrentFunctionPath` parameter, according to [DB2 documentation](https://www.ibm.com/support/producthub/db2/docs/content/SSEPGG_11.5.0/com.ibm.db2.luw.apdv.cli.doc/doc/r0008777.html), this parameter defines ...:

> ... the path used to resolve function references and data type references that are used in dynamic SQL statements. It contains a list of one or more schema-names, where schema-names are enclosed in double quotation marks and separated by commas.

So for example, if you define a UDF called `MyFunc` in schema `S1` and you do the following:

```
set schema S2; -- different schema
select MyFunc from sysibm.sysdummy1; -- you get an error that the function doesn't exist.
```

Now if you set this JDBC parameter in the connection string to refer to schema `S1`, you get the correct results:

```
connect 'jdbc:splice://localhost:1527/splicedb;user=<USER>;password=<PASSWORD>;CurrentFunctionPath="S1"';
set schema S2; -- different schema
select MyFunc from sysibm.sysdummy1; -- now works!
```

(you can find more tests in `CurrentFunctionPathIT.java`).

Note that according to DB2 documentation, this parameter can not override built-in procedures and functions because user-defined schemas are always preceded with `SYSIBM` and `SYSFUN`.

> The default value is "SYSIBM","SYSFUN",X where X is the value of the USER special register delimited by double quotation marks. The schema SYSIBM does not need to be specified. If it is not included in the function path, then it is implicitly assumed as the first schema.

During evaluation, we throw 42Y06 (LANG_CURRENT_FUNCTION_PATH_SCHEMA_DOES_NOT_EXIST) exception if a schema was not found. Notice that the evaluation of invalid schemas is done lazily, i.e. if the function we're looking for was found in a valid schema which precedes an invalid schema in order we do not throw an exception (check `CurrentFunctionPathIT.java` for concrete scenarios).


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/StaticMethodCallNode.java
```diff
@@ -222,54 +227,36 @@ public JavaValueNode bindExpression(FromList fromList,
...
-                setIsSystemFunction(true);
+            Object candidateSchemasProperty = getLanguageConnectionContext()
+                    .getSessionProperties()
+                    .getProperty(SessionProperties.PROPERTYNAME.CURRENTFUNCTIONPATH);
+            SchemaDescriptor sd = null;
+            if(candidateSchemasProperty != null) {
+                String[] candidateSchemas = (String[])(candidateSchemasProperty);
+                for(String candidateSchema : candidateSchemas) {
+                    try {
+                        sd = getSchemaDescriptor(candidateSchema, candidateSchema != null);
```
jyuanca, line 238: Is candidateSchema != null always true? A cleaner approach is not to raise exception and then check whether the schema exists
sd = getSchemaDescriptor(candidateSchema, false);
if (sd == null)
  throw StandardException.newException(SQLState.LANG_CURRENT_FUNCTION_PATH_SCHEMA_DOES_NOT_EXIST,Arrays.toString(candidateSchemas),candidateSchema);

jyuanca (approved):
approve with minor comments

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/SessionPropertiesImpl.java
```diff
@@ -104,11 +111,59 @@ public void setProperty(PROPERTYNAME property, Object value) {
...
+            return new String[] {"SYSIBM", "SYSFUN"};
+        } else {
+            // source: https://stackabuse.com/regex-splitting-by-character-unless-in-quotes/
+            List<String> candidateSchemas = new ArrayList<>(Arrays.asList(value.split(",(?=([^\"]*\"[^\"]*\")*[^\"]*$)")));
+            for(int i = 0; i < candidateSchemas.size(); ++i) {
+                candidateSchemas.set(i , normalizeAndUnquoteIdentifier(candidateSchemas.get(i)));
+            }
+            boolean sysibmSeen = false;
+            boolean sysfunSeen = false;
+            for(int i = 0; i < candidateSchemas.size(); ++i) {
```
ipraznik-splice, line 139: not sure if it's critical, but if for example SYSIBM will not on the first position, it will be added again to the first place.
input = [AAA, SYSIBM, BBB]
output = [SYSIBM, SYSFUN, AAA, SYSIBM, BBB]
there will be 2 items SYSIBM in the list, Maybe better to use indexOf method and move/add to the first place.

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/SessionPropertiesImpl.java
```diff
@@ -104,11 +111,59 @@ public void setProperty(PROPERTYNAME property, Object value) {
...
+            if(!sysfunSeen) {
+                candidateSchemas.add(1, "SYSFUN");
+            }
+            return candidateSchemas.toArray(new String[0]);
+        }
+    }
+
+    private String normalizeAndUnquoteIdentifier(String identifier) {
+        String normalized = StringUtil.normalizeSQLIdentifier(identifier);
+        if(!normalized.isEmpty() && normalized.charAt(0) == '"' && normalized.charAt(normalized.length() - 1) == '"') {
```
ipraznik-splice, line 161: another possibility, not sure if it work faster, but looks shorter:)
normalized.replaceAll("^\"|\"$", "");

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/StaticMethodCallNode.java
hatyo: Good point, thanks I fixed it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/SessionPropertiesImpl.java
hatyo: Thanks Igor for catching this, it is not clear to me from [DB2 documentation](https://www.ibm.com/support/producthub/db2/docs/content/SSEPGG_11.5.0/com.ibm.db2.luw.apdv.cli.doc/doc/r0008777.html) what's exactly the expected behavior in this case but I think it might be ok to leave it like this because if we are searching for function e.g. `cos` then we have two cases:

1. it exists in `SYSIBM` therefore we call it first regardless if `SYSIBM` was manually added by the user in some order.
2. it doesn't exist in `SYSIBM`, then we skip the first `SYSIBM` and any other occurance of it potentially specified by the user.

Therefore in both cases, it doesn't matter much if the user specified it that's why I just follow what DB2 says:

> The default value is "SYSIBM","SYSFUN",X where X is the value of the USER special register delimited by double quotation marks

What do you think?

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/SessionPropertiesImpl.java
hatyo: Interesting idea, I thought about this and I think it allows invalid schemas such as `"FOO` to be converted to `FOO` which is not correct, this is because of the OR in the Regex `"^\\\"|\\\"$"`, i.e.:
```
String normalized = "hello world\"";
System.out.println(normalized.replaceAll("^\\\"|\\\"$", ""));
// returns: hello world
```
To fix this, we could maybe use Regex's `()()`'s AND syntax as [explained here](https://stackoverflow.com/questions/6185438/and-operator-in-regular-expressions) to achieve this, then we end up with something like this `normalized.matches("(?=.*^\\\")(?=.*\\\"$).*")` which honestly looks a bit more clumsy than current version :D 

Let me know if there is a better way to do this

ipraznik-splice: sorry, I missed the fact, that it replaces only when quotes are on both sides. Then let's keep so, as it easier to read compare to regex:)

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/SessionPropertiesImpl.java
ipraznik-splice: it's ok for me to keep so

ipraznik-splice (approved):

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2350/

carolp-503 (approved):
+1

hatyo:

spliceci:
TEST SUCCEEDED +1

PR #5072, opened 2021-01-22
Add parameter, fix archiving and regex

bklo94:
Create a debugging node option for spliceengine-branch. 

-  Creates a non-spot instanced node that can be ssh'd into and debugged for issues. The ordinary standalone spot instances is under an internal VPC and will randomly get killed if the spot capacity is low (Possibly won't stay up long)

Switch spliceengine-platform pod to small ec2 node. 

- The spliceengine-platform flyweight pod currently dies after about an hour of idle time. The platform_it's take about 3ish hours to complete currently. While the tests still run, the flyweight job is killed, so even if the platform_it's all complete, the overall job will fail.

Remove ** from spliceengine-PR and spliceengine-platform

- There is not enough space in the disk to archive the entire workspace on jenkins master, since all job workspace are being pushed onto it. It will not scale effectively if the entire workspace is saved. 

Fix the jenkins job regex. 

- The `\` is not accepted by the job.

### DIFF
--- a/pipelines/spliceengine-PR/Jenkinsfile
+++ b/pipelines/spliceengine-PR/Jenkinsfile
@@ -110,7 +110,7 @@ node('spliceengine'){
         throw any
 
     } finally {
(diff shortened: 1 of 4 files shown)


ipraznik-splice (approved):

msirek (approved):

carolp-503 (approved):
+1

PR #5076, opened 2021-01-24
DB-11252 Fix JoinConditionVisitor childParentMap for VALUES clauses with subqueries.

msirek:
Fixes VALUES clause queries or triggers that use multiple joins and/or subqueries, e.g.

> CREATE TRIGGER FeatureTable_History_UPDATE_PREHISTORIC
> AFTER UPDATE ON FeatureTable
> REFERENCING OLD_TABLE AS OLDW NEW_TABLE AS NEWW
> FOR EACH STATEMENT
> WHEN (
>      (NOT EXISTS (SELECT 1 FROM FeatureTable_History H,NEWW N WHERE H.ENTITY_KEY=N.ENTITY_KEY AND N.LAST_UPDATE_TS >= H.ASOF_TS AND N.LAST_UPDATE_TS < H.UNTIL_TS))
>      AND
>      (EXISTS (SELECT 1 FROM OLDW,NEWW WHERE OLDW.ENTITY_KEY=NEWW.ENTITY_KEY AND OLDW.LAST_UPDATE_TS < NEWW.LAST_UPDATE_TS))
> )
> ...

The WHEN clause is converted to a VALUES clause internally.  Since the JoinConditionVisitor doesn't scan the entire RowResultSetNode tree for ResultColumns, an incomplete joinChainMap may be built leading to some column references not being found in the map, causing the statement to error out.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/ast/JoinConditionVisitor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/ast/JoinConditionVisitor.java
@@ -104,6 +104,12 @@ public class JoinConditionVisitor extends AbstractSpliceVisitor {
     }
 
     @Override
+    public RowResultSetNode visit(RowResultSetNode node) throws StandardException {
+        initializeMap(node);
+        return node;
+    }
+
+    @Override
     public JoinNode visit(JoinNode j) throws StandardException {
         if (LOG.isDebugEnabled())
             LOG.debug(String.format("visit joinNode=%s",j));
--- a/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Statement_IT.java
+++ b/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Statement_IT.java
@@ -16,6 +16,7 @@ package com.splicemachine.triggers;
 
 import com.splicemachine.db.client.am.ResultSet;
 import com.splicemachine.derby.test.framework.SpliceSchemaWatcher;
+import com.splicemachine.derby.test.framework.SpliceUnitTest;
 import com.splicemachine.derby.test.framework.SpliceWatcher;
 import com.splicemachine.derby.test.framework.TestConnection;
 import com.splicemachine.test.SerialTest;
@@ -40,7 +41,7 @@ import static org.junit.Assert.fail;
  */
 @Category(value = {SerialTest.class})
 @RunWith(Parameterized.class)
-public class Trigger_Statement_IT {
+public class Trigger_Statement_IT extends SpliceUnitTest {
 
     private static final String SCHEMA = Trigger_Statement_IT.class.getSimpleName();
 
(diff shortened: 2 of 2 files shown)


msirek:

cloudspliceci:
TEST FAILED
https://jenkins.build.splicemachine-dev.io/job/AutomatedTesting/job/spliceengine-PR/163/

spliceci:
TEST SUCCEEDED +1

ascend1 (approved):
Looks good. I can think of `RowResultSetNode` to be also in `insert/update/delete` statements, `execute` statement, and `values` statement. There should be no problem in `execute` and `values` statements. For the DMLs, however, I have a minor concern (minor since tests are all passed here):

ascend1 (commented):
After a second thought, I change it from approved to comment just to make sure that we double-check the `RowResultSetNode`s in DMLs.

ascend1 (commented):
Please see my comment above.

msirek:
> Looks good. I can think of `RowResultSetNode` to be also in `insert/update/delete` statements, `execute` statement, and `values` statement. There should be no problem in `execute` and `values` statements. For the DMLs, however, I have a minor concern (minor since tests are all passed here):
> 
> For example, we also possibly rewrite sub-select into `values ()` in `UpdateNode`. I have never dig into this part, but I've seen it in code and comment (`UpdateNode.java`, around line 799):
> 
> ```
>            /**
>              * When the underneath SelectNode is unsatisfiable, we rewrite it to
>              * a RowResultSetNode representing "values (null, null, ..., null) where 1=0"
>              * In such a case, it is possible that the target table does not participate in planning and thus
>              * its accessPath remains as null
>              */
> ```
> 
> Maybe we should check that this change doesn't bring any regression to this code path.

OK, I will take a look.  The reason I don't think there is an issue though, is that we only build the childParentMap once for a query, so we want to invoke the building of it at the highest node possible in the tree, so that this code can see the entire tree of joins:

>     public LongLongHashMap childParentMap()
>             throws StandardException {
>         // Lots of duplication, thus the HashSet
>         Set<ResultColumn> cols = new HashSet<>(RSUtils.collectNodes(this, ResultColumn.class));

.  If the RowResultSetNode exists below a ProjectRestrictNode, JoinNode, UnionNode, IntersectOrExceptNode, the code change in this Jira will have no effect because the building of it will be triggered when the top-level node of these other types is seen.  The change only has an effect when the RowResultSetNode is itself close to being the top level node, for example:

> VALUES (exists (select 1 from t1,t2 where t1.a = t2.a) and not exists (select 1 from t3,t4 where t3.a = t4.a));

ascend1:
> > Looks good. I can think of `RowResultSetNode` to be also in `insert/update/delete` statements, `execute` statement, and `values` statement. There should be no problem in `execute` and `values` statements. For the DMLs, however, I have a minor concern (minor since tests are all passed here):
> > For example, we also possibly rewrite sub-select into `values ()` in `UpdateNode`. I have never dig into this part, but I've seen it in code and comment (`UpdateNode.java`, around line 799):
> > ```
> >            /**
> >              * When the underneath SelectNode is unsatisfiable, we rewrite it to
> >              * a RowResultSetNode representing "values (null, null, ..., null) where 1=0"
> >              * In such a case, it is possible that the target table does not participate in planning and thus
> >              * its accessPath remains as null
> >              */
> > ```
> > 
> > 
> > Maybe we should check that this change doesn't bring any regression to this code path.
> 
> OK, I will take a look. The reason I don't think there is an issue though, is that we only build the childParentMap once for a query, so we want to invoke the building of it at the highest node possible in the tree, so that this code can see the entire tree of joins:
> 
> > ```
> > public LongLongHashMap childParentMap()
> >         throws StandardException {
> >     // Lots of duplication, thus the HashSet
> >     Set<ResultColumn> cols = new HashSet<>(RSUtils.collectNodes(this, ResultColumn.class));
> > ```
> 
> . If the RowResultSetNode exists below a ProjectRestrictNode, JoinNode, UnionNode, IntersectOrExceptNode, the code change in this Jira will have no effect because the building of it will be triggered when the top-level node of these other types is seen. The change only has an effect when the RowResultSetNode is itself close to being the top level node, for example:
> 
> > VALUES (exists (select 1 from t1,t2 where t1.a = t2.a) and not exists (select 1 from t3,t4 where t3.a = t4.a));
> 
> Also, for the Unsatisfiable query conversion to 'values (null, null, ..., null) where 1=0', in that case there will be no column references in the tree, so nothing to be added to childParentMap, and no columns that need to be looked up.

I see, that makes sense. I was not into that code path so just being cautious. I think the change is good.

ascend1 (approved):

jpanko1 (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #5079, opened 2021-01-26
DB-10273 progress bar

martinrupp:
# Short Description
This adds progress bar support in sqlshell. You can enable it with the command `progressbar on`.
This is done via the system procedure `SYSCS_UTIL.SYSCS_GET_PROGRESS(filter, isDRDA)`. filter can be either empty or the UUID for the operation to get. isDRDA can be 1 (using the drda connection token as UUID) or 0 (using the normal UUIDs). `SYSCS_GET_PROGRESS` could also be used by e.g. Cloud UI.

# Long Description

## progress bar looks

The progress bar looks like this
```
<job/stage information> (Job J, Stage X of Y)
[------------+----------25%----------+-----------50%-----------+----------75%-----------+-----------]
```
Depending on your progress, only part of that bar is displayed, e.g. for ~30%:
```
[------------+----------25%-----
```
Progressbar is forward-only, so not relying on ANSI Escape codes (like ESC[PnD - Cursor Backward). That way, it should work in all terminals and also when streaming into a file.

#### OlapExecutor
- Have a SparkListener and calculate a progressStr from the notifications we get
- Set progressStr in OlapStatus
- in intervals, we already send `OlapMessage.ProgressResponse` to indicate if the job is still alive.
- add a progressStr to OlapMessage.ProgressResponse by getting the one from Olap Status
- Send OlapMessage.ProgressResponse to server

#### Server
- receive OlapMessage.ProgressResponse and notify the current running job if we received a progressStr
- if the current job knows the UUID of the associated running operation, set the progressStr in the OperationManager
- system procedure SYSCS_GET_PROGRESS would return that progressStr for each running operation

#### sqlshell
- when executing a query, have a thread in the background check every second the SYSCS_GET_PROGRESS to get the progress of the just started query, print to console.

#### in sqlshell
In sqlshell's utilMain, before calling `ijParser.ijStatement` (which executes the current statement), we construct a `ProgressThread`. When started, this checks in the background the progress of the current sql. For this background check, we call `ijCommands.getProgress()`, which executes a `getProgress` on `currentStatement` (see ijImpl.getProgress). This effectively calls `Statement.getProgress`. In this, we have access to the DRDA connection token, which makes it possible to associate the pick the correct progress by its token: We execute a `call SYSCS_UTIL.SYSCS_GET_PROGRESS()` query, and if column 1 equals the DRDA connection toekn, it returns column 9 (state).

#### SYSCS_GET_PROGRESS
SYSCS_GET_PROGRESS is very similar to SYSCS_GET_RUNNING_OPERATIONS. In fact so similar it shares the same code, but just exchanges column 9 jobtype (`ro.getOperation().getScopeName()`) with state (`ro.getProgressString()`).
That RunningOperation.progressString is set by `RemoteQueryJob.notify`. For this, RemoteQueryJob will now be passed the  `runningOperationUUID` to access the RunningOperation via the OperationManager.
`RemoteQueryJob.notify` in turn will be called when AsyncOlapNIOLayer's ResultHandler receives a SubmittedResult.
SubmittedResult will be created from AsyncOlapNIOLayer's parseFromResponse, which gets a `OlapMessage.Response`, which is a protobuf message which already had the extension `ProgressResponse`, which we added a string `progressStr` to. This `ProgressResponse` is a Protobuf message that is sent from the Executor to the RegionServer to notify the server if the job is still running, finished or aborted.

#### Spark Listener
In `QueryJob.call`, around `dataset = root.getDataSet(dsp);`, we added the class `SparkProgressListener`. This adds with `sparkContext.addSparkListener(this)` itself to get spark notifications. Since we then get notifications of ALL running operations, we set a UUID to our spark context, identifying the job to the sparkContext. With this, we intercept messages `onJobStart` and `onTaskStart` and `onTaskEnd`, and setting a progress string to a `OlapStatus` object.
In regular intervals, the Executor will use `OlapSerializationUtils.buildResponse` to inform the server about the current job state (NOT_SUBMITTED/SUBMITTED/RUNNING/CANCELED/FAILED). For SUBMITTED/RUNNING, it constructs the already mentioned `OlapMessage.ProgressResponse`, in which the progressString is set by accessing the one form the OlapStatus.

## Todo
- the information is currently stored in a column of `SYSCS_GET_PROGRESS` as a string, which has to be parsed out again. I'm not sure we should leave it like this, but maybe rather change the columns of `SYSCS_GET_PROGRESS` to really have a column for jobs/stages/tasks .
- add some synchronized areas around get/setProgressStr calls.

## Example
```
splice> progressbar on;
ELAPSED TIME = 0 milliseconds
splice> select min(ws_sold_date_sk) from web_sales500;
[Running ...                                                                                        ]
Produce Result Set: Listing leaf files and directories for 499 paths:
s3a://*******/web_sales500/ws_sold_date_sk=2450816, ... (Job 2, Stage 1 of 1)
[------------+----------25%----------+---
```


martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2798/

martinrupp:
com.splicemachine.derby.utils.SpliceAdmin_OperationsIT failing.

db-shared/src/main/java/com/splicemachine/db/shared/ProgressInfo.java
```diff
@@ -0,0 +1,160 @@
...
+package com.splicemachine.db.shared;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ *  Information about the progress of a running operation.
+ *  JOB: exactly ONE job running at each time
+ *  JOB is sequence of STAGES:
+ *       STAGE exactly ONE stage running at each time
+ *       STAGE is a sequence of TASKS:
```
dgomezferro, line 24: There could be more than one job and/or stage running at the same time.
When we have multiple (native spark) broadcast joins we should have multiple concurrent jobs at once, one for each broadcast resultset.
For stages, some Spark operations introduce concurrency, for instance both branches of a mergeSortJoin can run concurrently

splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/OperationManagerImpl.java
```diff
@@ -36,8 +38,18 @@
...
+    {
+        return operations.get(uuid);
+    }
+
     public UUID registerOperation(SpliceOperation operation, Thread executingThread, Date submittedTime, DataSetProcessor.Type engine, String rdbIntTkn) {
         Random rnd = ThreadLocalRandom.current();
+        // Question is if we really need two different UUIDs here.
+        // seems like we don't always have a rdbIntTkn, but shouldn't we maybe only then create a custom UUID?
+        // like if(rdbIntTkn == null) rdbIntTkn = new UUID(rnd.nextLong(), rnd.nextLong()).toString(); ?
+        // HOWEVER, it might be that rdbIntTkn is not really unique, but only per connection?
```
dgomezferro, line 52: If I remember correctly you are right, rdbIntTkn is per connection.

hbase_sql/src/test/java/com/splicemachine/derby/utils/ProgressThreadIT.java
```diff
@@ -0,0 +1,112 @@
...
+import com.splicemachine.db.impl.tools.ij.ProgressThread;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+
+public class ProgressThreadIT {
```
dgomezferro, line 27: This should be a Test, not an IT (ITs depend on an instance of the DB running)

dgomezferro (changes requested):
This is great! I added some questions/suggestions. Can you also add an end-to-end IT?

db-shared/src/main/java/com/splicemachine/db/shared/ProgressInfo.java
dgomezferro: Just tested it with a mergeSortJoin and it appends both progress bars to the same line, resulting in something like this:
```
splice> select count(*) from a natural join a;
[Running ...                                                                                        ]
Produce Result Set: select count(*) from a natural join a (Job 2, Stage 1 of 4)
[-----------------+----------25%---+----------25%-----------+----------25%----------+-----------50%--------25%----------+-----------50%-----------+-----------+-----------50%-----------+-------------+-----------50%-----------+----------75%---+-----------50%-----------+----------75%---------------50%-----------+----------75%-----------+-0%-----------+----------75%-----------+------------+----------75%-----------+------
```

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ProgressThread.java
```diff
@@ -0,0 +1,140 @@
...
+                if(!hadException) {
+                    outputStream.println("\nException while parsing/updating ProgressInfo: " + e.toString());
+                    hadException = true; // ignore following to not spam
+                }
+            }
+        }
+    }
+
+    public void run() {
+        outputStream.println(firstRunning);
```
dgomezferro, line 115: Would it be possible to print this / start this thread only for Spark queries?

db-shared/src/main/java/com/splicemachine/db/shared/ProgressInfo.java
dgomezferro: With 2 jobs it fails with this exception:
```
create table a(i int); -- insert 12582912 rows into a
select * from a inner join (select count(*) as i from a
> ) b --splice-properties joinStrategy=broadcast
> on a.i = b.i
> inner join (select count(*) as i from a
> ) c --splice-properties joinStrategy=broadcast
> on a.i = c.i;
[Running ...                                                                                        ]

Exception while parsing/updating ProgressInfo: java.lang.NumberFormatException: For input string: ")"
```

martinrupp:
to test the behavior that daniel described:
```
create table a(i int);
insert into a values 1, 2, 3, 4, 5, 6, 7, 8, 9, 10;
insert into a select i+10 from a;
insert into a select i+20 from a;
insert into a select i+40 from a;
insert into a select i+80 from a;
insert into a select i+160 from a;
insert into a select i+320 from a;
insert into a select i+640 from a;
insert into a select i+1280 from a;
insert into a select i+2560 from a;
insert into a select i+5120 from a;
insert into a select i+10000 from a;
insert into a select i+20000 from a;
insert into a select i+40000 from a;
insert into a select i+80000 from a;
insert into a select i+160000 from a;
insert into a select i+320000 from a;
insert into a select i+640000 from a;
insert into a select i+1280000 from a;
insert into a select i+2560000 from a;
insert into a select i+5120000 from a;
insert into a select i+10240000 from a; -- will show progress bar

select count(*) from a; -- will show progress bar

-- this will result in one normal progress bar, and then one with "small jobs"
select * from a inner join (select count(*) as i from a ) b --splice-properties joinStrategy=broadcast
 on a.i = b.i inner join (select count(*) as i from a ) c --splice-properties joinStrategy=broadcast
 on a.i = c.i;
```

hbase_sql/src/test/java/com/splicemachine/derby/utils/ProgressThreadIT.java
martinrupp: done

splice_machine/src/main/java/com/splicemachine/derby/iapi/sql/execute/OperationManagerImpl.java
martinrupp: ok, comment removed

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ProgressThread.java
martinrupp: currently not

martinrupp:

dgomezferro (approved):
Those queries don't break anymore, looking good!

It still doesn't update for stages in a job, but this is already very useful as it is.

martinrupp:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3012/

splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
```diff
@@ -1712,30 +1713,14 @@ public static void SYSCS_GET_SESSION_INFO(final ResultSet[] resultSet) throws SQ
...
-        EmbedConnection conn = (EmbedConnection)getDefaultConn();
-        Activation lastActivation = conn.getLanguageConnection().getLastActivation();
-        IteratorNoPutResultSet resultsToWrap = new IteratorNoPutResultSet(rows, runningOpsDescriptors, lastActivation);
-        try {
-            resultsToWrap.openCore();
-        } catch (StandardException se) {
-            throw PublicAPI.wrapStandardException(se);
-        }
-        resultSet[0] = new EmbedResultSet40(conn, resultsToWrap, false, null, true);
+    public static void SYSCS_GET_PROGRESS(String drdaToken, Integer getDRDA, final ResultSet[] resultSet) throws SQLException{
```
ipraznik-splice, line 1721: If I understand correctly, SYSCS_GET_PROGRESS returns a formatted progress string. Is it possible to create another function or the same with a parameter to return a int value - progress in percents? It would easier to integrate with other systems, like cloud manager

ipraznik-splice (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
martinrupp: the problem is this
we need the following information:
```
jobNumber numCompletedStages numStages numCompletedTasks numActiveTasks
jobname (may include \n)
```
you can only calculate the percentage complete of **one stage**, by `(numCompletedTasks*100)/numTasks`.
You don't know how many  jobs there are, and often jobs only have one stage.

so, what spark does, it that it will display the following:
`[ Stage $numCompletedStages =======>--------------- ( $numCompletedTasks + $numActiveTasks ) / $numTasks`

e.g.
```
[Stage 2:============================================>              (3 + 1) / 4]
``` 
Now if you have multiple jobs/stages, this is just overwritten in the sparkshell, so there's no global indication of how long this command will take, just an approximation of how long the current stage is.

What GET_PROGRESS will currently do is to return back to the caller all the information we have (jobNumber numCompletedStages numStages numCompletedTasks numActiveTasks jobname), so from there you can generate a progress bar how you like it.
In a UI, where you can do overwrite, i would suggest printing all information like
`[ Job $jobNumber ($jobName), Stage $numCompletedStages / $numStages =======>--------------- ( $numCompletedTasks + $numActiveTasks ) / $numTasks`

You can't overwrite in sqlshell, so we print the `Job $jobNumber ($jobName), Stage $numCompletedStages / $numStages` only if they change, and then we would start a new progress bar.
In the example above, your first job is one where you get a nice progress bar (has a stage with a lot of tasks).
After the first job, you get 10 really small jobs, with each one stage and 2 Tasks. So this is why i added the "smallTasks" boolean. In this case, we don't print a new progress bar for each of these tasks, but combine them together into one progress bar (anyhow, a progressbar with only 2 tasks wouldn't have much information).

The result is 2 nicely progressing progressbars, instead of ~12 where 11 of them aren't really progressing but just finished immediately.

Note that if you're in a UI, you don't have to worry as much about this "smallTasks" problem, as when you can overwrite the UI, you don't have to worry about cluttering the console.

Both approaches require very different information, so this is why i chose to just hand over the raw information.

martinrupp:

splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
ipraznik-splice: Thank you for the detailed explanation!

ipraznik-splice (approved):

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3025/

martinrupp:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3039/

martinrupp:

cloudspliceci:
TEST SUCCEEDED +1

cloudspliceci:
TEST SUCCEEDED +1

PR #5091, opened 2021-01-29
DB-9898 fix drop column behavior.

hatyo:


hatyo:
Making the PR a draft since we still have issues with index scans after a column is dropped

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
```diff
@@ -333,16 +340,16 @@ public KVPair translate(KVPair mutation) throws IOException, StandardException {
...
                         accumulate(keyAccumulator, indexKeyPos,
                                 formatId,
-                                index.getDescColumns(indexKeyPos),
+                                index.getDescColumns(cnt),
                                 keyDecoder.array(), offset, length);
                     }
                 } else {
-                    decodedRow.setColumn(sourceKeyColumnPos + 1,
-                            DecodeBaseRowColumn(keyDecoder, offset, length, sourceKeyColumnPos, formatId));
+                    decodedRow.setColumn(pkPos + 1,
```
aimiebell, line 351: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

PR #5092, opened 2021-01-30
DB-11265 Remove SpliceClient check from BroadcastedActivation

jpanko1:


jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2365/

jyuanca (approved):

jpanko1:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BroadcastedActivation.java
```diff
@@ -98,9 +98,9 @@ public ActivationHolder getActivationHolder() {
         return baos.toByteArray();
     }
 
-    private boolean isSparkDriver() {
+    private boolean isOlapWorker() {
         String threadName = Thread.currentThread().getName();
-        return (threadName.startsWith("olap-worker-") || SpliceClient.isClient());
```
msirek, line 103: Do we need to replace SpliceClient.isClient() with something else instead of getting rid of it entirely?  In this code I was trying to avoid a similar problem as found in DB-10707 where the isOlapServer check was insufficient to detect we're in the Spark driver.  It looks like BroadcastedActivation is only created for the main driver of an operation, and not in a spark executor.  Therefore, it may be better to just get rid of isSparkDriver/isOlapWorker and its usage in readActivationHolder and just unconditionally call the code which records the hash of user jars.

msirek (changes requested):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BroadcastedActivation.java
msirek: However, if we remove the check altogether, we may hit the same error because we call SpliceSpark.getContext() -> getSession() which checks:
    if (!SpliceClient.isClient() && !threadName.startsWith("olap-worker-")) {

I don't know the problem scenario well enough.  Is there a different way to get the spark context when SpliceClient.isClient() can't be called?

msirek (approved):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BroadcastedActivation.java
msirek: Since this fix is urgent I opened [DB-11299](https://splicemachine.atlassian.net/browse/DB-11299) to handle NSDS support for DB-10759.

jpanko1: Mark entered https://splicemachine.atlassian.net/browse/DB-11299 to address these impacts.

carolp-503 (approved):
+1

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/BroadcastedActivation.java
jyuanca: @msirek @jpanko1 This is for adding custom jars to oalp server. Do we need it for NSDS? Custom jars can always be specified from command line or a script

PR #5095, opened 2021-01-30
DB-11278 More join planning time for triggers and hinted joinStrategy

msirek:
This Jira replaces the problematic prepared statement check of DB-10773 with the following:

1. A check if a stored prepared statement is currently being compiled.  If so, the join planning time limit lower bound is set to 10 ms.
2. A check if a joinStrategy hint is present in the SQL.  If so, the join planning time limit lower bound is set to 2 ms.

Also fixed as part of this Jira:

1. A bug in DB-11256, where the max timeout setting when using Level2OptimizerImpl is set to zero, which could cause queries to time out too early and use a suboptimal join plan.
2. The join planner is updated to only consider nested loop join when planning a single table.  Nested loop join is a special "join strategy" which also knows how to handle single-table access paths.  The planner should never pick a "join strategy" other than nested loop join when costing a single table, so the planner is updated to avoid considering the other join strategies.

See [DB-11278](https://splicemachine.atlassian.net/browse/DB-11278)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizer.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizer.java
@@ -407,4 +407,6 @@ public interface Optimizer{
     public void setForSpark(boolean forSpark);
 
     public boolean isForSpark();
+
+    int getJoinPosition();
 }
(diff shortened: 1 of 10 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2366/

msirek:

spliceci:
TEST SUCCEEDED +1

martinrupp (approved):

carolp-503 (approved):

carolp-503:
+1, once we have one more reviews

jpanko1 (approved):

PR #5096, opened 2021-02-01
DB-11249 Remove unused ConstraintConstantAction

arnaud-splice:


dgomezferro (approved):

ascend1 (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/AlterTableNode.java
```diff
@@ -524,22 +524,6 @@ public ConstantAction	makeConstantAction() throws StandardException
...
-                    // Check to see if a constraint of the same type
-                    // already exists
-                    ConstraintDescriptorList cdl =
-                        dd.getConstraintDescriptors(baseTable);
-
-                    if (cdl.getPrimaryKey() != null) {
-                        throw StandardException.newException(
-                            SQLState.LANG_ADD_PRIMARY_KEY_FAILED1,
-                            baseTable.getQualifiedName());
-                    }
```
martinrupp, line 540: why has this been removed? looks like a check, is that check superflous?

martinrupp (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/AlterTableNode.java
arnaud-splice: primaryKeyConstantActionCheck was checking if cca was of type ConstraintConstantAction, but that class was never used.

arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/AlterTableNode.java
martinrupp: aaaahhh!

martinrupp (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5097, opened 2021-02-01
DB-11296 fix upgrades from 1985 to 1996

martinrupp:
# Description

DB-8897 introduced a system table with conglomerate `SPLICE_CONGLOMERATE_SI`, and also an upgrade script `UpgradeConglomerateTable` to create this table. This table is used when other tables are created. However, that means it is are also used in some other upgrade scripts, specifically `UpgradeScriptToAddSysNaturalNumbersTable` from 1985. Since `UpgradeScriptToAddSysNaturalNumbersTable` runs before `UpgradeConglomerateTable`, this would fail.

Solution is to move `UpgradeConglomerateTable` script before `UpgradeScriptToAddSysNaturalNumbersTable` script.

# How to test
See also DB-11134 for some scripts.

Basic idea:
PROFILE=cdh6.3.0,core

1. create a "snapshot" of a standalone 1980 cluster:
a. `git checkout tags/3.1.0.1980`
b. `./start-splice-cluster -p${PROFILE}`
c. `./start-splice-cluster -k` # stop cluster
d. `mkdir my_snapshot && cp -r platform_it/target/hbase platform_it/target/zookeeper  my_snapshot`

2. get current version
a. `git checkout origin/master` # or other version
b. `mvn -T 16 clean install -P${PROFILE} -DskipTests=true` # or other profiles

3. start standalone with "snapshot"
a. `cd platform_it && git clean -dfx && cd ..` # clean up platform_it
b. `mkdir platform_it/target && cp my_snapshot/* platform_it/target/.`
c. `./start-splice-cluster -b`

4. check platform_it/splice.log for log messages containing `SpliceCatalogUpgradeScripts`.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
@@ -76,6 +76,10 @@ public class SpliceCatalogUpgradeScripts{
         this.tc=tc;
 
         scripts = new ArrayList<>();
+        // DB-11296: UpgradeConglomerateTable has to be executed first, because it adds a system table
+        // CONGLOMERATE_SI_TABLE_NAME that is from then on needed to create tables, e.g.
+        // in UpgradeScriptToAddSysNaturalNumbersTable. If UpgradeConglomerateTable is at the end,
+        // these upgrades would fail
         addUpgradeScript(baseVersion4, 1996, new UpgradeConglomerateTable(sdd, tc));
 
         addUpgradeScript(baseVersion1, 1901, new UpgradeScriptToRemoveUnusedBackupTables(sdd,tc));
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
@@ -76,6 +76,8 @@ public class SpliceCatalogUpgradeScripts{
         this.tc=tc;
 
         scripts = new ArrayList<>();
+        addUpgradeScript(baseVersion4, 1996, new UpgradeConglomerateTable(sdd, tc));
+
         addUpgradeScript(baseVersion1, 1901, new UpgradeScriptToRemoveUnusedBackupTables(sdd,tc));
         addUpgradeScript(baseVersion1, 1909, new UpgradeScriptForReplication(sdd, tc));
         addUpgradeScript(baseVersion1, 1917, new UpgradeScriptForMultiTenancy(sdd,tc));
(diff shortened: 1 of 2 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
```diff
@@ -76,6 +76,8 @@ public SpliceCatalogUpgradeScripts(SpliceDataDictionary sdd, TransactionControll
         this.tc=tc;
 
         scripts = new ArrayList<>();
+        addUpgradeScript(baseVersion4, 1996, new UpgradeConglomerateTable(sdd, tc));
```
msirek, line 79: Update 1996 -> 1997

dgomezferro, line 79: Please add a comment explaining why this one is out of order.

dgomezferro (approved):
Nice idea for a fix @martinrupp !

jyuanca:
Check existence of SPLICE_CONGLOMERATE_SI table during conglomerate creation. The upgrade script should still be in order

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
martinrupp: making this 1997 would make it run again on 1996 clusters, that already have executed that upgrade script (and if they already executed it, everything is fine). we are not *adding* an upgrade script, just moving it.

martinrupp: done

jyuanca (changes requested):
Check existence of SPLICE_CONGLOMERATE_SI table during conglomerate creation. The upgrade script should still be in order

dgomezferro (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/SpliceCatalogUpgradeScripts.java
msirek: OK, thanks for the explanation.  If an upgrade fails for some other reason, and we already copied rows from SPLICE_CONGLOMERATE into SPLICE_CONGLOMERATE_SI, would the upgrade try to copy the rows again and then fail the upgrade due to duplicate rows?

msirek: It looks like UpgradeConglomerateTable has a check to see if SPLICE_CONGLOMERATE_SI is already created, so I guess that's not a problem.

msirek (approved):

jyuanca (approved):

carolp-503 (approved):
+1

PR #5103, opened 2021-02-02
DB-10436 Rollback to Transaction

jpanko1:


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeManager.java
```diff
@@ -27,4 +27,6 @@ default void endRollingUpgrade() throws StandardException {}
     default List<String> unloadRegions(String hostAndPort) throws StandardException {return null;}
 
     default List<String> loadRegions(String serverName) throws StandardException { return null;}
+
+    default void restartOlapServer() throws StandardException {}
```
jpanko1, line 31: restartOlapServer() is implemented in [DB-10044](https://github.com/splicemachine/spliceengine-ee/pull/586/files#diff-bcaf7782ef4fbc67bdab5e571167ff48a0b9b624257571dba603d0166ca49a89R429-R440)

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2377/

hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
```diff
@@ -380,4 +391,45 @@ public static String getBackupPath() {
...
+        com.carrotsearch.hppc.BitSet fields  = new com.carrotsearch.hppc.BitSet(2);
+        fields.set(0,2);
+        com.carrotsearch.hppc.BitSet scalarFields = new com.carrotsearch.hppc.BitSet(2);
+        scalarFields.set(1, 2);
+        com.carrotsearch.hppc.BitSet floatFields = new com.carrotsearch.hppc.BitSet(0);
+        com.carrotsearch.hppc.BitSet doubleFields = new com.carrotsearch.hppc.BitSet(0);
+
+        return EntryEncoder.create(SpliceKryoRegistry.getInstance(),2,fields,scalarFields,floatFields,doubleFields);
+    }
 }
```
jyuanca, line 435: This was copied from Restore.setupIgnoreTxnTable. Please refactor. One option is to move this to HBaseBackupUtil

splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
```diff
@@ -657,4 +656,41 @@ public static void SYSCS_RESTORE_SCHEMA(String destSchema,
...
+            EngineDriver.driver().manager().getUpgradeManager().restartOlapServer();
+
+            SpliceAdmin.INVALIDATE_GLOBAL_DICTIONARY_CACHE();
+        } catch (Throwable t) {
+            resultSets[0] = ProcedureUtils.generateResult("Error", t.getLocalizedMessage());
+            SpliceLogUtils.error(LOG, "Database rollback error", t);
+            t.printStackTrace();
+        }
+    }
 }
```
jyuanca, line 696: What if there are running txns before database rollback, do we want to wait for them to complete, or ignore them and let them eventually fail?

jyuanca (changes requested):

hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
jpanko1: Moved it to HBaseBackupUtils in commits https://github.com/splicemachine/spliceengine/pull/5103/commits/b4d1fc496f39ae3153ce030b8211391232222313 and https://github.com/splicemachine/spliceengine-ee/pull/601/commits/c5d82a7efe3ff23f6b7b5400cd5200d659e11bdc

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2383/

jpanko1:

splice_machine/src/main/java/com/splicemachine/backup/BackupSystemProcedures.java
jpanko1: Wrote Jira for this issue: https://splicemachine.atlassian.net/browse/DB-11332

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
```diff
@@ -14,16 +14,26 @@
...
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.sql.dictionary.BackupDescriptor;
+import com.splicemachine.encoding.MultiFieldEncoder;
 import com.splicemachine.pipeline.Exceptions;
+import com.splicemachine.si.api.data.TxnOperationFactory;
 import com.splicemachine.si.constants.SIConstants;
 import com.splicemachine.si.impl.driver.SIDriver;
+import com.splicemachine.storage.DataPut;
+import com.splicemachine.storage.EntryEncoder;
+import com.splicemachine.storage.Partition;
```
dgomezferro, line 36: Cleanup imports, it doesn't look like they are needed.

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
```diff
@@ -734,6 +734,7 @@ public static void SYSCS_PERFORM_MAJOR_COMPACTION_ON_TABLE(String schemaName,Str
                 throw PublicAPI.wrapStandardException(Exceptions.parseException(e));
             }
             try (Partition partition = tableFactory.getTable(Long.toString(conglomID))) {
+                partition.flush();
```
dgomezferro, line 737: This looks unrelated, is it necessary? I think most of the times we want a MAJOR_COMPACTION we also want a flush, but we should be aware of the change.

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
```diff
@@ -436,6 +436,7 @@ protected boolean performCompaction(Compactor.FileDetails fd, InternalScanner sc
         List<Cell> cells = new ArrayList<>();
         boolean hasMore;
         ScannerContext scannerContext = ScannerContext.newBuilder().setBatchLimit(compactionKVMax).build();
+        SIDriver.driver().getIgnoreTxnSupplier().refresh();
```
dgomezferro, line 439: Is this necessary here? I don't see why.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
```diff
@@ -105,4 +141,18 @@ private void init() throws IOException {
...
+    @Override
+    public void refresh() {
+        synchronized (this) {
+            cache.clear();
+            try {
+                populateIgnoreTxnCache();
+            } catch (IOException e) {
+                LOG.error("Couldn't populate Ignore Cache", e);
+                // Force restart
+                // System.exit(-1);
```
dgomezferro, line 154: Remove comments

In any case, if this fails we need something more than just logging an error, can we fail the original ROLLBACK_TO_TXN call?

splice_si_api/src/main/java/com/splicemachine/si/impl/store/CompletedTxnCacheSupplier.java
```diff
@@ -35,8 +35,8 @@
  *         Date: 6/18/14
  */
 public class CompletedTxnCacheSupplier implements TxnSupplier{
+    private volatile TxnView[] cache;
```
dgomezferro, line 38: Is it possible to not make it volatile? Maybe by implementing `invalidate` like `Arrays.fill(cache, null);`

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
```diff
@@ -50,12 +56,14 @@ public boolean shouldIgnore(Long txnId) throws IOException{
...
 
         init();
-        
-        if (!cache.isEmpty()) {
-            for (Pair<Long, Long> range : cache) {
-                if (txnId > range.getFirst() && txnId < range.getSecond()) {
-                    ignore = true;
-                    break;
+
+        synchronized (this) {
```
dgomezferro, line 60: Can we get rid of this synchronized block? I think it should be possible to make `cache` volatile and make `refresh` atomically swap it.

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
```diff
@@ -87,7 +95,35 @@ private void populateIgnoreTxnCache() throws IOException {
                     }
                 }
             }
+            cache = combineOverlappingRanges(cache);
+        }
+    }
+
+    static Set<Pair<Long, Long>> combineOverlappingRanges(Set<Pair<Long, Long>> cache) {
+        BiPredicate<Pair<Long, Long>, Pair<Long, Long>> overlapping = (a, b) ->
+            (a.getFirst() < b.getFirst() && b.getFirst() < a.getSecond()) ||
```
dgomezferro, line 104: I think you are missing some equality checks:
a = 10, 20
b = 10, 30
should be overlapping

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
```diff
@@ -87,7 +95,35 @@ private void populateIgnoreTxnCache() throws IOException {
                     }
                 }
             }
+            cache = combineOverlappingRanges(cache);
+        }
+    }
+
+    static Set<Pair<Long, Long>> combineOverlappingRanges(Set<Pair<Long, Long>> cache) {
```
dgomezferro, line 102: I don't expect this to be a problem initially, but the algorithm could blow up since it has quadratic complexity. Add some logging information when populating the cache so that if we run into this problem at least it's clear where we are spending too much time.

dgomezferro (changes requested):

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
jpanko1: Agree that 10,20 and 10,30 should be overlapping.  What about 10,30 and 30,60 ?  The comparison in the `shouldIgnore` method doesn't use equality so it wouldn't include 30 as part of the range:
`if (txnId > range.getFirst() && txnId < range.getSecond())`
Should the beginning and ending values be included as part of the range?

splice_machine/src/main/java/com/splicemachine/derby/utils/SpliceAdmin.java
jpanko1: Major compaction for this new functionality wouldn't clean out everything without having a flush first.  Can we expect users to always flush before compaction?  Our online docs for compaction don't say that flush is required before compaction.

hbase_sql/src/main/java/com/splicemachine/hbase/BackupUtils.java
jpanko1: Fixed in commit https://github.com/splicemachine/spliceengine/pull/5103/commits/81376b4b31305ebc1dd7946beaccb4cc299844a1

hbase_sql/src/main/java/com/splicemachine/compactions/SpliceDefaultCompactor.java
jpanko1: Removed in commit https://github.com/splicemachine/spliceengine/pull/5103/commits/784a4dd0a6115e8fe949b71136075f0d49ef0336

splice_si_api/src/main/java/com/splicemachine/si/impl/store/CompletedTxnCacheSupplier.java
jpanko1: Changed in commit https://github.com/splicemachine/spliceengine/pull/5103/commits/74c7644dcfb2a08f831039cd2ed377d61fc3adb9

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
jpanko1: Changed in commit https://github.com/splicemachine/spliceengine/pull/5103/commits/19138b4bb9ba883e90b87b67324df4a83a7fd959

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
jpanko1: Added logs in commit https://github.com/splicemachine/spliceengine/pull/5103/commits/19138b4bb9ba883e90b87b67324df4a83a7fd959

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2397/

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
jpanko1: Was this question referring to the call to refresh in SpliceDefaultCompactor?  I've since removed that call based on the previous comment.  The only other call to refresh is in the LEAVE_RESTORE_MODE block of SpliceDatabase.registerDDL() which would occur late in the ROLLBACK_TO_TXN process.  Should ROLLBACK_TO_TXN still be failed at this point, as it's basically in cleanup phase with the main work already done?

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2401/

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeManager.java
jpanko1: Merged in restartOlapServer() with commit https://github.com/splicemachine/spliceengine/pull/5103/commits/0084db46bdb5a38d91dcb52fd9dda5749473d82f and updated function call in https://github.com/splicemachine/spliceengine/pull/5103/commits/30e1281c32782994f444bf3dc8ecc14d24e3c3df

jyuanca: This is no longer need.

jyuanca (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeManager.java
jpanko1: Removed UpgradeManager.restartOlapServer() in https://github.com/splicemachine/spliceengine/pull/5103/commits/3a7193d118fa5a2ea41e5ef123668becee5044fa

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
dgomezferro: Given the implementation of `shouldIgnore` I don't think [10,30] and [30,60] should be merged, but anyway, that case shouldn't happen often. I think coming back to the same transaction is more common (hence ending up with many overlapping ranges with the same start)

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
dgomezferro: No, it was a general comment, but you are right it's late in the execution already. I'm not sure how to handle a failure here to be honest, maybe we should really restart here, otherwise this could lead to inconsistencies if it's not populated correctly. Do you have an opinion? Otherwise we can open a JIRA to discuss it later.

dgomezferro (approved):
Thanks for the changes @jpanko1 ! Looks good to me. We can follow up some work in other JIRAs.

carolp-503 (approved):
+1

splice_si_api/src/main/java/com/splicemachine/si/impl/store/IgnoreTxnSupplierImpl.java
hatyo: I think we can reduce the complexity of the overlapping algorithm by maintaining the cache online, i.e. whenever we add a new element, we change the cache immediately, provided that we have a suitable equality / hash function and we use a TreeMap, we can probably reduce the complexity to from quadratic to logarithmic. If there is a JIRA open for this, please refer me to it so we can follow up there.

PR #5110, opened 2021-02-03
DB-11307 grammar for MERGE INTO

martinrupp:
first (simple) grammar implementation for MERGE INTO, taken from derby's implementation.
Tests only check parser, created execution is a dummy execution.


msirek (approved):

dgomezferro (approved):
Looks good!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNodeVector.java
```diff
@@ -46,7 +48,22 @@
  * lists of query tree nodes inherit from QueryTreeNodeVector.
  */
 
-abstract class QueryTreeNodeVector<T extends QueryTreeNode> extends QueryTreeNode implements Iterable<T>{
+class QueryTreeNodeVector<T extends QueryTreeNode> extends QueryTreeNode implements Iterable<T>{
+
+    final Class<T> eltClass; // needed for cast in #acceptChildren
```
arnaud-splice, line 53: What's the point of eltClass? The constructor that sets it does not seem to be used anywhere, and it looks kinda hacky.

arnaud-splice (changes requested):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5115, opened 2021-02-04
DB-11334 Refactor Unary/Binary/Ternary OperatorNodes

arnaud-splice:


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2395/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExtractOperatorNode.java
```diff
@@ -171,21 +171,21 @@ public ValueNode bindExpression(FromList fromList,
              && (extractField == DateTimeDataValue.SECOND_FIELD) ) {
             setType(new DataTypeDescriptor(
                             TypeId.getBuiltInTypeId(Types.DOUBLE),
-                            operand.getTypeServices().isNullable()
+                            getOperand().getTypeServices().isNullable()
                         )
                 );
         } else if (extractField == MONTHNAME_FIELD || extractField == DateTimeDataValue.WEEKDAYNAME_FIELD) {
             // name fields return varchar
             setType(new DataTypeDescriptor(
```
aimiebell, line 179: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowFunctionNode.java
```diff
@@ -87,6 +87,7 @@ public void init(Object arg1, Object arg2, Object arg3, Object arg4, Object arg5
      * @throws StandardException
      */
     public void init(Object arg1, Object arg2) throws StandardException {
+        super.init();
```
hatyo, line 90: I haven't check if such constructor exists, but should we call `super.init(arg1)` instead of `super.init()`?

ascend1 (approved):
Great job!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorNode.java
```diff
@@ -32,16 +32,319 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected boolean isSemanticallyEquivalent(ValueNode o) throws StandardException {
+        if (!isSignatureEquivalent(o))
+            return false;
+        OperatorNode other = (OperatorNode)o;
+        for (int i = 0 ; i < operands.size(); ++i) {
```
hatyo, line 289: a bit subtle to understand at first glance nevertheless looks good

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorNode.java
```diff
@@ -32,16 +32,319 @@
...
+     * underlying expression.
+     *
+     * @return ValueNode            The remapped expression tree.
+     *
+     * @exception StandardException            Thrown on error
+     */
+    public ValueNode remapColumnReferencesToExpressions()
+            throws StandardException
+    {
+        for (int i = 0; i < operands.size(); ++i) {
```
hatyo, line 186: does it make sense to check that `operands` is not null? (same in other methods such as `checkCRLevel` and `isConstantExpression` ... etc).

martinrupp (approved):

hatyo (changes requested):
LGTM, minor comment about assertion against null pointer.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ExtractOperatorNode.java
arnaud-splice: This seems to be a false positive

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/WindowFunctionNode.java
arnaud-splice: arg1 is used to initialise `this.window` which is `WindowFunctionNode` specific.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorNode.java
arnaud-splice: Done

arnaud-splice:

hatyo (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5116, opened 2021-02-04
DB-11258 DB-11328 cache EXPLAIN statement plan

hatyo:
In this ticket we cache `X` when running `EXPLAIN X` statement in the data dictionary statement cache. This guarantees to the user that the execution plan produced by `EXPLAIN X` is indeed the one we are going to execute next time we run `X` [1].

Some implementation details, after binding a statement we check if it is an `EXPLAIN X` statement, if so, then we grab the root node of the `EXPLAIN` plan (which is `X`) and then we bind, optimize, generate, and cache that one in the data dictionary statement cache. Afterwards, we plug-in the optimized plan of `X` back into `EXPLAIN`'s plan and turn a flag in `EXPLAIN` that causes the optimizer to skip optimizing its root node (since it is already optimized), finally we proceed with the execution of the `EXPLAIN X` as normal and return the execution plan of `X` to the user along with any extra information (such as tables missing stats).

To assist testing this new functionality I needed a way to observe what's in the region server's statement cache, so I ended up implementing a new system procedure called `SYSCS_GET_CACHED_STATEMENTS`, this procedure returns a result set comprising three columns:

1. `SERVER` a `VARCHAR` column containing the region server (RS) socket information such as `localhost:1527`.
2. `CACHED_STATEMENT` a `VARCHAR` column containing the cached statement (exactly as queried from the statement cache).
3. `ADDITION_TIMESTAMP` a `TIMESTAMP` column containing the timestamp information on when the statement was first added to the RS's statement cache.

This table allows us to easily check:
- whether a statement has been added to the cache.
- when was a statement added to the cache.
- whether the statement was compiled again by comparing the addition timestamp value before and after it is executed.

This was handy to write a number of ITs that could be found in `ExplainPlanIT.java` test suite.

Here is an example usage:

```
call syscs_util.SYSCS_GET_CACHED_STATEMENTS();
SERVER    |CACHED_STATEMENT                                   |ADDITION_TIMESTAMP
---------------------------------------------------------------------------------------
local:1527|call syscs_util.SYSCS_GET_CACHED_STATEMENTS()      |2021-02-04 16:45:47.5
local:1527|call SYSCS_UTIL.SYSCS_GET_CACHED_STATEMENTS_LOCAL()|2021-02-04 16:45:47.866
local:1528|call SYSCS_UTIL.SYSCS_GET_CACHED_STATEMENTS_LOCAL()|2021-02-04 16:45:48.119

explain select * from t;

call syscs_util.SYSCS_GET_CACHED_STATEMENTS();
SERVER    |CACHED_STATEMENT                                   |ADDITION_TIMESTAMP
---------------------------------------------------------------------------------------
local:1527|call syscs_util.SYSCS_GET_CACHED_STATEMENTS()      |2021-02-04 16:45:47.5
local:1527|call SYSCS_UTIL.SYSCS_GET_CACHED_STATEMENTS_LOCAL()|2021-02-04 16:45:47.866
local:1528|call SYSCS_UTIL.SYSCS_GET_CACHED_STATEMENTS_LOCAL()|2021-02-04 16:45:48.119
local:1527|select * from t                                    |2021-02-04 16:46:03.119
```

As we can see, after running the `explain select * from t` we now see `select * from t` cached in the RS's statement cache.

---
[1] Unless an explicit statement such as `ANALYZE` or stored procedure such as `SYSCS_EMPTY_STATEMENT_CACHE` is ran in-between `EXPLAIN X` and `X` causing the cache to be invalidated. Still caching `EXPLAIN X` makes guessing the actual execution plan of `X` more predictable.


db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -857,12 +881,66 @@ private void fourPhasePrepare(LanguageConnectionContext lcc,
...
+            // force invalidate the currently cached statement if it exists
+            // the assumption here is when running EXPLAIN, the user expects
+            // the plan to correspond to what the system currently thinks is
+            // the best plan, retrieving it from the cache instead could be
+            // misleading to say the least.
+            lcc.removeStatement(queryNode);
+
+            // add the statement to the statement cache
+            lcc.lookupStatement(queryNode);
+        }
```
dgomezferro, line 910: This needs to be discussed, I'm slightly in favor of reusing the already cached statement if there's one.

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
```diff
@@ -78,7 +79,7 @@
     private ManagedCache<Long,Conglomerate> conglomerateCache;
     private ManagedCache<Pair<Long, Long>, Conglomerate> txnAwareConglomerateCache;
     private ManagedCache<Long,ConglomerateDescriptor> conglomerateDescriptorCache;
-    private ManagedCache<GenericStatement,GenericStorablePreparedStatement> statementCache;
+    private ManagedCache<StatementKey,GenericStorablePreparedStatement> statementCache;
```
dgomezferro, line 82: To me it makes more sense to store the Timestmap in the value rather than the key, but I don't mind.

dgomezferro (approved):
Looks good! I would like to discuss the behaviour when the statement is already in the cache though.
Thanks for the very clear and descriptive comments!

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
hatyo: I added a comment on the corresponding JIRA ticket listing all the assumptions and asking for feedback, I will adapt it to whatever the consensus on the right behavior will be.

ascend1 (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryCache.java
hatyo: you're right, fixed it.

carolp-503 (commented):
+1

PR #5125, opened 2021-02-06
DB-11266 Fix trigger concurrency issues.

msirek:
Fixes issues with concurrent row triggers:

1. Aborts other running trigger threads when one thread errors out, so the main thread doesn't try to clean up resources actively being used by running triggers.
2. Increase the size of the stored prepared statement cache, to decrease likelihood two threads are trying to recompile the same trigger simultaneously.
3. Use a ConcurrentHashMap-backed HashSet for the "children" Set of transactions in AbstractTxn, to avoid ConcurrentModificationException errors.

Part 3 will alleviate customer issues in the interim until DB-10276 is complete.  In that Jira I'm working on an issue where the LanguageConnectionContext used to get the transaction to elevate when compiling a trigger may sometimes be the wrong one (is gotten from the ContextService instead of the Activation).  DB-10276 will remove the use of the ConcurrentHashMap in AbstractTxn.


msirek:

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

jyuanca (approved):

carolp-503 (approved):
+1

PR #5138, opened 2021-02-09
DB-11344: execute SQL from Olap server correctly

jyuanca:


jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2406/

jyuanca:

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
```diff
@@ -553,7 +553,7 @@ public ExecRow apply(@Nullable ExecRow execRow) {
     public void openCore() throws StandardException{
         DataSetProcessor dsp = EngineDriver.driver().processorFactory().chooseProcessor(activation, this);
         activation.getLanguageConnectionContext().getStatementContext().registerExpirable(this, Thread.currentThread());
-        if (dsp.getType() == DataSetProcessor.Type.SPARK && !isOlapServer() && !SpliceClient.isClient()) {
+        if (dsp.getType() == DataSetProcessor.Type.SPARK && !(isOlapServer() && activation.isSubStatement()) && !SpliceClient.isClient()) {
```
msirek, line 556: @jyuanca I'm wondering if this line of code needs re-work.  dsp.getType() tells how we're currently executing, but not what the new request uses.  
How about:

> if (activation.datasetProcessorType().isOlap() && !isOlapServer() && !SpliceClient.isClient())

jyuanca: @msirek Thanks for your suggestion. I'll give a try.

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
msirek: My last suggestion was incorrect, but I'm wondering if this piece is wrong:
!(isOlapServer() && activation.isSubStatement())  

This is the same as:  !isOlapServer() || !activation.isSubStatement()

isSubStatement is false for all regular statements (anything not a trigger), so this will essentially remove the !isOlapServer() check because the 2nd part is always true and we will now be able to start distributed operations from the Olap server.

jyuanca:

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
jyuanca: We should be able to do distributed operation from olap server, if olap server parses a SQL and determined it should executed from spark. In this case, the olap server is also a client. It needs to receive results streamed from spark.

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
msirek: OK, I didn't realize this is an intentional change, thanks.  Both paths could use spark.  I didn't know there is an advantage of streaming results through a 2nd RemoteQueryClient.

dgomezferro (dismissed):
It doesn't look like the UDT.jar is needed, is it?

This is making sure the trigger runs in Spark, is that right? It seems a bit wasteful for very simple triggers like the on in the original JIRA, was it not possible to execute it in OLTP mode?

jyuanca:
@dgomezferro I made some changes to UDT.jar, so it needs to be updated.

@msirek knows more about the second question. The trigger in the Jira is a statement trigger. Is olap server the right place to fire before/after statement triggers? Is it possible to fire at region server?

msirek (approved):
I had some discussions with @jyuanca.  Apparently openCore does support Spark execution, just not the collection of rows from executors.  So, for triggers it is OK to use openCore because it never streams rows back, and actually might be needed to use some local contexts.  But for any other type of query that collects rows, openDistributed is required.

msirek:
> @msirek One question - the trigger in the Jira is a statement trigger. Is olap server the right place to fire before/after statement triggers? Is it possible to fire at region server? If we can do this, we will remove some annoying problems in finding UDF jars in olap server.

@jyuanca Currently it is not possible.  The trigger is fired locally where the main statement is executing.  I think the same is true for anything else that we may try to run on the OLAP Server using ControlDataSetProcessor.

arnaud-splice (approved):

dgomezferro:
@jyuanca I meant running them on OLTP mode in the OlapServer, not in the RegionServer. Are we forcing all triggers running from OLAP queries to be executed in Spark too?

jyuanca:
@dgomezferro Yes, after my change, this query is allowed to run as OLTP query from olap server.
@msirek Do you have ideas about Daniel's question " Are we forcing all triggers running from OLAP queries to be executed in Spark too?"

msirek:
> @msirek Do you have ideas about Daniel's question " Are we forcing all triggers running from OLAP queries to be executed in Spark too?"

@dgomezferro @jyuanca  No, triggers can still use ControlDataSetProcessor, even if compiled and executed from the OlapServer.  Triggers do not, however, open up any JDBC connection to the region server to try and execute them remotely.  For row triggers, they would fire at the location the row is being inserted/updated/deleted, from the spark executor.  Statement triggers fire from the OlapServer.  Essentially activation.execute() is called to run the trigger code, which ends up calling openCore() and running the whole sequence of operations.

carolp-503 (approved):
+1

dgomezferro:
I've built the branch and indeed triggers from OLAP queries can un in OLTP mode. However the added test doesn't exercise the modified code, unless I did something wrong the test passes without the changes too. @jyuanca can you add a test that exemplifies why the changes are needed?

jyuanca:
@dgomezferro hmm, let me re-run the tests.

jyuanca:
@dgomezferro  can I file a followup jira to clean up the test? This is a blocker for Ben for a while.

jyuanca:
Create DB-11495 to add ITs

jyuanca:

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2485/

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2486/

jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2488/

jyuanca:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST SUCCEEDED +1

PR #5141, opened 2021-02-09
DB-11401 allow querying VTI without specifying the schema.

hatyo:
In this PR we allow querying a VTI without explicitly specifying the output result set provided the VTI class implements the newly introduced interface `CompileTimeSchema`.

The rationale behind this PR is that we want to be able to introduce a monitoring table-valued function called `MON_GET_CONNECTION`, which, [similar to DB2](https://www.ibm.com/support/knowledgecenter/SSEPGG_11.5.0/com.ibm.db2.luw.sql.rtn.doc/doc/r0053938.html), has hundreds of columns, VTIs provide a way to implement table-valued functions as [documented in Derby](https://db.apache.org/derby/docs/10.11/publishedapi/org/apache/derby/vti/package-summary.html), however with this limitation (having to explicitly specify the output schema) it becomes very difficult to expose `MON_GET_CONNECTION` as a table-valued function.

As a practical example, there is a VTI in SpliceMachine called `SpliceAllRolesVTI`, previously we were able to query it like this:

`select * From new com.splicemachine.derby.vti.SpliceAllRolesVTI() x(y varchar(12))`

removing the schema `(y varchar(12))` gives an error 42X81 'A query expression must return at least one column', with this PR this becomes possible (since we adapt `SpliceAllRolesVTI` such that it implements the interface `CompileTimeSchema`:

```
select * From new com.splicemachine.derby.vti.SpliceAllRolesVTI() x;
ROLENAME
----------
PUBLIC
SPLICE
```

To get more details on how this is implemented please have a look at the documentation of `CompileTimeSchema`, there we provide some details about how and when to use it.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
@@ -1787,9 +1787,5 @@ public class DataTypeDescriptor implements Formatable{
         return typeId.getStructField(columnName, getPrecision(), getScale(), childStructField);
 
     }
-
-    public TypeDescriptorImpl getTypeDescriptor() {
-        return typeDescriptor;
-    }
 }
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/DataTypeDescriptor.java
@@ -1787,5 +1787,9 @@ public class DataTypeDescriptor implements Formatable{
         return typeId.getStructField(columnName, getPrecision(), getScale(), childStructField);
 
     }
+
+    public TypeDescriptorImpl getTypeDescriptor() {
+        return typeDescriptor;
+    }
 }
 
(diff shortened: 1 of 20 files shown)


msirek (approved):

dgomezferro (commented):
I really like the ability to avoid specifying the schema, nice one!
Since we are accessing a static method of the classes, we don't really need the CompileTimeSchema interface. Do you want it just to have a marker and/or an example of the static methods? I think it might be a little bit confusing because static methods aren't inherited, but using it as a marker/example also adds clarity, so I'm ambivalent.

hatyo:
> I really like the ability to avoid specifying the schema, nice one!
> Since we are accessing a static method of the classes, we don't really need the CompileTimeSchema interface. Do you want it just to have a marker and/or an example of the static methods? I think it might be a little bit confusing because static methods aren't inherited, but using it as a marker/example also adds clarity, so I'm ambivalent.

Thanks @dgomezferro, yes I agree it's a little weird, my goal was to provide a type-safe way of defining these *static* methods when implementing a VTI. It _looks_ like we're overriding the static interface methods, but we're actually not since this is not allowed by design. To avoid this confusion, I would remove these methods from the interface and just provide an exemplary implementation of these methods in comments.

dgomezferro:
@hatyo I like this option better, and thanks for the nice documentation!

jpanko1:

arnaud-splice (approved):

carolp-503 (approved):
+1

spliceci:
TEST SUCCEEDED +1

PR #5149, opened 2021-02-10
DB-11124 LIKE predicate support for [long] [var]char for bit data

hatyo:
In this PR we extend `LIKE` predicate to support operations on `for bit data` data types such as `varchar for bit data`.

To do this, we check in `LIKE` operator's binding phase whether we have any `for bit data` operand, if so, then we add an implicit cast to `VARCHAR` in Single Byte Character Set (SBCS) mode so we can parse the characters back if necessary which is important for processing escapes correctly in later generation and execution phases.

As an example, in this PR the following queries work:

```
create table t(c long varchar for bit data)
insert into t values 'a';
select * from t where c like 'a';
C
-
61
```
We added quite many tests (240) in this PR to address different possibilities of using `LIKE` operator with different types in LIKE operands, also to test prepared vs. non-prepared, OLAP vs. OLTP.


msirek (approved):

jyuanca (approved):

carolp-503 (approved):
+1

PR #5158, opened 2021-02-11
DB-11405 Fix 'Unable to generate code' error with multiple subqueries

msirek:
### Description
DB-10774 modified the binding code to only add a SubqueryNode once to a subqueryList in an effort to bind the subquery only once in a given subtree.  The problem is there could be different subtrees that maintain their own subqueryList and they reference that SubqueryNode as well.  The SubqueryNode should be added to those other subqueryLists as well, but since the new flag thisSubqueryAddedToSubqueryList is now false, we fail to do so.

example query:

> insert into B values (
>     nullif(
>         case when
>             EXISTS (select C1 from C where EXISTS (select A1 from A))
>         then 14 end,
>         4)
> );

ERROR 42Z50: INTERNAL ERROR: Unable to generate code for
com.splicemachine.db.impl.sql.compile.SelectNode@45eb0df

### Fix
The fix is to add the SubqueryNode to subqueryList, but only if it is not already present in the list.  The thisSubqueryAddedToSubqueryList flag is removed.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNodeVector.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNodeVector.java
@@ -69,6 +69,11 @@ abstract class QueryTreeNodeVector<T extends QueryTreeNode> extends QueryTreeNod
         v.add(qt);
     }
 
+    public final void addElementIfNotPresent(T qt){
+        if (!v.contains(qt))
+            v.add(qt);
+    }
+
     public final void removeElementAt(int index){
         v.remove(index);
     }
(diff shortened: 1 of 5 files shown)


msirek:
jenkins please test branch@cdh6.3.0,skipTestsLongerThan2Minutes

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2441/

arnaud-splice (approved):

jyuanca (approved):

msirek:
jenkins please test branch@cdh6.3.0,skipTestsLongerThan2Minutes

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2442/

msirek:
jenkins please test branch@cdh6.3.0,skipTestsLongerThan2Minutes

spliceci:
TEST SUCCEEDED +1

hatyo (approved):

ascend1 (approved):

carolp-503 (approved):
+1

PR #5162, opened 2021-02-11
DB-11433 fix start-splice.sh use of BSD nc (3.0)

martinrupp:
This fixes nc hangs on Ubuntu 18.04 with the BSD version of nc.
We are using the same code we added in DB-9930 to ./start-splice-cluster. This runs on Mac, BSD and GNU versions of netcat (nc).


carolp-503 (approved):
Approved for 3.0

PR #5176, opened 2021-02-15
WIP SPLICE-1891/DB-10592 Additional mechanism to throttle the streaming(DRAFT) 

ipraznik-splice:
# Description

When rows are streamed from Spark execution to client, they go through two links: 1) from Spark to the region server where the query is submitted; 2) from the region server to client.And on the region server, we keep a buffer for each spark partition of 10K rows by default.

The throttling mechanism we have currently is to run 2 concurrent jobs at a time on the spark side, each sending rows from 2 partitions to the region server. If the buffer for a particular partition is full, it would cause back-pressure, and prevent Spark from sending more rows for that partition. However, if the number of rows in the partition is smaller than the buffer size, then there is no mechanism to hold Spark from sending all the rows for that partition to the region server.

What happens is that query1 has 900 partitions (due to the shuffle.partitions setting) with 5M rows evenly distributed across 900 partitions, so each partition only has ~5.5K rows which is less than the buffer size. When the speed for link 2 above is too slow to consume the rows in the buffer quickly, in theory all 5M rows could be accumulated on the region server, causing memory issue.

# What changes were proposed in this pull request?

There are two possible sollutions: another PR https://github.com/splicemachine/spliceengine/pull/5177

1. (This). The new ResultStreamer is blocked until StreamListener sends a Resume message. StreamListener checks the number of active partitions and via Resume message activates next one when there is the slot available.

2. ResultStreamer keeps the connection to StreamListener opened and does not return from the method call until the partition will be processed on StreamListener.

# How to test

For example with a standalone cluster:

1. Start a cluster, for example with cdh6.3.0 profile
2. Create a table with big number of records:
```
create table a ( i int);
insert into a values 1,2,3;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
```
3. Get an id of SpliceTestPlatform process:
```
ps -ef | grep 'com.splicemachine.test.SpliceTestPlatform' | less
```
4. Execute in sql client: 
```
select * from a order by i;
```
5. Monitore the number of PartitionState instances: 
```
jmap -histo:live [PS_ID] | grep 'PartitionState'
```

Without a fix the number of PartitionState will come soon to 200. With the fix it will be <=24 for ver1 or <=3 for ver2.
For version 1 the number of parallel partitions is managed by the field StreamListener.maxParallelPartitions = batches * 2;

**Only one from these PRs should be merged!**


hbase_sql/src/main/java/com/splicemachine/stream/ResultStreamer.java
```diff
@@ -75,7 +75,7 @@ public ResultStreamer(OperationContext<?> context, UUID uuid, String host, int p
         this.numPartitions = numPartitions;
         this.batches = batches;
         this.batchSize = batchSize;
-        this.permits = new Semaphore(batches - 1); // we start with one permit taken
+        this.permits = new Semaphore(0); // we start with one permit taken
```
dgomezferro, line 78: With this change we have to wait until the client gives us the go ahead to start sending data from the server, right? It adds a little bit of delay, maybe not a big deal but it could hurt us when dealing with those very big fetch jobs with thousands of partitions to fetch. In any case the comment has to be fixed.

dgomezferro (commented):
For this particular issue ( DB-10592 ) I think I'm a bit more in favor of version 2 because it's simpler, but I also like that version 1 is a bit more flexible and it could maybe help with the long standing SPLICE-1891 
For now I would go with version 2 and keep in mind version 1 when we come back to SPLICE-1891

PR #5180, opened 2021-02-16
DB-11119 Support TRANSLATE function

arnaud-splice:
### **Description**
This PR introduces TRANSLATE as a new SQL function. It is inspired from the DB2 implementation of translate that can be found here:

https://www.ibm.com/support/knowledgecenter/SSEPEK_10.0.0/sqlref/src/tpc/db2z_bif_translate.html

Usage:
`TRANSLATE(string_expression, outputTranslationTable, inputTranslationTable [, pad])
`

This function returns a new string, of type TYPEOF(string_expression)
If any operand is null, we return null

This new string is constructed the following way:

for each character c in string_expression:
if c is present in inputTranslationTable (position n), use outputTranslationTable[n]
else use c
 
For example:
```
translate('ABC', 'b', 'B') = 'AbC'
translate('ABC', 'ba', 'BA') = 'abC'

```
If outputTranslationTable is longer than inputTranslationTable, we ignore the excess characters
If inputTranslationTable is longer than outputTranslationTable, we pad outputTranslationTable with pad (or with ' ' if pad isn't specified)

string_expression must be a char / varchar
inputTranslationTable and outputTranslationTable must be char / varchar, or char / varchar for bit data. If 'for bit data', then we reinterpret it as varchar
pad must be a char / varchar of size 1 (else we throw SQLState 22022)

### **Testing**
This is tested with a variety of combinations and dimensions: 

OLTP / OLAP / OLAP with native spark

Prepared / Non prepared statements

NULL / not NULLS

Characters appearing more than once in the string expression

outputTranslationTable longer than inputTranslationTable

inputTranslationTable longer than outputTranslationTable

pad being ignored because both tables have equal size

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/ast/ISpliceVisitor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/ast/ISpliceVisitor.java
@@ -191,6 +191,7 @@ public interface ISpliceVisitor {
     Visitable visit(TernaryOperatorNode node) throws StandardException;
     Visitable visit(TestConstraintNode node) throws StandardException;
     Visitable visit(TimestampOperatorNode node) throws StandardException;
+    Visitable visit(TranslateFunctionNode node) throws StandardException;
     Visitable visit(TruncateOperatorNode node) throws StandardException;
     Visitable visit(TypeofOperatorNode node) throws StandardException;
     Visitable visit(UnaryArithmeticOperatorNode node) throws StandardException;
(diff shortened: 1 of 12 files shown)


arnaud-splice:

ipraznik-splice (approved):

martinrupp (changes requested):
- please provide a description and a "how to test"
- you should make clear (both in source and PR description) that this is the DB2 version of TRANSLATE, not the Postgres/SqlServer variant
- difference is both in order (DB2 is TRANSLATE(expr, out, in), while POSTGRES/SqlServer is TRANSLATE(expr, in, out)) and in behavior (DB2 replaces ALL occurances, and Postgres/SqlServer replace only the FIRST, and disallow different sizes in/out).

martinrupp (approved):

hatyo:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5181, opened 2021-02-17
DB-11345 Increase max map size of frequency sketch to 8192

ascend1:


ascend1:

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataType.java
```diff
@@ -1370,23 +1370,33 @@ public int compare(DataValueDescriptor o1, DataValueDescriptor o2) {
...
+     * https://datasketches.apache.org/docs/Frequency/FrequentItemsErrorTable.html .
+     * Memory consumption is (18 bytes * currentMapSize). currentMapSize starts from 8 and
+     * gets doubled when the number of items in map is bigger than 0.75 * currentMapSize.
+     * Also, currentMapSize is capped at maxMapSize.
+     */
     @Override
     public com.yahoo.sketches.frequencies.ItemsSketch getFrequenciesSketch() throws StandardException {
-        return new com.yahoo.sketches.frequencies.ItemsSketch(256);
+
+        return new com.yahoo.sketches.frequencies.ItemsSketch(8192);
```
msirek, line 1391: Maybe we can create a new static final field for this value?

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataType.java
```diff
@@ -1370,23 +1370,33 @@ public int compare(DataValueDescriptor o1, DataValueDescriptor o2) {
...
 
+    /* Bigger nomEntries leads to better accuracy. Reference:
+     * https://datasketches.apache.org/docs/Theta/ThetaErrorTable.html .
+     * Internal hash table size could also start small and grow as needed but it depends
+     * on the value of resizeFactor, or lgRF. Check Util.startingSubMultiple() in Sketch
+     * library for the logic determining the starting size.
+     */
     @Override
     public UpdateSketch getThetaSketch() throws StandardException {
         return UpdateSketch.builder().build(256);
```
carolp-503, line 1402: While in the neighborhood, can we change this one from a magic number as well?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
```diff
@@ -529,7 +533,8 @@ public void generateCost() throws StandardException {
             scanCost.setIndexLookupRows(-1.0d);
             scanCost.setIndexLookupCost(-1.0d);
         } else {
-            lookupCost = totalRowCount*filterBaseTableSelectivity*(openLatency+closeLatency);
+            double lookupCostPerRow = isOlap ? (openLatency+closeLatency) : ((openLatency+closeLatency) / 10);
```
carolp-503, line 536: Another magic number. Can we either turn this into a tunable parameter (latencyOLTPFactor?) or remove the explicit 10 so that people understand what it's about

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataType.java
ascend1: The idea is to make this parameter (together with other sketches' sizes) configurable. Jira: https://splicemachine.atlassian.net/browse/DB-11481 . So eventually we'd make the function accept a size argument from outside.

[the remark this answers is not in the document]
ascend1: Yes, we want the method to accept a size argument. Jira: https://splicemachine.atlassian.net/browse/DB-11481.

db-engine/src/main/java/com/splicemachine/db/iapi/types/DataType.java
ascend1: Sure, the scope of DB-11481 will cover all the three sketch sizes.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
ascend1: Ideally, `openLatency` and `closeLatency` can be sampled for OLTP and OLAP periodically when running queries. Then this artificial factor can be removed. For now, I changed it to a constant from a literal. Making it tunable is possible. I record this parameter in DB-11481, too.

arnaud-splice (approved):

msirek (approved):

PR #5185, opened 2021-02-18
DB-10872 consider QUOTE_ALWAYS when exporting FOR BIT DATA field 

arnaud-splice:
### Description
`QUOTE_MODE` was ignored when running `EXPORT` on `[VAR]CHAR FOR BIT DATA` fields. We now properly surround those fields with `QUOTE_CHARACTER` when `QUOTE_MODE` is set to `ALWAYS`

Example:

```
DROP TABLE foo if exists;
CREATE TABLE foo (a int, b varchar(5) , c varchar(5) for bit data, d char(5) for bit data );

insert into foo values (4, 'abcde', 'abcde', 'abcde');
insert into foo values (0, '', '', '');
insert into foo values (null, null, null, null);

EXPORT to '/tmp/DB-10872'
AS CSV
FIELD_SEPARATOR ','
QUOTE_CHARACTER '"'
QUOTE_MODE ALWAYS
```
 
```
$ cat /tmp/DB-10872/part-r-00000.csv
,,,
4,"abcde","6162636465","6162636465"
0,"","","2020202020"

``` 

### Testing
Tests were added to ExportOperationIT . We now test that properly for BIT DATA columns. We also added empty strings and NULL tests.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportCSVWriterBuilder.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/export/ExportCSVWriterBuilder.java
@@ -16,6 +16,7 @@ package com.splicemachine.derby.impl.sql.execute.operations.export;
 
 import com.splicemachine.db.iapi.sql.ResultColumnDescriptor;
 import com.splicemachine.db.iapi.types.DataTypeDescriptor;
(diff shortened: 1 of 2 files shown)


arnaud-splice:

dgomezferro (approved):

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

carolp-503 (approved):
+1

PR #5189, opened 2021-02-18
DB-11412 disable preserveLineEndings if oneLineRecords is set

martinrupp:
# Description

Before the fix, preserveLineEndings didn't use parallel readings in SpliceFileVTI. This is because we use another
code path which would not preserve line endings. This is correct, however If we have oneLineRecords, it doesn't make sense to preserve line endings, since we don't read the line endings into the data (for this, we would need multiline-line varchars). So the case oneLineRecords we can set preserveLineEndings to false, which then allows parallel reads in this case.

So, simple solution is to set preserveLineEndings to false if we have oneLineRecords (in Code).
Workaround for customer is to do this manually with global options.

# How to test
This is a performance improvement, see https://splicemachine.atlassian.net/browse/DB-11412 .

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/vti/SpliceFileVTI.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/vti/SpliceFileVTI.java
@@ -161,15 +161,20 @@ public class SpliceFileVTI implements DatasetProvider, VTICosting {
                 quotedEmptyIsNull = vtiOperation.getQuotedEmptyIsNull();
             }
 
-            boolean preserveLineEndings = getPreserveLineEndings(op);
-            if (oneLineRecords && !preserveLineEndings
-                    && (charset==null || charset.toLowerCase().equals("utf-8"))) {
+            // if we have oneLineRecords, we don't read the line endings into the data
+            // for this, we would need multiline-line varchars). so in that cases, we set preserveLineEndings = false
+
+            if (oneLineRecords && (charset==null || charset.toLowerCase().equals("utf-8"))) {
+                // full parallel execution
                 DataSet<String> textSet = dsp.readTextFile(fileName, op);
                 operationContext.pushScopeForOp("Parse File");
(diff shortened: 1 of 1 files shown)


arnaud-splice (approved):

splice_machine/src/main/java/com/splicemachine/derby/vti/SpliceFileVTI.java
```diff
@@ -161,7 +161,10 @@ boolean getPreserveLineEndings(SpliceOperation op) throws StandardException {
                 quotedEmptyIsNull = vtiOperation.getQuotedEmptyIsNull();
             }
 
-            boolean preserveLineEndings = getPreserveLineEndings(op);
+            // if we have oneLineRecords, it doesn't make sense to preserve line endings, since
+            // we don't read the line endings into the data (for this, we would need multiline-line varchars)
+            boolean preserveLineEndings = oneLineRecords ? false : getPreserveLineEndings(op);
+
             if (oneLineRecords && !preserveLineEndings
                     && (charset==null || charset.toLowerCase().equals("utf-8"))) {
```
jyuanca, line 169: can this be "if (oneLineRecords && && (charset==null || charset.toLowerCase().equals("utf-8")))", ignoring preserveLineEndings ?

jyuanca (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/vti/SpliceFileVTI.java
martinrupp: yes, of course

martinrupp:

dgomezferro (approved):

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

carolp-503 (approved):
+1

jyuanca:
Do we also need this for 3.0 and 3.1?

PR #5196, opened 2021-02-19
DB-11498 enable JMX for mem profile.

hatyo:
In this PR we enable JMX for running in mem-profile, this helps reproducing mem-profile tests which was passing on Jenkins but failing locally because in Jenkins we set the JMX when running these ITs under mem-profile, but when starting a local mem-profile cluster we do not.


martinrupp (approved):

ipraznik-splice (approved):

carolp-503 (approved):
+1

PR #5200, opened 2021-02-19
DB-11302/DB-11266 Performance improvement for statement triggers on Spark (3.1)

msirek:
# DB-11302

Performance issues fixed:
1.  Creates a persisted rdd for reading trigger referencing rows instead of reading from the temporary conglomerate each time.
2. Avoids merge join on spark if the target of an update or delete is involved.  Frequently updated tables tend to generate uneven splits unless flushed to HDFS first.  Other joins like merge sort join or broadcast join repartition the data, and create more even partitions, making these joins perform better.  More work is needed in the area of picking split points for operations which don't repartition the data set.
3. Convert the trigger rows rdd to a native spark Dataset, and apply the Dataset limit function on it when processing WHEN clause predicates.  The native spark Dataset is reused instead of building a new splice DataSet for each reference to NEW TABLE or OLD TABLE rows.

This PR should merge after DB-10410 because it fixes some serialization issues needed for ITs to pass.

# DB-11266

Fixes issues with concurrent row triggers:

1. Aborts other running trigger threads when one thread errors out, so the main thread doesn't try to clean up resources actively being used by running triggers.
2. Increase the size of the stored prepared statement cache, to decrease likelihood two threads are trying to recompile the same trigger simultaneously.
3. Use a ConcurrentHashMap-backed HashSet for the "children" Set of transactions in AbstractTxn, to avoid ConcurrentModificationException errors.

Part 3 will alleviate customer issues in the interim until DB-10276 is complete.  In that Jira I'm working on an issue where the LanguageConnectionContext used to get the transaction to elevate when compiling a trigger may sometimes be the wrong one (is gotten from the ContextService instead of the Activation).  DB-10276 will remove the use of the ConcurrentHashMap in AbstractTxn.


msirek:
Making this a draft PR for now.  DB-10410 needs to merge to 3.1 before this PR can merge.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2590/

msirek:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
Approved for 3.1

PR #5206, opened 2021-02-19
DB-11500 Stop throwing when using getKeyColumnPosition with indexes on expr

arnaud-splice:
### Description

```
splice> create table foo (a int primary key, b int, c int);
ELAPSED TIME = 313 milliseconds
splice> create index bar on foo(b+c);
ELAPSED TIME = 630 milliseconds
splice> --select * from sysibm.systables;
splice> create index barz on foo(a+1);
ELAPSED TIME = 141 milliseconds
splice> select * from sysibm.systables;
ERROR XJ001: Java exception: 'java.lang.IllegalArgumentException: Cannot retrieve ordinal position of a base table column in an index defined on expressions because it may appear in multiple expressions.: java.io.IOException'.
ERROR XJ001: Java exception: 'Cannot retrieve ordinal position of a base table column in an index defined on expressions because it may appear in multiple expressions.: java.lang.IllegalArgumentException'.
```

The problem is visible in SYSIBM.SYSTABLES, but not limited to that one. Many views use 

```
SYSCONGLOMERATES.DESCRIPTOR.getKeyColumnPosition()
```

This function currently throws an exception if called on an index on expression.
This was mitigated for some views by using 

```
SYSCONGLOMERATES.DESCRIPTOR.isOnExpression()
```
However, there is no guarantee in the order of execution of predicates in SQL, so 

```
NOT SYSCONGLOMERATES.DESCRIPTOR.isOnExpression()
AND SYSCONGLOMERATES.DESCRIPTOR.getKeyColumnExpression()
```
could be evaluated in any order, causing sporadic exceptions.

Moreover, we also usually test the nullability of the column before running this :

```
SYSCONGLOMERATES IS NOT NULL AND
NOT SYSCONGLOMERATES.DESCRIPTOR.isOnExpression() AND
SYSCONGLOMERATES.DESCRIPTOR.getKeyColumnExpression()
```

Testing the nullability is (currently) not sporadic, because the null test predicate can be pushed down to table scan, but not the other two predicates. This is nevertheless pretty brittle and should not be done this way.

This PR does 3 things:

1. getKeyColumnPosition() should return -1 if isOnExpression() instead of throwing and the code of each view should be adapted to test getKeyColumnPosition() > 0 instead of getKeyColumnPosition() <> 0

2. Use CASE WHEN to force evaluation order between the nullability check and the function call: CASE WHEN DESCRIPTOR IS NULL THEN 0 ELSE DESCRIPTOR.getKeyColumnPosition()

3. Refactor upgrade scripts to be able to update views easily during an upgrade

### Testing:
A test was added to ensure that the described problem is fixed.
The test that checked that getKeyColumnPosition should throw now tests that it should return -1


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2487/

hatyo:
test failure seems to be sporadic as reported in DB-11165. Retrying CI.

hatyo:

msirek (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2490/

arnaud-splice:
Turned this into a draft until I fix an error in IndexIT

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

carolp-503 (approved):
+1

PR #5207, opened 2021-02-19
DB-11504 Add instructions for startup for hdp

bklo94:
# Short Description
Documentation update to match Cloudera documentation. 

# Long Description
Encountered this issue with olapserver not starting up in HDP due to these steps missing from the HDP documentation. The steps are currently in the CDH documentation. The main important part of the documentation is the HDFS commands to add the user and splice directories with the correct permission.

# How to test
New clusters should be able to startup and run OlapServer queries. A simple hinted query should run
```
SELECT * FROM SYS.SYSTABLES --splice-properties useSpark=true
;
```


hatyo (approved):

arnaud-splice (approved):

martinrupp (approved):

carolp-503 (approved):
+1

PR #5217, opened 2021-02-20
DB-11511 Fix logical conflict between DB-11121 & DB-10968

arnaud-splice:
DB-11121 introduced a new call to executeUpdate(CALL SYSCS_UTIL) but DB-10968 changed those to execute instead of executeUpdate


jyuanca (approved):

arnaud-splice:
Not necessary anymore since DB-10968 will have to be redone

PR #5227, opened 2021-02-22
DB-11517 DB2 compatible varchar for (Native)SparkDataSet.distinct()

ascend1:
This change fixes distinct operation when result set contains varchar columns on native Spark. According to SQL standard, character string comparison are done by padding the shorter string to the longer string's length according to collation in use. On native Spark, however, there is no such padding. 'abc' and 'abc  ' are different strings. Method `distinct()` or `dropDuplicates()` do not remove any of them since they are different.

To make Spark's `distinct()` and `dropDuplicates()` work as SQL standard defines, all StringType columns are rtrimmed before running the distinct operation. rtrimmed columns are added as extra columns to the data frame so that in the end we return a row from original data for each group, not a row with rtrimmed values.

This fix is currently guarded under DB2 compatible varchar mode. Although it should be standard compliant, exposing it unconditionally would lead to inconsistent results on OLTP and OLAP. Moreover, Postgresql doesn't seem to be standard compliant on this point, either.

Current tests include
1. explicit distinct keyword in select list, and
2. union distinct semantic.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
@@ -223,20 +223,28 @@ public class NativeSparkDataSet<V> implements DataSet<V> {
         boolean varcharDB2CompatibilityMode = Boolean.parseBoolean(varcharDB2CompatibilityModeString);
         pushScopeIfNeeded(context, pushScope, scopeDetail);
         try {
+            // Do not replace string columns using rtrimmed columns but add them as extra columns.
+            // After distinct, for each group, we need to return a row originally in that group, not
+            // a row with rtrimmed values.
+            List<String> compColNames = new ArrayList<>();
             if (varcharDB2CompatibilityMode) {
-                List<String> stringColNames = new ArrayList<>();
+                List<String> strColNames = new ArrayList<>();
                 Tuple2<String, String>[] colTypes = dataset.dtypes();
                 for (Tuple2<String, String> colType : colTypes) {
                     if (colType._2().equals("StringType")) {
-                        stringColNames.add(colType._1());
+                        strColNames.add(colType._1());
+                    } else {
+                        compColNames.add(colType._1());
                     }
                 }
-                for (String colName : stringColNames) {
-                    dataset = dataset.withColumn(colName, rtrim(col(colName)));
+                for (String colName : strColNames) {
+                    String newColName = colName+"_rtrimmed";
+                    dataset = dataset.withColumn(newColName, rtrim(col(colName)));
+                    compColNames.add(newColName);
                 }
             }
-            Dataset<Row> result = dataset.distinct();
 
+            Dataset<Row> result = dataset.dropDuplicates(compColNames.toArray(new String[0]));
             return new NativeSparkDataSet<>(result, context);
         } finally {
             if (pushScope) context.popScope();
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
@@ -15,9 +15,9 @@
 
 package com.splicemachine.derby.stream.spark;
 
-import splice.com.google.common.base.Function;
-import splice.com.google.common.collect.Iterators;
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.Property;
+import com.splicemachine.db.iapi.services.property.PropertyUtil;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.types.SQLLongint;
 import com.splicemachine.db.impl.sql.compile.ExplainNode;
@@ -41,12 +41,18 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.spark.SparkContext;
 import org.apache.spark.api.java.JavaRDD;
 import org.apache.spark.api.java.function.FlatMapFunction;
-import org.apache.spark.scheduler.*;
+import org.apache.spark.scheduler.SparkListener;
+import org.apache.spark.scheduler.SparkListenerJobStart;
+import org.apache.spark.scheduler.SparkListenerTaskEnd;
+import org.apache.spark.scheduler.StageInfo;
 import org.apache.spark.sql.*;
 import org.apache.spark.sql.types.DataType;
 import org.apache.spark.sql.types.StructType;
 import org.apache.spark.storage.StorageLevel;
+import scala.Tuple2;
 import scala.collection.JavaConverters;
+import splice.com.google.common.base.Function;
+import splice.com.google.common.collect.Iterators;
 
 import javax.annotation.Nullable;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -205,17 +211,40 @@ public void unpersistIt() {
...
+                    }
+                }
+                for (String colName : strColNames) {
+                    String newColName = colName+"_rtrimmed";
+                    dataset = dataset.withColumn(newColName, rtrim(col(colName)));
+                    compColNames.add(newColName);
+                }
+            }
 
+            Dataset<Row> result = dataset.dropDuplicates(compColNames.toArray(new String[0]));
```
msirek, line 247: The fix looks good, but when varcharDB2CompatibilityMode is false I think we need to return dataset.distinct().

msirek (changes requested):

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DB2VarcharCompatibilityIT.java
```diff
@@ -480,4 +495,30 @@ public void testParameterizedMultiProbeScan() throws Exception {
         testPreparedQuery1(sqlTemplate, expected, false, true);
 
     }
+
+    @Test
+    public void testSparkDistinct() throws Exception {
+        String query = format("select distinct v1, v2 from z --splice-properties useSpark=true", useSpark);
```
msirek, line 501: Can you please change useSpark=true to useSpark=%s so we test control also?

martinrupp (changes requested):
please update description:
- Description (what has been done and why)
- How to test (if bugfix: old behavior or sql that causes bug, vs. new behavior)

ascend1:

ascend1:
> please update description:
> 
>     * Description (what has been done and why)
> 
>     * How to test (if bugfix: old behavior or sql that causes bug, vs. new behavior)

Done

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
hatyo, line 247: I think we should drop newly added columns `XXX_rtrimmed` because we're changing the DF structure which is not part of `distinct` API contract.

hatyo (commented):
I think we could have an alternative implementation approach which might be better in terms of memory consumption.

In the current approach we do copy the string columns (rtrimmed) which could be very expensive, maybe a better approach would be:

1. define a new class e.g. `PaddedString` with custom-defined `equal` and `hashcode` that does proper comparisons w.r.t. distinct.
2. in `NativeSparkDataSet.distinct` we map all string columns to the newly created type `PaddedString`, call distinct, and map back `PaddedString` to `String`.

I haven't investigated the implementation complexity of this, but I reckon it is not very easy, e.g. we must implement serialization/deserialization of `PaddedString`.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Yes, that's right. For clarity, I still changed it to use `distinct()` when `compColNames` is empty, partly also because I was not able to verify it based on Spark's source code. Their `dropDuplicates()` implementation (no args) passes `df.columns` in.

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/DB2VarcharCompatibilityIT.java
ascend1: Sure, it's a copy-pasting error and fixed in the latest commit.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Yes, extra columns are removed after distinct operation returns in the latest commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2516/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2518/

msirek (approved):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -205,17 +211,50 @@ public void unpersistIt() {
...
+                for (Tuple2<String, String> colType : colTypes) {
+                    if (colType._2().equals("StringType")) {
+                        strColNames.add(colType._1());
+                    } else {
+                        compColNames.add(colType._1());
+                    }
+                }
+                for (String colName : strColNames) {
+                    String newColName = colName+"_rtrimmed";
+                    dataset = dataset.withColumn(newColName, rtrim(col(colName)));
```
dgomezferro, line 243: We've run into performance issues when using `.withColumn` repeatedly, it'd be better to use `.withColumns` to add all columns in bulk. See https://splicemachine.atlassian.net/browse/DB-9599 for reference.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -205,17 +211,50 @@ public void unpersistIt() {
...
 
+            Dataset<Row> result;
+            if (compColNames.isEmpty()) {
+                result = dataset.distinct();
+            } else {
+                result = dataset.dropDuplicates(compColNames.toArray(new String[0]));
+            }
+            for (String extraColName : extraColNames) {
+                result = result.drop(extraColName);
+            }
```
dgomezferro, line 257: Same here, use bulk `.drop(colNames: String*)` instead

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -205,17 +211,50 @@ public void unpersistIt() {
...
+    public DataSet<V> distinct(OperationContext context) throws StandardException {
         return distinct("Remove Duplicates", false, context, false, null);
     }
 
     @SuppressWarnings({ "unchecked", "rawtypes" })
     @Override
-    public DataSet<V> distinct(String name, boolean isLast, OperationContext context, boolean pushScope, String scopeDetail) {
+    public DataSet<V> distinct(String name, boolean isLast, OperationContext context, boolean pushScope, String scopeDetail) throws StandardException {
+        String varcharDB2CompatibilityModeString =
+        PropertyUtil.getCachedDatabaseProperty(context.getActivation().getLanguageConnectionContext(), Property.SPLICE_DB2_VARCHAR_COMPATIBLE);
```
dgomezferro, line 222: Indentation is wrong here

dgomezferro (changes requested):
I have some concerns regarding performance for dataset with many Varchar columns, but great fix otherwise!

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Done.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Done.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Done.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Done.

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
ascend1: Done.

ascend1:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

carolp-503 (approved):
+1

martinrupp (approved):

PR #5234, opened 2021-02-24
DB-11342 Fix ArrayIndexOutOfBoundsException with NLJ and triggers (3.1)

msirek:
# Short Description
Fixes ArrayIndexOutOfBoundsException and NPE during execution of a triggers which accesses trigger referencing rows via nested loop join by making sure the correct TriggerExecutionContext is always being used.
# Long Description
- Gets rid of the problematic pushTriggerExecutionContextFromActivation.  The TriggerExecutionContext should come from the local LCC from the thread's context manager.
- Updates the trigger rowHolder to always point to the correct TriggerExecutionContext upon deserialization.
- In TriggerNewTransitionRows.getDataSet, use the TriggerExecutionContext from the class field, as it should always be correct.  The object is instantiated when the operation tree's init functions are called and allocates a new TriggerExecutionContext for the thread.

# How to test
The specific test from LiveWire was added to the ITs.


carolp-503 (approved):
Approved for 3.1

PR #5236, opened 2021-02-25
DB-11535 Fix trigger performance regression from DB-10276 (3.1)

msirek:


carolp-503 (approved):
Approved for 3.1

carolp-503 (commented):
Approved for 3.1

PR #5238, opened 2021-02-26
DB-11412 disable preserveLineEndings if oneLineRecords is set (3.1)

martinrupp:
## Description

Before the fix, preserveLineEndings didn't use parallel readings in SpliceFileVTI. This is because we use another
code path which would not preserve line endings. This is correct, however If we have oneLineRecords, it doesn't make sense to preserve line endings, since we don't read the line endings into the data (for this, we would need multiline-line varchars). So the case oneLineRecords we can set preserveLineEndings to false, which then allows parallel reads in this case.

So, simple solution is to set preserveLineEndings to false if we have oneLineRecords (in Code).
Workaround for customer is to do this manually with global options.

## How to test
This is a performance improvement, see https://splicemachine.atlassian.net/browse/DB-11412 .


carolp-503 (approved):
Approved for 3.1

PR #5240, opened 2021-02-26
DB-10968 fix Trigger_Exec_Stored_Proc_IT

martinrupp:
## Short Description
fixing IT failure in Trigger_Exec_Stored_Proc_IT

## Long Description
1. fixing IT failure in Trigger_Exec_Stored_Proc_IT
2. `SYSCS_SET_LOGGER_LEVEL` is now set to only allow admins to change this - normal user shouldn't set log level to DEBUG and slow down the system. Execution prevention is currently implemented with  `RoutineAliasInfo.MODIFIES_SQL_DATA` (we should change this, see DB-11542), which also prevents this function then to be used in BEFORE triggers (`java.sql.SQLSyntaxErrorException: Procedures that modify SQL data are not allowed in BEFORE triggers`). Therefore we're now using .after() in `Trigger_Exec_Stored_Proc_IT.testStatementTriggerSysStoredProc`.

## How to test
(fix for IT)

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Exec_Stored_Proc_IT.java
+++ b/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Exec_Stored_Proc_IT.java
@@ -39,7 +39,7 @@ import splice.com.google.common.collect.Lists;
  * Note the dependency on user-defined stored procedures in this test class.<br/>
  * See {@link TriggerProcs} for instructions on adding/modifying store procedures.
  */
-@Category({SerialTest.class})
+@Category({SerialTest.class, LongerThanTwoMinutes.class})
 @RunWith(Parameterized.class)
 public class Trigger_Exec_Stored_Proc_IT  extends SpliceUnitTest {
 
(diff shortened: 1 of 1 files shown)


hatyo:
Thanks Martin for the quick fix.

> 3. removing LongerThanTwoMinutes Category since it takes 3s on mem and 45s on cdh6.3.0

Could you please revert that? Jenkins is much slower than our Laptops and it could take more than two minutes.

arnaud-splice:
> Thanks Martin for the quick fix.
> 
> > 1. removing LongerThanTwoMinutes Category since it takes 3s on mem and 45s on cdh6.3.0
> 
> Could you please revert that? Jenkins is much slower than our Laptops and it could take more than two minutes.

Last time we gathered statistics, this took 164 seconds on Jenkins, so I would also be in favour of removing that.

arnaud-splice (approved):

hatyo (approved):

hatyo:
Tested locally and it passes, bringing it in immediately since it fixes master branch.

PR #5242, opened 2021-02-26
DB-10948 use dep. mgr to invalidate statements when properties change.

hatyo:
## Short Description
Use dependency manager to invalidate only the relevant cached SQL statement when setting a property.

## Long Description
In a [previous PR](https://github.com/splicemachine/spliceengine/pull/4740/files#diff-2865a28e52939a573ffd050822d006469ec18ae401825d18d982db8a93cd1427R1287) we changed how `SYSCS_SET_GLOBAL_DATABASE_PROPERTY` works s.t. it invalidates all cached statements, that was necessary because we introduced a new property that changes the way a SQL statement processes its timestamp fields, therefore, when changing this property (and other newly-introduced ones) we should invalidate the cache to make sure that SQL statements pick up the new value of a system property.

In this PR, we improve our approach by:
1. Introduce a new `SPSProperty` which acts as new `Provider` in the `DependencyManager` subsystem.
2. Model properties that affect query execution as `SPSProperty` classes.
3. Introduce new hook in `QueryTreeNode` called `addSPSPropertyDependency`, we use this hook during binding of certain nodes that we know are affected by some SPS properties (such as `CastNode` and others), for more information on why we do this, read the subsection below.
4. Once we (re)set a system property, we check if it is an `SPSProperty`, if so, then we use the `DependencyManager` subsystem to look up dependent SPS objects and invalidate them across all region servers.

As a result, the overall robustness is slightly improved when setting a system property because we do not blindly invalidate all statements, we invalidate exactly those affected by the property we're setting and other irrelevant cached statements are kept untouched.

### Processing SPS properties during binding

In Derby/SpliceMachine we do not have a dedicated binding phase which only resolves references without altering the AST, binding phase is a bit misleading, since it changes the AST besides binding its symbols, for example, it collapses constant casts, so if we had a node such as `char(timestamp('2021-03-21 17:27:29.12345'))` it will be collapsed after binding to `char('2021-03-21 17:27:29.12345')`, if we analyze the plan looking for SPS timestamp property after this transformation we will not associate that statement with the timestamp property which is incorrect.

Doing the SPS property analysis after parsing and before binding is also incorrect because at this point many nodes such as column references are not resolved to their proper data types, therefore we could e.g. miss linking a floating-point SPS property with a (still unresolved) floating-point column reference.

The original design of the SPS property assumed we had such a phase allowing us to visit the plan looking for dependencies. Now, since we don't have such phase, we would use the new hook in certain locations inside each node to expose it the `SPSPropertyRegistry`.

Maybe if we have such a phase in the future, we could refactor it again.

## How to test

Testing the new behavior for regressions is done by ITs introduced before for the (now) SPS properties such as `TimestampIT`, `SpliceDateFunctionsIT`, and `NumericTypesToStringTypesConversionIT.java`.

I also did another synthetic benchmark and noticed some performance improvement when calling `SYSCS_SET_GLOBAL_DATABASE_PROPERTY('splice.function.timestampFormat',...)` after warming up the statement cache with one SELECT statement containing timestamp field and another 25,000 statements that do not have any timestamp fields. Here are the results:

| Command | Description |
| --- | --- |
| [master](d67c3c45d1210a7880525d5b976bf5e8d32fbfad) | 259ms |
| DB-10948 | 167ms |


hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2542/

hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2547/

ascend1 (approved):
With a minor comment.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertySecondFunction.java
```diff
@@ -0,0 +1,42 @@
...
+        super(uuid, name);
+    }
+
+    @Override
+    protected void checkAndAddDependency(final StatementNode statementNode, CompilerContext cc) throws StandardException {
+        assert statementNode != null;
+        assert cc != null;
+        final CollectNodesVisitor v = new CollectNodesVisitor(SecondFunctionNode.class);
+        statementNode.accept(v);
+        if(!v.getList().isEmpty()) {
```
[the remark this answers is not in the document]
hatyo: Thank you, I didn't know such visitor exists, I will use it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A static registry that keeps track of all SPS properties and checks whether a dependency should be created
+ * from a statement plan to a specific SPS property.
+ *
+ * If you add a new SPS property, make sure to register it here so it is checked when analyzing a statement plan.
+ */
+public class SPSPropertyRegistry {
+    private static Set<SPSProperty> propertySet = ConcurrentHashMap.newKeySet();
+    static {
+        propertySet.add(new SPSPropertyTimestampFormat(new BasicUUID("91916f24-fa35-493d-9048-e41ffbe004d7"), Property.SPLICE_TIMESTAMP_FORMAT));
```
dgomezferro, line 34: Is this still a bit too aggressive? This should only affect Timestamp -> Char conversions, but this invalidates all statements with a timestamp reference if I understand it correctly. Would it be possible to restrict it further?

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+        propertySet.add(new SPSPropertyTimestampFormat(new BasicUUID("b2decb1d-acc2-4402-9278-d4aab1fc6431"), Property.SPLICE_CURRENT_TIMESTAMP_PRECISION));
```
dgomezferro, line 35: Similarly here, I think this property only affects CURRENT_TIMESTAMP, can we restrict the invalidation to statements that contain CurrentDatetimeOperatorNode ?

dgomezferro (changes requested):
I have some comments regarding the accuracy of the invalidations, but this is great change overall!

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
arnaud-splice, line 34: How were those UUIDs defined?

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertySecondFunction.java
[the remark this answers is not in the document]
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
hatyo: done. good catch, BTW, I think `splice.function.currentTimestampPrecision` is already deprecated, but still.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
hatyo: Generated manually using a simple program.

martinrupp:
this looks like great work, but... to be honest i don't really understand how we get that something depends on e.g. currentTimestampPrecision, so i'm not sure how you want to guarantee that the dependency information is always correct. I guess you checked this, but are we sure this is also correct if someone writes a bit more stupid code, or a statement is only getting a config when some other config is set, would that also work? To me it looks a bit risky, if we only save 100ms each time we set a config value (which shouldn't happen often, no?)

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/SPSPropertyRegistry.java
hatyo: I did some refactoring, please revise my PR message for more information about the new approach. Now, we scan exactly the `CAST` pattern from timestamp to string type.

hatyo:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
```diff
@@ -1620,6 +1620,8 @@ public boolean collectExpressions(Map<Integer, Set<ValueNode>> exprMap) {
         return collectSingleExpression(exprMap);
     }
 
+
+
```
ascend1, line 1624: Extra newlines.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
```diff
@@ -300,6 +300,8 @@ public ValueNode bindExpression(FromList fromList,
             castOperand.setType(getTypeServices());
         }
 
+        addSPSPropertyDependency();
```
ascend1, line 303: 1. I'm not sure if at this point we know the target type of the cast. We bind only the operand and not yet the cast node itself.
2. bindExpression() is called only for explicit casts users write in queries. If a `CastNode` is added in our binding phase as an implicit cast, we will just call `((CastNode) newNode).bindCastNodeOnly()`. If such a cast happen to trigger a dependency, we will miss it. Question is if that should be how it works by design.

ascend1 (commented):
A few questions:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2555/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ValueNode.java
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CastNode.java
hatyo: > 1. I'm not sure if at this point we know the target type of the cast. We bind only the operand and not yet the cast node itself.
> 2. bindExpression() is called only for explicit casts users write in queries. If a CastNode is added in our binding phase as an implicit cast, we will just call ((CastNode) newNode).bindCastNodeOnly(). If such a cast happen to trigger a dependency, we will miss it. Question is if that should be how it works by design.

yes such a cast could cause a dependency trigger, for example `upper(timestamp('2020-10-10 01:02:03.123456'))` will create such a cast and it is important to create a dependency to e.g. timestamp format SPS property. I moved the invocation of `addSPSPropertyDependency()` from `bindExpression` to `bindCastNodeOnly` and removed it from `com.splicemachine.db.impl.sql.compile.SimpleStringOperatorNode#bindExpression` since it is redundant.

hatyo:
> this looks like great work, but... to be honest i don't really understand how we get that something depends on e.g. currentTimestampPrecision, so i'm not sure how you want to guarantee that the dependency information is always correct. I guess you checked this, but are we sure this is also correct if someone writes a bit more stupid code, or a statement is only getting a config when some other config is set, would that also work? To me it looks a bit risky, if we only save 100ms each time we set a config value (which shouldn't happen often, no?)

@martinrupp, thanks for the comment, according to DB-10994 it could take much longer, also as @msirek hinted out in DB-10948:

> If a customer uses SYSCS_INVALIDATE_STORED_STATEMENTS frequently, SQL and triggers may get parsed and compiled more often instead of using the statement cache.

So, overall we could use such improvement. Regarding verifying dependencies, please note that all the checks are done by pattern matching, for example, format timestamp, we pattern match cast from timestamp to string value, we have some tests for that in `TimestampIT` that should detect potential future problems, same goes for other properties such as `second` function and so on, but I can not guarantee 100% that it is error proof against programming mistakes, I think this approach could've been simplified if we had a clear binding step -which we don't, since binding involves some optimization rules as I explained above- but still, this is the closest I could get keeping it clean and easy to extend, any suggestions to improve it are welcome!

martinrupp:
ok. maybe you can give me a run-down next days how this works ... i'm just afraid someone (like me) would do sth stupid and the dependency system doesn't work anymore... can the customer somehow FORCE complete invalidation?

hatyo:
> ok. maybe you can give me a run-down next days how this works ... i'm just afraid someone (like me) would do sth stupid and the dependency system doesn't work anymore... can the customer somehow FORCE complete invalidation?

Sure, I will be happy to, let's find some time tomorrow to go over this PR. The customer can always explicitly call [SYSCS_INVALIDATE_STORED_STATEMENTS()](https://docstest.splicemachine.com/sqlref_sysprocs_invalidatestoredstmts.html) and [SYSCS_EMPTY_GLOBAL_STATEMENT_CACHE()](https://docstest.splicemachine.com/sqlref_sysprocs_emptyglobalcache.html) to invalidate the caches.

hatyo:

spliceci:
TEST SUCCEEDED +1

martinrupp:
> How to test

Since the QA team doesn't run ITs it would be great - if possible - to provide a small SQL example or pseudo-code SQL: A stored procedure, a table, then run 25k this query, observed behavior before fix, after fix.

hatyo:
> Since the QA team doesn't run ITs it would be great - if possible - to provide a small SQL example or pseudo-code SQL: A stored procedure, a table, then run 25k this query, observed behavior before fix, after fix.

Good point, I haven't written yet a test plan for the QA on the JIRA ticket, will do now and upload my test script there, thanks!

dgomezferro (approved):

PR #5244, opened 2021-03-02
DB-11559 Improve conglomerate cost estimation performance

ascend1:
## Short Description
This change improves conglomerate cost estimation performance.

## Long Description
After frequency sketch size is increased to 8192, some queries get slower on first run. The time spent in optimizer is increased drastically.

There are three issues:
1. String literals representing date/time/timestamp are used to compare with column statistics values directly. The string literals are converted to date/time/timestamp values over and over again. Example predicate is `DATE_COL > '2010-11-05'
`.
2. `frequenciesSketch.getFrequentItems(ErrorType.NO_FALSE_POSITIVES)` sorts items internally. With sketch size increased to 8192, this sort involves a lot of `DataType.compare()`, making it slow.
3. `ColumnStatisticsImpl.getSkewedRowCountInRange()` compares every skewed value to start value and stop value. With sketch size increased to 8192, the number of skewed values becomes big and the loop doing comparison becomes slow.

In this change, issue 2 is addressed by caching `NO_FALSE_POSITIVES` in `ColumnStatisticsImpl`. Issue 1 is addressed by evaluating the implicit cast in binary comparison operator after the implicit cast node is added if the cast operand is a constant. Doing so is consistent with explicit cast on constant literals, which gets evaluated by default.

After doing this two items, issue 3 doesn't seem to be a serious issue for now because comparing between `SQLDate`s without converting is pretty fast.

## How to test
1. Restore the customer backup.
2. Explain the query shown in Jira DB-11559.
Before the change, explaining the query takes no shorter than 2000 ms in hot runs (typically 2500 ms).
After the change, explaining the query should take no longer than 1000 ms in hot runs (typically around 500 ms).

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/stats/ColumnStatisticsImpl.java
@@ -507,7 +507,7 @@ public class ColumnStatisticsImpl implements ItemStatistics<DataValueDescriptor>
          * it with the lower bound of rpv and upper bound of total not-null rows.
          */
 
-        long qualifiedRows = 0;
+        long qualifiedRows;
         /* 1. range selectivity returned by CDF */
         if (!includeStart && start != null && !start.isNull())
             start = new StatsExcludeStartDVD(start);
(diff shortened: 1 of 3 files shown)


jyuanca (approved):

msirek (approved):

carolp-503 (approved):
+1

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #5246, opened 2021-03-02
DB-11552 Fix DB2VarcharCompatibility mode for parameterized queries (3.1)

msirek:
## Short Description
Queries involving a varchar index and IndexLookup step return wrong results with database property splice.db2.varchar.compatible = true.

## Long Description
In non-parameterized queries, a SQLVarcharDB2Compatible object is created in the gen phase (query compilation) for the value used in an index lookup (as opposed to the normal SQLVarchar).  For parameterized queries, we built an array of DataTypeDescriptors and end up calling DataTypeDescriptor.getNull() which checks the column data type and builds a null SQLVarchar.  

The fix is to introduce a new data type, VARCHAR_DB2_COMPATIBLE_TYPE_ID, to be used in place of VARCHAR_TYPE_ID in the parameter row data types array.  It will be used for internal purposes only.  The only usage of the parameter data types is to get a new null, so it should not require updating the code to support VARCHAR_DB2_COMPATIBLE_TYPE_ID wherever VARCHAR_TYPE_ID is used.

## How to test

```
call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', true);
drop schema DB_11552 cascade;
create schema DB_11552;
set schema DB_11552;
create table a ( v1 varchar(10) not null default '', v2 varchar(10) not null default '');
create index idx1 on a (v2);
create index idx2 on a (v2, v1);
create index idx3 on a (v1);

insert into a values ('hello', 'there');

prepare f1 as 'select v1 from a --splice-properties index=IDX1
    where v2=?';
-- Should return 'hello', before fix returned nothing
execute f1 using 'values (''there '')';

prepare f2 as 'select v1 from a --splice-properties index=IDX3
    where v2=?';
-- Should return 'hello', before fix returned nothing
execute f2 using 'values (''there '')';
```


msirek:
To merge after DB-10410 is available.

carolp-503 (approved):
Approved for 3.1

spliceci:
TEST SUCCEEDED +1

PR #5248, opened 2021-03-03
DB-11564 Support hinting cursor '?' 

arnaud-splice:
## Short Description
Introduced `splice.bind.cursorUntypedExpressionType` to hint `select ?`'s type.

## Long Description
This PR introduces a new global database property: `splice.bind.cursorUntypedExpressionType`.
As of now, the only supported value (beside null) for that parameter is `'varchar'`.
If that parameter is set, we assume the type of a '?' at top select level to be `VARCHAR(254)`.
This is a workaround introduced until we provide proper deferred prepare statement support, at which point, this parameter can be discontinued once DB-11582 is addressed.

## How to test
```
splice> call syscs_util.syscs_set_global_database_property('splice.bind.cursorUntypedExpressionType', null);
name                |value
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PROPERTY_NAME       |splice.bind.cursorUntypedExpressionType
NEW VALUE           |NULL
PREVIOUS VALUE      |varchar
INFO                |

4 rows selected
ELAPSED TIME = 27 milliseconds
splice> prepare foo as 'select ?';
ERROR 42X34: There is a ? parameter in the select list.  This is not allowed.


splice> call syscs_util.syscs_set_global_database_property('splice.bind.cursorUntypedExpressionType', 'varchar');
name                |value
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PROPERTY_NAME       |splice.bind.cursorUntypedExpressionType
NEW VALUE           |varchar
PREVIOUS VALUE      |int
INFO                |

4 rows selected
ELAPSED TIME = 43 milliseconds
splice> prepare foo as 'select ?';
ELAPSED TIME = 6 milliseconds
splice> execute foo using 'values ''abc''';
Splice WARNING: Autocommit may close using result set
1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
abc

1 row selected
ELAPSED TIME = 11 milliseconds       

```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
@@ -46,8 +46,10 @@ import com.splicemachine.db.iapi.sql.depend.Dependency;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.sql.dictionary.SchemaDescriptor;
 import com.splicemachine.db.iapi.sql.execute.ExecutionContext;
+import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.types.FloatingPointDataType;
 import com.splicemachine.db.iapi.types.SQLTimestamp;
+import com.splicemachine.db.iapi.types.TypeId;
 import com.splicemachine.db.iapi.util.ByteArray;
 import com.splicemachine.db.iapi.util.InterruptStatus;
 import com.splicemachine.db.impl.ast.JsonTreeBuilderVisitor;
@@ -67,6 +69,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.Timestamp;
+import java.sql.Types;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -491,6 +494,7 @@ public class GenericStatement implements Statement{
         setSecondFunctionCompatibilityMode(lcc, cc);
         setFloatingPointNotation(lcc, cc);
         setOuterJoinFlatteningDisabled(lcc, cc);
+        setCursorUntypedExpressionType(lcc, cc);
 
         if (!cc.isSparkVersionInitialized()) {
             setSparkVersion(cc);
@@ -765,6 +769,15 @@ public class GenericStatement implements Statement{
                 timestampFormatString = CompilerContext.DEFAULT_TIMESTAMP_FORMAT;
             }
             cc.setTimestampFormat(timestampFormatString);
+        }
+    }
+
+    private void setCursorUntypedExpressionType(LanguageConnectionContext lcc, CompilerContext cc) throws StandardException {
+        String type = PropertyUtil.getCachedDatabaseProperty(lcc, Property.CURSOR_UNTYPED_EXPRESSION_TYPE);
+        if(type != null && "varchar".equals(type.toLowerCase())) {
+            cc.setCursorUntypedExpressionType(DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.VARCHAR, true, 254));
+        } else {
+            cc.setCursorUntypedExpressionType(null);
         }
     }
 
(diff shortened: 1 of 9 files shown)


hatyo:
Please remove
> (e.g. short description of the PR)
from your PR message.

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -768,6 +772,15 @@ private void setTimestampFormat(LanguageConnectionContext lcc, CompilerContext c
         }
     }
 
+    private void setCursorUntypedExpressionType(LanguageConnectionContext lcc, CompilerContext cc) throws StandardException {
+        String type = PropertyUtil.getCachedDatabaseProperty(lcc, Property.CURSOR_UNTYPED_EXPRESSION_TYPE);
+        if(type != null && "varchar".equals(type.toLowerCase())) {
+            cc.setCursorUntypedExpressionType(DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.VARCHAR, true, 254));
+        } else {
+            cc.setCursorUntypedExpressionType(null);
```
hatyo, line 780: Instead of silently setting it to `null`, would it make more sense to throw e.g. UnsupportedTypeException or something similar?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumn.java
```diff
@@ -1510,6 +1511,13 @@ void rejectParameter() throws StandardException
             throw StandardException.newException(SQLState.LANG_PARAM_IN_SELECT_LIST);
     }
 
+    void setUnknownParameterType(DataTypeDescriptor type) throws StandardException
+    {
+        if (expression != null && expression.isParameterNode()) {
```
hatyo, line 1516: This looks a bit strange to me, should this be an assertion instead, i.e. isn't a programming error to set a parameter type to a non-parameter column?

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/PreparedStatementIT.java
```diff
@@ -35,6 +37,7 @@
  * @author Jeff Cunningham
  *         Date: 6/19/13
  */
+@Category(SerialTest.class) // Not run in parallel since it is changing global DB configs
```
hatyo, line 40: Instead of marking the entire suite as serial, potentially increasing the build time, does it make sense to move this system-property test another serial test suite (or have a dedicated serial test suite only for those properties).

hatyo (changes requested):
Could you please check whether it is necessary to invalidate a cached statement when this property changes? If so, you should integrate with the SPS Property API introduced in [DB-10948](https://github.com/splicemachine/spliceengine/pull/5242).

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ResultColumn.java
arnaud-splice: This function is modelled after rejectParameterColumn. I could rename it to setPotentialUnknownParameterType if that makes more sense. The idea being that we want to set the type iff it's a parameterNode

arnaud-splice:
> Could you please check whether it is necessary to invalidate a cached statement when this property changes? If so, you should integrate with the SPS Property API introduced in [DB-10948](https://github.com/splicemachine/spliceengine/pull/5242).
We do not need to invalidate cached statements for this one.

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/PreparedStatementIT.java
arnaud-splice: Good idea!

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
arnaud-splice: Will do, thanks!

dgomezferro (approved):
+1 once the reported issues are addressed

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
arnaud-splice: Tracked by DB-11617

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/PreparedStatementIT.java
arnaud-splice: Tracked by DB-11618

carolp-503 (approved):
+1

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #5249, opened 2021-03-03
DB-11558 Update sketch library to Apache DataSketches 2.0.0

ascend1:
## Short Description
This change solves the assertion failure on drift >= 1024 in sketch library so that analyzing a table after sketch size is increased to 8192 returns successfully.

## Long Description
In `com.yahoo.sketches.frequencies.ReversePurgeItemHashMap`, `adjustOrPutValue()`, there is an assertion:
```
assert (drift < DRIFT_LIMIT) : "drift: " + drift + " >= DRIFT_LIMIT";
```
This is to assert that the number of drifts in linear probing the hash map should never exceeds 1024 times.

When sketch size is increased to 1024, this assertion gets hit when the data stream is not really skewed. Removing the assertion would not affect the correctness of the algorithm. But it indicates a potential performance issue because looking up for one key could have many linear probing steps and thus, slow.

Further analysis showed that this issue is a combination of two things mixed together:
1. Data stream has uniform distribution. In extreme case, every value is unique (count = 1);
2. The way how sketch merging is implemented in sketch library. Each value is inserted into the other map in order.

When merging goes on and the map hit load factor (size is already at maximum 8192), algorithm decreases all values count by 1 and remove all non-positive count values. Note that for the case above (count = 1 for all values), it removes all values from the map and essentially leave an empty map.

Now the rest items in source map has to be inserted into the target map. Also note that the two maps have the same hash function. So for all rest items, they tend to be hashed to a similar index range in the target map. From this point on, values in the target map is not balanced anymore. After a few merging process, target map has a very dense area full of values, and another area full of free slot. The assertion is bound to be hit at this point.

## Why upgrading to the latest sketch library solves the issue?

The problem is solved in the latest sketch library by addressing point 2 above.

In the latest implementation, when merging to sketches, items are not inserted into the other map in index order. Instead, the iterator on map has a stride added:

```
    boolean next() {
-      i++;
-      while (i < iKeys.length) {
-        if (iStates[i] > 0) { return true; }
-        i++;
+      i_ = (i_ + stride_) & mask_;
+      while (count_ < numActive_) {
+        if (states_[i_] > 0) {
+          count_++;
+          return true;
+        }
+        i_ = (i_ + stride_) & mask_;
      }
      return false;
    }
```

In this way, after all items are kicked out from target map after load factor is hit, rest items in source map are still roughly evenly distributed over index range.

## How to test

A new test `StatisticsAdminIT.testNoAssertionFailureInSketch` is added. This test should fail on platform_it before this change and pass after this change.


platform_it/pom.xml
```diff
@@ -1043,6 +1043,7 @@
                                     -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=4020
                                     -Djava.security.krb5.conf=${project.build.directory}/krb5.conf
                                     -enableassertions
+                                    -disableassertions:com.yahoo.sketches.frequencies.ReversePurgeItemHashMap
```
msirek, line 1046: Should we disable these in the mem_sql pom file too?

ascend1: Actually, I don't know if it's needed. I didn't see this problem on mem platform. When looking into `mem_sql/pom.xml`, I didn't find `-enableassertions` arg sent to jvm. Do we generally have assertions enabled on mem platform?

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2560/

ascend1:

spliceci:
TEST SUCCEEDED +1

platform_it/pom.xml
msirek: You're right, it looks like mem does not have assertions enabled.

msirek (approved):

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2570/

ascend1:

platform_it/pom.xml
```diff
@@ -1057,6 +1058,7 @@
                                     -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=4025
                                     -Dsplice.olapServerOnStandAlone=true
                                     -enableassertions
+                                    -disableassertions:org.apache.datasketches.frequencies.ReversePurgeItemHashMap
```
dgomezferro, line 1061: Do we still want this if we assume we shouldn't hit the assertion anymore?

dgomezferro (commented):
@ascend1 have you tested upgrades? Is the new version using a compatible serialization format?

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2572/

ascend1:

platform_it/pom.xml
ascend1: I had the same question. Since this assertion doesn't affect correctness, I thought it's safe to just disable it. If we encounter some performance issue, we could enable it locally and try if it gets hit. But I'm also fine with enabling it in general, no strong opinion here.

splice_machine/src/test/java/com/splicemachine/derby/utils/StatisticsAdminIT.java
```diff
@@ -1242,6 +1267,19 @@ public void testCollectStatsWithMoreSplits() throws Exception {
         methodWatcher4.execute("drop table TAB_WITH_MORE_SPLITS");
     }
 
+    @Test
+    public void testNoAssertionFailureInSketch() throws Exception {
```
arnaud-splice, line 1271: This test does not test anything anymore, since assertions are disabled, right?
We should either remove it, @Ignore it with a comment, or enable assertions I suppose

arnaud-splice (changes requested):

platform_it/pom.xml
ascend1: The assertion is enabled in the latest commit.

splice_machine/src/test/java/com/splicemachine/derby/utils/StatisticsAdminIT.java
ascend1: That's right. I enabled the assertion in the latest commit.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2581/

ascend1:

ascend1:
> @ascend1 have you tested upgrades? Is the new version using a compatible serialization format?

Yes, I tested upgrading and it worked.

ascend1:
Test builds failed with different tests every time...Please trace back to the first full cdh6.3.0 test build that is successful.

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2594/

dgomezferro (approved):

arnaud-splice (approved):

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2619/

hatyo (approved):

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2629/

PR #5251, opened 2021-03-04
DB-10258 Create parameter to disallow multidatabase creation

arnaud-splice:
## Short Description
This disables multidatabase as a default and introduces a new system procedure to enable it.

## Long Description
This introduces two new system procedures: 
```
SYSCS_UTIL.SYSCS_ENABLE_MULTIDATABASE()
SYSCS_UTIL.SYSCS_DISABLE_MULTIDATABASE()
```
Creating a new database is forbidden if multidatabase is disabled. Disabling multidatabase after the creation of a new database does not destroy that database, but only prevents the creation of new ones.
This parameter is persisted in zookeeper, because we need to be able to access during connection time.

 
## How to test
```
splice> show databases;
DATABASENAME
------------------------------------
SPLICEDB
splice> connect 'jdbc:splice://localhost:1527/splicedb3;user=foo;password=pass;create=true';
ERROR 42Y19: DERBY SQL error: SQLCODE: -1, SQLSTATE: 42Y19, SQLERRMC: Multidatabase is disabled
splice> call syscs_util.syscs_enable_multidatabase();
Statement executed.
ELAPSED TIME = 190 milliseconds
splice> connect 'jdbc:splice://localhost:1527/splicedb3;user=foo;password=pass;create=true';
ELAPSED TIME = 67 milliseconds
splice> show databases;
DATABASENAME
------------------------------------
SPLICEDB
SPLICEDB3
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
@@ -71,8 +71,8 @@ public class UnaryOperatorNode extends OperatorNode
      */
     private int operatorType;
 
-    String		resultInterfaceType;
-    String		receiverInterfaceType;
+    String        resultInterfaceType;
+    String        receiverInterfaceType;
 
     // At the time of adding XML support, it was decided that
     // we should avoid creating new OperatorNodes where possible.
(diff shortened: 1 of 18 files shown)


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
```diff
@@ -740,7 +740,7 @@ public void setOperand(ValueNode operand) {
     public void setMatchIndexExpr(int tableNumber, int columnPosition, ConglomerateDescriptor conglomDesc) {
         this.operandMatchIndexExpr = tableNumber;
         this.operandMatchIndexExprColumnPosition = columnPosition;
-        this.operandMatchIndexExprConglomDesc = conglomDesc;
+        //this.operandMatchIndexExprConglomDesc = conglomDesc;
```
martinrupp, line 743: can this be removed completely?

martinrupp (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryOperatorNode.java
arnaud-splice: Done

martinrupp (approved):

jyuanca (approved):

carolp-503 (approved):
+1

PR #5256, opened 2021-03-08
DB-11270 some refactorings in Write Pipeline

martinrupp:
## Description
Some refactorings in Write Pipeline

## How to test
No test necessary, functionality and behavior should stay the same and is covered by existing ITs / QA tests.


pipeline_api/src/main/java/com/splicemachine/pipeline/client/MergingWriteStats.java
```diff
@@ -22,19 +22,19 @@
...
-    private final Counter partialRows;
-    private final Counter partialThrownErrorRows;
-    private final Counter partialRetriedRows;
-    private final Counter partialIgnoredRows;
-    private final Counter partialWrite;
-    private final Counter ignoredRows;
-    private final Counter catchThrownRows;
-    private final Counter catchRetriedRows;
-    private final Counter regionTooBusy;
+    public final Counter writtenCounter;
```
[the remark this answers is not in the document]
martinrupp: this way it's accessible from BulkWriteAction. Otherwise we have to re-create the same variables over at BulkWriteReader again. Actually i think we only need MergingWriteStats, not WriteStats and SimpleWriteStats. It's a bit overengineered for something that is basically just a collection of counters in my opinion.

martinrupp:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2624/

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2633/

pipeline_api/src/test/java/com/splicemachine/pipeline/PipelineTest.java
```diff
@@ -50,6 +51,7 @@
  * @author Scott Fines
  *         Date: 12/23/15
  */
+@Ignore // doesn't work (No PipelineTestDataEnv found)
```
hatyo, line 54: I think you can run it by specifying either cdh or mem in the classpath of the test run configuration (if you're using IntelliJ), please make sure that the test is not running in Jenkins before ignoring it otherwise this PR decreases test coverage.

pipeline_api/src/test/java/com/splicemachine/pipeline/writeconfiguration/DefaultWriteConfigurationTest.java
```diff
@@ -35,6 +36,7 @@
  * @author David Winters
  * Created on: 5/10/15
  */
+@Ignore // doesn't work (No PipelineTestDataEnv found)
```
hatyo, line 39: ditto.

splice_machine/src/main/java/com/splicemachine/pipeline/IndexWriteHandler.java
```diff
@@ -33,6 +33,7 @@
  *         Created on: 5/1/13
  */
 public class IndexWriteHandler extends RoutingWriteHandler{
+    // todo: IndexWriteHandler is the only usage of RoutingWriteHandler. Merge into one class
```
hatyo, line 36: not sure we want to do this, there is probably a design decision behind that, we have other  few examples where an interface / abstract class is implemented by a single concrete class.

hatyo (changes requested):
Nice refactoring, please cleanup `SITransactor.java` (remove commented out code). This is a hot codepath, so it probably makes sense to do some measurements to make sure that replacing the old way of using `IntObjectHashMap` by simple `ArrayList` for caching the mutation does not e.g. cause more pressure on GC.

pipeline_api/src/test/java/com/splicemachine/pipeline/PipelineTest.java
dgomezferro, line 54: This is supposed to run using the classpath of `hbase_pipeline` or `mem_pipelin` (it's an architecture specific test).

I just noticed this tests are not running on jenkins or as part of the normal maven tests, so we need to fix that.

pipeline_api/src/test/java/com/splicemachine/pipeline/client/BulkWriteActionTest.java
```diff
@@ -61,9 +60,7 @@
 
     @Before
     public void setUp() throws Exception{
-        PipelineTestDataEnv pipelineTestDataEnv =PipelineTestEnvironment.loadTestDataEnvironment();
-
-        this.pef = pipelineTestDataEnv.pipelineExceptionFactory();
```
dgomezferro, line 66: Same here, we don't need this change, this has to run using the classpath from `hbase_pipeline`

pipeline_api/src/test/java/com/splicemachine/pipeline/writeconfiguration/DefaultWriteConfigurationTest.java
dgomezferro, line 39: Remove

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -197,7 +208,11 @@ private MutationStatus getCorrectStatus(MutationStatus status,MutationStatus old
                     SIConfigurations.DEFAULT_ACTIVE_TRANSACTION_MAX_CACHE_SIZE;
             supplier = new ActiveTxnCacheSupplier(txnSupplier, initialSize, maxSize);
         }
-        @SuppressWarnings("unchecked") final LongHashSet[] conflictingChildren=new LongHashSet[size];
+        // todo: maybe checkConflictsForKvBatch should return a ArrayList of objects with
+        // int index
+        // LongHashSet conflictingChildren
+        // DataPut data
+        //@SuppressWarnings("unchecked") final LongHashSet[] conflictingChildren=new LongHashSet[size];
```
dgomezferro, line 215: remove?

splice_machine/src/main/java/com/splicemachine/pipeline/IndexWriteHandler.java
martinrupp: i have to remove this comment. Reason is that the IndexWriteHandler is in another component where the RoutingWriteHandler hasn't access to.

dgomezferro (changes requested):
I added some minor comments/suggestions, but I'm marking it as changes requested due to the changes in pipeline_api tests, we should revert those (and make sure the tests run from maven)

pipeline_api/src/test/java/com/splicemachine/pipeline/PipelineTest.java
martinrupp: The test is not running nor working on current master. We should make them run again and then enable in jenkins/pom (DB-11686).

martinrupp: ok, i removed this commit

martinrupp: ok, i removed this commit

pipeline_api/src/test/java/com/splicemachine/pipeline/writeconfiguration/DefaultWriteConfigurationTest.java
martinrupp: removed this commit

martinrupp: removed this commit

pipeline_api/src/test/java/com/splicemachine/pipeline/client/BulkWriteActionTest.java
martinrupp: removed this commit

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
martinrupp: done

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/RoutingWriteHandler.java
```diff
@@ -135,4 +137,9 @@ else if(t instanceof WriteConflict)
             ctx.failed(mutation, WriteResult.failed(t.getClass().getSimpleName() + ":" + t.getMessage()));
         }
     }
+
+    @Override
+    public String toString() {
+        return "RoutingWriteHandler { destination = " + Arrays.toString(destination) + " keepState = " + keepState + "}";
```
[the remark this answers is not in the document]
martinrupp: done

pipeline_api/src/test/java/com/splicemachine/pipeline/PipelineExceptionFactoryForTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+import com.splicemachine.pipeline.api.PipelineExceptionFactory;
+import com.splicemachine.pipeline.client.WriteResult;
+import com.splicemachine.pipeline.constraint.ConstraintContext;
+import com.splicemachine.si.api.txn.Txn;
+
+import java.io.IOException;
+
+// somehow we can't use DirectPipelineExceptionFactory (No PipelineTestDataEnv found)
+public class PipelineExceptionFactoryForTest implements PipelineExceptionFactory {
```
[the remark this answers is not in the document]
martinrupp: removed

pipeline_api/src/main/java/com/splicemachine/pipeline/client/BulkWriteAction.java
```diff
@@ -452,28 +405,39 @@ private boolean shouldWrite(BulkWrites nextWrite){
         return true;
     }
 
-    private Exception parseIntoException(BulkWriteResult response){
-        IntObjectHashMap<WriteResult> failedRows=response.getFailedRows();
+    /** this is used to mark exceptions that are from failed rows (throwExceptionFromFailedRows)
+     * (in contast to pipeline exceptions)
```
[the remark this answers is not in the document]
martinrupp: done

db-shared/src/main/java/com/splicemachine/db/shared/common/sql/Utils.java
```diff
@@ -25,7 +28,14 @@ public static String escape(String in, char escapeCharacter) {
...
-                 .replace("%", escapeCharacter + "%");
+                .replace("_", escapeCharacter + "_")
+                .replace("%", escapeCharacter + "%");
+    }
+
+    public static <T> String listToString(List<T> list) {
+        if(list == null )
+            return "null";
+        else
+            return list.stream().map(Object::toString).collect(Collectors.joining(", "));
```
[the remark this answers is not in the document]
martinrupp: done, replaced

martinrupp:
> Nice refactoring, please cleanup `SITransactor.java` (remove commented out code). This is a hot codepath, so it probably makes sense to do some measurements to make sure that replacing the old way of using `IntObjectHashMap` by simple `ArrayList` for caching the mutation does not e.g. cause more pressure on GC.

Do you have a suggestion for a Microbenchmark to run?

Some comments why i went from IntObjectHashMap to ArrayList:
- there's only one place we use use `IntObjectHashMap.put`, and that is `finalMutationsToWrite.put(i)`. This one is always with a different `i` (since `for(int i = 0; i<dataAndLocks.length; i++) {`), so we we don't make use of the `map` characteristics when inserting
- whenever we access the data (through `IntObjectHashMap<DataPut> writes`) we iterate over all data. So we also do not make use of the `map` characteristics when accessing data.

Both together means we don't actually need a `Map<Integer, DataPut>`, but more sth like a `List<Pair<Integer, DataPut>`.
Performance wise i would say a ArrayList is much faster inserting + iterating over than a Map.

Also, old code is complicated and due to that sometimes even wrong: e.g. there's this here `finalStatus[i] = operationStatusLib.failure(e);` which is wrong because `i` is just the index in the list `conflictingChildren`, but NOT the index in `finalStatus`. That index would have been `write.value` (= the integer from `IntObjectHashMap<DataPut> writes`).

martinrupp:
wait i think then also `conflictingChildren[i]` is wrong

hatyo:
> > Nice refactoring, please cleanup `SITransactor.java` (remove commented out code). This is a hot codepath, so it probably makes sense to do some measurements to make sure that replacing the old way of using `IntObjectHashMap` by simple `ArrayList` for caching the mutation does not e.g. cause more pressure on GC.
> 
> Do you have a suggestion for a Microbenchmark to run?
> 
> Some comments why i went from IntObjectHashMap to ArrayList:
> 
>     * there's only one place we use use `IntObjectHashMap.put`, and that is `finalMutationsToWrite.put(i)`. This one is always with a different `i` (since `for(int i = 0; i<dataAndLocks.length; i++) {`), so we we don't make use of the `map` characteristics when inserting
> 
>     * whenever we access the data (through `IntObjectHashMap<DataPut> writes`) we iterate over all data. So we also do not make use of the `map` characteristics when accessing data.
> 
> 
> Both together means we don't actually need a `Map<Integer, DataPut>`, but more sth like a `List<Pair<Integer, DataPut>`.
> Performance wise i would say a ArrayList is much faster inserting + iterating over than a Map.
> 
> Also, old code is complicated and due to that sometimes even wrong: e.g. there's this here `finalStatus[i] = operationStatusLib.failure(e);` which is wrong because `i` is just the index in the list `conflictingChildren`, but NOT the index in `finalStatus`. That index would have been `write.value` (= the integer from `IntObjectHashMap<DataPut> writes`).

Thanks for the clarification, it makes sense to me choosing an `ArrayList` instead for these reasons, if you're still interested in pursuing a benchmark, I think TPCC would be a good start!

hatyo (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

ipraznik-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2736/

ipraznik-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2738/

martinrupp:
note to self: fails in 
com.splicemachine.pipeline.client.BulkWriteActionTest.testCorrectlyRetriesPartialResults
com.splicemachine.pipeline.client.BulkWriteActionTest.testCorrectlyRetriesWhenOneRegionStopsButReturnsResult
com.splicemachine.pipeline.client.BulkWriteActionTest.testCorrectlyRetriesWhenOneRegionStops

PR #5257, opened 2021-03-08
DB-11570 Use SQLDB2Varchar in SQLChar.getNewVarchar when applicable

arnaud-splice:
## Short Description
When `splice.db2.varchar.compatible` is set to true, SQLChar.{replace,strip,etc.} should return a `SQLDB2Varchar`

## Long Description
Functions in SQLChar (substring, split_part, right, left, replace, translate, ansiTrim, db2Trim) need to create a new `SQLVarchar` to store the result. If `splice.db2.varchar.compatible` is set to true, we should make sure that the result will be stored in a `SQLVarcharDB2Compatible` instead.

Comparisons done on `replace(SQLChar)` would not behave according to that parameter before that change.

## How to test
```
call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', false);
values 'abcdefgh            '=strip(replace(TRANSLATE('abcdefgh            ',' ',X'00'),' ',''));
1
-----
false

call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', true);
values 'abcdefgh            '=strip(replace(TRANSLATE('abcdefgh            ',' ',X'00'),' ',''));
1
-----
true


```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/SQLChar.java
@@ -191,6 +191,8 @@ public class SQLChar extends DataType implements StringDataValue, StreamStorable
     /* Locale info (for International support) */
     private LocaleFinder localeFinder;
 
+    private boolean ignoreTrailingWhitespacesInVarcharComparison = false;
+
 
     /**************************************************************************
      * Constructors for This class:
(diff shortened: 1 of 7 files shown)


arnaud-splice:

spliceci:
TEST SUCCEEDED +1

msirek (approved):

jpanko1 (approved):

carolp-503 (approved):
+1

PR #5258, opened 2021-03-08
verify Maven profile before starting the cluster.

hatyo:
## Short Description
verify Maven profile(s) before starting a cluster

## Long Description
It is common to mistype a Maven profile, e.g. `cdh6,3.0` instead of `cdh6.3.0`, to prevent this from happening, we add a profile verification step that checks each of the user-defined profiles against available profile and fails if a profile is not found.

### Example usage:

```
./start-splice-cluster -p skywalker
unknown profile 'skywalker', available profiles: noJacoco default ee codecoverage skipTestsLongerThan2Minutes hdp3.1.5 jacoco hdp3.1.0 splicemachine-internal dbaas3.0 it secure core mem insecure dbaas cdh6.3.0 splicemachine-external staticanalysis generate-docs
```


martinrupp (approved):

OlegMazurov (approved):
The change affects only the start_splice-cluster script and has no effect on ITs. My previous comment on comparing marginal convenience benefit for developers with not insignificant cost of extra dependencies thus has little merit.

carolp-503 (approved):
+1

PR #5265, opened 2021-03-08
DB-11476 Apply computeSqrtLevel selectivity to the correct level

msirek:
## Short Description
A cross join of two tables is picking the large table as the right table and broadcasting it, leading to out of memory errors or query execution that runs seemingly forever.

## Long Description
Broadcast costs are dominating the execution of cross join when the broadcast table is large, but the cost formula estimates that the join costs are 3 orders of magnitude larger than remote cost of doing the broadcast.  

Remote cost formula:
> remoteCost = openLatency + closeLatency + totalRowCount * totalSelectivity * remoteLatency * (1+colSizeFactor/1024d)
remoteLatency is 10.0.

Join row cost formula:
> joiningRowCost = numOfJoinedRows * localLatency;
localLatency is 1.0, but the number of joined rows is so high that joiningRowCost dominates.

For a query of the form:

> select count(*) from
>        t1 BH, t1 BU
>        WHERE (BH.b='h' or BH.c='h') and
>              (BH.b=BU.b OR BH.c=BU.b)
>          AND BU.b <> 'h';

The join condition "(BH.b=BU.b OR BH.c=BU.b)" is included part of BH when considering (BU join BH), and as part of BU when considering (BH join BU).  When calculating scan selectivity, ScanCostFunction.computeSelectivity calls shouldApplySelectivity and correctly skips the join condition, however the argument to parameter "level" is passed as a value 1 higher than when BU is the inner table because we have one extra predicate:

>         for (int i = 0; i< holders.size();i++) {
>             // Do not include join predicates unless join strategy is nested loop.
>             if (holders.get(i).shouldApplySelectivity()) {
>                 selectivity = computeSqrtLevel(selectivity, i, holders.get(i));
>             }
>         }

This affects the exponential backoff in computeSqrtLevel and BU join BH appears to have more join rows than BH join BU.  Since the joiningRowCost dominates costs, we pick the join permutation that appears to join less rows, and end up broadcasting the large table.  The long-term solution may be to adjust the cost formulas to more accurately estimate broadcast costs.  The solution to this Jira is to fix exponential backoff to ignore levels that don't have their selectivity applied:
>             if (holders.get(i).shouldApplySelectivity()) {
>                 selectivity = computeSqrtLevel(selectivity, level, holders.get(i));
>                 level++;
>             }

Also, a change is made in isTheRightJoinPredicate when picking join predicates in estimateJoinSelectivity.  It currently skips predicates of type JoinPredicateType.ALL, probably because these predicates used to be counted as part of the scan selectivity.  But since we skip them in scan selectivity estimates now, they must be counted as part of join selectivity.  Function isJoinPredicate is renamed to isHashableJoinPredicate as it is applicable to hash-based joins for calls where JoinPredicateType is MERGE_SEARCH or HASH_SEARCH.  When JoinPredicateType is ALL, isTheRightJoinPredicate returns true if more than one table is referenced in the predicate (the opposite of the shouldApplySelectivity check in scan selectivity estimation).

## How to test

> drop table t1;
> create table t1 (a int, b varchar(10), c varchar(10));
> 
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> insert into t1 values (1, 'a', 'a');
> 
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> insert into t1 select * from t1;
> analyze table t1;
> explain
> select count(*) from
>        t1 BH, t1 BU
>        WHERE (BH.b='h' or BH.c='h') and
>              (BH.b=BU.b OR BH.c=BU.b)
>          AND BU.b <> 'h';
> 
The explain should pick BH as the right table:
> Plan
> ----
> Cursor(n=9,rows=1,updateMode=READ_ONLY (1),engine=OLAP (session hint))
>   ->  ScrollInsensitive(n=9,totalCost=3436036397.562,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
>     ->  ProjectRestrict(n=8,totalCost=3435996965.562,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
>       ->  GroupBy(n=7,totalCost=3435996965.562,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
>         ->  ProjectRestrict(n=6,totalCost=3435992564.653,outputRows=2123366,outputHeapSize=12.15 MB,partitions=1,parallelTasks=1)
>           ->  CrossJoin(n=4,totalCost=3435992564.653,outputRows=2123366,outputHeapSize=12.15 MB,partitions=1,parallelTasks=1,preds=[((BH.B[4:2] = BU.B[4:1]) or ((BH.C[4:3] = BU.B[4:1]) or false))])
>             ->  TableScan[T1(2672)](n=2,totalCost=2782.726,scannedRows=2621440,outputRows=1310720,outputHeapSize=12.15 MB,partitions=1,parallelTasks=1,preds=[((BH.B[2:1] = h) or ((BH.C[2:2] = h) or false))])
>             ->  TableScan[T1(2672)](n=0,totalCost=2782.726,scannedRows=2621440,outputRows=2621440,outputHeapSize=5 MB,partitions=1,parallelTasks=1,preds=[(BU.B[0:1] <> h)])

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
@@ -666,10 +666,13 @@ public class ScanCostFunction{
      * @throws StandardException
      */
     public static double computeSelectivity(double selectivity, List<SelectivityHolder> holders) throws StandardException {
+        int level = 0;
         for (int i = 0; i< holders.size();i++) {
             // Do not include join predicates unless join strategy is nested loop.
-            if (holders.get(i).shouldApplySelectivity())
-                selectivity = computeSqrtLevel(selectivity,i,holders.get(i));
+            if (holders.get(i).shouldApplySelectivity()) {
+                selectivity = computeSqrtLevel(selectivity, level, holders.get(i));
+                level++;
+            }
         }
         return selectivity;
     }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
@@ -84,7 +84,10 @@ public class SelectivityUtil {
 
 
     private static boolean isTheRightJoinPredicate(Predicate p, JoinPredicateType predicateType) {
-        if (p == null || !p.isJoinPredicate())
+        if (p == null || predicateType != JoinPredicateType.ALL && !p.isHashableJoinPredicate())
+            return false;
+
+        if (p.referencedSet.cardinality() < 2)
             return false;
 
         // only equality join conditions can be used for hashable joins to search for matching rows
@@ -298,7 +301,7 @@ public class SelectivityUtil {
     // Look for equality predicate that is not a join predicate
     private static boolean existsNonJoinPredicate(List<Predicate> predList) {
         for (Predicate predicate : predList) {
-            if (!predicate.isJoinPredicate() && !predicate.isFullJoinPredicate()) {
+            if (!predicate.isHashableJoinPredicate() && !predicate.isFullJoinPredicate()) {
                 return true;
             }
         }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/ast/PredicateUtils.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/ast/PredicateUtils.java
@@ -47,7 +47,7 @@ public class PredicateUtils {
         @Override
         public boolean apply(Predicate p) {
             return p != null &&
-                    p.isJoinPredicate() &&
+                    p.isHashableJoinPredicate() &&
                     p.getAndNode().getLeftOperand().isBinaryEqualsOperatorNode();
         }
     };
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DefaultPredicateSelectivity.java
@@ -32,6 +32,7 @@
 package com.splicemachine.db.impl.sql.compile;
 
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.sql.compile.AccessPath;
 import com.splicemachine.db.iapi.sql.compile.Optimizable;
 import com.splicemachine.db.impl.ast.RSUtils;
 
@@ -63,8 +64,15 @@ public class DefaultPredicateSelectivity extends AbstractSelectivityHolder {
         return
             getNumReferencedTables() < 2             ||
             baseTable.getCurrentAccessPath() == null ||
-            RSUtils.isNLJ(baseTable.getCurrentAccessPath());
+            canPushJoinPredAsInnerTablePred(baseTable.getCurrentAccessPath());
     }
+
+    private boolean canPushJoinPredAsInnerTablePred(AccessPath accessPath) {
+        if (accessPath == null)
+            return false;
+        return accessPath.getJoinStrategy().getJoinStrategyType().isAllowsJoinPredicatePushdown();
+    }
+
 
     public Predicate getPredicate() { return p; }
 }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
@@ -981,7 +981,7 @@ public class FromBaseTable extends FromTable {
                     if(baseTableRestrictionList.isRedundantPredicate(i)) continue;
                 }
 
-                if(!p.isJoinPredicate()&& !p.isFullJoinPredicate() || currentJoinStrategy.allowsJoinPredicatePushdown()) //skip join predicates unless they support predicate pushdown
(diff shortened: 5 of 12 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2577/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ScanCostFunction.java
```diff
@@ -666,10 +666,13 @@ private static void collectSelectivityHolders(List<SelectivityHolder>[] from, Li
...
     public static double computeSelectivity(double selectivity, List<SelectivityHolder> holders) throws StandardException {
+        int level = 0;
         for (int i = 0; i< holders.size();i++) {
             // Do not include join predicates unless join strategy is nested loop.
-            if (holders.get(i).shouldApplySelectivity())
-                selectivity = computeSqrtLevel(selectivity,i,holders.get(i));
+            if (holders.get(i).shouldApplySelectivity()) {
+                selectivity = computeSqrtLevel(selectivity, level, holders.get(i));
+                level++;
+            }
```
ascend1, line 675: Great find! This part should be so despite of other changes.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
```diff
@@ -84,7 +84,10 @@ else if (outerCost.isAntiJoin())
 
 
     private static boolean isTheRightJoinPredicate(Predicate p, JoinPredicateType predicateType) {
-        if (p == null || !p.isJoinPredicate())
+        if (p == null || predicateType != JoinPredicateType.ALL && !p.isHashableJoinPredicate())
+            return false;
+
+        if (p.referencedSet.cardinality() < 2)
```
ascend1, line 90: Is it possible that a predicate has `predicateType == JoinPredicateType.ALL` but its `referencedSet.cardinality() < 2` ? I mean, if it references only one table or even no table, probably it shouldn't be `JoinPredicateType.ALL` at the first place. I'd like to understand in which situation we have such cases.

ascend1 (approved):
Great change!

I read the long description and have a question. Why join condition `(BH.b=BU.b OR BH.c=BU.b)` is included as part of `FromBaseTable.baseTableRestrictionList` when considering (BU join BH), but not when considering (BH join BU)? It's a symmetric join, and it's a join predicate. Also, I thought it can only be pushed to inner table if case of NLJ. From the description, it looks like there is something to be fixed here.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectivityUtil.java
msirek: JoinPredicateType.ALL isn't an attribute of the predicate, but the type of predicates we're considering.  This rule is already built into Predicate.isHashableJoinPredicate:

        return (leftUsableAsJoinColumn &&
                rightUsableAsJoinColumn &&
                leftTableNumber!=rightTableNumber) &&        <<-----------
                (!checkScope  || !isScopedForPush());

So, now we are considering more than hashable predicates, but we still need to enforce the rule that we are only looking for predicates referencing more than one table.

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:
> Great change!
> 
> I read the long description and have a question. Why join condition `(BH.b=BU.b OR BH.c=BU.b)` is included as part of `FromBaseTable.baseTableRestrictionList` when considering (BU join BH), but not when considering (BH join BU)? It's a symmetric join, and it's a join predicate. Also, I thought it can only be pushed to inner table if case of NLJ. From the description, it looks like there is something to be fixed.

@ascend1 This part of the long description is incorrect.  baseTableRestrictionList always includes the join predicate, and it is included due to the following code in FromBaseTable:

>         else {
>             int predListSize = predList!=null?baseTableRestrictionList.size():0;
>             for(int i=0;i<predListSize;i++){
>                 Predicate p = (Predicate)baseTableRestrictionList.getOptPredicate(i);
>                 if(!p.isStartKey()&&!p.isStopKey()){
>                     if(baseTableRestrictionList.isRedundantPredicate(i)) continue;
>                 }
> 
>                 if(!p.isHashableJoinPredicate()&& !p.isFullJoinPredicate() || currentJoinStrategy.allowsJoinPredicatePushdown()) //skip join predicates unless they support predicate pushdown
>                     scf.addPredicate(p, defaultSelectivityFactor);   <<<-------------------
>             }

it's not a hashable join predicate so it is always pushed.  The difference in estimates comes because when it's the inner table it picks up the predicate, but does not when it's the outer table.

jyuanca (approved):

carolp-503 (approved):
+1

PR #5267, opened 2021-03-09
DB-11445 SI table scan refactoring (WIP)

hatyo:
## Short Description

## Long Description

## How to test


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/scanner/SITableScanner.java
```diff
@@ -328,7 +325,7 @@ public ReturnCode filterCell(DataCell keyValue) throws IOException {
                              * but we have to transactionally resolve anyway, for instance if the field we are actually interested
                              * in is on the rowkey.
                              */
-                            if (keyValue.dataType() == CellType.USER_DATA  && (!accumulator.isInteresting(keyValue) || accumulator.isFinished()) && keyIncluded){
+                            if (keyValue.dataType() == CellType.USER_DATA  && keyIncluded && (accumulator.isFinished() || !accumulator.isInteresting(keyValue))){
```
hatyo, line 328: @dgomezferro why are we returning `INCLUDE` if the `keyValue` is deemed uninteresting by the accumulator?

hatyo:
Refactoring is causing issues, will redo it.

PR #5269, opened 2021-03-09
DB-11133: persist DDL demarcation point to zookeeper(3.1)

jyuanca:
## Short Description

## Long Description

## How to test


carolp-503 (approved):
Approved for 3.1

carolp-503 (approved):
Approved for 3.1

PR #5270, opened 2021-03-09
DB-11133: persist DDL demarcation point to zookeeper

jyuanca:
## Short Description
This PR fixes an issue only happens if a region server restarts and lost most recent DDL demarcation point. The workload initiated(from a different region server) before  most recent DDL demarcation point may try to use data dictionary cache and read data that's not supposed to be visible to it.

## Long Description

For every DDL, we set a DDL demarcation point(transaction id) in DDL refresher. Any txns started earlier than this demarcation point is forbidden from reading/writing data dictionary cache, because the DDL may have changed data dictionary. The data dictionary cache can only contain data that's visible after demarcation point. However, demarcation point is not persisted. If a region server restarts, it loses the most recent DDL transaction id.

We have found in a rare case, a truncated, insert and region server restart can cause the write workload to be written to old base table and new index, thus causing a data inconsistency.

In the following case, a table is hosted by RS2. A insert was executed after the table is truncated from a different region server.

     
1  (RS1)insert into table 
2  (RS1)truncate a table with index                   (RS2) Restart

3   (RS1) keep retrying insert

4                                                                     (RS2)Restarted

5                                                                     (RS2)another workload queries the table. Table descriptor is read into  cache 

6                                                                     (RS2)write pipeline receives insert workload, queries cache, and sees new descriptor

At timeline 6, no demarcation point for truncate DDL was lost, so write pipeline uses data dictionary cache to find out indexes of the table. However, the descriptor in cache is for the table/index after truncate. In normal case without a restart, write pipeline can check demarcation point to find itself started earlier than most recent DDL, and decide to read data dictionary table(not cache).

This PR addresses this issue by persisting the most recent DDL transaction ID in ZooKeeper. If a region server restarts, it fetches the transaction ID from zookeeper.

## How to test
Extremely hard to reproduce manually. Please run Dmitry's test suite on srv056 repetitively to confirm this is fixed. Turn on debug trace for ZooKeeperDDLWatchChecker to confirm DDL demarcation point is set in zookeeper and retrieved every time a region server restarts.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
+++ b/hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
@@ -18,6 +18,10 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
 
+import com.splicemachine.access.HConfiguration;
+import com.splicemachine.primitives.Bytes;
+import com.splicemachine.si.impl.driver.SIDriver;
+import com.splicemachine.utils.SpliceLogUtils;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import splice.com.google.common.collect.Lists;
 import org.apache.log4j.Logger;
@@ -67,11 +71,32 @@ public class ZooKeeperDDLWatchChecker implements DDLWatchChecker{
                 }
             }
         };
-
-
+        initZKDemarcationPoint();
         return true;
     }
 
+    private void initZKDemarcationPoint() throws IOException{
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String ddlPath = rootPath+HConfiguration.DDL_PATH;
+        List<String> children = ZkUtils.getChildren(ddlPath, null);
+        long max = 0;
+        for (String child : children) {
+            String demarcationPath = ddlPath + "/" + child;
+            byte[] data = ZkUtils.getData(demarcationPath);
+            long demarcationPoint = data != null ? Bytes.toLong(data) : 0;
+            if (demarcationPoint > max) {
+                max = demarcationPoint;
+            }
+        }
+        String demarcationPointPath = ddlPath+"/"+id;
+        try {
+            ZkUtils.recursiveSafeCreate(demarcationPointPath, Bytes.toBytes(max), ZooDefs.Ids.OPEN_ACL_UNSAFE,
+                    CreateMode.EPHEMERAL);
+            SpliceLogUtils.info(LOG, "Initialize DDL demarcation point to %d", max);
+        } catch (Exception e) {
+            throw new IOException(e);
+        }
+    }
     @Override
     public Collection<String> getCurrentChangeIds() throws IOException{
         try{
@@ -159,4 +184,27 @@ public class ZooKeeperDDLWatchChecker implements DDLWatchChecker{
        zkClient.deleteChangeNode(key);
     }
 
+
+    @Override
+    public void assignDDLDemarcationPoint(long txnId) throws Exception {
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String demarcationPath = rootPath+HConfiguration.DDL_PATH+"/"+id;
+        byte[] data = Bytes.toBytes(txnId);
+        ZkUtils.setData(demarcationPath, data, -1);
+        if (LOG.isDebugEnabled()) {
+            SpliceLogUtils.debug(LOG, "Set DDL demarcation point to %d", txnId);
+        }
+
+    }
+
+    @Override
+    public long initDemarcationPoint() throws IOException{
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String demarcationPointPath = rootPath+HConfiguration.DDL_PATH+"/"+id;
+        byte[] data = ZkUtils.getData(demarcationPointPath);
+        long demarcationPoint = data != null ? Bytes.toLong(data) : 0;
+        SpliceLogUtils.info(LOG, "Get DDL demarcation point %d from zookeeper", demarcationPoint);
+
+        return demarcationPoint;
+    }
 }
--- a/splice_machine/src/main/java/com/splicemachine/derby/ddl/AsynchronousDDLWatcher.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/ddl/AsynchronousDDLWatcher.java
@@ -76,6 +76,7 @@ public class AsynchronousDDLWatcher implements DDLWatcher,CommunicationListener{
         if(!checker.initialize(this))
             return; //we aren't a server, so do nothing further
 
+        refresher.initDemarcationPoint();
         try {
             // run refresh() synchronously the first time
             if (!refresher.refreshDDL(ddlListeners)) return;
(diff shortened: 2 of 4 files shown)


jyuanca:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
```diff
@@ -159,4 +180,37 @@ public void killDDLTransaction(String key){
...
+    }
+
+    @Override
+    public long getDemarcationPoint() throws IOException{
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String ddlPath = rootPath+HConfiguration.DDL_PATH;
+        byte[] data = ZkUtils.getData(ddlPath);
+        long demarcationPoint = data != null ? Bytes.toLong(data) : 0;
+        if (LOG.isDebugEnabled()) {
+            SpliceLogUtils.debug(LOG, "Get DDL demarcation point %d from zookeeper", demarcationPoint);
```
dgomezferro, line 212: This only happens during startup right? Please increase the logging level to INFO

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
```diff
@@ -46,6 +55,18 @@
...
+    static {
+        String ensemble = ZKConfig.getZKQuorumServersString(HConfiguration.unwrapDelegate());
+        CuratorFramework client = CuratorFrameworkFactory.newClient(ensemble, new ExponentialBackoffRetry(1000,3));
+        client.start();
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String ddlPath = rootPath+HConfiguration.DDL_PATH;
+        lock = new InterProcessMutex(client, ddlPath);
+    }
+
+
```
dgomezferro, line 69: I think this synchronization mechanism is too heavyweight. Could each RS persist it in its own ephemeral ZNode and when starting up we take the maximum demarcation point?

dgomezferro (changes requested):
See above for requested changes

jyuanca:

spliceci:
TEST SUCCEEDED +1

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
```diff
@@ -159,4 +189,30 @@ public void killDDLTransaction(String key){
...
     }
 
+
+    @Override
+    public void assignDDLDemarcationPoint(long txnId) throws Exception {
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String demarcationPath = rootPath+HConfiguration.DDL_PATH+"/"+id;
+        byte[] data = ZkUtils.getData(demarcationPath);
+        long demarcationPoint = data != null ? Bytes.toLong(data) : 0;
+        if (txnId > demarcationPoint) {
```
dgomezferro, line 199: I don't think we need this check, we already made sure it's the most recent one in `DDLWatchRefresher.assignDDLDemarcationPoint` We can write blindly into ZK I'd say

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
```diff
@@ -159,4 +189,30 @@ public void killDDLTransaction(String key){
...
+    @Override
+    public long getDemarcationPoint() throws IOException{
+        String rootPath = SIDriver.driver().getConfiguration().getSpliceRootPath();
+        String demarcationPointPath = rootPath+HConfiguration.DDL_PATH+"/"+id;
+        byte[] data = ZkUtils.getData(demarcationPointPath);
+        long demarcationPoint = data != null ? Bytes.toLong(data) : 0;
+        SpliceLogUtils.info(LOG, "Get DDL demarcation point %d from zookeeper", demarcationPoint);
+
+        return demarcationPoint;
+    }
```
dgomezferro, line 217: I'd just return the locally stored version from the init method

splice_machine/src/main/java/com/splicemachine/derby/ddl/AsynchronousDDLWatcher.java
```diff
@@ -76,6 +76,7 @@ public void start() throws IOException {
         if(!checker.initialize(this))
             return; //we aren't a server, so do nothing further
 
+        refresher.getDemarcationPoint();
```
dgomezferro, line 79: I think we can remove this

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
```diff
@@ -67,11 +76,32 @@ public void process(WatchedEvent watchedEvent){
...
+        }
+        String demarcationPointPath = ddlPath+"/"+id;
+        try {
+            ZkUtils.recursiveSafeCreate(demarcationPointPath, Bytes.toBytes(max), ZooDefs.Ids.OPEN_ACL_UNSAFE,
+                    CreateMode.EPHEMERAL);
+            SpliceLogUtils.info(LOG, "Initialize DDL demarcation point to %d", max);
+        } catch (Exception e) {
+            throw new IOException(e);
+        }
+    }
```
dgomezferro, line 104: Rather than storing it back into ZK, we can store it locally and return it in `getDemarcationPoint()`

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLWatchRefresher.java
```diff
@@ -72,6 +71,19 @@ public DDLWatchRefresher(DDLWatchChecker watchChecker,
         ddlDemarcationPoint = new AtomicReference<>();
     }
 
+    public void getDemarcationPoint() {
```
dgomezferro, line 74: This should be called `initDemarcationPoint` I think

dgomezferro (approved):

dgomezferro (changes requested):
I meant to request changes, see above for the request

splice_machine/src/main/java/com/splicemachine/derby/ddl/DDLWatchRefresher.java
jyuanca: Right, it is renamed

splice_machine/src/main/java/com/splicemachine/derby/ddl/AsynchronousDDLWatcher.java
jyuanca: This is needed to initialize a local copy of demarcation point in DDLWatchRefresher

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
jyuanca: This is invoked during initialization, when there is no local copy. I renamed it to initDemarcationPoint

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
jyuanca: Right, this check is duplicate. Removed.

hbase_sql/src/main/java/com/splicemachine/ddl/ZooKeeperDDLWatchChecker.java
dgomezferro: Ok, I think that makes sense.

dgomezferro (approved):

PR #5277, opened 2021-03-11
DB-11641 Only disable NLJ on Spark if it enables no index access

msirek:
## Short Description
DB-11006 disables some good performing nested loop joins on Spark, which should be enabled.

## Long Description
Modify the fix of DB-11006 to only impose a high cost penalty for nested loop join on Spark if it does not have any join predicates that get pushed to the inner table which enable index access with both a start key and a stop key.  A new session property is added to allow fallback to the old behavior of always applying a cost penalty to nested loop join on Spark:

> set session_property olapAlwaysPenalizeNLJ=true;

## How to test
Create a nested loop join case with a join predicate on the first index column of the inner table, and test that the optimizer picks the join if running on spark.


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2605/

msirek:

spliceci:
TEST SUCCEEDED +1

dgomezferro (commented):
@msirek I see DB-11521 is not fixed yet, I understand this could lead to us choosing NLJ even though we are dealing with multi-million rows join, isn't that a risk?

msirek:
> @msirek I see DB-11521 is not fixed yet, I understand this could lead to us choosing NLJ even though we are dealing with multi-million rows join, isn't that a risk?

@dgomezferro Yes, but it is also a risk to disable good nested loop joins and get other similarly long-running multi-million row joins that now can't use an index.  This Jira is providing a new session property to control which behavior to take depending on the exposure of the customer.  We could perhaps choose to merge this to master only to give more time to fix cardinality underestimate issues (not only DB-11521).  Or, we could make the default in 3.1 be olapAlwaysPenalizeNLJ=true until those issues are fixed.

arnaud-splice (approved):

martinrupp:
> How to test
> Create a nested loop join case with a join predicate on the first index column of the inner table, and test that the optimizer  > picks the join if running on spark.

If this is not too complicated or lengthy, can you maybe provide a concrete example of such a join? (SQL). 
I guess we would have to run a EXPLAIN then, right?

dgomezferro:
@msirek I was mostly worried about large scans on the left of the NLJ, or a large complex query on the right hand side. I see this PR makes sure #2 doesn't happen and after playing with it for a while I'm mostly convinced #1 shouldn't be a problem either because having a large amount of rows on the left should penalize NLJ enough to use something else if possible. 
I'd rather only merge this one to master though, just to be sure, if everyone is OK with it.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
```diff
@@ -319,14 +321,17 @@ private double getNljOnSparkPenalty(Optimizable table,
...
-//            return NLJ_ON_SPARK_PENALTY * multiplier;
-//        if (hasJoinPredicateWithIndexKeyLookup(predList) && outerCost.rowCount() <= 10)
-//            return retval;
+        QueryTreeNode queryTreeNode = (QueryTreeNode) table;
+        LanguageConnectionContext lcc = queryTreeNode.getLanguageConnectionContext();
+        Boolean olapAlwaysPenalizeNLJ = (Boolean)
+            lcc.getSessionProperties().getProperty(SessionProperties.PROPERTYNAME.OLAPALWAYSPENALIZENLJ);
+
+        if (olapAlwaysPenalizeNLJ == null || !olapAlwaysPenalizeNLJ.booleanValue()) {
+            if (!isBaseTable(table))
```
dgomezferro, line 330: @msirek Should we allow IndexLookups here too? We could also have ProjectRestricts before the Scan.

msirek: @dgomezferro IndexLookups are built automatically.  A FromBaseTable node still describes the access.  I have added support for ProjectRestricts on top of the scan.

msirek:
> @msirek I was mostly worried about large scans on the left of the NLJ, or a large complex query on the right hand side. I see this PR makes sure #2 doesn't happen and after playing with it for a while I'm mostly convinced #1 shouldn't be a problem either because having a large amount of rows on the left should penalize NLJ enough to use something else if possible.
> I'd rather only merge this one to master though, just to be sure, if everyone is OK with it.

@dgomezferro OK, though we may eventually want it in 3.1.  I know Feature Store uses nested loop joins on OLAP and they will be using 3.1.

msirek:
@dgomezferro I've removed the ProjectRestrict change as discussed

dgomezferro (approved):

jyuanca (approved):

carolp-503 (approved):
+1

PR #5278, opened 2021-03-11
DB-11639 Native spark cross join for DB2 mode

msirek:
## Short Description
Implements native spark cross join for DB2 varchar compatibility mode.

## Long Description
When cross join cannot run natively in spark it has a broadcast join implementation to fill the gap.  This implementation does not honor the flag indicating the right table should not be cached.  Previously, when splice.db2.varchar.compatible is set to true, a cross join selected by the optimizer does not qualify for native spark, so if the join planner ever mistakenly picks the large table as the right table of cross join we can hit an out of memory condition if the table is too large to be broadcast.

Fix:
Implement native spark cross join for DB2 mode by adding trimmed join columns so trailing spaces are ignored.  Also add IS NOT NULL conditions on the left and right join keys.  The remaining case that can't use native spark cross join is equijoin on columns of the REAL data type.  Possibly this case could be handled in native spark in the future as well, and the non-native spark path can be eliminated entirely. 

## How to test
Create different tables with varchar columns and join the together with the following property enabled:
> call syscs_util.syscs_set_global_database_property('splice.db2.varchar.compatible', true);

Check for incorrect results.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
@@ -926,6 +926,11 @@ public class NativeSparkDataSet<V> implements DataSet<V> {
     @Override
     public DataSet<V> crossJoin(OperationContext context, DataSet<V> rightDataSet, Broadcast type) {
         try {
+            String varcharDB2CompatibilityModeString =
+                PropertyUtil.getCachedDatabaseProperty(context.getActivation().getLanguageConnectionContext(),
+                                                       Property.SPLICE_DB2_VARCHAR_COMPATIBLE);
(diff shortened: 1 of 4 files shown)


msirek:

dgomezferro (approved):
This is great @msirek !

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2608/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2615/

msirek:

arnaud-splice (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5284, opened 2021-03-12
DB-11052 Correct calulation of MaxExecutorCores on cloud clusters

msirek:
## Short Description
Corrects the calculation of maxExecutorCores on cloud clusters.

## Long Description
On Kubernetes, rmClient is null in OlapServerMaster, so we cannot find the number of nodes from that source.  Also, there is no YARN, so the max memory allocation: yarn.scheduler.maximum-allocation-mb is not set.  
The fix is to derive the number of "nodes" from the number of spark executor instances because cloud uses one instance per pod.  Also, use spark.kubernetes.executor.request.memory as the memory size as Kubernetes tries to honor this value.

## How to test
Create a large table on a cloud cluster and select from it using the hint --splice-properties splits=24
or a higher number of splits.
The explain should show a parallelTasks value of 24.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/olap/OlapServerMaster.java
+++ b/hbase_sql/src/main/java/com/splicemachine/olap/OlapServerMaster.java
@@ -417,18 +417,30 @@ public class OlapServerMaster {
 
         try {
             Configuration yarnConf = HConfiguration.unwrapDelegate();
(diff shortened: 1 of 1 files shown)


msirek:

spliceci:
TEST SUCCEEDED +1

hatyo (approved):

dgomezferro (approved):

ascend1 (approved):

carolp-503 (approved):
+1

PR #5286, opened 2021-03-12
DB-11636: fix avg aggregator regression

jyuanca:
## Short Description
A regression due to serialization changes to avg aggregator

## Long Description
The SumAggregator included in a AvgAggregator can be one of LongBufferedSumAggregator, DecimalBufferedSumAggregator, or DoubleBufferedSumAggregator. When it is deserialized, we should find out its correct data type and construct the right aggregator.

## How to test
This is only reproducible if a grouped aggregation is executed a normal spark processor (not native spark). To disable native spark, set the following database property

```
call syscs_util.syscs_set_global_database_property('splice.execution.nativeSparkAggregationMode', 'off');
create table AG (a int, b real, c bigint, d decimal(15,2));
insert into AG values(1, 1, 2, 3);
insert into AG values(2, 1, 2, 3);
insert into AG values(3, 1, 2, 3);

insert into AG select a, b+1, c+1, d+1 from AG
insert into AG select a, b+1, c+1, d+1 from AG
insert into AG select a, b+1, c+1, d+1 from AG
insert into AG select a, b+1, c+1, d+1 from AG
```

This query should not fail with NPE
```
select a, avg(b), avg(c), avg(d) from ag --splice-properties useSpark=true
group by a order by a;
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
@@ -56,17 +56,21 @@ public class DecimalBufferedSumAggregator extends SumAggregator {
 
 	private BigDecimal sum = BigDecimal.ZERO;
 	private boolean isNull = true;
+	private int bufferSize;
 
 	public DecimalBufferedSumAggregator() { // SERDE
 
 	}
 
 	public DecimalBufferedSumAggregator(CatalogMessage.SystemAggregator agg) throws IOException, ClassNotFoundException {
-		this(64);
 		init(agg);
 	}
 
 	public DecimalBufferedSumAggregator(int bufferSize) {
+		init(bufferSize);
+	}
+
+	private void init(int bufferSize) {
 		int s = 1;
 		while(s<bufferSize){
 			s<<=1;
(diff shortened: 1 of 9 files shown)


msirek (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
```diff
@@ -61,6 +61,11 @@ public DecimalBufferedSumAggregator() { // SERDE
 
 	}
 
+	public DecimalBufferedSumAggregator(CatalogMessage.SystemAggregator agg) throws IOException, ClassNotFoundException {
+		this(64);
```
martinrupp, line 65: why 64?

martinrupp (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
jyuanca: I'm copying the size from else where. Not sure why it was decided to be 64

carolp-503 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/DecimalBufferedSumAggregator.java
carolp-503: Jun, can you research this one offline?  No need to hold the submit, but would be nice to have an explanation for the 64.

dgomezferro: Please make this part of the serialized message

dgomezferro (changes requested):

dgomezferro (approved):

PR #5292, opened 2021-03-13
DB-11622 Fix cost accumulation and NLJ cost function

ascend1:
## Short Description
This change fixes performance regression of the query reported in DB-11622.

## Long Description
Before this change, we have two issues in cost estimation:
1. Cost accumulation is generally done as per parallel task except for the first table access, from which total cost is accumulated.
2. NLJ cost function always multiply outer rows by inner total cost. Essentially, this means we believe inner table access can be perfectly parallelized in case of broadcast join, sort merge join, and merge join, but not NLJ and cross join. When inner table access cost is high and number of parallelism is also high, this difference makes NLJ lose immediately.

In this change, 1 is fixed by accumulating per task cost for the first table access. 2 is improved by making NLJ uses inner table per task cost so that its cost is estimated under the same assumption as other join strategies.

## How to test
On 1TB data set, DB-11622 query should finish within 2 s after the fix. Without the fix and on optimizer branch, the query doesn't finish in 7 hours.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
@@ -967,6 +967,18 @@ public class OptimizerImpl implements Optimizer{
 
                 CostEstimate addend = getAddendCost(bestCostEstimateSortAvoidance, currentCost);
                 addCost(addend, currentSortAvoidanceCost);
+                /* DB-11622 note
+                 * For join position 0, bestCostEstimate is just the scan cost of
+                 * optimizable 0 because there is no join yet. For a scan cost,
+                 * localCost = localCostPerParallelTask * parallelism. However,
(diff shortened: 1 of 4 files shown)


jyuanca:
Is it possible to add an IT to prevent regression?

msirek (approved):
Looks good.

carolp-503 (approved):
+1 once we have 2 other reviews

arnaud-splice (approved):

arnaud-splice:
> Is it possible to add an IT to prevent regression?

@ascend1 can you add an IT in a follow-up commit?

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2647/

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #5302, opened 2021-03-15
Redo of DB-11558 Update sketch library to Apache DataSketches 2.0.0

arnaud-splice:
[Original PR](https://github.com/splicemachine/spliceengine/pull/5249)

## Short Description

This change solves the assertion failure on drift >= 1024 in sketch library so that analyzing a table after sketch size is increased to 8192 returns successfully.

## Long Description

In `com.yahoo.sketches.frequencies.ReversePurgeItemHashMap`, `adjustOrPutValue()`, there is an assertion:

```
assert (drift < DRIFT_LIMIT) : "drift: " + drift + " >= DRIFT_LIMIT";
```

This is to assert that the number of drifts in linear probing the hash map should never exceeds 1024 times.

When sketch size is increased to 1024, this assertion gets hit when the data stream is not really skewed. Removing the assertion would not affect the correctness of the algorithm. But it indicates a potential performance issue because looking up for one key could have many linear probing steps and thus, slow.

Further analysis showed that this issue is a combination of two things mixed together:

1. Data stream has uniform distribution. In extreme case, every value is unique (count = 1);
2. The way how sketch merging is implemented in sketch library. Each value is inserted into the other map in order.

When merging goes on and the map hit load factor (size is already at maximum 8192), algorithm decreases all values count by 1 and remove all non-positive count values. Note that for the case above (count = 1 for all values), it removes all values from the map and essentially leave an empty map.

Now the rest items in source map has to be inserted into the target map. Also note that the two maps have the same hash function. So for all rest items, they tend to be hashed to a similar index range in the target map. From this point on, values in the target map is not balanced anymore. After a few merging process, target map has a very dense area full of values, and another area full of free slot. The assertion is bound to be hit at this point.

## Why upgrading to the latest sketch library solves the issue?

The problem is solved in the latest sketch library by addressing point 2 above.

In the latest implementation, when merging to sketches, items are not inserted into the other map in index order. Instead, the iterator on map has a stride added:

```
    boolean next() {
-      i++;
-      while (i < iKeys.length) {
-        if (iStates[i] > 0) { return true; }
-        i++;
+      i_ = (i_ + stride_) & mask_;
+      while (count_ < numActive_) {
+        if (states_[i_] > 0) {
+          count_++;
+          return true;
+        }
+        i_ = (i_ + stride_) & mask_;
      }
      return false;
    }
```

In this way, after all items are kicked out from target map after load factor is hit, rest items in source map are still roughly evenly distributed over index range.

## How to test

A new test StatisticsAdminIT.testNoAssertionFailureInSketch is added. This test should fail on platform_it before this change and pass after this change.


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2639/

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2640/

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2645/

ascend1:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

hatyo (approved):

PR #5303, opened 2021-03-16
DB-11625 Support both operands of concat to be parameters

arnaud-splice:
## Short Description

Assume operands types of `? || ?` to be VARCHAR

## Long Description

Two things are introduced by this PR:
1. We assume both operands of the concatenation operator to be VARCHAR if they are both unkwnon
2. We cast LONG VARCHAR back to VARCHAR when trying to compare LONG VARCHAR to another string type

The second point is necessary because the result of `? || ?` is of type LONG VARCHAR

## How to test

```
prepare foo as 'select ? || ?';
execute foo using 'values (''abc'', ''def'')';
1
-
abcdef

prepare foo as 'select ''abcdef'' = ? || ?';
execute foo using 'values (''abc'', ''def'')';
1
-
true
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ConcatenationOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ConcatenationOperatorNode.java
@@ -138,148 +138,60 @@ public class ConcatenationOperatorNode extends BinaryOperatorNode {
          * the other parameter, with maximum length for that type.
          */
 
-        if (getLeftOperand().requiresTypeFromContext()) {
-            if (getRightOperand().requiresTypeFromContext()) {
-                throw StandardException.newException(
-                        SQLState.LANG_BINARY_OPERANDS_BOTH_PARMS, operator);
-            }
-
-            TypeId leftType;
-
-            /*
-             * * A ? on the left gets its type from the right. There are eight *
-             * legal types for the concatenation operator: CHAR, VARCHAR, * LONG
-             * VARCHAR, CLOB, BIT, BIT VARYING, LONG BIT VARYING, and BLOB. * If
-             * the right type is BLOB, set the parameter type to BLOB with max
-             * length. * If the right type is one of the other bit types, set
-             * the parameter type to * BIT VARYING with maximum length. * * If
-             * the right type is CLOB, set parameter type to CLOB with max
-             * length. * If the right type is anything else, set it to VARCHAR
-             * with * maximum length. We count on the resolveConcatOperation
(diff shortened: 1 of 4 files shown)


arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ConcatenationOperatorNode.java
```diff
@@ -138,156 +138,67 @@ public ValueNode bindExpression(FromList fromList,
...
-             * length. * If the right type is anything else, set it to VARCHAR
-             * with * maximum length. We count on the resolveConcatOperation
-             * method to * catch an illegal type. * * NOTE: When I added the
-             * long types, I could have changed the * resulting parameter types
-             * to LONG VARCHAR and LONG BIT VARYING, * but they were already
-             * VARCHAR and BIT VARYING, and it wasn't * clear to me what effect
-             * it would have to change it. - Jeff
-             */
-            if (getRightOperand().getTypeId().isBitTypeId()) {
-                if (getRightOperand().getTypeId().isBlobTypeId())
```
martinrupp, line 166: how can this all be removed?

splice_machine/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/PreparedStatementIT.java
```diff
@@ -625,4 +625,31 @@ public void testCursorUntypedExpressionType() throws Exception {
             methodWatcher.execute("call syscs_util.syscs_set_global_database_property('splice.bind.cursorUntypedExpressionType', null)");
         }
     }
+
+    @Test
+    public void testConcatUnknownParameters() throws Exception {
+        testParamsHelper("select ? || ?",
+                new Object[] {"abc", "def"},
```
ipraznik-splice, line 632: In IT there are tests only for strings. Does it make sense to add other types or the type casting is covered enough by other tests?

ipraznik-splice (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ConcatenationOperatorNode.java
arnaud-splice: This code is duplicated for left and right, i removed it all and refactored it below

spliceci:
TEST SUCCEEDED +1

martinrupp (approved):

carolp-503 (approved):
+1

PR #5309, opened 2021-03-16
DB-11687 Improve NE+Range estimates

ascend1:
## Short Description
This change boost the performance of query with hash `SGgiYFr5`.

## Long Description
Three issues:
* Index XPFVEZUASERLZP is not picked up probably because high number of estimated output rows. Currently working on this.
* Join between VE and VZ is not NLJ probably due to the same issue above.
* We don’t have bitmap IXAND.

To improve the first two points, this change improves the following:
* A not equal predicate and a range predicate are combined if the not equal value equals to an inclusive boundary value of the range predicate.

## How to test
`SGgiYFr5` should have plan on OLTP after this change. It's execution time should below 30 seconds on a k8s cluster. Before this change, the plan is on OLAP and takes at least 350 seconds.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NotEqualsSelectivity.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/NotEqualsSelectivity.java
@@ -41,7 +41,7 @@ import com.splicemachine.db.iapi.types.DataValueDescriptor;
  *
  */
 public class NotEqualsSelectivity extends AbstractSelectivityHolder {
-    private final DataValueDescriptor value;
+    public final DataValueDescriptor value;
     private final StoreCostController storeCost;
     private final double selectivityFactor;
     private final boolean useExtrapolation;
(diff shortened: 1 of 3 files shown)


ascend1:

msirek (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
```diff
@@ -367,53 +367,15 @@ private boolean hasJoinPredicateWithIndexKeyLookup(OptimizablePredicateList pred
      */
 
     public static double nestedLoopJoinStrategyLocalCost(CostEstimate innerCost, CostEstimate outerCost,
-                                                         double numOfJoinedRows, boolean useSparkCostFormula) {
+                                                         double numOfJoinedRows) {
```
dgomezferro, line 370: I think this is dangerous. I'm not too happy about the different formulas for control and Spark, but the current Spark penalty only applies in some cases, so we could end up choosing NLJ in Spark where it would be very costly.

dgomezferro (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
ascend1: Right. I thought about it yesterday again for some time and I think it makes sense to revert back to two formulae. One reason is to avoid regression on OLAP, the other reason is that we probably do have special treatments for OLTP later.

ascend1: Reverted and rebased.

ascend1:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

dgomezferro (approved):

PR #5310, opened 2021-03-16
DB-11675 Fix NPE in IndexLookup: SPLICE-2399 regression

msirek:
## Short Description
During index lookup the wrong base table field is used to locate the column side of the predicate and no matching column is found, leading to NPE.

## Long Description
In `PredicateList.orderUsefulPredicates`, when an index access is identified as not using the first index column AccessPath.numUnusedLeadingIndexFields is set to 1.  The logic should only kick in when examining the first column position of an index, but it sometimes gets invoked after the first index column is already found:

>                     if (indexPrefixIterationAllowed &&
>                         (!thisPred.isHashableJoinPredicate() || considerJoinPredicateAsKey) &&
>                          (thisIndexPosition-currentStartPosition) == numColsInStartPred + 1) {
>                         accessPath.setNumUnusedLeadingIndexFields(1);
>                     }

currentStartPosition is zero at this point (it starts out as -1, and is set to the first index column position:
> currentStartPosition=thisIndexPosition;

The condition:
> (thisIndexPosition-currentStartPosition) == numColsInStartPred + 1) 

is not sufficient to limit the logic to only kick in on the first index column.

Since currentStartPosition is initialized to firstColumnIdx - 1 by this line:
> currentStartPosition += firstColumnIdx;

The first is to add a test for that condition:
>                     if (indexPrefixIterationAllowed &&
>                         !rowIdScan                  &&
>                         currentStartPosition == (firstColumnIdx - 1) &&   <<<-----------
>                         (!thisPred.isHashableJoinPredicate() || considerJoinPredicateAsKey) &&
>                          (thisIndexPosition-currentStartPosition) == numColsInStartPred + 1) {
>                         accessPath.setNumUnusedLeadingIndexFields(1);
>                     }

A new session property is added to force selection of indexPrefixIteration mode, if legal, for a given index, to allow more test coverage of SPLICE-2399:
> set session_property favorIndexPrefixIteration=true;

Statistics are required for SPLICE-2399 to be active, but for testing purposes, the following session property can bypass this requirement:
> set session_property alwaysAllowIndexPrefixIteration=true;

## How to test
The following test case should not hit a NullPointerException:
```
CREATE TABLE Table1 (
"COL1" CHAR(36) NOT NULL
,"COL2" CHAR(5) NOT NULL DEFAULT ''
,"COL3" TIMESTAMP NOT NULL
,"COL4" CHAR(1) NOT NULL DEFAULT ''
,"COL5" CHAR(1) NOT NULL DEFAULT ''
,"COL6" TIMESTAMP NOT NULL
,"COL7" CHAR(8) NOT NULL DEFAULT ''
,"COL8" CHAR(36) NOT NULL
,"COL9" CHAR(36) NOT NULL
,"COL10" VARCHAR(3560) NOT NULL DEFAULT ''
, CONSTRAINT MY_PK PRIMARY KEY("COL1")
);
create index Table1_Idx1 on Table1 (COL2, COL5, COL3, COL8, COL9);

analyze table Table1;

set session_property favorIndexPrefixIteration=true;

explain SELECT DP.COL1,
       DP.COL3,
       DP.COL9,
       DP.COL8,
--       GF.sprache,
       DP.COL10
FROM --splice-properties joinOrder=fixed
       Table1 DP --splice-properties index=Table1_Idx1
WHERE
       DP.COL2 = 'ABCDE'
   AND DP.COL5 ^= 'J'
   AND DP.COL3 < '2011-11-01'
                    || '-00.00.00.000000'
ORDER BY DP.COL1;
```

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/IndexPrefixIterationIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/IndexPrefixIterationIT.java
@@ -302,5 +302,34 @@ public class IndexPrefixIterationIT  extends SpliceUnitTest {
 
     }
 
+    @Test
+    public void testIndexLookupNoNPE() throws Exception {
+        if (isMemPlatform && useSpark)
+            return;
+        String expected =
+        "COL1 |           COL3            |COL9 |COL8 | COL10 |\n" +
+        "------------------------------------------------------\n" +
+        "  a  |2010-12-31 15:59:59.321111 |  f  |  e  |   g   |";
 
+        String query = format("SELECT DP.COL1,\n" +
+                              "       DP.COL3,\n" +
+                              "       DP.COL9,\n" +
+                              "       DP.COL8,\n" +
+                              "       DP.COL10\n" +
+                              "FROM --splice-properties joinOrder=fixed\n" +
+                              "       Table1 DP --splice-properties index=Table1_Idx1\n" +
+                              "WHERE\n" +
+                              "       DP.COL2 = 'ABCDE'\n" +
+                              "   AND DP.COL5 ^= 'J'\n" +
+                              "   AND DP.COL3 < '2011-11-01'\n" +
+                              "                    || '-00.00.00.000000'\n" +
+                              "ORDER BY DP.COL1", useSpark);
+
+        methodWatcher.execute("set session_property favorIndexPrefixIteration=true");
+        List<String> containedStrings = Arrays.asList("IndexLookup");
+        List<String> notContainedStrings = Arrays.asList("IndexPrefixIteratorMode");
+        testQuery(query, expected, methodWatcher);
+        testExplainContains(query, methodWatcher, containedStrings, notContainedStrings);
+        methodWatcher.execute("set session_property favorIndexPrefixIteration=false");
+    }
 }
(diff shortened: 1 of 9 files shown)


msirek:

arnaud-splice (approved):

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/IndexPrefixIterationIT.java
```diff
@@ -302,5 +302,34 @@ public void testParameterized() throws Exception {
 
     }
 
-
+    @Test
+    public void testIndexLookupNoNPE() throws Exception {
```
ascend1, line 306: Does this test case test both start key and stop key cases? I mean, would this test pass if we don't fix the stop key part?

ascend1 (approved):

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

splice_machine/src/test/java/com/splicemachine/db/impl/sql/compile/IndexPrefixIterationIT.java
msirek: @ascend1 Yes.  With the stop key portion of the fix commented out, the test will fail, so both portions are needed.

PR #5313, opened 2021-03-17
DB-11693 Skip purging in memstore flushes if region is closing 

arnaud-splice:
## Short description

We should not try to purge any data when flushing the memstore if the region being flushed is closing

## Long description

When flushing the memstore to disk, we call `getTxnLowWatermark` to understand what can be purged. 
If a region that is necessary for `getTxnLowWatermark` is closing, we will hang until that region is back online. However, if that region is the very one we try to flush, this will result in a deadlock.

We now make sure that a region isn't closing before trying to call `getTxnLowWatermark`

## How to test

An easy way to reproduce in standalone is:

```
create table a(i int); -- add some data to SYSTABLES
CALL SYSCS_UTIL.INVALIDATE_GLOBAL_DICTIONARY_CACHE(); -- invalidate caches so we have to go to HBase to retrieve TableDescriptor
```

then from hbase shell

```
disable 'splice:48' # SYSTABLE conglomerate, trigger flush as part of closing the region
```


arnaud-splice:

dgomezferro (approved):

carolp-503 (approved):
+1

jyuanca (approved):

spliceci:
TEST SUCCEEDED +1

PR #5319, opened 2021-03-17
DB-11626 Authorize '?' in with clauses

arnaud-splice:
## Short Description

We do not support `?` in `create view` but we want to support it in with clauses

## Long Description

This adds support for `?` in with clauses. We currently implement with clauses via the `CREATE_VIEW_NODE`, with `isDynamic` set to `true`. This moves the parameter check from parsing phase to binding phase, so that we can allow such parameters for with clauses.

## How to test

```
create table a (i int);
insert into a values 1,2,3,4,5;
create table b (j int);
insert into b values 1,2,3,4,5;
with cte_a as (select i from a where i = 3) select * from b where j = (select i from cte_a);
J          
-----------
3          

prepare t1 as 'with cte_a as (select i from a where i = ?) select * from b where j = (select i from cte_a)';
execute t1 using 'values 3';
J          
-----------
3          

```


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2656/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

dgomezferro (approved):
@arnaud-splice do we have a negative test for the view case?

jyuanca (approved):

PR #5329, opened 2021-03-19
DB-11666 Timeout of testStatementTriggersRuntimeUnderLimit: 10s -> 15s

msirek:
## Short Description
Increase statement trigger IT timeout

## Long Description
Increase the timeout of testStatementTriggersRuntimeUnderLimit from 10 seconds to 15 seconds.  ITs run much slower on colo clusters, and testStatementTriggersRuntimeUnderLimit runs well under 10 seconds on a laptop often takes 11 or 12 seconds on a colo system.

## How to test
Run testStatementTriggersRuntimeUnderLimit on a colo system and verify it passes.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Performance_IT.java
+++ b/splice_machine/src/test/java/com/splicemachine/triggers/Trigger_Performance_IT.java
@@ -203,7 +203,7 @@ public class Trigger_Performance_IT extends SpliceUnitTest {
 
         long endTime = System.currentTimeMillis();
         long runTime = endTime - startTime;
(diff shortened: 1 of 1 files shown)


ascend1:

ascend1 (approved):
It's pretty hard to assert on execution time especially when the test is going to run on cloud. We may want to move these tests out of IT suite and make a separate perf test suite. We can always increase this upper bound but at some point, it may loses meaning to have such a test.

arnaud-splice (approved):

spliceci:
TEST SUCCEEDED +1

PR #5336, opened 2021-03-20
DB-11677:fix JDBC timeout(3.0)

jyuanca:
## Short Description
JDBC timeout is not respected by some queries.

## Long Description
JDBC query timeout is not respected by a query with many stages, because the timeout applies for a stage, not for the whole query. One example is select * query that runs on spark. If each spark task can complete within timeout, the query will succeed, even though it runs much longer than JDBC query timeout.

## How to test
Write a JDBC program and set query timeout to a small number. Run select * SQL on a big table. If the query runtime is expected to longer than timeout, it should fail.


carolp-503 (approved):
Approved for 3.0

PR #5338, opened 2021-03-22
DB-10772 DB2-compatible getColumnDisplaySize() for decimals

ascend1:
## Short Description
This change makes JDBC `getColumnDisplayWidth` method returns the same value as DB2 for decimal and numeric type columns.

## Long Description
Comparison of return values of `getColumnDisplayWidth` between splice and DB2:
|  | splice | splice example | DB2 | DB2 example |
|------|-------|-------|-------|-------|
| precision > scale > 0 | precision + 2 | -11.12 | precision + 2 | -11.12 |
| precision == scale | precision + 3 | -0.13 | precision + 2 | -.13 |
| scale == 0 | precision + 1 | -12 | precision + 2 | -12. |

Examples above show why `precision + 2` is always sufficient in DB2.

This change introduces a new JDBC connection attribute `jdbcDb2CompatibleMode=[true|false]` to set the JDBC `getColumnDisplayWidth` method.

Note that the string representation is purely a client thing. `getString()` of `SQLDecimal` is not modified. Decimal values are returned to client as they are. The difference can only be observed if
* client connects to database with a connection string that contains `jdbcDb2CompatibleMode=true`, and
* client calls `ResultSet.getString(int column)` on a decimal or numeric column.

The client can of course retrieve the original decimal values by calling `ResultSet.getBigDecimal(int column)` and perform its own string conversion.

## How to test
Setup:
```
create table test (dc1 decimal(4,2), dc2 decimal(2,0), dc3 decimal(2,2));
insert into test values (11.11, -12, -0.13);
```
On Java application side, do the following:
```
String dbUrl = "jdbc:splice://localhost:[port]/[dbName];user=[username];password=[password];jdbcDb2CompatibleMode=true";
try(Connection conn = DriverManager.getConnection(dbUrl)){
            try(Statement statement = conn.createStatement()){
                try (ResultSet rs = statement.executeQuery("select * from test")) {
                    for (int i = 1; i <= rs.getMetaData().getColumnCount(); i++) {
                        int colDispSize = rs.getMetaData().getColumnDisplaySize(i);
                        System.out.println(colDispSize);
                    }
                }
            }
```
The code snippet above should print 
```
6
4
4
```

Before this change, it should print
```
6
3
5
```
A new test `testJdbcDb2CompatibleMode` is added to JdbcApiIT.java.

### DIFF
--- a/db-client/src/main/java/com/splicemachine/db/client/am/ClientConnection.java
+++ b/db-client/src/main/java/com/splicemachine/db/client/am/ClientConnection.java
@@ -74,6 +74,7 @@ public abstract class ClientConnection
     protected final String user_;
     public boolean retrieveMessageText_;
     protected boolean jdbcReadOnly_;
+    protected boolean jdbcDb2CompatibleMode_ = false;
     /**
      * Holdabilty for created statements.
      * Only access through the holdability method
@@ -334,6 +335,8 @@ public abstract class ClientConnection
 
         principal = ClientDataSource.getClientPrincipal(properties);
         keytab =  ClientDataSource.getClientKeytab(properties);
+
+        jdbcDb2CompatibleMode_ = ClientDataSource.getJdbcDb2CompatibleMode(properties);
 
         agent_ = newAgent_(logWriter,
                 loginTimeout_,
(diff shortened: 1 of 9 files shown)


ascend1:

spliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

msirek (approved):

PR #5341, opened 2021-03-23
DB-11637: abort broadcast join cache loading if spark task has been killed

jyuanca:
## Short Description
If a broadcast join task is killed when it is loading the inner table to a cache, it does not terminate immediately. 

## Long Description
If a broadcast join task is killed when it is loading the inner table to a cache, it does not terminate immediately. Broadcast join does not check whether the task has been killed until it has completed cache loading. If the inner table is big, it will waste time and resources executing cache loading.

To fix the issue, the inner table rows are transformed to an InterruptibleIterator. When each row is read from the iterator, a task context is checked to find out whether the task has been killed. If so, an exception will be raised to terminate cache loading.

## How to test

Run TPCH1 query without using native spark dataset. Cancel the job during the 1st stage. The task should be terminated immediately. You will also find an exception raised from ValueRowMapTableLoader

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinCache.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinCache.java
@@ -15,6 +15,7 @@
 package com.splicemachine.derby.impl.sql.execute.operations;
 
 import com.splicemachine.db.iapi.sql.conn.ResubmitDistributedException;
+import org.apache.spark.TaskContext;
 import splice.com.google.common.cache.Cache;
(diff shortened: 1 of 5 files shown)


jyuanca:

spliceci:
TEST SUCCEEDED +1

ascend1 (approved):
Is it possible to have an IT for this? I think we have a TPCHIT. Maybe borrow the query from there and kill operation?

arnaud-splice (approved):

jyuanca:
@ascend1 It is hard to verify whether broadcast join aborts during cache loading. This needs to be tested manually

carolp-503 (approved):
+1

PR #5342, opened 2021-03-23
DB-11024 Remove XXX comments to prepare multidb merge

arnaud-splice:
## Short description

Remove or address multidb todo's 

## Long description

Since the creation of the multi database epic branch (DB-10193), I've added `// XXX(multidb)` comments throughout the code for lines to potentially revisit before we want to merge DB-10193 back to main.
With this PR, I revisited some of them and did one of the following:
- address it
- remove it because it no longer applies
- file a separate jira for it

Additionally, I removed some outdated upgrade code that we inherited from derby


db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
```diff
@@ -1594,74 +1594,6 @@ public static void SYSCS_UPDATE_METADATA_STORED_STATEMENTS()
...
-     * Create a system stored procedure.
-     * PLEASE NOTE:
-     * This method is currently not used, but will be used when Splice Machine has a SYS_DEBUG schema available
-     * with tools to debug and repair databases and data dictionaries.
-     *
-     * @param schemaName name of the system schema
-     * @param procName   name of the system stored procedure
-     * @throws SQLException
-     */
-    public static void SYSCS_CREATE_SYSTEM_PROCEDURE(String schemaName,String procName)
```
hatyo, line 1607: I understand this is not used at the moment, but looking at the comment, it looks like we would need it once we have `SYS_DEBUG` schema available, is this still relevant or not?

hatyo (approved):
LGTM, regarding authentication / ranger I would consult with @ipraznik-splice

db-engine/src/main/java/com/splicemachine/db/catalog/SystemProcedures.java
arnaud-splice: I don't think there are plans to bring sys_debug at the moment.

ipraznik-splice (approved):

PR #5360, opened 2021-03-31
DB-11737 Add microbenchmarks for index lookup and join cost

ascend1:
## Short Description
This change adds two microbenchmarks for cost estimation tuning.

## Long Description
This change adds the following microbenchmarks:
- index lookup performance w.r.t. number of rows on OLTP/OLAP
- index scan performance w.r.t. number of rows on OLTP/OLAP
- joining on key column performance w.r.t.
    - left table size
    - right table size
    - join strategy
    - OLTP / OLAP
- joining on non-key column performance with the same set of parameters as above

## How to test
The microbenchmarks do not run by default anywhere. To run any of them against a standalone cluster, simply do the same as running an IT the way you like. To run any of them against a cluster on cloud, do the following:

1. Prepare the necessary jar files as a tarball.
```
$ mvn -Pcore,cdh6.3.0 clean install -DskipTests
$ mkdir benchmark
$ cp ./platform_it/src/main/resources/info-log4j.properties /Users/<your username>/.m2/repository/junit/junit/4.12/junit-4.12.jar /Users/<your username>/.m2/repository/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar ./splice_machine/target/splice_machine-3.2.0.2007-SNAPSHOT-tests.jar ./platform_it/target/dependency/log4j-1.2.17.jar ./platform_it/target/dependency/db-client-3.2.0.2007-SNAPSHOT.jar benchmark
$ tar czvf benchmark.tar.gz ./benchmark
$ 
```
2. Upload the tarball to your cluster. You need to be authorized to the k8s environment before doing this step.
```
$ kubectl cp benchmark.tar.gz splicedb-hregion-0:/home/hbase/ -n <your cluster namespace>
```
3. Log into the `splicedb-hregion-0` pod of your cluster:
```
$ kubectl exec -it splicedb-hregion-0 -n <your cluster namespace> -- bash
```
4. Once in pod, run the benchmark:
```
$ cd home/hbase
$ tar zxvf benchmark.tar.gz
$ java -Dlog4j.configuration=file:./benchmark/info-log4j.properties -cp 'benchmark/*' org.junit.runner.JUnitCore com.splicemachine.benchmark.<benchmark name>
```

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/test/Benchmark.java
+++ b/splice_machine/src/test/java/com/splicemachine/test/Benchmark.java
@@ -16,17 +16,19 @@
 package com.splicemachine.test;
 
 import com.splicemachine.derby.test.framework.SpliceNetConnection;
+import com.splicemachine.derby.test.framework.SpliceWatcher;
+import com.splicemachine.derby.test.framework.TestConnection;
 import org.apache.log4j.Logger;
+import org.junit.Assert;
 
-import java.sql.Connection;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
+import java.sql.*;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicLongArray;
+
+import static org.junit.Assert.fail;
 
 public class Benchmark {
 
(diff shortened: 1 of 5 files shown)


splice_machine/src/test/java/com/splicemachine/test/Benchmark.java
```diff
@@ -146,4 +148,28 @@ public static void runBenchmark(int numThreads, Runnable runnable) {
         }
         resetStats();
     }
+
+    public static void rowContainsQuery(int[] levels, String query, Connection conn, String[]... contains) throws Exception {
```
hatyo, line 152: I find the method name and signature somehow confusing, is it possible to change it to something like this:

```

/***
 * Executes `query` using `conn` and verifies `matchLines` against the query's result set.
 * 
 * @param matchLines a map of the expected String occurance(s) in each 
 * line of the result set, e.g. { {1,['abc','def']}, {10, 'ghi'}} expects 'abc'
 * and 'def' top appear at line 1 and 'ghi' at line 10.
 */
public static void executeQueryAndMatchLines(Connection conn,
                                             String query,
                                             Map<Integer,String[]> matchLines);
```

dgomezferro (approved):

hatyo (changes requested):

splice_machine/src/test/java/com/splicemachine/test/Benchmark.java
ascend1: Sure, I'm fine with either way. The method was copied from our unit test framework. But we can improve its interface for benchmarks.

hatyo (approved):

carolp-503 (approved):
+1

ipraznik-splice:

spliceci:
TEST SUCCEEDED +1

PR #5361, opened 2021-03-31
DB-11701 Update to set false mirror state zookeeper properties

bklo94:
## Short Description
Update documentation for cdh and for hdp. Also update the hdp scripts to add the new properties

## Long Description
Add these two settings which should allow for faster restarts for a large cluster that is running on HBase 2.0+

## How to test
N/A


jyuanca (approved):

arnaud-splice (approved):

dgomezferro (changes requested):
Marking as changes requested to discuss how to address this issue, even though I'm +1 on merging:
`hbase.migrate.table.state.from.zookeeper` should be True when migrating from HBase 1.X to 2.X, so our upgrade procedure should make sure that's the case. After that first migration we should set it to false. Where should we make sure of that?

carolp-503 (approved):
+1

PR #5363, opened 2021-03-31
DB-11701 Update to set false mirror state zookeeper properties (3.0)

bklo94:
Short Description
Update documentation for cdh and for hdp. Also update the hdp scripts to add the new properties

Long Description
Add these two settings which should allow for faster restarts for a large cluster that is running on HBase 2.0+

How to test
N/A


carolp-503 (approved):
Approved for 3.0

PR #5369, opened 2021-04-02
DB-11809 Implement Feature Store benchmark

OlegMazurov:

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreBenchmark.java
+++ b/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreBenchmark.java
@@ -1,0 +1,451 @@
+package com.splicemachine.benchmark;
+
+import com.splicemachine.derby.test.framework.SpliceNetConnection;
+import com.splicemachine.derby.test.framework.SpliceSchemaWatcher;
+import com.splicemachine.test.Benchmark;
+import org.apache.log4j.Logger;
+import org.junit.*;
+import org.junit.experimental.categories.Category;
+
+import java.sql.*;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Category(Benchmark.class)
+public class FeatureStoreBenchmark extends Benchmark {
+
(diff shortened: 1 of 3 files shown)


ascend1 (changes requested):
The name of this benchmark is too general. Is it possible to make it more specific? Or if we plan to make it a whole suite for feature store later, maybe we should document it in source at least regarding the following points:

- which queries are benchmarked (and potentially why)
- benchmark setup
- what numbers we are interested in

hatyo (approved):

OlegMazurov:
@ascend1 This micro-benchmark was developed according to the spec provided outside GitHub/JIRA. Indeed, there are plans to extend it to become a general Feature Store Benchmark.
I have added comments addressing your points above.

splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreBenchmark.java
```diff
@@ -0,0 +1,477 @@
...
+     *    In a table with 10 columns, select columns c2, c4, c6, c8, c10, with 100 columns: c20, c40, c60, c80, c100 and with 1000: c200, c400, c600, c800, c1000.
+     *
+     *    SELECT <tableName>.<columnName>, ...
+     *    FROM <tableName>, ...
+     *    WHERE (<tableName>.<primaryKey>, ...) = (<random value>, ...)
+     *
+     *    Prepared queries are run from multiple concurrent connections for the specified number of operations per connection.
+     *    Query results are read and hashed for possible validation (not implemented).
+     *    Query execution times are collected to compute statistics provided by the general framework.
+     */
```
martinrupp, line 289: good information. we might want to have this at the top of the file?

OlegMazurov: Although it's part of the benchmark spec, provided in a separate document from which this description was borrowed, this comment really explains what _runQueries()_ does. It's similar in that regard to the _createTables()_ comment above, which explains how tables are created. I thought separating comments from what they explain would decrease clarity.

ascend1 (approved):
Awesome, these comments make it much easier to understand what's going on in this benchmark. Thanks!

PR #5376, opened 2021-04-07
DB-11738 Fix feasible() of HashableJoinStrategy

ascend1:
## Short Description
This change fixes an issue that some irrelevant equijoin predicates could make optimizer believes that a hashable join strategy (broadcast, merge sort, merge) is feasible but later before code generation, no hash key columns can actually be found for such a join.

### Problem description
The problem can be described using the following query:
```
select D.c1, D.c2, G.c2 
from D, G
where G.c2 = D.c2
  and D.ts <= '2000-01-01'
  and not (exists(select * from D_LZ, L
                  where D_LZ.c1 = D.c1
                    and L.c2 = D.c2
                    and L.e = 'X'));
```
Let's focus on the subquery. It has three predicates, but none of them is for the join between `D_LZ` and `L`. Also note that these predicates belong to the subquery at the very beginning. They are not pushed down from outer block. Also, note that the subquery cannot be flattened and `L` can only be joined with `D_LZ`.

Eventually, we want to estimate the cost of joining table `L` with predicates `L.c2 = D.c2` and `L.e = 'X'`. For hashable join strategies, we will call `feasible()` in `HashableJoinStrategy`, which does the following:
```
...
/* Look for equijoins in the predicate list */
hashKeyColumns = findHashKeyColumns(innerTable, cd, predList);
```

`predList` contains both predicates related to `L`. At this point, `findHashKeyColumns()` believes `L.c2 = D.c2` is a useful hash join predicate because it's an equijoin predicate and one side is a column in `L`, although the join is actually between `L` and `D_LZ`. In the end, the logic will see that we can find hash key columns on `L`, so a hashable join strategy like broadcast join or merge sort join is feasible.

Suppose a hashable join strategy for `L` is chosen as the best and we proceed. Later in modifying access paths, we want to find the hash key columns of `L` again for code generation. But this time, we do the following:
```
...
originalRestrictionList.transferPredicates(storeRestrictionList, innerTable.getReferencedTableMap(), innerTable, joinedTableSet);
...
originalRestrictionList.copyPredicatesToOtherList(nonStoreRestrictionList);
...
int[] hashKeyColumns = findHashKeyColumns(hashTableFor, cd, nonStoreRestrictionList);
```

The extra step `transferPredicates()` figures out that `L.c2 = D.c2` is not relevant to the join between `D_LZ` and `L`, and it removes it from `originalRestrictionList`. In the end, `nonStoreRestrictionList` is empty, and we cannot find any hash key columns for a table that was chosen to have a hashable join.

### Solution

The idea is to enhance the logic in `feasible()` so that irrelevant predicates are also excluded, just as in modifying access path.

`OptimizerImpl` has an `assignedTableMap` field tracking the set of table numbers that have been placed in current join order up to the current join position. This is exactly what we need to check if a predicate is relevant for a join in an inner query block.

There is one problem of using this field though. In our optimizing process, each query block has its own optimizer instance, which is exactly what we need. However, there are also cases in which we construct a new optimizer instance for a node that is not necessarily another query block. These cases happen in `TableOperatorNode` and `JoinNode`.

`TableOperatorNode` cases (union, intersect, except) are fine because we don't join their children. For `JoinNode`, we must manually include left child reference map when optimizing the right child using a new optimizer instance. Otherwise, there is no way to tell if a predicate referencing a table number that is not tracked by the current optimizer instance is relevant to the current join node or not.

## How to test
A new IT `testInfeasibleHashableJoin` is added. Setup tables as the way this IT does and run the query without this fix should fail with error message like the following:
```
ERROR 42Y63: Hash join requires an optimizable equijoin predicate on a column in the selected index or heap.  An optimizable equijoin predicate does not exist on any column in table or index ...
```
With this change, the query should run through.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
@@ -35,7 +35,6 @@ import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.services.cache.ClassSize;
 import com.splicemachine.db.iapi.services.compiler.MethodBuilder;
-import com.splicemachine.db.iapi.services.context.ContextManager;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.sql.compile.*;
 import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
@@ -44,7 +43,6 @@ import com.splicemachine.db.iapi.sql.dictionary.IndexRowGenerator;
 import com.splicemachine.db.iapi.store.access.TransactionController;
 import com.splicemachine.db.iapi.util.JBitSet;
 
-import javax.naming.Context;
 import java.util.Vector;
 
 /**
@@ -137,26 +135,8 @@ public abstract class HashableJoinStrategy extends BaseJoinStrategy {
             cd = innerTable.getCurrentAccessPath().getConglomerateDescriptor();
         }
 
-        /* Exclude predicates that are not useful in probing hash table built from right table.
-         * Make sure the same logic applies here and in divideUpPredicateLists(). Otherwise,
-         * this routine could return a hash-based join strategy is feasible but later in
-         * modifying access paths, no hash columns is found.
-         */
-        if (predList != null) {
-            JBitSet joinedTablesInCurrentQueryBlock = optimizer.getAssignedTableMap();
-            PredicateList predListCopy = new PredicateList();
-            PredicateList nljPreds = new PredicateList();
-            ContextManager cm = ((QueryTreeNode) innerTable).getContextManager();
-            predListCopy.setContextManager(cm);
-            nljPreds.setContextManager(cm);
-            predList.copyPredicatesToOtherList(predListCopy);
-            predListCopy.transferPredicates(nljPreds, innerTable.getReferencedTableMap(), innerTable, joinedTablesInCurrentQueryBlock);
-
-            /* Look for equijoins in the predicate list */
-            hashKeyColumns = findHashKeyColumns(innerTable, cd, predListCopy);
-        } else {
-            hashKeyColumns = null;
-        }
+        /* Look for equijoins in the predicate list */
+        hashKeyColumns = findHashKeyColumns(innerTable, cd, predList, optimizer.getAssignedTableMap());
 
         if (SanityManager.DEBUG) {
             if (hashKeyColumns == null) {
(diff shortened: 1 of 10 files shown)


ascend1:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
```diff
@@ -135,8 +137,26 @@ public boolean feasible(Optimizable innerTable,
...
+         */
+        if (predList != null) {
+            JBitSet joinedTablesInCurrentQueryBlock = optimizer.getAssignedTableMap();
+            PredicateList predListCopy = new PredicateList();
+            PredicateList nljPreds = new PredicateList();
+            ContextManager cm = ((QueryTreeNode) innerTable).getContextManager();
+            predListCopy.setContextManager(cm);
+            nljPreds.setContextManager(cm);
+            predList.copyPredicatesToOtherList(predListCopy);
+            predListCopy.transferPredicates(nljPreds, innerTable.getReferencedTableMap(), innerTable, joinedTablesInCurrentQueryBlock);
```
msirek, line 153: It may cause extra memory and processing overhead to build new predicate lists every time the optimizer considers a new hashable join strategy.  It would be better to just pass joinedTableSet into findHashKeyColumns and hasOptimizableEquijoin.  Then we can write a new function that both transferPredicates and hasOptimizableEquijoin can use to test a predicate:

> JBitSet getMaskedReferenceSet(Predicate predicate, JBitSet joinedTableSet) {
>     JBitSet maskedReferencedSet = (JBitSet)predicate.getReferencedSet().clone();
>     maskedReferencedSet.and(joinedTableSet);
>     return maskedReferencedSet
> }

In hasOptimizableEquijoinHelper, add:

>             // skip NLJ predicates
>             JBitSet maskedReferencedSet = getMaskedReferenceSet(predicate, joinedTableSet);
>             if (optTable.getReferencedTableMap().contains(maskedReferencedSet)) {
>                 continue;
>             }

The test: `if (optTable.getReferencedTableMap().contains(maskedReferencedSet))`
could also be refactored into a function to be shared between hasOptimizableEquijoinHelper and transferPredicates so that we won't ever modify the logic in one location without modifying the logic in the other location.  If we could find out a way to avoid the call to predicate.getReferencedSet().clone() we could avoid memory to an even greater extent, but I don't know if this is possible.

msirek (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/HashableJoinStrategy.java
ascend1: OK, I refactored the code a bit to remove the need of constructing new predicate lists.

ascend1:

spliceci:
TEST SUCCEEDED +1

msirek (approved):

hatyo (approved):

PR #5377, opened 2021-04-07
DB-11795 Improve index lookup cost estimation

ascend1:
## Short Description
This change improves index lookup cost estimation process.

## Long Description
In investigating query hash `SMopSK5W`, three problems are identified (for detailed explanation, see DB-11718). This PR addresses the second issue.

Currently, our index lookup cost has a penalty of 4000 per row. There are two issues about this. Firstly, the penalty is too high compared to actual execution times. Secondly, it's an over simplification because index lookup happens in a batched manner and could have some parallelism. Batch size and the maximum number of batches that can run concurrently are controlled by settings in `SQLConfiguration`. This change tunes the index lookup cost to consider these parameters and lower lookup penalty.

Concrete coefficients and separate formulae for OLTP and OLAP are based on the index lookup benchmark results obtained from DB-11737.

## How to test
After this change, index lookup cost should be proportional to index lookup execution time with roughly a fixed coefficient. Specifically, on OLTP we should see index lookup costs also have the "flat" phase when there are still capacity for running index lookup batches concurrently, and suddenly bumped up when one more serial stage needed. On OLAP, we should see that index lookup cost is roughly linear but with a high startup cost, making looking up small amount of rows less attractive.

To test index lookup costs really share the same trend as the execution times obtained from running the microbenchmark, expected costs are extracted from the microbenchmark results and we assert the estimated costs are close enough. Currently, without tuning anything else (scan costs, parallelism, etc.), we see that estimated costs fall into 11% range of expected costs on OLTP, and 14% range on OLAP for all test cases.

See ITs in `IndexLookupCostEstimationIT` for details.


db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerFactoryImpl.java
```diff
@@ -52,12 +52,13 @@
...
 	implements ModuleControl, OptimizerFactory {
 
-	protected String optimizerId = null;
 	protected boolean ruleBasedOptimization = false;
 	protected boolean noTimeout = false;
 	protected boolean useStatistics = true;
 	protected int maxMemoryPerTable = Integer.MAX_VALUE;//may need to be long, but this also can be defined in configuration
 	protected long determineSparkRowThreshold; //The default value is defined in SQLConfiguration
+	protected int indexBatchSize;              //The default value is defined in SQLConfiguration
+	protected int indexLookupBlocks;           //The default value is defined in SQLConfiguration
```
martinrupp, line 61: as far as i understand, this can be configured via `splice.index.batchSize` (default 4000) and `splice.index.numConcurrentLookups` (default 5), and these can't be configured by `SYSCS_SET_GLOBAL_DATABASE_PROPERTY`, but by site.xml ?
Do we have to change these settings or is this using the settings better?

martinrupp (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerFactoryImpl.java
ascend1: The idea is to consider these two configuration values in cost estimation. They decide how much parallelism we can have in executing index lookup, but we don’t estimate the cost accordingly. The goal of this change is to make index lookup costs reflect the same trend as real execution times.

ascend1:

spliceci:
TEST SUCCEEDED +1

msirek (approved):
The changes look good.  I am wondering why OLAP_INDEXLOOKUP_COST_PER_ROW is so much smaller than the OLTP cost per row, and if the number of executor cores in the cluster affects this number ( for example, if each parallel spark task does its lookups in parallel with other spark tasks, maybe this is increasing the throughput and we can model this better instead of using a contant).

dgomezferro (approved):

ascend1:
> The changes look good. I am wondering why OLAP_INDEXLOOKUP_COST_PER_ROW is so much smaller than the OLTP cost per row, and if the number of executor cores in the cluster affects this number ( for example, if each parallel spark task does its lookups in parallel with other spark tasks, maybe this is increasing the throughput and we can model this better instead of using a contant).

Yes, this is something we would like to test as well but it heavily rely on multiple k8s clusters with different olap executors setup. We didn’t managed to make much progress there. OLAP coefficient is likely to related to number of executors, but I think we can only do this later.

martinrupp (commented):
+1

PR #5378, opened 2021-04-07
DB-11829 Evaluate constant String -> String conversion in binding

ascend1:
## Short Description
This change allows constant string to string conversion to be evaluated in binding.

## Long Description
In investigating query hash `SMopSK5W`, three problems are identified (for detailed explanation, see Jira DB-11718). This PR addresses the first issue.

So far, we don't evaluate a cast node if it's from string to string, for example, `cast('abc' as char(4))`. Predicate selectivity estimation is affected if a predicate contains such a cast node as one of its operand. Since it's not a constant node, we will go for default estimation rather than ask statistics for a concrete value.

The reason we don't evaluate this kind of cast might be that we could cast a constant string to a narrower type, leading to data truncation. The truncation itself is totally fine, but according to standard, we should raise a warning in this case and send this warning together with the final result set to client. This is something not really possible in binding phase. However, early evaluation should be possible for casts whose target types are wider.

## How to test
Casting a string literal to a not-narrower type should be evaluated right in binding phase. However, this difference is not observable on client side. See `StringTypesToStringTypesConversionIT` for more details about this.


martinrupp (approved):

carolp-503 (approved):
+1

ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2812/

jpanko1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2892/

ascend1:

spliceci:
TEST SUCCEEDED +1

PR #5382, opened 2021-04-09
DB-10765 Update only affected indexes

dgomezferro:
## Short Description
Currently we are updating all indices even when updating a non-indexed column, we should only update indexes that contain the updated column

## Long Description
We used to avoid updating indices that were not affected by an update, but this had problems when these indices overlapped affected indices. https://splicemachine.atlassian.net/browse/DB-10746 fixed it by always updating all indices

This PR restores the code that tries to update only the affected indices (in `DMLModStatementNode`) and improves the logic in `IndexTransformer` to detect when an index write has to be generated (from "one of the indexed columns is in the write" to "all indexed columns are in the write"). We need to ignore PK columns because those are not included in the value part of the write, but in the rowkey. 

This still would cause spurious writes to indices that are fully contained on other indices (see comment in `IndexWriteHandler` for more details) To fix this we just ignore index updates where we are deleting and inserting the same row. (We'd need to revisit this if we ever support "indexes with included/non-key columns" like https://docs.microsoft.com/en-us/sql/relational-databases/indexes/create-indexes-with-included-columns?view=sql-server-ver15 )

## How to test
IT was already part of https://splicemachine.atlassian.net/browse/DB-10746
Create indices with overlapping indexed columns, run inserts, deletes, etc, make sure they don't become inconsistent.
Verify HBase data to make sure we are not writing more entries than necessary.


dgomezferro:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2748/

arnaud-splice (approved):

martinrupp (approved):

dgomezferro:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5383, opened 2021-04-12
DB-11845 sqlshell cancel queries with CTRL+C

martinrupp:
## Description

Currently, when in hitting Ctrl+C in sqlshell, it directly terminates, without cancelling the current running query. With this fix, we don’t terminate sqlshell, but cancel the current query.

On Mac and Linux, If no query is running, Ctrl+C does nothing. You can quit with Ctrl+D.

On Windows, Ctrl+C still terminates the application if no query is running.

## How to test

General, execute a long query and try to terminate with Ctrl+C. Note that before this fix, a CTRL+C would cause the sqlshell to get closed, but the query still to run (check with call `SYSCS_UTIL.SYSCS_GET_RUNNING_OPERATIONS();`).

Detailed test description with testdata in https://splicemachine.atlassian.net/browse/DB-11845 .

### DIFF
--- a/db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ijImpl.java
+++ b/db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ijImpl.java
@@ -4,8 +4,6 @@ import com.splicemachine.db.iapi.tools.i18n.LocalizedOutput;
 import com.splicemachine.db.iapi.tools.i18n.LocalizedResource;
 import com.splicemachine.db.tools.JDBCDisplayUtil;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import sun.misc.Signal;
-import sun.misc.SignalHandler;
 
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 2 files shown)


bklo94 (approved):

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ijImpl.java
```diff
@@ -31,6 +36,30 @@
...
+                if(currentStatement == null) {
+                    out.println("\nCTRL+C -> exit...\n");
+                    originalSigIntHandler.handle(sigInt);
+                }
+                else {
+                    out.println("\nCTRL+C -> Canceling Statement...\n");
+                    currentStatement.cancel();
+                }
+            } catch (SQLException e) {
+                out.println("Error Canceling Statement: " + e.toString());
```
martinrupp, line 59: i just looked this up, seems like Canceling is American English and Cancelling is British English, see e.g. https://en.wiktionary.org/wiki/canceling#English

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ijImpl.java
```diff
@@ -1,7 +1,11 @@
 package com.splicemachine.db.impl.tools.ij;
 
+import com.splicemachine.db.iapi.tools.i18n.LocalizedOutput;
 import com.splicemachine.db.iapi.tools.i18n.LocalizedResource;
 import com.splicemachine.db.tools.JDBCDisplayUtil;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import sun.misc.Signal;
```
hatyo, line 7: Using `sun.misc.Signal` is a bit "controversial" to say the least (the API contains a lot of non-standard OS-specific code and it was supposed to be removed but somehow survived as [listed here](https://stackoverflow.com/a/34188676)).

Maybe that's OK considering the benefit this PR brings, I am guessing you tested it already on Mac/Linux and maybe Windows? if not, then could you please test the client there, there is a [guide](https://splicemachine.atlassian.net/wiki/spaces/PD/pages/976289824/How+to+use+Windows+SQL+Shell) on how to do so.

hatyo (changes requested):

db-tools-ij/src/main/java/com/splicemachine/db/impl/tools/ij/ijImpl.java
martinrupp: i didn't tested it on linux and windows, but now i did. Works on both Windows and Linux. What can happen is that sun.misc.Signal is not allowed to be executed, that's why i added a try/catch block around this.
However, it seems like on some systems there is no sun.misc.Signal so i will change this by using reflection.

martinrupp: done

hatyo (approved):

martinrupp:

spliceci:
TEST SUCCEEDED +1

PR #5384, opened 2021-04-13
DB-11189 Avoid base table lookups for index updates

dgomezferro:
## Short Description

## Long Description

### DIFF
--- a/pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
+++ b/pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
@@ -9,9 +9,36 @@ import com.splicemachine.storage.index.BitIndex;
 import com.splicemachine.storage.index.BitIndexing;
 import com.splicemachine.utils.ByteSlice;
 
-public abstract class AbstractWriteHandler implements WriteHandler {
+public class UpdateUtils {
 
-    protected KVPair getBaseUpdateMutation(KVPair mutation) {
+    public static KVPair deleteFromUpdate(KVPair mutation) {
+        EntryDecoder rowDecoder = new EntryDecoder();
+        rowDecoder.set(mutation.getValue());
+        BitIndex index = rowDecoder.getCurrentIndex();
+        MultiFieldDecoder fieldDecoder = rowDecoder.getEntryDecoder();
+        int cardinality = index.cardinality();
+        for(int i=index.nextSetBit(0), c=0; c<cardinality/2; i=index.nextSetBit(i+1), c++){
+            rowDecoder.seekForward(fieldDecoder, i);
+        }
+        byte[] data = fieldDecoder.slice(fieldDecoder.length() - fieldDecoder.offset());
+        BitIndex resultIndex = BitIndexing.getBestIndex(
+                halveSet(index.getFields()),
+                halveSet(index.getScalarFields()),
+                halveSet(index.getFloatFields()),
+                halveSet(index.getDoubleFields()));
+        byte[] bitData = resultIndex.encode();
+        byte[] result = new byte[bitData.length+data.length+1];
+        System.arraycopy(bitData, 0, result, 0, bitData.length);
+        result[bitData.length] = 0;
+        System.arraycopy(data,0,result,bitData.length+1,data.length);
+
+        ByteSlice rowSlice = mutation.rowKeySlice();
+        return new KVPair(
+                rowSlice.array(),rowSlice.offset(),rowSlice.length(),
+                result,0,result.length,KVPair.Type.DELETE);
+    }
+
+    public static KVPair getBaseUpdateMutation(KVPair mutation) {
         EntryDecoder rowDecoder = new EntryDecoder();
         rowDecoder.set(mutation.getValue());
         BitIndex index = rowDecoder.getCurrentIndex();
@@ -40,7 +67,7 @@ public abstract class AbstractWriteHandler implements WriteHandler {
                 result,0,result.length,KVPair.Type.UPDATE);
     }
 
-    protected BitSet halveSet(BitSet bitSet) {
+    public static BitSet halveSet(BitSet bitSet) {
         BitSet result = new BitSet(bitSet.capacity()/2);
         long cardinality = bitSet.cardinality();
         BitSetIterator it = bitSet.iterator();
--- a/splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
@@ -38,6 +38,7 @@ import com.splicemachine.pipeline.callbuffer.RecordingCallBuffer;
 import com.splicemachine.pipeline.config.RollforwardWriteConfiguration;
 import com.splicemachine.pipeline.config.WriteConfiguration;
 import com.splicemachine.pipeline.utils.PipelineUtils;
+import com.splicemachine.pipeline.writehandler.UpdateUtils;
 import com.splicemachine.primitives.Bytes;
 import com.splicemachine.si.api.txn.TxnView;
 import com.splicemachine.si.constants.SIConstants;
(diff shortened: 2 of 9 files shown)


dgomezferro:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2754/

splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
```diff
@@ -95,10 +97,17 @@ public void open(TriggerHandler triggerHandler, SpliceOperation operation, boole
...
-            colPositionMap[i]=pos;
+        for(int i=heapList.anySetBit(), pos=heapList.getNumBitsSet();i!=-1;i=heapList.anySetBit(i),pos++) {
+            colPositionMap[i] = pos;
+            colPositionMap[i+heapList.size()] = pos - heapList.getNumBitsSet();
+        }
+        valuesList = (FormatableBitSet) heapList.clone();
+        valuesList.grow(heapList.size()*2);
+        for (int i=heapList.anySetBit();i!=-1;i=heapList.anySetBit(i)) {
+            valuesList.set(i + heapList.size());
+        }
```
martinrupp, line 110: i don't understand this code.
please add a comment (in the code) to explain what's going on here.

martinrupp, line 110: i don't understand this code.
please add a comment (in the code) to explain what's going on here.
should that be part of UpdateUtils ?

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        result[bitData.length] = 0;
+        System.arraycopy(data,0,result,bitData.length+1,data.length);
+
+        ByteSlice rowSlice = mutation.rowKeySlice();
+        return new KVPair(
+                rowSlice.array(),rowSlice.offset(),rowSlice.length(),
+                result,0,result.length,KVPair.Type.UPDATE);
+    }
+
+    public static BitSet halveSet(BitSet bitSet) {
```
martinrupp, line 70: add comment and unit test

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+
+import com.carrotsearch.hppc.BitSet;
+import com.carrotsearch.hppc.BitSetIterator;
+import com.splicemachine.encoding.MultiFieldDecoder;
+import com.splicemachine.kvpair.KVPair;
+import com.splicemachine.storage.EntryDecoder;
+import com.splicemachine.storage.index.BitIndex;
+import com.splicemachine.storage.index.BitIndexing;
+import com.splicemachine.utils.ByteSlice;
+
```
martinrupp, line 11: or add a description here. PR descriptions are good but we should also copy the necessary information to understand this code over here.

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+public class UpdateUtils {
+
+    public static KVPair deleteFromUpdate(KVPair mutation) {
```
martinrupp, line 14: this looks like it can be easily unit-tested. please add a unit test and a comment what this should do.

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        result[bitData.length] = 0;
+        System.arraycopy(data,0,result,bitData.length+1,data.length);
+
+        ByteSlice rowSlice = mutation.rowKeySlice();
+        return new KVPair(
+                rowSlice.array(),rowSlice.offset(),rowSlice.length(),
+                result,0,result.length,KVPair.Type.DELETE);
+    }
+
+    public static KVPair getBaseUpdateMutation(KVPair mutation) {
```
martinrupp, line 41: add comment and unit test

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        EntryDecoder rowDecoder = new EntryDecoder();
+        rowDecoder.set(mutation.getValue());
+        BitIndex index = rowDecoder.getCurrentIndex();
+        MultiFieldDecoder fieldDecoder = rowDecoder.getEntryDecoder();
+        int offset = fieldDecoder.offset();
+        for(int i=index.nextSetBit(0), c=0; c<index.cardinality()/2; i=index.nextSetBit(i+1), c++){
+            rowDecoder.seekForward(fieldDecoder, i);
+        }
+        int end = fieldDecoder.offset();
```
martinrupp, line 50:
```
int offset = fieldDecoder.offset();
int end = fieldDecoder.offset();
byte[] data = fieldDecoder.slice(end - offset - 1);
```
this looks like
`byte[] data = fieldDecoder.slice( - 1);`
????
am i missing something?

why is this so different from line 23?
`byte[] data = fieldDecoder.slice(fieldDecoder.length() - fieldDecoder.offset());`

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        byte[] bitData = resultIndex.encode();
+        byte[] result = new byte[bitData.length+data.length+1];
+        System.arraycopy(bitData, 0, result, 0, bitData.length);
+        result[bitData.length] = 0;
+        System.arraycopy(data,0,result,bitData.length+1,data.length);
+
+        ByteSlice rowSlice = mutation.rowKeySlice();
+        return new KVPair(
+                rowSlice.array(),rowSlice.offset(),rowSlice.length(),
+                result,0,result.length,KVPair.Type.UPDATE);
```
martinrupp, line 67: except for line 23/52 and KVPair.Type.UPDATE and KVPair.Type.DELETE, functions
deleteFromUpdate and getBaseUpdateMutation are the same.
Please unify.

martinrupp (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
dgomezferro: I refactored it a bit and added comments, please let me know if it's more understandable now.

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: Done

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: Added explanation

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: Added explanation and unit test

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: Done

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: Addressed in the other PR

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: You are right, consolidated in single method. Thanks for the review!

dgomezferro:
Closing PR, superseded by https://github.com/splicemachine/spliceengine/pull/5391

PR #5386, opened 2021-04-13
DB-11189 Avoid base table lookups for index updates (3.0)

dgomezferro:
## Short Description

## Long Description


dgomezferro:
Master only for now

PR #5388, opened 2021-04-13
DB-11860 Prefer indexes with more selective predicates

ascend1:
## Short Description
This change provides a temporary fix for DB-11860.

## Long Description
Main reason is again that statistics are collected when the table is empty. Long-term solution would be again automatically updated statistics.

## How to test
Check `testPreferIndexOnSpecialColumns` for details.

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
@@ -19,6 +19,7 @@ import com.splicemachine.access.api.SConfiguration;
 import com.splicemachine.derby.test.framework.SpliceSchemaWatcher;
 import com.splicemachine.derby.test.framework.SpliceUnitTest;
 import com.splicemachine.derby.test.framework.SpliceWatcher;
+import com.splicemachine.derby.test.framework.TestConnection;
 import com.splicemachine.homeless.TestUtils;
 import com.splicemachine.test_tools.TableCreator;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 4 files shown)


ascend1:

hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
```diff
@@ -460,4 +461,44 @@ public void testPreferIndexAccessEqualCosts() throws Exception {
...
+                ",\"K_A\" CHAR(1) NOT NULL DEFAULT ''\n" +
+                ",\"ST5\" DECIMAL(5,0) NOT NULL DEFAULT 0.0\n" +
+                ",\"ST\" CHAR(1) NOT NULL DEFAULT ''\n" +
+                ",\"VAK\" CHAR(1) NOT NULL DEFAULT ''\n" +
+                ",\"FK#\" CHAR(36) NOT NULL\n" +
+                ",\"D\" VARCHAR (32000) FOR BIT DATA NOT NULL DEFAULT X''\n" +
+                ", CONSTRAINT PK PRIMARY KEY(\"BP#\"))");
+
+        methodWatcher.execute("create index IDX1 on TP(GF#, TRS, ST, VAK, GFAK, SN, A_G#)");
+        methodWatcher.execute("create index IDX2 on TP(SN, VAK, GF#, ST, TRS, GFAK, A_G#)");
```
dgomezferro, line 484: @ascend1 how does this work? Does IDX1 have 1 specialColumn (because of the predicate on GF#) and IDX2 0 because there's no prefix up to GF#? If IDX2 was (SN, GF#,...) would they be tied or would IDX1 win because it's first?

ascend1: The code counts how many scan start/stop keys are on special columns. If IDX2 is defined as (SN, GF#) then they will tie, because the predicate on GF# would also be a scan key. Order of column is not considered because generally, there is no relation between how selective a predicate is and on which index column ordinal position the predicate applies.

dgomezferro: That wasn't the relation I was going for, if you have two predicates P0 on C0 and P1 on C1, and you have idx0 (C0, C1) and idx1 (C1, C0), you want to choose the index that puts the most selective predicate first (that means you end up scanning less rows). That's what I meant.

dgomezferro (approved):
I think the heuristic is good enough for further experimentation

martinrupp (approved):

hbase_sql/src/test/java/com/splicemachine/hbase/CostEstimationIT.java
ascend1: I see. Then I think our selectivity estimate to apply for calculating the number of scanned rows is not really correct. As far as I understand, if we have P0 on C0 and P1 on C1, both are scan start/stop keys, then they are pushed to the same scan selectivity phase. That means scan selectivity is s(P0) * sqrt(s(P1))  (suppose s(P0) is smaller), and it's the same for both idx0 and idx1. According to what you just mentioned, we should probably calculate this only based on predicates applied on the leading column?

Back to this ticket, unfortunately the selectivitiy of predicates are not available when we compare costs. They are calculated in `ScanCostFunction` and then thrown away. But again, this is not a proper fix anyway and we need a special treatment for the 0-stat case.

ascend1:

PR #5390, opened 2021-04-14
DB-11865 Enable native spark broadcast join with non-equijoin predicates

msirek:
## Short Description
Enable native spark broadcast join with non-equijoin predicates.

## Long Description
[SPLICE-2362](https://splice.atlassian.net/browse/SPLICE-2362) added support for general expressions in native spark join predicates.  It was enabled for combinations of equijoin plus non-equijoin predicates, but not non-equijoin predicates by themselves.  This Jira enables the latter by removing the check in BroadcastJoinOperation which checks that we have at least one left hash key before choosing native spark join.

## How to test
A test case such as the following should indicate BroadcastNestedLoopJoin in the sparkexplain output:

> create table t1(a1 int not null, b1 int, c1 int, primary key(c1));
> create table t2 (a2 int not null, b2 int, c2 int, primary key(c2));
> sparkexplain select * from t1 a, t2 b --splice-properties joinStrategy=broadcast
>             where a.a1 between b.a2 and b.a2+1;
> 
> Native Spark Execution Plan
> ----------------------------
> -> NativeSparkDataSet
>     *(3) Project [c0#4365, c1#4366, c2#4367, c0#4380 AS c3#4431, c1#4381 AS c4#4432, c2#4382 AS c5#4433]
>     +- BroadcastNestedLoopJoin BuildRight, Inner, ((c0#4365 >= c0#4380) && (cast(c0#4365 as bigint) <= (cast(c0#4380 as bigint) + 1)))
>        :- *(1) Filter isnotnull(c0#4365)
>        :  +- Scan ExistingRDD[c0#4365,c1#4366,c2#4367]
>              -> TableScan[T1(7728)](RS=0,totalCost=4.04, scannedRows=20,outputRows=20, outputHeapSize=60 B,partitions=1, parallelTasks=8)
>        +- BroadcastExchange IdentityBroadcastMode
>           +- *(2) Filter isnotnull(c0#4380)
>              +- Scan ExistingRDD[c0#4380,c1#4381,c2#4382]
>                 -> TableScan[T2(7744)](RS=2,totalCost=4.04, scannedRows=20,outputRows=20, outputHeapSize=97 B,partitions=1, parallelTasks=8)

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
@@ -180,7 +180,7 @@ public class BroadcastJoinOperation extends JoinOperation{
          * so route to the rdd implementation for now for SSQ.  Also, spark join can't handle a zero length
          * hash key, so it will use rdd as well.
          */
-        if (rightFromSSQ || leftHashKeys.length == 0)
+        if (rightFromSSQ || (leftHashKeys.length == 0 && !hasSparkJoinPredicate()))
             useDataset = false;
 
         DataSet<ExecRow> result;
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
@@ -180,7 +180,7 @@ public class BroadcastJoinOperation extends JoinOperation{
          * so route to the rdd implementation for now for SSQ.  Also, spark join can't handle a zero length
          * hash key, so it will use rdd as well.
          */
-        if (rightFromSSQ || (leftHashKeys.length == 0 && !hasSparkJoinPredicate()))
+        if (rightFromSSQ)
             useDataset = false;
 
         DataSet<ExecRow> result;
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
@@ -177,8 +177,7 @@ public class BroadcastJoinOperation extends JoinOperation{
         boolean useDataset = true;
 
         /** TODO don't know how to let spark report SQLState.LANG_SCALAR_SUBQUERY_CARDINALITY_VIOLATION error,
-         * so route to the rdd implementation for now for SSQ.  Also, spark join can't handle a zero length
-         * hash key, so it will use rdd as well.
+         * so route to the rdd implementation for now for SSQ.
          */
         if (rightFromSSQ)
             useDataset = false;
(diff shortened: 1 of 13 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2757/

msirek:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
```diff
@@ -178,7 +178,7 @@ public SpliceOperation getLeftOperation(){
          * so route to the rdd implementation for now for SSQ.  Also, spark join can't handle a zero length
          * hash key, so it will use rdd as well.
          */
-        if (rightFromSSQ || leftHashKeys.length == 0)
+        if (rightFromSSQ || (leftHashKeys.length == 0 && !hasSparkJoinPredicate()))
```
dgomezferro, line 181: Do we need to check for `varcharDB2CompatibilityMode` here? If it's enabled we won't apply the sparkJoinPredicate to the native join (it'll be applied as a non-native flat-map function after the full join is generated). Is that OK or not? I guess that's still what we'd be doing if we set `useDataset = true`, so maybe this is fine.

dgomezferro (changes requested):
Marking as changes requested for potential problem, PR looks good otherwise. If my concern is not a real problem I'm OK with merging.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/BroadcastJoinOperation.java
msirek: @dgomezferro This is a great point.  I have added code to transform the sparkJoinPredicate into an column expression with rtrimmed columns, so varcharDB2CompatibilityMode can use full native spark too.

msirek: Also, I have reworked the code in HashableJoinStrategy to allow broadcast join to be picked in more cases (which triggers BroadCastNestedLoopJoin in Spark).  Previously the rules were added because BroadcastJoinStrategy.feasible was called even for single-table access, so we ended up pick broadcast join for things which weren't joins.  
The new check:

> if (!isSingleTableScan(optimizer))

should avoid those cases without all the convoluted rules.

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2874/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2876/

msirek:

spliceci:
TEST SUCCEEDED +1

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2905/

dgomezferro (approved):

ascend1 (approved):

ascend1:

arnaud-splice (approved):

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2921/

msirek:

msirek:
IT run:
http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/2106/default/

carolp-503 (approved):
+1

PR #5391, opened 2021-04-14
DB-11189 Avoid base table lookups for index updates (with DB-10765)

dgomezferro:
## Short Description

## Long Description


dgomezferro:

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/WriteHandler.java
```diff
@@ -14,8 +14,13 @@
 
 package com.splicemachine.pipeline.writehandler;
 
+import com.splicemachine.encoding.MultiFieldDecoder;
 import com.splicemachine.kvpair.KVPair;
 import com.splicemachine.pipeline.context.WriteContext;
+import com.splicemachine.storage.EntryDecoder;
+import com.splicemachine.storage.index.BitIndex;
+import com.splicemachine.storage.index.BitIndexing;
+import com.splicemachine.utils.ByteSlice;
```
dgomezferro, line 23: Cleanup

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2762/

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
```diff
@@ -477,11 +483,13 @@ public boolean areIndexKeysModified(KVPair mutation, BitSet indexedColumns) {
...
-            if (indexedColumns.get(i))
-                return true;
+        BitSetIterator iterator = nonPkIndexedCols.iterator();
+        int nextBit;
+        while((nextBit = iterator.nextSetBit()) != -1) {
+            if (!updateIndex.isSet(nextBit))
+                return false;
         }
-        return false;
+        return true;
```
ascend1, line 492: It looks like parameter `indexedColumns` is not needed after the change.

splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
```diff
@@ -95,10 +96,12 @@ public void open(TriggerHandler triggerHandler, SpliceOperation operation, boole
...
         for(int i=0;i<columnOrdering.length;++i)
             kdvds[i]=LazyDataValueFactory.getLazyNull(formatIds[columnOrdering[i]]);
-        colPositionMap=new int[heapList.size()];
+        colPositionMap=new int[heapList.size()*2];
         // Map Column Positions for encoding
-        for(int i=heapList.anySetBit(), pos=heapList.getNumBitsSet();i!=-1;i=heapList.anySetBit(i),pos++)
-            colPositionMap[i]=pos;
+        for(int i=heapList.anySetBit(), pos=heapList.getNumBitsSet();i!=-1;i=heapList.anySetBit(i),pos++) {
+            colPositionMap[i] = pos;
+            colPositionMap[i+heapList.size()] = pos - heapList.getNumBitsSet();
```
arnaud-splice, line 103: Can you add a comment about this part? The map being twice as long as before to store old and new values

arnaud-splice (approved):

martinrupp (changes requested):
see comments on https://github.com/splicemachine/spliceengine/pull/5384

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -9,9 +9,36 @@
 import com.splicemachine.storage.index.BitIndexing;
 import com.splicemachine.utils.ByteSlice;
 
-public abstract class AbstractWriteHandler implements WriteHandler {
+public class UpdateUtils {
 
-    protected KVPair getBaseUpdateMutation(KVPair mutation) {
+    public static KVPair deleteFromUpdate(KVPair mutation) {
```
ascend1, line 14: I think it makes sense to have a comment about the content of mutation passed in to this function. I feel the logic in this function that takes half of the fields can be confusing to someone who doesn't have the background of this problem. Just a suggestion.

ascend1 (approved):
I didn't get what problem the commit "DB-11189 Fix constraints and Mem platform" fixes. Otherwise it looks good.

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        rowDecoder.set(mutation.getValue());
+        BitIndex index = rowDecoder.getCurrentIndex();
+        MultiFieldDecoder fieldDecoder = rowDecoder.getEntryDecoder();
+        int offset = fieldDecoder.offset();
+        for(int i=index.nextSetBit(0), c=0; c<index.cardinality()/2; i=index.nextSetBit(i+1), c++){
+            rowDecoder.seekForward(fieldDecoder, i);
+        }
+        int end = fieldDecoder.offset();
+        fieldDecoder.seek(offset);
+        byte[] data = fieldDecoder.slice(end - offset - 1); // -1 to remove the NULL separator
```
martinrupp, line 52: (copied from https://github.com/splicemachine/spliceengine/pull/5384)

```
int offset = fieldDecoder.offset();
int end = fieldDecoder.offset();
byte[] data = fieldDecoder.slice(end - offset - 1);
```
this looks like
`byte[] data = fieldDecoder.slice( - 1);`
????
am i missing something?

why is this so different from line 23?
`byte[] data = fieldDecoder.slice(fieldDecoder.length() - fieldDecoder.offset());`

Apart from this, this function looks like a copy&paste of the function deleteFromUpdate, except for KVPair.Type.UPDATE

martinrupp (changes requested):

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
martinrupp: ah ok `rowDecoder.seekForward(fieldDecoder, i);` would change this. But why is that code so different from deleteFromUpdate?

carolp-503 (approved):
+1

dgomezferro:

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: In order to decode the value we need to skip over one field at a time. In general fields are separated by a null value `0x00` but depending some types of fields (scalar, float & double) can contain null bytes, and the type information is encoded in the BitIndex structure.
So if we know we have 4 fields, we skip over 2 fields and get that offset. If we want the second half we take everything from that point till the end, if we want the first half remember that position, we `seek` the decoder back to the beginning and take everything from the beginning to the middle position.
I renamed some variables to hopefully make it more understandable.

pipeline_api/src/main/java/com/splicemachine/pipeline/writehandler/UpdateUtils.java
dgomezferro: I added that explanation to `extractWriteFromUpdate` after I refactored `deleteFromUpdate` and `getBaseUpdateMutation` as @martinrupp suggested.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
dgomezferro: Good catch, removed.

splice_machine/src/main/java/com/splicemachine/derby/stream/output/update/UpdatePipelineWriter.java
dgomezferro: I added a short comment and a reference to `UpdateUtils.extractWriteFromUpdate`

dgomezferro:
> I didn't get what problem the commit "DB-11189 Fix constraints and Mem platform" fixes. Otherwise it looks good.

@ascend1 the constraint fix has to do with exiting early in `UpdateNode.getUpdateReadMap`. Since I was including all columns on the select (if we update the PK) I thought we shouldn't keep checking constraints etc for more columns to include, but there were some side effects that were needed (for instance `baseTable.getAllRelevantConstraints` updates `relevantConstraints` )

The Mem platform fix was related to DB-10765, the `get()` method in MPartition (used when deleting an index entry) wasn't applying filters correctly. If it needed to merge multiple KVs to generate the result it would return the first KV. For instance if we are asking for columns {1, 3} and there's two KVs one with columns {0, 1} and another with columns {0,1,2,3} instead of taking column 0 from the first and column 3 from the second to materialize the result correctly, it would return the first KV with values {0, 1} so we'd delete the wrong index entry (in practice we wouldn't delete the index entry). That was behind this jenkins failure http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2748/

spliceci:
TEST SUCCEEDED +1

dgomezferro:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2766/

dgomezferro:

martinrupp (approved):

spliceci:
TEST SUCCEEDED +1

PR #5392, opened 2021-04-14
DB-11224 Don't set exclude null/default flags of an index when they have no effect

ascend1:
## Short Description
This change improves the logic of setting `exclude null/default keys` flags of an index.

## Long Description
From splice SQL reference:
> Excluding NULL and Default Values
>
> You can include the optional EXCLUDE clause to specify that you want to exclude from the index either default values or NULL values for the column. This can be desirable for a large table in which the index column is largely populated with the same (default or NULL) value.
> Note: For an index with multiple columns, only rows with NULL or default values on the leading index column are excluded.

For the NPE issue described in Jira, root cause is that an index with `exclude default keys` clause is defined on a set of columns but the leading column has no default value defined. When optimizer checks if this index is eligible for a certain query, it tries to retrieve the default value of the leading column because the index has `excludeDefault` flag set to true. Default value returned is null, leading to NPE for the following checks on the default value.

However, in the case above, `exclude default keys` has no effect because the column has no default values, thus nothing to exclude.

A similar issue happens to `exclude null keys` clause, too. Though the symptom is less severe. Optimizer could consider the index not eligible for a query, missing an alternative access path.

In detail, we could also define an index with `exclude null keys` on a column that is `not null`. Effectively, the clause also has no effect because there is nothing to exclude.

The change fixes both problems by adding a check in binding phase of creating an index. The check applies to the leading column of the index only. The check contains two parts:

- If `exclude null keys` is specified but the leading column is `not null`, unset the `excludeNulls` flag so that the `exclude null keys` clause is effectively removed;
- If `exclude default keys` is specified but the leading column has no default value defined, unset the `excludeDefault` flag so that the `exclude default keys` clause is effectively removed.

## How to test
Two new ITs `createIndexExcludeNullNotApplicable` and `createIndexExcludeDefaultNotApplicable` are added. They should fail without the fix but pass otherwise. In particular, without the fix, 

- `createIndexExcludeNullNotApplicable` would fail with an error message that no valid plan is found;
- `createIndexExcludeDefaultNotApplicable` would fail with NPE.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/CreateIndexNode.java
@@ -255,6 +255,18 @@ public class CreateIndexNode extends DDLStatementNode
                 throw StandardException.newException(SQLState.LANG_COLUMN_NOT_ORDERABLE_DURING_EXECUTION,
                         columnDescriptor.getType().getTypeId().getSQLTypeName());
             }
+
+            // unset exclude flags that has no effect
+            // consider only the first index column
+            if (i == 0) {
+                if (excludeNulls && !columnDescriptor.getType().isNullable()) {
+                    excludeNulls = false;
+                }
+
+                if (excludeDefaults && columnDescriptor.getDefaultValue() == null) {
+                    excludeDefaults = false;
+                }
+            }
         }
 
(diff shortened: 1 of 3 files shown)


ascend1:

spliceci:
TEST SUCCEEDED +1

ipraznik-splice (approved):

dgomezferro (approved):

PR #5395, opened 2021-04-15
DB-11863 Defer DDL notification in collecting statistics

ascend1:
## Short Description
This change move `ddlNotification()` call in `COLLECT_SCHEMA_STATISTICS()` and `doStatsCollectionForTables()` from right after dropping old statistics to right after new statistics are collected.

This problem is not a serious one. As of current implementation, a prepared statement would never miss newly collected statistics as long as the transaction of collecting statistics is committed. This change brings a minor improvement that a prepared statement is not invalidated too early because of statistics are being re-collected, saving some time in invoking optimizer to generate a plan that is identical to the already cached one.

## Long Description

The following figure shows an example of concurrent execution of `analyze` and `ps.executeQuery()`.
```
                                                      currChangeCount++,                                             currChangeCount--
                                                      stmt cache disabled                                            stmt cache enabled
 T_DDL   |                                                    v                                                              v
Time     |---------------------------------------------------------------------------------------------------------------------------->
    T1   |         ^                  ^                      ^                       ^                                      ^
               analyze start    drop old stats           ddl notify            collect new                                commit
    T2   |                                       ^                         ^                                                     ^
                                            exec ps (1)                 exec ps (2)                                      exec ps (3)
```

T2 executes the prepared statement three times. The first time happens before T1 calls `ddlNotification` so the prepared statement is not invalidated yet. This time, `ps` just run with the cached plan.

The second execution of `ps` happens after T1 calls `ddlNotification`, which invalidates all objects depending on the table on which statistics are going to be re-collected, including `ps`. For this reason, this execution of `ps` goes through the optimizing process for a new plan. Since `analyze` is also transactional, this run of `ps` won't see the stats are dropped. So the new plan would be exactly the same as the previously cached one, because they are both based on the old statistics (let's not think about strange non-determinism here).

One question is, whether this execution of `ps` would cache this old plan and validates itself again. If that happens, following execution of `ps` wouldn't pick up newly collected statistics because it's already valid and up to date, no compilation is needed later. That would be a problem.

Fortunately, this is not the case. When a `DDLChange` is sent to region servers, each region server has an `AsynchronousDDLWatcher` that picks this `DDLChange` up and update `currChangeCount` in `DDLWatchRefresher`. Every incoming `DDLChange` increases `currChangeCount` by 1. However, if `currChangeCount != 0`, data dictionary cache is disabled:
```
public boolean canWriteCache(TransactionManager xact_mgr) {
        return cacheIsValid() && canSeeDDLDemarcationPoint(xact_mgr);
    }

public boolean cacheIsValid() {
        return currChangeCount.get() ==0;
    }
```

That means the second execution of `ps` will run through, but it cannot cache its plan because the cache is not available. As of current implementation, `currChangeCount` is only decreased when the transaction notifying `DDLChange` commits. That means before the `analyze` in T1 finishes, we can run as many times of `ps` as we want but all of them will get recompiled, not because dependency is invalidated, but because they don't have a cached plan.

Finally, T2 executes `ps` after the `analyze` in T1 commits. It still got recompiled because previous executions of `ps` didn't cache a plan. Suppose we didn't run `ps` after `ddlNotification`, then this execution of `ps` also got recompiled because it's cached plan got invalidated by `ddlNotification`. So in both cases, it got recompiled.

Recompilation is exactly what we want here because it needs to pick up the new statistics.

This change put `ddlNotification` right after new statistics are collected (shortly before commit in the figure above) to minimize the time period during which running `ps` requires a recompilation.

## How to test
A new IT `testInvalidateStoredPlanTimePointInCollectingStats` is added.

In short, the IT starts collecting statistics in a separate thread and run `ps` repetitively in main thread before statistics collection is done. All these executions of `ps` should see old statistics. Unfortunately, it's not possible to see if they are actually recompiled (being called before `ddlNotification` or after).

In the end, the IT run `ps` again right after statistics are collected and committed. This run of `ps` should see new statistics and produces a new plan. `scannedRows` of the plan is examined to verify that it sees new statistics.

### Problem of dependency on explain

In building this IT, one problem emerged. If one prepares an explain statement, `ddlNotification` would not invalidate that prepared statement because there is no dependency built between the explain statement and the tables involved in the statement to be explained.

This is a minor problem introduced by DB-11258. I don't want to explain it fully here but effectively, if we have `explain X`, then we build dependency between `X` and the tables. Before that change, dependencies are built between `explain X` and the tables.

This is generally not a problem at all, since explain statement are never cached. The problem only has an effect when we prepare an explain statement, which is a rare use case in practice.

The problem does affect testing, though. For this reason, a fix is included in this PR.


ascend1:
enkins please test branch @cdh6.3.0,skipTestsLongerThan2Minutes

ascend1:
enkins please test branch @cdh6.3.0,skipTestsLongerThan2Minutes

martinrupp:

martinrupp (approved):

dgomezferro (approved):

spliceci:
TEST SUCCEEDED +1

PR #5400, opened 2021-04-16
DB-11755 Add a new frequencySketchSize column to SYS.SYSCOLUMNS

mush-zhang:
## Short Description
We need a new column in SYS.SYSCOLUMNS to support different sketch sizes per column. 

## Long Description
- Modifying SYSCOLUMNSRowFactory to add a new column.
- Providing an upgrade script to upgrade our data dictionary from a previous version

## How to test
manual testing with `describe sys.systables;`

test upgrade script:
```
cd spliceengine
git checkout 3.2.0.2000
./start-splice-cluster -k 
mvn -T 4 -B clean install -Pcore,cdh6.3.0 -DskipTests 
./start-splice-cluster -b -p cdh6.3.0
git checkout DB-11755
./start-splice-cluster -l -p cdh6.3.0
```


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1437,4 +1444,85 @@ public TabInfoImpl getTableInfo(int catalogNum) throws StandardException{
...
+                    TransactionController.ISOLATION_REPEATABLE_READ) ) {
+                // If upgrade has already been done, and we somehow got here again by
+                // mistake, don't re-add the columns to the conglomerate descriptor.
+                if (heapCC instanceof HBaseController) {
+                    HBaseController hCC = (HBaseController) heapCC;
+                    if (hCC.getConglomerate().getFormat_ids().length >= lastCol) {
+                        return;
+                    }
+                }
+            }
```
jyuanca, line 1469: TableDescriptor.columnDescriptorList stores its columns. You don't need to fetch from SPLICE_CONGLOMERATE table

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1437,4 +1444,85 @@ public TabInfoImpl getTableInfo(int catalogNum) throws StandardException{
...
+            while (sc.fetchNext(fetchedRow)) {
+                for (int i = 0; i < fetchedRow.length; ++i) {
+                    if (ArrayUtils.contains(colIds, i + 1)) {
+                        newRow[i] = templateRow.getColumn(i + 1).cloneValue(false);
+                    } else {
+                        newRow[i] = fetchedRow[i] == null ? null : fetchedRow[i].cloneValue(false);
+                    }
+                }
+                sc.replace(newRow, columnToUpdateSet);
+            }
```
jyuanca, line 1519: It is unnecessary to rewrite all rows in sys.syscolumns table

jyuanca (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/SYSCOLUMNSRowFactory.java
```diff
@@ -507,6 +515,9 @@ public	int	getPrimaryKeyIndexNumber() {
                   null,null,view,viewId,0,0,0),
                 new ColumnDescriptor("SCHEMANAME"               ,15,15,
                         DataTypeDescriptor.getBuiltInDataTypeDescriptor(Types.VARCHAR, false, 128),
+                        null,null,view,viewId,0,0,0),
+                new ColumnDescriptor("SKETCHSIZE"               ,8,8,
```
ascend1, line 519: I think we should have `16,16` here because column 8 is already `AUTOINCREMENTVALUE`.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1203,7 +1202,7 @@ public void addMinRetentionPeriodColumn(TransactionController tc) throws Standar
                     SYSTABLESRowFactory.MIN_RETENTION_PERIOD));
 
             // now upgrade the views if necessary
-            createOrUpdateSystemView(tc, "SYSVW", SYSTABLESRowFactory.SYSTABLE_VIEW_NAME);
+            createOrUpdateSystemView(tc, "SYSVW", "SYSCOLUMNSVIEW");
```
ascend1, line 1205: Was this a bug? Great that you found it.

The `SYSCOLUMNVIEW` name is probably also available from `SYSCOLUMNSRowFactory`. Can you also check it? We want to avoid hard coded names as much as possible.

ascend1 (changes requested):

arnaud-splice (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
arnaud-splice: Good point, it's code that was moved from `UpgradeScriptForTriggerMultipleStatements.java` though, so I would not change it in the frame of this PR

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
arnaud-splice: Tracked by DB-11947

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
arnaud-splice: Tracked by DB-11947

arnaud-splice (approved):

jyuanca (approved):

arnaud-splice:

ascend1 (approved):

carolp-503 (approved):
+1

PR #5401, opened 2021-04-19
DB-11895 DB-11899 fixes for add column behavior

hatyo:
## Short Description
In this PR we fix two issues:
* DB-11895 NPE when adding a column with a foreign key reference.
* DB-11899 Incorrectly allowing `null` default value even if the newly added column is not-nullable.

## Long Description
The first issue was related to having a code path that attempts to add a foreign key when a new column is added without initializing the FK graph properly before starting the FK validation process. The code path was not tested correctly, now it is fixed and new ITs were added to cover it.

The second issue was related to a missing check when validating a default value for a newly added column, a check was added and an IT.

## How to test

For DB-11895:
```
create table p(c1 int primary key);
create table c(c1 int);
insert into p values 1,2,3;
insert into c values 100;
alter table c add column c2 int references p(c1); -- previously it gave NPE, now it should work.
```

For DB-11899:
```
create table t1(c1 int);
alter table t1 add column c2 int not null default null; –- previously it worked, now it should fail with a proper error message
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnDefinitionNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/ColumnDefinitionNode.java
@@ -557,6 +557,12 @@ public class ColumnDefinitionNode extends TableElementNode
     void validateDefault(DataDictionary dd, TableDescriptor td)
         throws StandardException
     {
+        if(defaultValue != null && defaultValue.isNull() && type != null && !type.isNullable()) {
+            throw StandardException.newException(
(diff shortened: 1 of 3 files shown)


hatyo:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2777/

hatyo:

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

ipraznik-splice (approved):

arnaud-splice (approved):

PR #5407, opened 2021-04-21
DB-11534: Filter out rolled back txn during WW-conflict detection

jyuanca:
## Short Description
We only check WW-conflict with latest version. In case 

- [ ]  @latest version is rolled back, WW-conflict check can miss a conflict or constraint violation with previous version.

## Long Description
We only check WW-conflict with latest version. In case latest version is rolled back, WW-conflict check can miss a conflict or constraint violation with previous version. 

Here is an example.

        Session 0                                               Session 1
        autocommit off;                                    autocommit off;
        create table a(i int primary key, j int);   
        insert into a values (1,1);                        
        commit;                                                   
        update a set j= j+ 1;                                
        (sleep 20s)                                              
        rollback;                                                 commit;
                                                             insert into a values (1,5);
                                                             commit;
For row (1,5), splice checks WW-conflict with row (1,2) and finds it is rolled back. Splice does not try to check ww-conflict with older versions in this case, and missed PK violation between (1,5) and (1,1). 

To fix the problem, a filter is added to filter out all rolled back rows when getting latest version for the row.  

## How to test
Run previous test to confirm it is fixed. Run Jenkins elle-list test.

### DIFF
--- a/hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FailuresSparkIT.java
+++ b/hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FailuresSparkIT.java
@@ -93,6 +93,7 @@ public class FailuresSparkIT {
     @Test
     public void testPKViolationIsRolledback() throws Throwable {
         try(Statement s =methodWatcher.getOrCreateConnection().createStatement()) {
+            s.execute("delete from a");
             String sql = "insert into a (col1) select * from ( " +
                     "values (1) union all values (1) " +
                     "union all select * from a --splice-properties useSpark=true\n" +
@@ -129,15 +130,8 @@ public class FailuresSparkIT {
                     }
                     con1.rollback();
 
-                    // force WW conflict to rollfoward (by deleting) the previous write
-                    try {
-                        s2.executeUpdate("insert into a values 10");
-                        fail("Expected WW conflict");
-                    } catch (SQLException se) {
-                        assertEquals("SE014", se.getSQLState());
-                    }
-
-
+                    // insert a row to roll forward (by deleting) the previous write
+                    s2.executeUpdate("insert into a values 10");
                     con2.commit();
 
 
--- a/hbase_storage/src/main/java/com/splicemachine/storage/RangedClientPartition.java
+++ b/hbase_storage/src/main/java/com/splicemachine/storage/RangedClientPartition.java
@@ -21,6 +21,7 @@ import com.splicemachine.kvpair.KVPair;
 import com.splicemachine.metrics.MetricFactory;
 import com.splicemachine.primitives.ByteComparator;
 import com.splicemachine.primitives.Bytes;
+import com.splicemachine.si.api.txn.TxnSupplier;
 import com.splicemachine.si.data.HExceptionFactory;
 import com.splicemachine.si.impl.HNotServingRegion;
 import com.splicemachine.si.impl.HRegionTooBusy;
@@ -36,6 +37,7 @@ import org.apache.log4j.Logger;
 
 import java.io.IOException;
 import java.io.InterruptedIOException;
+import java.nio.charset.Charset;
 import java.util.BitSet;
 import java.util.Collections;
 import java.util.Iterator;
@@ -110,7 +112,7 @@ public class RangedClientPartition implements Partition, Comparable<RangedClient
     public void compact(boolean isMajor) throws IOException {
         Connection connection = HBaseConnectionFactory.getInstance(HConfiguration.getConfiguration()).getConnection();
         HExceptionFactory exceptionFactory = HExceptionFactory.INSTANCE;
-        byte[] encodedRegionName = regionInfo.getEncodedName().getBytes();
+        byte[] encodedRegionName = regionInfo.getEncodedName().getBytes(Charset.defaultCharset().name());
 
         try(Admin admin = connection.getAdmin()){
             if (isMajor)
@@ -150,7 +152,7 @@ public class RangedClientPartition implements Partition, Comparable<RangedClient
     @Override
     public void flush() throws IOException {
         Connection connection = HBaseConnectionFactory.getInstance(HConfiguration.getConfiguration()).getConnection();
-        byte[] encodedRegionName = regionInfo.getEncodedName().getBytes();
+        byte[] encodedRegionName = regionInfo.getEncodedName().getBytes(Charset.defaultCharset().name());
         try(Admin admin = connection.getAdmin()) {
             admin.flushRegion(encodedRegionName);
         }
@@ -297,8 +299,8 @@ public class RangedClientPartition implements Partition, Comparable<RangedClient
     }
 
     @Override
-    public DataResult getLatest(byte[] key, DataResult previous) throws IOException {
-        return delegate.getLatest(key, previous);
+    public DataResult getLatest(byte[] key, DataResult previous, Object obj) throws IOException {
+        return delegate.getLatest(key, previous, obj);
     }
 
     @Override
--- a/hbase_storage/src/main/java/com/splicemachine/storage/RegionPartition.java
+++ b/hbase_storage/src/main/java/com/splicemachine/storage/RegionPartition.java
@@ -14,9 +14,10 @@
 
 package com.splicemachine.storage;
 
+import com.splicemachine.si.impl.*;
+import com.splicemachine.si.impl.server.ConflictRollForward;
 import org.apache.hadoop.hbase.exceptions.ConnectionClosingException;
 import splice.com.google.common.base.Function;
-import com.splicemachine.si.impl.HRegionTooBusy;
 
 import org.apache.hadoop.hbase.DroppedSnapshotException;
 import splice.com.google.common.collect.Iterators;
@@ -24,8 +25,6 @@ import com.splicemachine.kvpair.KVPair;
 import com.splicemachine.metrics.MetricFactory;
 import com.splicemachine.metrics.Metrics;
 import com.splicemachine.si.constants.SIConstants;
-import com.splicemachine.si.impl.HNotServingRegion;
-import com.splicemachine.si.impl.HWrongRegion;
 import com.splicemachine.storage.util.MeasuredListScanner;
 import com.splicemachine.utils.Pair;
 import org.apache.hadoop.hbase.NotServingRegionException;
@@ -44,7 +43,7 @@ import java.util.concurrent.locks.Lock;
  * A lot of these methods shouldn't (and generally won't) be called directly, because they will
  * result in an excessive number of objects being created. For example, instead of relying on
  * {@link #get(DataGet,DataResult)} to get the latest version of all the columns, instead call
- * {@link #getLatest(byte[],DataResult)}, which will save on the creation of a DataGet object, while still
+ * {@link #getLatest(byte[],DataResult, Object)}, which will save on the creation of a DataGet object, while still
  * allowing re-use of the DataResult object.
  * <p/>
  * Nonetheless, all methods are properly implemented in a Thread-safe manner (they must, in order to pass SI
@@ -176,9 +175,14 @@ public class RegionPartition implements Partition{
     }
 
     @Override
-    public DataResult getLatest(byte[] key,DataResult previous) throws IOException{
+    public DataResult getLatest(byte[] key,DataResult previous, Object obj) throws IOException{
         Get g=new Get(key);
-        g.setMaxVersions(1);
+        g.readVersions(1);
+
+        assert obj instanceof ConflictRollForward;
+        ConflictRollForward conflictRollForward = (ConflictRollForward)obj;
+        RollbackTxnFilter filter = new RollbackTxnFilter(conflictRollForward.getTxnSupplier(), conflictRollForward.getMutations());
+        g.setFilter(filter);
 
         try{
             Result result=region.get(g);
--- a/mem_storage/src/main/java/com/splicemachine/storage/MPartition.java
+++ b/mem_storage/src/main/java/com/splicemachine/storage/MPartition.java
@@ -16,6 +16,12 @@ package com.splicemachine.storage;
 
 import com.splicemachine.access.util.ByteComparisons;
 import com.splicemachine.si.api.filter.TxnFilter;
+import com.splicemachine.si.api.txn.Txn;
+import com.splicemachine.si.api.txn.TxnStore;
+import com.splicemachine.si.api.txn.TxnSupplier;
+import com.splicemachine.si.api.txn.TxnView;
+import com.splicemachine.si.impl.driver.SIDriver;
+import com.splicemachine.si.impl.server.ConflictRollForward;
 import splice.com.google.common.base.Predicate;
 import splice.com.google.common.collect.BiMap;
 import splice.com.google.common.collect.HashBiMap;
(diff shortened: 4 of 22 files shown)


jyuanca:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2786/

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public Filter.ReturnCode filterCell(Cell keyValue){
+        DataCell data = new HCell(keyValue);
+        long txnId = data.version();
+        try {
+            SIDriver driver = SIDriver.driver();
+            if (driver != null) {
+                TxnSupplier supplier = driver.getTxnSupplier();
+                if (supplier != null) {
+                    TxnView txn = supplier.getTransaction(txnId);
```
dgomezferro, line 65: One issue here is that we only have the CompletedTxnCache, not an ActiveTxnCache. If we have a transaction that wrote 10K records and then updates them, each update will try to resolve the active transaction independently, becoming very costly. Would it be possible to pass in a `txnSupplier` into the filter so that we can reuse an ActiveTxnCache?

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -298,33 +300,64 @@ private void releaseLocksForKvBatch(Pair<KVPair, Lock>[] locks){
...
+                            if (ioe instanceof WriteConflict) {
+                                finalStatus[i] = operationStatusLib.failure(ioe);
+                                shouldFail = true;
+                                break;
+                            } else throw ioe;
+                        }
+
+                        if (constraintChecker != null) {
+                            List<DataCell> visibleColumns = filterCells(possibleConflicts, constraintStateFilter);
+                            if (visibleColumns.size() == 0 && !conflictResults.hasAdditiveConflicts()) {
```
dgomezferro, line 330: Given that we already filtered out the Rolledback txns I think we don't need to filter them out again. If some active txn just became rolledback, so be it, raising a WW conflict (or PK violation) is fine.

dgomezferro (changes requested):

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,112 @@
...
+                    if (txn != null) {
```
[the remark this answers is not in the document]
jyuanca: They could be null for Unit tests.

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
jyuanca: That will report some false positives

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FailuresSparkIT.java
```diff
@@ -129,15 +130,8 @@ public void testDeletedCellsInMemstoreDontHangSparkScans() throws Throwable {
                     }
                     con1.rollback();
 
-                    // force WW conflict to rollfoward (by deleting) the previous write
```
ipraznik-splice, line 132: The IT code to test a WW conflict is removed here. Should another test be added for this case, or it's not needed any more?

ipraznik-splice (changes requested):

jyuanca:

hbase_sql/src/test/java/com/splicemachine/derby/impl/sql/execute/operations/FailuresSparkIT.java
jyuanca: The purpose was to rollforward data(removing a rolled back row). Code change for this PR already handle that. When the RollbackTxnFiter finds a rolled back row, it rolls it forward(delete it)

spliceci:
TEST SUCCEEDED +1

ipraznik-splice (approved):

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
```diff
@@ -261,7 +261,8 @@ private void releaseLocksForKvBatch(Pair<KVPair, Lock>[] locks){
         }
     }
 
-    private IntObjectHashMap<DataPut> checkConflictsForKvBatch(Partition table,
+    private IntObjectHashMap<DataPut> checkConflictsForKvBatch(TxnView txn,
```
dgomezferro, line 264: It looks like this new parameter is not used, is it?

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
dgomezferro: I don't understand this, if we've already filtered out rolledback rows we don't need to check again. Do you have an example for those false positives?

hbase_storage/src/main/java/com/splicemachine/storage/RegionPartition.java
```diff
@@ -187,6 +189,10 @@ public DataResult getLatest(byte[] key,DataResult previous) throws IOException{
             else{
                 ((HResult)previous).set(result);
             }
+            List<DataMutation> mutations = filter.getMutations();
+            if (mutations.size() > 0) {
+                batchMutate(mutations);
+            }
```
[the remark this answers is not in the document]
jyuanca: Due to module dependencies, it's hard to return these mutations back to SITransactor

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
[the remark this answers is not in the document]
jyuanca: Removed some of them

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: Passed in a ActiveTxnSupplier

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
jyuanca: It could happen that the latest row returned was rolled back after it was returned. Either we report a WW conflict or get next row that is visible to check WW conflict.

hbase_storage/src/main/java/com/splicemachine/storage/RegionPartition.java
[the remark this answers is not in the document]
jyuanca: jenkins please test branch @cdh6.3.0,skipTestsLongerThan2Minutes

hatyo (changes requested):

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,102 @@
...
+import com.splicemachine.utils.SpliceLogUtils;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.FilterBase;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.log4j.Logger;
+
+import java.util.List;
+
+public class RollBackTxnFilter extends FilterBase {
```
hatyo, line 36: Why are we inheriting from `org.apache.hadoop.hbase.filter.FilterBase` instead of `com.splicemachine.si.api.filter.TxnFilter`?

hatyo, line 36: very minor, I would name the class `RollbackTxnFilter` instead of `RollBackTxnFilter` .

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,102 @@
...
+                    bs.set(keyValue.getRowArray(), keyValue.getRowOffset(), keyValue.getRowLength());
+                    DataDelete delete = opFactory.newDelete(bs);
+                    delete.deleteColumn(SIConstants.DEFAULT_FAMILY_BYTES,SIConstants.FK_COUNTER_COLUMN_BYTES, keyValue.getTimestamp());
+                    delete.deleteColumn(SIConstants.DEFAULT_FAMILY_BYTES,SIConstants.FIRST_OCCURRENCE_TOKEN_COLUMN_BYTES, keyValue.getTimestamp());
+                    delete.deleteColumn(SIConstants.DEFAULT_FAMILY_BYTES,SIConstants.PACKED_COLUMN_BYTES, keyValue.getTimestamp());
+                    delete.deleteColumn(SIConstants.DEFAULT_FAMILY_BYTES,SIConstants.TOMBSTONE_COLUMN_BYTES, keyValue.getTimestamp());
+                    delete.deleteColumn(SIConstants.DEFAULT_FAMILY_BYTES,SIConstants.COMMIT_TIMESTAMP_COLUMN_BYTES, keyValue.getTimestamp());
+                    delete.addAttribute(SIConstants.SUPPRESS_INDEXING_ATTRIBUTE_NAME,SIConstants.SUPPRESS_INDEXING_ATTRIBUTE_VALUE);
+                    delete.setDurability(Durability.NONE);
+                    mutations.add(delete);
```
hatyo, line 78: this duplicates the logic in `com.splicemachine.si.impl.server.ConflictRollForward#handle`, instead:
- pass in the `ConflictRollForward` object to this class constructor (instead of the `mutation`)
- let it generate the deletion mutation itself by calling `conflictRowForward.handle(data)`.

IMHO this makes the intention clearer and reduces code duplication.

hatyo (changes requested):

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+            return ReturnCode.INCLUDE;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public List<DataMutation> getMutations() {
+        return mutations;
+    }
```
hatyo, line 101: is this used?

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
```diff
@@ -0,0 +1,102 @@
...
+                        byte[] k = data.key();
+                        byte[] v = data.value();
+                        SpliceLogUtils.debug(LOG, "Include %s, k=%s, v=%s", type.toString(),
+                                Bytes.toStringBinary(k), Bytes.toStringBinary(v));
+
+                    }
+                    return ReturnCode.INCLUDE;
+                }
+            }
+            return ReturnCode.INCLUDE;
```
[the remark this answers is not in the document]
jyuanca: it is needed because of 
if (txnSupplier != null)

hbase_storage/src/main/java/com/splicemachine/storage/RegionPartition.java
```diff
@@ -176,9 +176,14 @@ public DataResult getFkCounter(byte[] key,DataResult previous) throws IOExceptio
     }
 
     @Override
-    public DataResult getLatest(byte[] key,DataResult previous) throws IOException{
+    public DataResult getLatest(byte[] key,DataResult previous, Object obj) throws IOException{
         Get g=new Get(key);
-        g.setMaxVersions(1);
+        g.readVersions(1);
+
+        assert obj instanceof ConflictRollForward;
```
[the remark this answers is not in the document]
jyuanca: Due to module dependencies, ConflictRollForward  is not visible. We may need some refactoring to make this clean

splice_si_api/src/main/java/com/splicemachine/si/impl/server/SITransactor.java
jyuanca: removed

hbase_storage/src/main/java/com/splicemachine/storage/RangedClientPartition.java
```diff
@@ -297,8 +299,8 @@ public DataResult getFkCounter(byte[] key, DataResult previous) throws IOExcepti
     }
 
     @Override
-    public DataResult getLatest(byte[] key, DataResult previous) throws IOException {
-        return delegate.getLatest(key, previous);
+    public DataResult getLatest(byte[] key, DataResult previous, Object obj) throws IOException {
```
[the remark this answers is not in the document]
jyuanca: I'll document it in Partition interface

mem_storage/src/main/java/com/splicemachine/storage/MPartition.java
```diff
@@ -604,4 +620,8 @@ public BitSet getBloomInMemoryCheck(boolean hasConstraintChecker, Pair<KVPair, L
     public PartitionDescriptor getDescriptor() throws IOException {
         throw new UnsupportedOperationException("Operation not supported in mem storage engine");
     }
+
+    public void setTxnStore(TxnStore txnStore) {
+        this.txnStore = txnStore;
+    }
```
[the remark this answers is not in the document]
jyuanca: It's no longer needed. I'll remove it.

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: removed

jyuanca:

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: Renamed

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: It's not exactly same as ConflictRollForward#handle

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: It is an HBase filter

spliceci:
TEST SUCCEEDED +1

hbase_storage/src/main/java/com/splicemachine/si/impl/RollBackTxnFilter.java
jyuanca: TxnFilter does not work with a HBase Get.

hatyo (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #5411, opened 2021-04-21
DB-11449 Unioned Index Scans AccessPath

msirek:
## Short Description
Optimize queries of the form:
select * from t1 where index1_col1 = 1 or index2_col1 = 'a' or index3_col1 >= 99.0;

by scanning index1, index2 and index3 in separate scans, one predicate per scan, then UNIONing the matching base table RowIDs together and joining the result back to the base table to get the final result set.  This is called a Unioned Index Scans access path, because we are unioning different scans against different indexes (or the primary key).  If the predicates are highly selective and the base table is large, we bypass reading many rows.

## Long Description
Consider a table, t1:
create table t1 (a int, b int, primary key(a,b));
create index t1_idx2 on t1(b);

... and a query,
 select * from t1 where a=1 or b=2;

Column a is the leading column of the primary key and column b is the leading column of the index.  So, considering a=1 alone, we could do a keyed PK access, and considering b=2 alone we could do a keyed index access.  The Unioned Index Scans plan for this query is:

```
Cursor(n=18,rows=720,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=18,totalCost=0,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=16,totalCost=4.04,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=14,totalCost=451.876,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
        ->  NestedLoopJoin(n=12,totalCost=451.876,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
          ->  ProjectRestrict(n=11,totalCost=4.04,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
            ->  TableScan[T1(2096)](n=10,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1,keys=[(dnfPathDT_###_T1.BASEROWID[9:1] = T1.BASEROWID[11:3])])
          ->  Distinct(n=8,totalCost=4.036,outputRows=36,outputHeapSize=54 B,partitions=1,parallelTasks=1)
            ->  Union(n=6,totalCost=4.036,outputRows=36,outputHeapSize=54 B,partitions=2,parallelTasks=1)
              ->  ProjectRestrict(n=5,totalCost=4.027,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=4,totalCost=4.027,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
                  ->  IndexScan[T1_IDX2(2113)](n=3,totalCost=4.027,scannedRows=18,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,baseTable=T1(2096),keys=[(B[3:1] = 2)])
              ->  ProjectRestrict(n=2,totalCost=4.036,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=1,totalCost=4.036,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1)
                  ->  TableScan[T1(2096)](n=0,totalCost=4.036,scannedRows=18,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1,keys=[(A[0:1] = 1)])
```

This feature modifies the explain text to indicate 'keys=' to signify any predicates which cause an index access which skips over rows.  Normal qualifier predicates are still indicated as 'preds='.
It also adds the BASEROWID derived column name.  If you query a table, 'select BASEROWID from t1;' this will return back the rowids of the base table rows, whether or not the scan uses an index to satisfy the query (the base table rowids are stored in the index).  Derived column name ROWID is pre-existing, but it may return back the rowid of the base table row, or the index row, depending on whether the base table or index was scanned (or if an IndexLookup operation was done).
Note in the explain the BASEROWID = BASEROWID join between the UNION derived table and the base table T1.  This predicate is also indicated as a key, meaning the join back to T1 looks up those base rowids directly instead of scanning the entire T1.

This is a cost-based optimization, meaning when costing table T1 as a single-table scan, or as the inner table of a join, we will first cost the scan/join using the normal method, then we look for any OR'ed predicates that may enable UIS (Unioned Index Scans) access path, and cost the scan/join with UIS access path.  The cheaper of the two is selected.  

DNF to CNF conversion of predicates previously does not apply the distributive property to derive new terms.  This is being added by this PR.  For example, given the query:

```
select * from BH, BU where ( BH.a = 1 AND BH.b = BU.a ) 
                     OR ( BH.a = BU.a AND BH.b = 1);
```

We can convert to CNF as follows:

```
(BH.a = 1 OR BH.a = BU.a)   AND 
(BH.a = 1 OR BH.b = 1)     AND 
(BH.b = BU.a OR BH.a = BU.a) AND 
(BH.b = BU.a OR BH.a = BU.a)
```

This conversion enables two separate UIS access paths on the two middle terms.  
First, (BH.a = 1 OR BH.b = 1)  can be applied to table BH in two scans.
Next, (BH.b = BU.a OR BH.a = BU.a)  can be applied between the result from above in 2 unioned joins with these index-enabling join predicates.

The number of predicate we can derive in this manner is limited to 100, by default, but may be tuned lower or higher (as high as 10000) using a new database property, splice.optimizer.maxDerivedCNFPredicates:

> call syscs_util.syscs_set_global_database_property('splice.optimizer.maxDerivedCNFPredicates', '200');

A CloneCRsVisitor is added to clone any BinaryRelationalOperatorNodes and underlying ColumnReferences during DNF to CNF term expansion.  This is required because we don't want remapping of a ColumnReference in one predicate to also remap a ColumnReference in a copied operator in another predicate.  Each predicate can be pushed separately so must not share memory.  If this cloning operation fails, so does DNF to CNF conversion because it is unsafe to proceed using shared memory in separate predicates.  Currently not many operators support cloning, so building up cloning support could extend CNF conversion potential to the full range of boolean expressions Splice supports.

UIS access paths can be disabled via database property splice.optimizer.disableUnionedIndexScans:
> call syscs_util.syscs_set_global_database_property('splice.optimizer.disableUnionedIndexScans', true);

UIS access paths can be favored by the optimizer for testing purposes via database property splice.optimizer.favorUnionedIndexScans:
> call syscs_util.syscs_set_global_database_property('splice.optimizer.favorUnionedIndexScans', true);

ROWID and BASEROWID cannot be column names in CREATE TABLE or RENAME statements, and will result in an error.

When costing UIS access paths, we first separate OR'ed predicates into separate lists and categorize them one-at-a-time.  If each separate predicate is either a start key or stop key, we attempt UIS access path.  The UNION tree which collects rowids and removes any duplicates is built up by function buildUnionedScans.  The FromBaseTable cloned objects in this tree are marked as indexFriendlyJoinsOnly, so that only nested loop join or merge join are used, to skip over unqualified join rows.
If there are no errors, an AccessPath object is returned from buildUnionedScans with the UNION derived table to optimize.  bindAndOptimizeUnionedIndexScansPath is then called, which builds a BASEROWID = BASEROWID join predicate between the base table and the UNION derived table and constructs a new SELECT statement:  

> SELECT (original_T1_projection_list) from T1, UNION_derived_table where T1.BASEROWID = UNION_derived_table.BASEROWID;

This statement is then bound and optimized (bindStatement and optimizeStatement are called).  
If the table is being joined with an outer table, the UNION_derived_table will include that join, but also we need to join the entire result back to the outer table in the main join.  buildRowIdPredWithOuterTable is called to generate a UNION_derived_table_plus_RowID_Join.BASEROWID2 = outer_table.BASEROWID predicate to match the proper rows.  It is possible to eliminate this join entirely by manually manipulating the list of optimizables at the end of join planning, but this may be a high-risk change, so is not being done in the first pass of this feature.  The UIS derived table plus rowid join result set is marked as skipBindAndOptimize, which is a new flag to indicate no more binding or optimization should be done on the result set if it is joined or referenced elsewhere in the AST.  It is also marked as outerTableOnly so it is only considered as the outer table of any join it participates in (since it has a reduced set of RowIDs, we want to drive the join from these instead of scan all RowIDs in the base table).

## How to test
create table t1 (a int, b int, primary key(a,b));
create index t1_idx2 on t1(b);
call syscs_util.syscs_set_global_database_property('splice.optimizer.favorUnionedIndexScans', 'true');
// The following should contain 'BASEROWID' in the explain output:
explain select * from t1 where a=1 or b=2;

### DIFF
--- a/db-client/src/main/java/com/splicemachine/db/client/am/ResultSet.java
+++ b/db-client/src/main/java/com/splicemachine/db/client/am/ResultSet.java
@@ -39,6 +39,8 @@ import java.sql.Timestamp;
 import java.util.Arrays;
 import java.util.Calendar;
 
+import static com.splicemachine.db.client.am.LOBStateTracker.NO_OP_TRACKER;
+
 @SuppressFBWarnings(value = {"NM_SAME_SIMPLE_NAME_AS_INTERFACE", "URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD"})
 public abstract class ResultSet implements java.sql.ResultSet,
         ResultSetCallbackInterface {
@@ -241,6 +243,9 @@ public abstract class ResultSet implements java.sql.ResultSet,
         resultSetHoldability_ = resultSetHoldability;
         fetchDirection_ = statement_.fetchDirection_;
         suggestedFetchSize_ = statement_.fetchSize_;
+        // Making sure the following is always initialized avoids NullPointerExceptions
+        // that sometimes hide the original Exception.
+        lobState = NO_OP_TRACKER;
 
         maxRows_ = statement_.maxRows_;
         
@@ -6220,7 +6225,7 @@ public abstract class ResultSet implements java.sql.ResultSet,
      */
     final void createLOBColumnTracker() {
         if (SanityManager.DEBUG) {
-            SanityManager.ASSERT(this.lobState == null,
+            SanityManager.ASSERT(this.lobState == null || this.lobState == NO_OP_TRACKER,
                     "LOB state tracker already initialized.");
         }
         if (this.connection_.supportsSessionDataCaching() &&
@@ -6244,7 +6249,7 @@ public abstract class ResultSet implements java.sql.ResultSet,
             this.lobState = new LOBStateTracker(lobIndexes, isBlob, true);
         } else {
             // Use a no-op state tracker to simplify code expecting a tracker.
-            this.lobState = LOBStateTracker.NO_OP_TRACKER;
+            this.lobState = NO_OP_TRACKER;
         }
     }
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
@@ -36,6 +36,9 @@ import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
+import com.splicemachine.db.impl.sql.compile.FromTable;
+import com.splicemachine.db.impl.sql.compile.Predicate;
+import com.splicemachine.db.impl.sql.compile.ResultSetNode;
 
 /**
  * AccessPath represents a proposed access path for an Optimizable.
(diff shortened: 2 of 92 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2810/

msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2817/

msirek:
Github is not seeing my latest force-push.  Closing this PR and will open a new one.

PR #5414, opened 2021-04-22
DB-11919 pluggable join estimation cost model

hatyo:
In this PR we make join cost estimation pluggable such that we can set dynamically at runtime.

## Short Description

Currently the join cost estimation logic is hard coded in the different `JoinStrategy` classes, this makes it difficult to tweak the cost model and fix potential bugs, i.e. for that we need basically to recompile and use a new version of SpliceMachine with the fixes.

With this PR we make it possible to dynamically switch the join cost estimation model via session properties allowing us to easily alter the behavior of the optimizer to fix a bug, test a new cost estimation model, or simply regress to old behaviors, it is slightly similar to SQLServer's [Compatibility Level](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-compatibility-level?view=sql-server-ver15) feature. 

## Long Description

In this PR we introduce mainly a interface `JoinCostEstimationModel`, the interface is pretty simple, it has one method `estimateCost` that, for a given join strategy, calculates the inner and outer join costs.

Implementations of this interface are registered in a `JoinCostEstimationModelRegistry`, which is basically a static hash map that keeps track of all join cost models in the system.

The current join cost estimation logic was spread across different `JoinStrategy` implementations, the logic is refactored and made accessible in a new `V1JoinCostEstimationModel : JoinCostEstimationModel` implementation that is registered under the name "v1" in the `JoinCostEstimationModelRegistry`.

When the database boots up, we instantiate a new object of type `V1JoinCostEstimationModel` and we use it for future (concurrent) invocations of the `Optimizer` instances (which is fine, since cost estimation logic is stateless).

A new session property is added `joinCostEstimationModel`, this property allows us to specify the join cost estimation model to use for the current session. With this PR, we have only "v1" available, but in the future we could have more.

### Outlook

There are multiple things that we improve:

- add a new stored procedure that allows the user to dynamically register a new join cost estimation model at runtime, similar to dynamically loading a VTI, by uploading the estimation model's code to the cluster and making it visible to the class loader, then simply registering it in the `JoinCostEstimationModelRegistry`, this gives maximum flexibility in tweaking cost estimation model without recompilation.
- currently we only cover join cost estimation, we could also add scan cost estimation models and selectivity estimation algorithms making them also pluggable.

## How to test

This PR does not introduce nor alter any behavior, it just refactors the join cost estimation model and makes it pluggable, there is a new session property that allows us to specify the model, we can set it as the following:

```
set session_property joinCostEstimationModel='v1';
```

or in the connection string:

```
connect 'jdbc:splice://localhost:1527/splicedb;user=<USER>;password=<PASSWORD>;joinCostEstimationModel=v1';
```

once set, all queries in that session will be optimized from now on using the model 'v1', we could easily verify this in the future once we have more cost estimation models in place.


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MergeSortJoinOperation.java
```diff
@@ -198,7 +198,7 @@ public String prettyPrint(int indentLevel) {
             SpliceLogUtils.debug(LOG, "getDataSet Performing MergeSortJoin type=%s, antiJoin=%s, hasRestriction=%s",
                     getJoinTypeString(), isAntiJoin(), restriction != null);
         }
-        rightDataSet1.map(new CountJoinedRightFunction(operationContext));
+        //rightDataSet1.map(new CountJoinedRightFunction(operationContext));
```
msirek, line 201: What's the reason for commenting this out?

msirek (approved):

arnaud-splice (changes requested):
+1 once Mark's comment has been addressed

Fantastic work!

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/MergeSortJoinOperation.java
hatyo: That was a left-over from some debugging session. Thanks for catching it, I will remove it and push a new commit.

arnaud-splice (approved):

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5419, opened 2021-04-23
DB-11907 Make FakeColumnStats selectivity should return rowsPerValue

dgomezferro:
## Short Description
FakeColumnStats were wrong in some cases, specifically for selectivity of columns with a default value

## How to test
- Create a table with a column with default value
- Set fake column statistics with some rowsPerValue
- Prepare statement with equality predicate on that column, the estimated scanned rows should be rowsPerValue

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
@@ -134,7 +134,7 @@ public class FakeColumnStatisticsImpl extends ColumnStatisticsImpl implements Ex
 
     @Override
     public long selectivityExcludingValueIfSkewed(DataValueDescriptor value) {
-        return totalCount;
+        return rpv;
     }
 
     @Override
--- a/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/JoinOrderJumpModeIT.java
+++ b/splice_machine/src/test/java/com/splicemachine/derby/impl/sql/compile/JoinOrderJumpModeIT.java
@@ -247,14 +247,14 @@ public class JoinOrderJumpModeIT extends SpliceUnitTest {
                 new String[] {"Join"},                                                        // 14
                 new String[] {"Scan["},                                                       // 16, IndexScan on mem but TableScan on cdh
                 new String[] {"Join"},                                                        // 17
-                new String[] {"TableScan[TABLE_5", "scannedRows=800,outputRows=800"},         // 18
+                new String[] {"TableScan[TABLE_5", "scannedRows=1,outputRows=1"},             // 18
                 new String[] {"Join"},                                                        // 19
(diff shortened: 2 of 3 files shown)


dgomezferro:

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
```diff
@@ -134,7 +134,7 @@ public void update(DataValueDescriptor item) {
 
     @Override
     public long selectivityExcludingValueIfSkewed(DataValueDescriptor value) {
-        return totalCount;
+        return rpv;
```
msirek, line 137: Selectivity should be in units of values per row (fraction of rows all having the same value) and be between 0 and 1 as it's multiplied by number of rows (rows on top and bottom should cancel out):

> scanCost.setEstimatedRowCount(Math.round(totalRowCount*totalSelectivity));

Perhaps we should be using 1/rpv here.  I haven't looked in detail though.  What exactly does selectivityExcludingValueIfSkewed aim to estimate?  There are also methods selectivity and rangeSelectivity in this file which return rpv.  They should be updated as well.  Also, rangeSelectivity has:

        // Use Null Data
        if (element == null || element.isNull())
            return nullCount;

I think this should return nullCount/totalCount

msirek (changes requested):

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/stats/FakeColumnStatisticsImpl.java
dgomezferro: The result of this method is being divided by the rowcount, so we are eventually computing `1/cardinality`. Perhaps the name of the method is wrong, here's the definition in `ItemStatistics.java`:
`    /**
     * get the row count excluding the specified value
     * @param value
     * @return
     */
    long selectivityExcludingValueIfSkewed(T value);`

This seems wrong too, because the 'real' statistics implementation is returning `return (long) (((double)nonSkewCount)/nonSkewedNum);` which is basically `rpv` excluding skewed values.

I think a better name would be `rowsPerValueExcludingValueIfSkewed` , but the other methods you mentioned also seem to be using `selectivity` in the same way this method is, should we update all of them?

msirek: @dgomezferro I see.  All of these estimates eventually get divided by rowCount in a higher-level method.  I think it is confusing that the lower-level methods are named the same, because they are not really returning selectivity.  They are returning rows per value.  So, I think your fix is correct and those other methods don't need updates, but we might want to consider renaming them at some point, replacing "selectivity" with "rowsPerValue".

msirek (approved):

ascend1 (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5420, opened 2021-04-23
DB-11449 Unioned Index Scans AccessPath

msirek:
## Short Description
Optimize queries of the form:
`select * from t1 where index1_col1 = 1 or index2_col1 = 'a' or index3_col1 >= 99.0;`

by scanning index1, index2 and index3 in separate scans, one predicate per scan, then UNIONing the matching base table RowIDs together and joining the result back to the base table to get the final result set.  This is called a Unioned Index Scans access path, because we are unioning different scans against different indexes (or the primary key).  If the predicates are highly selective and the base table is large, we bypass reading many rows.

## Long Description
Consider a table, t1:
```
create table t1 (a int, b int, primary key(a,b));
create index t1_idx2 on t1(b);
```

... and a query,
` select * from t1 where a=1 or b=2;`

Column a is the leading column of the primary key and column b is the leading column of the index.  So, considering a=1 alone, we could do a keyed PK access, and considering b=2 alone we could do a keyed index access.  The Unioned Index Scans plan for this query is:

```
Cursor(n=18,rows=720,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=18,totalCost=0,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=16,totalCost=4.04,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=14,totalCost=451.876,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
        ->  NestedLoopJoin(n=12,totalCost=451.876,outputRows=720,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
          ->  ProjectRestrict(n=11,totalCost=4.04,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1)
            ->  TableScan[T1(2096)](n=10,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=3.164 KB,partitions=1,parallelTasks=1,keys=[(dnfPathDT_###_T1.BASEROWID[9:1] = T1.BASEROWID[11:3])])
          ->  Distinct(n=8,totalCost=4.036,outputRows=36,outputHeapSize=54 B,partitions=1,parallelTasks=1)
            ->  Union(n=6,totalCost=4.036,outputRows=36,outputHeapSize=54 B,partitions=2,parallelTasks=1)
              ->  ProjectRestrict(n=5,totalCost=4.027,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=4,totalCost=4.027,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
                  ->  IndexScan[T1_IDX2(2113)](n=3,totalCost=4.027,scannedRows=18,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,baseTable=T1(2096),keys=[(B[3:1] = 2)])
              ->  ProjectRestrict(n=2,totalCost=4.036,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=1,totalCost=4.036,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1)
                  ->  TableScan[T1(2096)](n=0,totalCost=4.036,scannedRows=18,outputRows=18,outputHeapSize=36 B,partitions=1,parallelTasks=1,keys=[(A[0:1] = 1)])
```

This feature modifies the explain text to indicate 'keys=' to signify any predicates which cause an index access which skips over rows.  Normal qualifier predicates are still indicated as 'preds='.
It also adds the BASEROWID derived column name.  If you query a table, 'select BASEROWID from t1;' this will return back the rowids of the base table rows, whether or not the scan uses an index to satisfy the query (the base table rowids are stored in the index).  Derived column name ROWID is pre-existing, but it may return back the rowid of the base table row, or the index row, depending on whether the base table or index was scanned (or if an IndexLookup operation was done).
Note in the explain the BASEROWID = BASEROWID join between the UNION derived table and the base table T1.  This predicate is also indicated as a key, meaning the join back to T1 looks up those base rowids directly instead of scanning the entire T1.

This is a cost-based optimization, meaning when costing table T1 as a single-table scan, or as the inner table of a join, we will first cost the scan/join using the normal method, then we look for any OR'ed predicates that may enable UIS (Unioned Index Scans) access path, and cost the scan/join with UIS access path.  The cheaper of the two is selected.  

DNF to CNF conversion of predicates previously does not apply the distributive property to derive new terms.  This is being added by this PR.  For example, given the query:

```
select * from BH, BU where ( BH.a = 1 AND BH.b = BU.a ) 
                     OR ( BH.a = BU.a AND BH.b = 1);
```

We can convert to CNF as follows:

```
(BH.a = 1 OR BH.a = BU.a)   AND 
(BH.a = 1 OR BH.b = 1)     AND 
(BH.b = BU.a OR BH.a = BU.a) AND 
(BH.b = BU.a OR BH.b = 1)
```

This conversion enables two separate UIS access paths on the two middle terms.  
First, (BH.a = 1 OR BH.b = 1)  can be applied to table BH in two scans.
Next, (BH.b = BU.a OR BH.a = BU.a)  can be applied between the result from above in 2 unioned joins with these index-enabling join predicates.

The number of predicate we can derive in this manner is limited to 100, by default, but may be tuned lower or higher (as high as 10000) using a new database property, splice.optimizer.maxDerivedCNFPredicates:

> call syscs_util.syscs_set_global_database_property('splice.optimizer.maxDerivedCNFPredicates', '200');

A CloneCRsVisitor is added to clone any BinaryRelationalOperatorNodes and underlying ColumnReferences during DNF to CNF term expansion.  This is required because we don't want remapping of a ColumnReference in one predicate to also remap a ColumnReference in a copied operator in another predicate.  Each predicate can be pushed separately so must not share memory.  If this cloning operation fails, so does DNF to CNF conversion because it is unsafe to proceed using shared memory in separate predicates.  Currently not many operators support cloning, so building up cloning support could extend CNF conversion potential to the full range of boolean expressions Splice supports.

UIS access paths can be disabled via database property splice.optimizer.disableUnionedIndexScans:
> call syscs_util.syscs_set_global_database_property('splice.optimizer.disableUnionedIndexScans', true);

UIS access paths can be favored by the optimizer for testing purposes via database property splice.optimizer.favorUnionedIndexScans:
> call syscs_util.syscs_set_global_database_property('splice.optimizer.favorUnionedIndexScans', true);

ROWID and BASEROWID cannot be column names in CREATE TABLE or RENAME statements, and will result in an error.

When costing UIS access paths, we first separate OR'ed predicates into separate lists and categorize them one-at-a-time.  If each separate predicate is either a start key or stop key, we attempt UIS access path.  The UNION tree which collects rowids and removes any duplicates is built up by function buildUnionedScans.  The FromBaseTable cloned objects in this tree are marked as indexFriendlyJoinsOnly, so that only nested loop join or merge join are used, to skip over unqualified join rows.
If there are no errors, an AccessPath object is returned from buildUnionedScans with the UNION derived table to optimize.  bindAndOptimizeUnionedIndexScansPath is then called, which builds a BASEROWID = BASEROWID join predicate between the base table and the UNION derived table and constructs a new SELECT statement:  

> SELECT (original_T1_projection_list) from T1, UNION_derived_table where T1.BASEROWID = UNION_derived_table.BASEROWID;

This statement is then bound and optimized (bindStatement and optimizeStatement are called).  
If the table is being joined with an outer table, the UNION_derived_table will include that join, but also we need to join the entire result back to the outer table in the main join.  buildRowIdPredWithOuterTable is called to generate a UNION_derived_table_plus_RowID_Join.BASEROWID2 = outer_table.BASEROWID predicate to match the proper rows.  It is possible to eliminate this join entirely by manually manipulating the list of optimizables at the end of join planning, but this may be a high-risk change, so is not being done in the first pass of this feature.  The UIS derived table plus rowid join result set is marked as skipBindAndOptimize, which is a new flag to indicate no more binding or optimization should be done on the result set if it is joined or referenced elsewhere in the AST.  It is also marked as outerTableOnly so it is only considered as the outer table of any join it participates in (since it has a reduced set of RowIDs, we want to drive the join from these instead of scan all RowIDs in the base table).

## How to test
```
create table t1 (a int, b int, primary key(a,b));
create index t1_idx2 on t1(b);
call syscs_util.syscs_set_global_database_property('splice.optimizer.favorUnionedIndexScans', 'true');
// The following should contain 'BASEROWID' in the explain output:
explain select * from t1 where a=1 or b=2;
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
@@ -36,6 +36,9 @@ import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.error.StandardException;
 import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
+import com.splicemachine.db.impl.sql.compile.FromTable;
+import com.splicemachine.db.impl.sql.compile.Predicate;
+import com.splicemachine.db.impl.sql.compile.ResultSetNode;
 
 /**
  * AccessPath represents a proposed access path for an Optimizable.
@@ -198,4 +201,52 @@ public interface AccessPath {
     void setNumUnusedLeadingIndexFields(int numUnusedLeadingIndexFields);
 
     FirstColumnOfIndexStats getFirstColumnStats();
+
+	/**
+	 * Store in the access path the predicate used to enable unioned index scans.
+	 */
+    void setUisPredicate(Predicate uisPredicate);
+
+	/**
+	 * The predicate used to enable unioned index scans, if any
+	 */
+    Predicate getUisPredicate();
+
+	/**
+	 * Store in the access path the RowId predicate used to join back
+	 * to the outer base table for unioned index scan query plans, when
+	 * the UIS tree has index enabling predicates which are join predicates.
+	 */
+    void setUisRowIdPredicate(Predicate uisPredicate);
+
+	/**
+	 * The RowId predicate used to join back to the outer base table
+	 * for unioned index scan query plans, when the UIS tree has
+	 * index enabling predicates which are join predicates
+	 */
+    Predicate getUisRowIdPredicate();
+
+	/**
+	 * The tree of UnionNodes that combines results of the OR'ed index scans,
+	 * for a unioned index scans access path
+	 */
+    FromTable getUnionOfIndexes();
+
+	/**
+	 * Store in the access path the tree of UnionNodes that combines results
+	 * of OR'ed index scans for a unioned index scans access path.
+	 */
+    void setUnionOfIndexes(FromTable unionOfIndexes);
+
+	/**
+	 * The entire optimized result set tree of the unioned index scans
+	 * plus the rowid join back to the base table
+	 */
+    ResultSetNode getUisRowIdJoinBackToBaseTableResultSet();
+
+	/**
+	 * Store in the access path the entire optimized result set tree of
+	 * the unioned index scans plus the rowid join back to the base table.
+	 */
+    void setUisRowIdJoinBackToBaseTableResultSet(ResultSetNode uisRowIdJoinBackToBaseTableResultSet);
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
@@ -35,10 +35,7 @@ import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.DataDictionary;
 import com.splicemachine.db.iapi.error.StandardException;
-import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
-import com.splicemachine.db.impl.sql.compile.FromTable;
-import com.splicemachine.db.impl.sql.compile.Predicate;
-import com.splicemachine.db.impl.sql.compile.ResultSetNode;
+import com.splicemachine.db.impl.sql.compile.*;
 
 /**
  * AccessPath represents a proposed access path for an Optimizable.
@@ -230,13 +227,13 @@ public interface AccessPath {
 	 * The tree of UnionNodes that combines results of the OR'ed index scans,
 	 * for a unioned index scans access path
 	 */
-    FromTable getUnionOfIndexes();
+    UnionNode getUnionOfIndexes();
 
 	/**
 	 * Store in the access path the tree of UnionNodes that combines results
 	 * of OR'ed index scans for a unioned index scans access path.
 	 */
-    void setUnionOfIndexes(FromTable unionOfIndexes);
+    void setUnionOfIndexes(UnionNode unionOfIndexes);
 
 	/**
 	 * The entire optimized result set tree of the unioned index scans
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
@@ -214,7 +214,7 @@ public interface AccessPath {
 	 * to the outer base table for unioned index scan query plans, when
 	 * the UIS tree has index enabling predicates which are join predicates.
 	 */
-    void setUisRowIdPredicate(Predicate uisPredicate);
+    void setUisRowIdPredicate(Predicate uisRowIdPredicate);
 
 	/**
 	 * The RowId predicate used to join back to the outer base table
@@ -246,4 +246,15 @@ public interface AccessPath {
 	 * the unioned index scans plus the rowid join back to the base table.
 	 */
     void setUisRowIdJoinBackToBaseTableResultSet(ResultSetNode uisRowIdJoinBackToBaseTableResultSet);
+
+	/**
+	 * Set all the fields in the access path related to Unioned Index Scans (UIS)
+	 * from the corresponding fields in the "other" access path.
+	 * This fields currently include:
+	 *    uisPredicate
+	 *    uisRowIdPredicate
+	 *    unionOfIndexes
+	 *    uisRowIdJoinBackToBaseTableResultSet
+	 */
+    void setUisFields(AccessPath copyFromAccessPath);
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
@@ -470,4 +470,22 @@ public interface Optimizable {
      * the first index column.
      */
     default boolean indexPrefixIteratorAllowed(AccessPath accessPath) { return false; }
+
+    /**
+     * If true, this optimizable must be the outer table of a join, and
+     * join plans which use it as the inner table will not be feasible.
+     * For a given binary join involving an <code>Optimizable</code> marked as
+     * <code>outerTableOnly</code>, the optimizer will only allow that
+     * <code>Optimizable</code> to participate as the outer table of the join.
+     */
+    default boolean outerTableOnly() { return false; }
+
+    /**
+     * If true, when this <code>Optimizable</code> is joined as the inner table,
+     * only index-friendly join strategies may be used:
+     *     Merge join or Nested Loop join.
+     * Index-friendly means that join predicates involving primary
+     * key or index columns can be used to limit the scan.
+     */
+    default boolean indexFriendlyJoinsOnly() { return false; }
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
@@ -477,6 +477,10 @@ public interface Optimizable {
      * For a given binary join involving an <code>Optimizable</code> marked as
      * <code>outerTableOnly</code>, the optimizer will only allow that
      * <code>Optimizable</code> to participate as the outer table of the join.
+     * This will only return true in limited scenarios when a table may
+     * only join with a single other table and no public interface to
+     * mark an optimizable as outerTableOnly is provided, on purpose,
+     * as this is an internal limited-use mechanism.
      */
     default boolean outerTableOnly() { return false; }
 
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
@@ -88,7 +88,7 @@ public interface OptimizablePredicate
 	boolean isStopKey();
 
 	/** Is this predicate a start or stop key? */
-	boolean isKey();
+	boolean isScanKey();
 
 	/**
 	 * Tell the predicate that it is to be used as a qualifier in an index
@@ -197,7 +197,7 @@ public interface OptimizablePredicate
 	boolean isFullJoinPredicate();
 
 	// Is this an OR of two or more relational operators which all are start keys or stop keys?
-	boolean isIndexEnablingORedPredicate(FromBaseTable optTable, AccessPath accessPath, Optimizer optimizer) throws StandardException;
+	boolean isDisjunctionOfScanKeys(FromBaseTable optTable, AccessPath accessPath, Optimizer optimizer) throws StandardException;
 
 	// Collect the OR'ed conditions in this Predicate into a list of PredicateLists,
 	// with each PredicateList containing a single Predicate which is the left child
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
@@ -37,7 +37,10 @@ import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
 import com.splicemachine.db.iapi.types.DataValueDescriptor;
 
 import com.splicemachine.db.iapi.util.JBitSet;
+import com.splicemachine.db.impl.sql.compile.FromBaseTable;
 import com.splicemachine.db.impl.sql.compile.SelectivityUtil;
+
+import java.util.List;
 
 /**
  * OptimizablePredicate provides services for optimizing predicates in a query.
@@ -83,6 +86,9 @@ public interface OptimizablePredicate
 
 	/** Is this predicate a stop key? */
 	boolean isStopKey();
+
+	/** Is this predicate a start or stop key? */
+	boolean isKey();
 
 	/**
 	 * Tell the predicate that it is to be used as a qualifier in an index
@@ -189,4 +195,13 @@ public interface OptimizablePredicate
 	void markFullJoinPredicate(boolean isForFullJoin);
 
 	boolean isFullJoinPredicate();
+
+	// Is this an OR of two or more relational operators which all are start keys or stop keys?
+	boolean isIndexEnablingORedPredicate(FromBaseTable optTable, AccessPath accessPath, Optimizer optimizer) throws StandardException;
+
+	// Collect the OR'ed conditions in this Predicate into a list of PredicateLists,
+	// with each PredicateList containing a single Predicate which is the left child
+	// of one of the OrNodes (OrNode chains are right-deep, with each left child being
+	// something other than an OrNode, if normalized properly).
+	List<OptimizablePredicateList> separateOredPredicates() throws StandardException;
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
@@ -62,7 +62,8 @@ public interface SessionProperties {
         ALWAYSALLOWINDEXPREFIXITERATION(17),
         OLAPALWAYSPENALIZENLJ(18),
         FAVORINDEXPREFIXITERATION(19),
-        COSTMODEL(20);
+        COSTMODEL(20),
+        JOINSTRATEGY(21);
 
         public static final int COUNT = PROPERTYNAME.values().length;
 
@@ -96,7 +97,7 @@ public interface SessionProperties {
             property = SessionProperties.PROPERTYNAME.valueOf(propertyNameString);
         } catch (IllegalArgumentException e) {
             throw StandardException.newException(SQLState.LANG_INVALID_SESSION_PROPERTY,propertyNameString,
-                "useOLAP, useSpark (deprecated), defaultSelectivityFactor, skipStats, olapQueue, recursiveQueryIterationLimit, tableLimitForExhaustiveSearch, minPlanTimeout, currentFunctionPath, disablePredsForIndexOrPkAccessPath, alwaysAllowIndexPrefixIteration, olapAlwaysPenalizeNLJ, favorIndexPrefixIteration, costModel");
+                "useOLAP, useSpark (deprecated), defaultSelectivityFactor, skipStats, olapQueue, recursiveQueryIterationLimit, tableLimitForExhaustiveSearch, minPlanTimeout, currentFunctionPath, disablePredsForIndexOrPkAccessPath, alwaysAllowIndexPrefixIteration, olapAlwaysPenalizeNLJ, favorIndexPrefixIteration, costModel, joinStrategy");
         }
 
         String valString = pair.getSecond();
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
@@ -753,14 +753,9 @@ public class GenericStatement implements Statement{
         String disablePrefixIteratorModeString =
             PropertyUtil.getCachedDatabaseProperty(lcc, Property.DISABLE_INDEX_PREFIX_ITERATION);
         boolean disablePrefixIteratorMode = CompilerContext.DEFAULT_DISABLE_INDEX_PREFIX_ITERATION;
-        try {
-            if (disablePrefixIteratorModeString != null)
-                disablePrefixIteratorMode =
-                Boolean.parseBoolean(disablePrefixIteratorModeString);
-        } catch (Exception e) {
-            // If the property value failed to convert to a boolean, don't throw an error,
-            // just use the default setting.
-        }
+        if (disablePrefixIteratorModeString != null)
+            disablePrefixIteratorMode =
+            Boolean.parseBoolean(disablePrefixIteratorModeString);
         cc.setDisablePrefixIteratorMode(disablePrefixIteratorMode);
     }
 
@@ -768,14 +763,9 @@ public class GenericStatement implements Statement{
         String disableUnionedIndexScansString =
             PropertyUtil.getCachedDatabaseProperty(lcc, Property.DISABLE_UNIONED_INDEX_SCANS);
         boolean disableUnionedIndexScans = CompilerContext.DEFAULT_DISABLE_UNIONED_INDEX_SCANS;
-        try {
-            if (disableUnionedIndexScansString != null)
-                disableUnionedIndexScans =
-                Boolean.parseBoolean(disableUnionedIndexScansString);
-        } catch (Exception e) {
-            // If the property value failed to convert to a boolean, don't throw an error,
-            // just use the default setting.
-        }
+        if (disableUnionedIndexScansString != null)
+            disableUnionedIndexScans =
+            Boolean.parseBoolean(disableUnionedIndexScansString);
         cc.setDisableUnionedIndexScans(disableUnionedIndexScans);
     }
 
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
@@ -43,14 +43,12 @@ import com.splicemachine.db.iapi.sql.compile.*;
 import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
 import com.splicemachine.db.iapi.sql.conn.StatementContext;
 import com.splicemachine.db.iapi.sql.depend.Dependency;
-import com.splicemachine.db.iapi.sql.depend.Provider;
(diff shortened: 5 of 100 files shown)


msirek:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2818/

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -740,6 +763,36 @@ private void setDisablePrefixIteratorMode(LanguageConnectionContext lcc, Compile
...
+            PropertyUtil.getCachedDatabaseProperty(lcc, Property.DISABLE_UNIONED_INDEX_SCANS);
+        boolean disableUnionedIndexScans = CompilerContext.DEFAULT_DISABLE_UNIONED_INDEX_SCANS;
+        try {
+            if (disableUnionedIndexScansString != null)
+                disableUnionedIndexScans =
+                Boolean.parseBoolean(disableUnionedIndexScansString);
+        } catch (Exception e) {
+            // If the property value failed to convert to a boolean, don't throw an error,
+            // just use the default setting.
+        }
```
ascend1, line 777: `Boolean.parseBoolean()` never throws, but it's minor anyway.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -555,4 +555,11 @@ public boolean collectExpressions(Map<Integer, Set<ValueNode>> exprMap) {
         }
         return result;
     }
+
+    public void copy(BinaryComparisonOperatorNode other) throws StandardException
+    {
+    	super.copy(other);
```
ascend1, line 561: indentation

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryOperatorNode.java
```diff
@@ -977,5 +979,18 @@ public void castLeftOperandAndBindCast(DataTypeDescriptor type) throws StandardE
...
+    	super.copy(other);
+        this.operatorType = other.operatorType;
+        this.leftMatchIndexExpr = other.leftMatchIndexExpr;
+        this.rightMatchIndexExpr = other.rightMatchIndexExpr;
+        this.leftMatchIndexExprConglomDesc = other.leftMatchIndexExprConglomDesc;
+        this.rightMatchIndexExprConglomDesc = other.rightMatchIndexExprConglomDesc;
+        this.leftMatchIndexExprColumnPosition = other.leftMatchIndexExprColumnPosition;
+        this.rightMatchIndexExprColumnPosition = other.rightMatchIndexExprColumnPosition;
+        this.xmlQuery = other.xmlQuery;
+    }
```
ascend1, line 994: I don't know what's the convention in Java for copying, but a name like `copyFrom()` looks better to me.
Also, indentation is not consistent for `super.copy(other)`, this applies to other `copy()` method as well.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -64,6 +64,7 @@
 import com.splicemachine.db.impl.sql.catalog.SYSTOKENSRowFactory;
 import com.splicemachine.db.impl.sql.catalog.SYSUSERSRowFactory;
```
ascend1, line 65: I stopped at this file due to time constraint. Will try to continue in another free time slot.

ascend1 (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -902,18 +1322,61 @@ public CostEstimate estimateCost(OptimizablePredicateList predList,
...
+        if (uisAccessPath != null &&
+            (getCompilerContext().getFavorUnionedIndexScans() ||
+            uisAccessPath.getCostEstimate().compare(firstPassCostEstimate) < 0)) {
+
+            hintedUISAccessPath = getCompilerContext().getFavorUnionedIndexScans();
+            if (hintedUISAccessPath)
+                reduceEstimatedCosts(uisAccessPath);
+            currentAccessPath.copy(uisAccessPath);
+            finalCostEstimate = firstPassCostEstimate = uisAccessPath.getCostEstimate();
+        }
```
ascend1, line 1372: This is a very hacky process. If I understand it correctly, the UIS access path is compared only to current access path, meaning current cd + current join strategy. Suppose we are in cost estimation of (table cd + NLJ) and then UIS has lower cost. Then we directly do `currentAccessPath.copy(uisAccessPath)`, meaning when calling `nextDecoratedPermutation()`, it would be (UIS + SortMerge join). So we skip all combinations of table cd with other join strategies! What if, for example, (table cd + Merge join) is has better cost than (UIS + NLJ)? My understanding is we would miss that chance.

Since we make an `AccessPath` for UIS, why don't we just integrate this directly into the framework:
```
while(optimizer.nextJoinOrder()){
    while(optimizer.getNextDecoratedPermutation()){
        optimizer.costPermutation();
    }
 }
```
Basically, in enumerating all conglomerates of this table, we could add one more fake conglomerate indicating the UIS access path. Internally, it could be just table cd + UIS flag. Then we would have all combinations of UIS fake cd with all join strategies enumerated, fitting into the current framework of deciding better plan.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -926,11 +1389,279 @@ public CostEstimate estimateCost(OptimizablePredicateList predList,
...
+        accessPath.setCostEstimate(accessPath.getCostEstimate().cloneMe());
+
+        double costScaleFactor = 1e-9;
+        CostEstimate estimate = accessPath.getCostEstimate();
+
+        estimate.setLocalCost(estimate.getLocalCost() * costScaleFactor);
+        estimate.setRemoteCost(estimate.getRemoteCost() * costScaleFactor);
+        estimate.setLocalCostPerParallelTask(estimate.getLocalCostPerParallelTask() * costScaleFactor);
+        estimate.setRemoteCostPerParallelTask(estimate.getRemoteCostPerParallelTask() * costScaleFactor);
+    }
```
ascend1, line 1406: If we could implement the fake UIS cd idea, then this is not needed. We simply add one more condition in deciding better plan to check whether we favor UIS code path. The code would be much less hacky, I think.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -926,11 +1389,279 @@ public CostEstimate estimateCost(OptimizablePredicateList predList,
...
+                   }
+                   tempAnd.setRightOperand(andNodeToAdd);
+               }
+           }
+       }
+       return true;
+    }
+
+    // Build a UNION node between 2 branches in a Unioned Index Scans access path.
+    private FromTable getUnionNode(ResultSetNode leftSide, FromTable rightSide, Optimizer optimizer) throws StandardException {
```
ascend1, line 1646: Function signature is strange. Why `leftSide` is `ResultSetNode` but `rightSide` is a `FromTable`?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -926,11 +1389,279 @@ public CostEstimate estimateCost(OptimizablePredicateList predList,
...
+                C_NodeTypes.RESULT_COLUMN,
+                BASEROWID2,
+                columnReference,
+                getContextManager());
+
+        resultColumnList.addResultColumn(rowIdResultColumn);
+    }
+
+    // Build one branch of the UNION tree for UIS access path.
+    ResultSetNode buildSelectNode(ResultSetNode source, Optimizer optimizer) throws StandardException {
```
ascend1, line 1504: I didn't look into this method.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -1633,6 +2369,35 @@ public void bindResultColumns(FromList fromListParam)
         /* Nothing to do, since RCL bound in bindNonVTITables() */
     }
 
+    private void buildRowIdColumn(boolean isRowId) throws StandardException {
+        String colName = isRowId ? ROWID : BASEROWID;
```
ascend1, line 2373: What's the difference between ROWID and BASEROWID?

ascend1 (changes requested):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/AccessPath.java
```diff
@@ -198,4 +201,52 @@
...
+	 * The tree of UnionNodes that combines results of the OR'ed index scans,
+	 * for a unioned index scans access path
+	 */
+    FromTable getUnionOfIndexes();
+
+	/**
+	 * Store in the access path the tree of UnionNodes that combines results
+	 * of OR'ed index scans for a unioned index scans access path.
+	 */
+    void setUnionOfIndexes(FromTable unionOfIndexes);
```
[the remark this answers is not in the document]
msirek: Sure, this may be more convenient and accurate.  Refactored it.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
```diff
@@ -470,4 +470,22 @@ default void setHasJoinPredicatePushedDownFromOuter(boolean value) {
...
     default boolean indexPrefixIteratorAllowed(AccessPath accessPath) { return false; }
+
+    /**
+     * If true, this optimizable must be the outer table of a join, and
+     * join plans which use it as the inner table will not be feasible.
+     * For a given binary join involving an <code>Optimizable</code> marked as
+     * <code>outerTableOnly</code>, the optimizer will only allow that
+     * <code>Optimizable</code> to participate as the outer table of the join.
+     */
+    default boolean outerTableOnly() { return false; }
```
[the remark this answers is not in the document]
msirek: This setting is only intended to be used in the simple case of joining one table with another, like in bindAndOptimizeUnionedIndexScansPath.  Complex join scenarios are not targeted or considered.  To clarify this, I have added the following comment to the above:

     * This will only return true in limited scenarios when a table may
     * only join with a single other table and no public interface to
     * mark an optimizable as outerTableOnly is provided, on purpose,
     * as this is an internal limited-use mechanism.

Also, I have made setOuterTableOnly protected, and added the following comment:
    // This is a special-purpose method for tagging an optimizable to be used
    // as the outer table of a join in very limited scenarios, usually when
    // we are planning only a single join between it and another table.
    // Please do not use this for queries of all forms, and only
    // use it if you have a targeted, specific and limited use case.
    protected void setOuterTableOnly(boolean outerTableOnly) {

The only Optimizable that is planned outside of bindAndOptimizeUnionedIndexScansPath with other Optimizables is the uisRowIdJoinBackToBaseTableResultSet.  The flag is set in initializeUnionedIndexScanAccessPath.  I'm thinking this may not actually be required.  So, to limit the scope of usage of this setting I will remove the call to setOuterTableOnly in this method.

ascend1:
I'm OK to have it merged into side branch in its current status. To master, though, let's have a discussion about details.

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
msirek: OK, removed the try/catch

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
msirek: Fixed

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryOperatorNode.java
msirek: Fixed the indentation and changed copy -> copyFrom

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: OK

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: ROWID doesn't always retrieve the base table row id.  If we do an IndexScan without IndexLookup, it is the row id of the index row.
BASEROWID always retrieves the base table row id.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: Changed the function signature to:

private UnionNode getUnionNode(ResultSetNode leftSide, ResultSetNode rightSide, Optimizer optimizer) throws StandardException

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
```diff
@@ -84,6 +87,9 @@
 	/** Is this predicate a stop key? */
 	boolean isStopKey();
 
+	/** Is this predicate a start or stop key? */
+	boolean isKey();
```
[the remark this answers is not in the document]
msirek: Changed it to isScanKey

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
```diff
@@ -189,4 +195,13 @@ int hasEqualOnColumnList(int[] baseColumnPositions,
 	void markFullJoinPredicate(boolean isForFullJoin);
 
 	boolean isFullJoinPredicate();
+
+	// Is this an OR of two or more relational operators which all are start keys or stop keys?
+	boolean isIndexEnablingORedPredicate(FromBaseTable optTable, AccessPath accessPath, Optimizer optimizer) throws StandardException;
```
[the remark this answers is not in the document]
msirek: Renamed to isDisjunctionOfScanKeys

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: There should be no skipping of combinations in the current approach.  The UIS access path and associated cost is recalculated for each new join strategy we consider.  In your example, when switching from NLJ + UIS to SortMerge join via nextDecoratedPermutation(), all of the UIS-related fields in the access path are set to null and we begin again fresh with a new comparison.  I did notice a bug, however, where those access path fields were being set to null in the wrong location, so I moved them up to the top of FromBaseTable.estimateCost:

        currentAccessPath.setUisPredicate(null);
        currentAccessPath.setUisRowIdPredicate(null);
        currentAccessPath.setUnionOfIndexes(null);
        currentAccessPath.setUisRowIdJoinBackToBaseTableResultSet(null);

When looking at the access path enumeration flow, this is not as clean as it could be.  Really we should be clearing these fields out as soon as we set to a new access path.  Therefore I will also initialize these fields in FromTable.nextAccessPath.  Let me know if you know of more locations where new access paths are considered, where we should also reset these fields.

On the pass through the access paths which looks at the base table conglomerate we cost each join strategy, both with and without UIS.  If UIS wins for any given join strategy, it is saved into the best access path if it is cheaper than the current best access path.  As the next join is considered, UIS fields are reset, and we again cost normal access path vs. UIS access path, and also vs. IndexPrefixIteratorMode.  On subsequent passes, when only index conglomerates are considered, there is no need to cost UIS access path again because the same set of index conglomerates and primary keys used by UIS would be chosen again and any best cost UIS path is already stored in the best access path.

The alternative you mentioned to used a fake conglomerate and drive the combinations from the outside would also work, and I can see the appeal if your preference is to drive the index combinations from the outside.  But it seems a little like we'd be trying to fit this access path information into a ConglomerateDescriptor structure, which describes something entirely different.  We might not have useful information for every field, whereas building only the access path creates no made-up structures.  It seems like this is just a matter of preference.  Maybe we can have a discussion about this to compare the two approaches.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: Maybe we can discuss this as well.

msirek:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizable.java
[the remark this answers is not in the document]
ascend1: Thanks. I think it really makes sense to document it and prevent it from being misused. Change is good.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
ascend1: Since the access path enumeration question is answered, I think it's OK to do so. However, I feel a bit strange about reducing cost with a hard-coded factor. Is it possible to cost uis access path in a way that it's a fair comparison with costs of other access paths?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -894,6 +894,24 @@ public boolean nextJoinOrder() throws StandardException{
...
+                    FromTable fromTable = (FromTable)optimizableResultSet;
+                    AccessPath bestAP = fromTable.getBestAccessPath();
+                    fromTable.startOptimizing(this,currentRowOrdering);
+                    if (nextOptimizable == proposedJoinOrder[0] )
+                        bestAP.setCostEstimate(fromTable.getTrulyTheBestAccessPath().getCostEstimate());
+                    AccessPath currentAP = fromTable.getCurrentAccessPath();
+                    CostEstimate costEstimate=fromTable.getCostEstimate(this);
+                    costEstimate.setCost(Double.MAX_VALUE,Double.MAX_VALUE,Double.MAX_VALUE);
+                    currentAP.setCostEstimate(costEstimate);
+                }
```
ascend1, line 910: Can you please add more comment about under which condition we enter this code path?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2404,7 +2426,10 @@ private void ruleBasedCostOptimizable(Optimizable optimizable,
                 bestAp.setLockMode(currentAccessPath.getLockMode());
                 bestAp.setMissingHashKeyOK(currentAccessPath.isMissingHashKeyOK());
                 bestAp.setNumUnusedLeadingIndexFields(currentAccessPath.getNumUnusedLeadingIndexFields());
-
+                bestAp.setUisPredicate(currentAccessPath.getUisPredicate());
+                bestAp.setUisRowIdPredicate(currentAccessPath.getUisRowIdPredicate());
+                bestAp.setUnionOfIndexes(currentAccessPath.getUnionOfIndexes());
+                bestAp.setUisRowIdJoinBackToBaseTableResultSet(currentAccessPath.getUisRowIdJoinBackToBaseTableResultSet());
```
ascend1, line 2432: It looks like that it makes sense to have a smaller helper function like `setUisFields()`.

ascend1 (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
msirek: I think the point is we don't want it to be fair.  When splice.optimizer.favorUnionedIndexScans is set, we want UIS to be picked all the time.  Decreasing all of the UIS access paths by the same factor would mean only another UIS access path with a similarly scaled cost estimate would beat the current one.  I'm not sure how else to get this same effect.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromTable.java
```diff
@@ -1200,6 +1248,52 @@ protected void markUpdatableByCursor(List<String> updateColumns){
...
+            fromSSQ                  = other.fromSSQ;
+            outerJoinLevel           = other.outerJoinLevel;
+            hasJoinPredicatePushedDownFromOuter = other.hasJoinPredicatePushedDownFromOuter;
+
+            protected CostEstimate bestCostEstimate;
+            protected CostEstimate accumulatedCost;
+            protected CostEstimate accumulatedCostForSortAvoidancePlan;
+            private Map<Object,AccessPath> bestPlanMap;
+            PredicateList postJoinPredicates;
+        */
```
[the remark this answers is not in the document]
msirek: I don't know.  I like listing them as intentionally not added so if someone thinks they need to start copying one of these fields they can see it was excluded on purpose, and think twice about adding it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: OK, added the comment:

>             // If we want to re-use this Optimizable without re-planning it, return now.
>             // This is the case where nextOptimizable uses a Unioned Index Scans access path,
>             // and it is being joined with another table which itself might use a UIS access path.
>             // We want to set the join cost estimate to an uninitialized state and set assignedTableMap
>             // to include the proper tables, but we don't want to push any new predicates to this
>             // table because it has already been materialized into a sequence of statements that
>             // can't be modified.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: Yes, that would make it easier to read.  Added setUisFields.

dgomezferro (commented):
@msirek can you elaborate a bit more about this:
> It is possible to eliminate this join entirely by manually manipulating the list of optimizables at the end of join planning, but this may be a high-risk change, so is not being done in the first pass of this feature.

Would it make sense to have a hint that forces this access path?

I love the `keys` entry in the explain plan!

dgomezferro (approved):
I have made a surface level review only, but it looks great! I'd like to test this extensively since it's a big change

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/SessionProperties.java
```diff
@@ -62,7 +62,8 @@
         ALWAYSALLOWINDEXPREFIXITERATION(17),
         OLAPALWAYSPENALIZENLJ(18),
         FAVORINDEXPREFIXITERATION(19),
-        COSTMODEL(20);
+        COSTMODEL(20),
+        JOINSTRATEGY(21);
```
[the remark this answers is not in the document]
msirek: This is not a common-use item, but can be useful if it difficult in the current processing to hint joinStrategy on the table, like when a subquery is flattened or for UIS access path.

db-engine/src/main/java/com/splicemachine/db/impl/sql/GenericStatement.java
```diff
@@ -730,15 +753,35 @@ private void setDisablePrefixIteratorMode(LanguageConnectionContext lcc, Compile
...
+        String disableUnionedIndexScansString =
+            PropertyUtil.getCachedDatabaseProperty(lcc, Property.DISABLE_UNIONED_INDEX_SCANS);
+        boolean disableUnionedIndexScans = CompilerContext.DEFAULT_DISABLE_UNIONED_INDEX_SCANS;
+        if (disableUnionedIndexScansString != null)
+            disableUnionedIndexScans =
+            Boolean.parseBoolean(disableUnionedIndexScansString);
+        cc.setDisableUnionedIndexScans(disableUnionedIndexScans);
+    }
+
+    private void setfavorUnionedIndexScans(LanguageConnectionContext lcc, CompilerContext cc) throws StandardException {
```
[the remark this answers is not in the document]
msirek: Changed it to setFavorUnionedIndexScans

carolp-503 (approved):
+1

PR #5428, opened 2021-04-28
DB-11955 Add scan cost estimation to pluggable cost estimation model. 

hatyo:
## Short Description

In this PR, we make scan cost estimation part of the pluggable cost model.

## Long Description

Until now, we only have join cost estimation in the pluggable cost model, after doing some experiments with base table scan and index lookup, we found that the cost estimation of these operations is also quite different from the actual execution time and they have to be adapted.

We refactor the scan cost estimation relevant functions:  `addPredicat`,  `generateCost`, and `generateOneRowCost` from `ScanCostFunction` into a separate interface.

Please note that `ScanCostFunction` is not a stateless function as the name might imply, it basically holds a state of the scan such as predicates, base table, ... and computes the cost from this state. We didn't change the design of the class, instead we just renamed to `AbstractScanCostEstimator` and moved the implementation of the above three methods to `V1ScanCostEstimator`.

## How to test

Nothing is changed in terms of functionality, this PR is only about refactoring, therefore having it only passing existing ITs is a good enough measure of its quality.


ascend1 (approved):

msirek (approved):

carolp-503 (approved):
+1

PR #5429, opened 2021-04-28
DB-11657 Include DB Query Statement Hash Signature in statement logging messages

martinrupp:
## Short Description

## Long Description

## How to test


db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -3986,6 +3987,15 @@ public void logEndFetching(String uuid, String statement, long fetchedRows) {
...
 
+    String getStatmentId(String stmt)
+    {
+        try {
+            return SQLStatement.getSqlStatement(stmt).getId();
+        } catch (Exception e) {
+            return "X#ERROR#";
+        }
+    }
+
```
dgomezferro, line 3998: This should be constant for any prepared statement, is it possible to compute it during query compilation only once? If we do it early enough it can also be reused by DB-11957

dgomezferro (changes requested):

utilities/src/main/java/com/splicemachine/qpt/SQLSignature.java
```diff
@@ -0,0 +1,218 @@
...
+        switch (token.value) {
+            case SQLTokenizer.TOKEN_NULL:
+            case SQLTokenizer.TOKEN_STRINGS:
+            case SQLTokenizer.TOKEN_NUMBER:
+                return true;
+        }
+        return false;
+    }
+
+    /* An attempt to automate parameterization based on changing values
```
ipraznik-splice, line 129: please remove if it's not needed

utilities/src/main/java/com/splicemachine/qpt/SQLSignature.java
```diff
@@ -0,0 +1,218 @@
...
+            if (isExcluded(token) && !token.string.equals(newTokens[i].string)) {
+                parameter[i] = true;
+            }
+        }
+    }
+    */
+
+    public static SQLSignature getSignature(Token[] tokens) {
+        SQLSignature signature = signatures.computeIfAbsent(computeHash(tokens), (key) -> new SQLSignature(tokens));
+//        signature.apply(tokens);
```
ipraznik-splice, line 144: please remove if it's not needed

utilities/src/main/java/com/splicemachine/qpt/SQLSignature.java
```diff
@@ -0,0 +1,218 @@
+/*
```
ipraznik-splice, line 1: Please add a description of the hash signature algorithm. Maybe some information on why it was chosen, if the signature is unique enough etc.

The same for Encoding.java. Is it possible maybe to chose another name for this class, that the name says which kind of encoding is used here?

ipraznik-splice (changes requested):

PR #5430, opened 2021-04-28
DB-11941 Fix optimizer selectivity issues with db2varchars

arnaud-splice:
## Short Description

This PR fixes two selectivity issues with db2 compatible SQLVarchar

## Long Description

1. DB2 strings lookup in sketches:
To estimate selectivity, we lookup the key from a predicate in statistics sketches. If that key is Varchar and DB2 strings compatibility mode is enabled, we should not search for that precise key in the sketches, but for a range of equivalent keys. Luckily, we already support a startKey and a stopKey when probing sketches, so instead of passing the same value twice, we need to pass a trimmed version of the key as a startKey, and a padded version of the key as a stopKey

2. Multi-column indexes cannot be used beyond a DB2 varchar column with startKey/stopKey
Say we have an index on three columns (VARCHAR, INT, INT)
If DB2 strings are not used, the first three predicates of that index can have startKey and stopKey.
However, given the fact that the first column of that index is a DB2Varchar, its startKey and stopKey will cover an entire range, and subsequent columns would fraction that range in separate subranges instead of refining it.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
@@ -17,13 +17,13 @@ import com.splicemachine.EngineDriver;
 import com.splicemachine.access.api.FileInfo;
 import com.splicemachine.access.api.SConfiguration;
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.services.io.StoredFormatIds;
 import com.splicemachine.db.iapi.sql.compile.CostEstimate;
 import com.splicemachine.db.iapi.sql.dictionary.*;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.stats.*;
 import com.splicemachine.db.iapi.store.access.StoreCostController;
-import com.splicemachine.db.iapi.types.DataValueDescriptor;
-import com.splicemachine.db.iapi.types.RowLocation;
+import com.splicemachine.db.iapi.types.*;
 import com.splicemachine.db.impl.sql.catalog.DataDictionaryCache;
 import com.splicemachine.db.impl.sql.catalog.SYSTABLESTATISTICSRowFactory;
 import com.splicemachine.db.impl.sql.compile.FirstColumnOfIndexStats;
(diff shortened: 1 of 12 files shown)


spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2838/

arnaud-splice:

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
```diff
@@ -360,8 +359,37 @@ public RowLocation newRowLocationTemplate() {
...
+
     @Override
-    public double getSelectivity(boolean fromExprIndex, int columnNumber, DataValueDescriptor start, boolean includeStart, DataValueDescriptor stop, boolean includeStop, boolean useExtrapolation) {
+    public double getSelectivity(boolean fromExprIndex, int columnNumber, DataValueDescriptor start, boolean includeStart, DataValueDescriptor stop, boolean includeStop, boolean useExtrapolation) throws StandardException {
+        if (useDb2CompatibleVarchars && baseTableRow.getColumn(columnNumber).getTypeName().equals(TypeId.VARCHAR_NAME)) {
+            int varcharLength = ((SQLVarchar)baseTableRow.getColumn(columnNumber)).getSqlCharSize();
+            if (includeStart) {
+                start = trimKey((StringDataValue)start);
+            } else {
+                start = padKey((StringDataValue) start, varcharLength);
```
msirek, line 385: I don't think includeStart means that the start key value should change.  It just means when comparing with stats, do we use a > or a >= comparison (from ColumnStatisticsImpl.java):

> start != null && !start.isNull() && start.compare(includeStart ? ORDER_OP_GREATERTHAN : ORDER_OP_GREATEROREQUALS,

It seems we should always trim the start and always pad the end.  What do you think?

msirek (changes requested):

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/stats/StoreCostControllerImpl.java
arnaud-splice: Hum, I'm not sure. For example, if `startKey == 'a    '` and should be included, then startKey should be `'a'`. However, if it should not be included, then it means that `'a'` to `'a    ' should also not be included, so we need to pad it to the highest value not to be included.

msirek: > Hum, I'm not sure. For example, if `startKey == 'a '` and should be included, then startKey should be `'a'`. However, if it should not be included, then it means that `'a'` to `'a '` should also not be included, so we need to pad it to the highest value not to be included.

Ah, right.  Everything matches up to the longest startKey, so to compute greater than it has to be greater than the longest key, and likewise less than end key has to be smaller than the shortest possible match.  Thanks.

msirek (approved):

ascend1 (approved):

carolp-503 (approved):
+1

PR #5437, opened 2021-04-30
DB-11957 Add query hinting mechanism (AZ-2011)

dgomezferro:
## Short Description
Add query hinting mechanism


carolp-503 (commented):
Approved for this branch (belated)

PR #5443, opened 2021-04-30
DB-11966: set selectivityEstimationIncludingSkewedDefault for all servers

jyuanca:
## Short Description
`selectivityEstimationIncludingSkewedDefault` was defined as a service property, which cannot be set globally by `syscs_set_global_database_property`.

## Long Description
`selectivityEstimationIncludingSkewedDefault` was defined as a service property, which cannot be set globally by `syscs_set_global_database_property`. To fix it, remove it from service properties, and fix the code to retrieve it from service property.

## How to test

Run 

`call syscs_util.syscs_set_global_database_property('derby.database.selectivityEstimationIncludingSkewedDefault', 'true');`

to set it to true. Verify it is set for all servers

```
call syscs_util.syscs_get_global_database_property('derby.database.selectivityEstimationIncludingSkewedDefault');
HOST_NAME                     |PROPERTY_VALUE
-----------------------------------------------
Juns-MacBook-Pro-2.local:1527 | true  
Juns-MacBook-Pro-2.local:1528 | true

2 rows selected
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
@@ -533,7 +533,6 @@ public abstract class DataDictionaryImpl extends BaseDataDictionary{
                     }
                 }
             }
-            startParams.setProperty("catalogVersion", dictionaryVersion.toString());
             assert authorizationDatabaseOwner!=null:"Failed to get Database Owner authorization";
 
             // Update (or create) the system stored procedures if requested.
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.ZooDefs;
 
 import com.splicemachine.access.HConfiguration;
@@ -93,11 +94,11 @@ public class ZkPropertyManager implements PropertyManager{
     }
 
     @Override
-    public void removeProperty(String propertyName) throws StandardException {
+    public void removeProperty(String propertyName){
         try {
             ZkUtils.safeDelete(zkSpliceDerbyPropertyPath + "/" + propertyName, -1);
-        } catch (Exception e) {
-            throw StandardException.plainWrapException(e);
+        } catch (KeeperException | InterruptedException e) {
+            throw new RuntimeException(e);
         }
     }
 }
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
@@ -67,8 +67,14 @@ public class ZkPropertyManager implements PropertyManager{
 
     @Override
     public void addProperty(String propertyName,String propertyValue) throws StandardException{
+        String path = zkSpliceDerbyPropertyPath + "/" + propertyName;
         try{
-            ZkUtils.safeCreate(zkSpliceDerbyPropertyPath+"/"+propertyName,Bytes.toBytes(propertyValue),ZooDefs.Ids.OPEN_ACL_UNSAFE,CreateMode.PERSISTENT);
+            if (ZkUtils.getRecoverableZooKeeper().exists(path, false) == null) {
+                ZkUtils.safeCreate(path, Bytes.toBytes(propertyValue), ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
+            }
+            else {
+                ZkUtils.setData(path, Bytes.toBytes(propertyValue), -1);
+            }
         }catch(Exception e){
             throw Exceptions.parseException(e);
         }
--- a/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
+++ b/hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
@@ -85,4 +85,13 @@ public class ZkPropertyManager implements PropertyManager{
             throw Exceptions.parseException(e);
         }
     }
+
+    @Override
+    public void removeProperty(String propertyName) throws StandardException {
+        try {
+            ZkUtils.safeDelete(zkSpliceDerbyPropertyPath + "/" + propertyName, -1);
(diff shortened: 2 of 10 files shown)


arnaud-splice (approved):

carolp-503 (approved):
+1

jyuanca:

spliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/catalog/DataDictionaryImpl.java
```diff
@@ -533,7 +533,6 @@ public void boot(boolean create,Properties startParams) throws StandardException
                     }
                 }
             }
-            startParams.setProperty("catalogVersion", dictionaryVersion.toString());
```
dgomezferro, line 536: Why is this removed?

jyuanca: dictionaryVersion is always derby version 10.x. We should set splice build version. It is set somewhere else

splice_machine/src/main/java/com/splicemachine/derby/impl/SpliceService.java
```diff
@@ -61,8 +62,10 @@ public Properties getServiceProperties(String serviceName, Properties defaultPro
 			Set<String> properties = propertyManager.listProperties();
 //			List<String> children = ZkUtils.getChildren(zkSpliceDerbyPropertyPath, false);
 			for (String property: properties) {
-				String value = propertyManager.getProperty(property);
-				service.setProperty(property, value);
+				if (!property.equals(Property.SELECTIVITY_ESTIMATION_INCLUDING_SKEWED)) {
```
martinrupp, line 65: please add a comment why we need special handling of SELECTIVITY_ESTIMATION_INCLUDING_SKEWED here. if this is because SELECTIVITY_ESTIMATION_INCLUDING_SKEWED is now not stored in propertyManager anymore, then i think we should have sth like `PropertyUtil.deletedPropertiesList`

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/PropertyConglomerate.java
```diff
@@ -119,6 +118,10 @@ public PropertyConglomerate(TransactionController tc, boolean create, Properties
         try {
             Set<String> props = propertyManager.listProperties();
             for (String child: props) {
+                if (child.equals(Property.SELECTIVITY_ESTIMATION_INCLUDING_SKEWED)) {
```
martinrupp, line 121: `PropertyUtil.deletedPropertiesList.contains(child)`

dgomezferro (approved):

martinrupp (changes requested):

hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
```diff
@@ -67,8 +67,14 @@ public String getProperty(String propertyName) throws StandardException{
...
     public void addProperty(String propertyName,String propertyValue) throws StandardException{
+        String path = zkSpliceDerbyPropertyPath + "/" + propertyName;
         try{
-            ZkUtils.safeCreate(zkSpliceDerbyPropertyPath+"/"+propertyName,Bytes.toBytes(propertyValue),ZooDefs.Ids.OPEN_ACL_UNSAFE,CreateMode.PERSISTENT);
+            if (ZkUtils.getRecoverableZooKeeper().exists(path, false) == null) {
+                ZkUtils.safeCreate(path, Bytes.toBytes(propertyValue), ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
+            }
+            else {
+                ZkUtils.setData(path, Bytes.toBytes(propertyValue), -1);
+            }
```
hatyo, line 77: This is not an atomic operation anymore, for a good reason, as `safeCreate` does [not](https://github.com/sfines/menagerie/blob/master/src/main/java/org/menagerie/ZkUtils.java#L206) alter the data if the path exist.

Does it make sense to make it atomic, e.g. by obtaining a distributed lock on the `path` itself?

By the way, is setting a property a transactional operation? at first glance it seems not, right, from the looks of this method, it looks like last-writer-wins, causing this anomaly:

```
client 1                                                       client 2

autocommit off                                                 autocommit off
call syscs_util.syscs_set_global_database_property('a','b');
call syscs_util.syscs_set_global_database_property('c','e');
                                                               call syscs_util.syscs_set_global_database_property('a','z');
commit;
                                                               commit;
```

End result:
```
PropertyName,PropertyValue
a,z
c,e
```

hatyo (commented):
PR looks good me given that @martinrupp's comment is addressed.

My comment was meant mainly to spark a discussion about transactionality of properties, and maybe revisiting the property model.

Let us have a follow up discussion to revisit the following topics:
- Is it expected that global database properties are not transactional? if so, is this properly documented?
- Should we encourage devs to think about having the property defined in a session if possible, and only add it as a database property if and only if the property rarely changes and is meant to change the behavior of the entire system.
- global database properties do affect all database in a multi-db setup, should we have a database-specific layer allowing us to e.g. set min. retention period for DB1 to be 4 days and for DB2 to be 15 days?
- some properties look like a session property but could have cross-session side-effects, e.g. it is possible to set a property that affect a statement's plan, if that plan is cached and used by other sessions then this property has a cross-session side-effect. Should we revisit this model? (e.g. cache statements affected by such properties in a special session-specific statement-cache?)

hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
jyuanca: property is not transactional

jyuanca:

jyuanca:

spliceci:
TEST SUCCEEDED +1

spliceci:
TEST SUCCEEDED +1

jyuanca:
@martinrupp please review latest changes.

martinrupp (approved):

hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
jyuanca: If the path exists, the data is modified.

jyuanca:
Filed  DB-12091 to followup Youssef's comments

hbase_sql/src/main/java/com/splicemachine/derby/impl/sql/ZkPropertyManager.java
hatyo: > Filed DB-12091 to followup Youssef's comments

Thank you @jyuanca

hatyo (approved):

PR #5449, opened 2021-05-05
DB-11925 Flatten EXISTS queries with non-equijoin predicates

msirek:
## Short Description
This feature improves the performance of EXISTS and NOT EXISTS correlated subqueries where the correlating predicate is an inequality predicate (any predicate other than '=' which connects outer and inner tables) by flattening them.

## Long Description

Given the following setup and query:

> create table t1 (a char(2), b char(2), c char(2));
> create table t2 (a char(2), b char(2), c char(2));
> create table t3 (a char(2), b char(2), c char(2));
> 
> explain select * from t1 where not exists (select 1 from t2, t3 where t2.a = t3.a and locate(t1.a,t2.a) > 0);

The presence of the predicate 'locate(t1.a,t2.a) > 0' prevents the subquery from being flattened, so for every outer row in t1 we must execute the entire inner query.  Correlated equality predicates such as  t1.a = t2.a already allow flattening and the flattening logic takes advantage of the fact that matching rows between inner and outer tables share the same values between t1.a and t2.a, so avoids pushing the join with t1 into the subquery.  We just execute the subquery without correlation and then group by t2.a, removing duplicate values.  Then a left outer join with t1 is constructed on t1.a = t2.a.  Any result rows that have a null value on the result t2.a column qualify for the NOT EXISTS predicate:

> explain
> select * from t1 where not exists (select 1 from t2, t3 where t2.a = t3.a and t1.a = t2.a);

Plan
//----
Cursor(n=14,rows=20,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=14,totalCost=41.396,outputRows=20,outputHeapSize=121 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=13,totalCost=28.796,outputRows=20,outputHeapSize=121 B,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=12,totalCost=28.796,outputRows=20,outputHeapSize=121 B,partitions=1,parallelTasks=1,preds=[**is null(ExistsFlatSubquery-0-1.A[11:4])**])
        ->  **BroadcastLeftOuterJoin**(n=11,totalCost=28.796,outputRows=20,outputHeapSize=121 B,partitions=1,parallelTasks=1,preds=[**(T1.A[11:1] = ExistsFlatSubquery-0-1.A[11:4])**])
          ->  **Distinct**(n=9,totalCost=12.296,outputRows=16,outputHeapSize=31 B,partitions=1,parallelTasks=1)
            ->  ProjectRestrict(n=8,totalCost=12.296,outputRows=16,outputHeapSize=32 B,partitions=1,parallelTasks=1)
              ->  BroadcastJoin(n=6,totalCost=12.296,outputRows=16,outputHeapSize=32 B,partitions=1,parallelTasks=1,preds=[(T2.A[6:1] = T3.A[6:2])])
                ->  TableScan[T3(5760)](n=4,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=32 B,partitions=1,parallelTasks=1)
                ->  TableScan[T2(5744)](n=2,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=20 B,partitions=1,parallelTasks=1)
          ->  ProjectRestrict(n=1,totalCost=4.04,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)
            ->  TableScan[T1(5728)](n=0,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)

Inequality joins can't avoid the join, but they can still be flattened by pushing the join with t1 into the subquery, and then instead of grouping by t2.a, group by t1.a, then remove duplicates for a subsequent left outer join back to the outer t1 table on outer_t1.a = inner_t1.a.  This is the functionality added by this pull request:

> explain select * from t1 where not exists (select 1 from t2, t3 where t2.a = t3.a and locate(t1.a,t2.a) > 0);

Plan
//----
Cursor(n=18,rows=20,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=18,totalCost=329.335,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=17,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=16,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1,preds=[**is null(ExistsFlatSubquery-0-1.A[15:4])**])
        ->  BroadcastLeftOuterJoin(n=15,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1,preds=[**(ExistsFlatSubquery-0-1.A[15:4] = T1.A[15:1])**])
          ->  **Distinct**(n=13,totalCost=279.807,outputRows=113,outputHeapSize=319 B,partitions=1,parallelTasks=1)
            ->  ProjectRestrict(n=12,totalCost=279.807,outputRows=113,outputHeapSize=321 B,partitions=1,parallelTasks=1)
              ->  NestedLoopJoin(n=10,totalCost=279.807,outputRows=113,outputHeapSize=321 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=9,totalCost=4.04,outputRows=7,outputHeapSize=321 B,partitions=1,parallelTasks=1,preds=[**(LOCATE(T1.A[8:1], T2.A[7:1], 1)  > 0)**])
                  ->  TableScan[**T1**(5728)](n=8,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=321 B,partitions=1,parallelTasks=1)
                ->  BroadcastJoin(n=6,totalCost=12.296,outputRows=16,outputHeapSize=32 B,partitions=1,parallelTasks=1,preds=[(T2.A[6:1] = T3.A[6:2])])
                  ->  TableScan[T3(5760)](n=4,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=32 B,partitions=1,parallelTasks=1)
                  ->  ProjectRestrict(n=3,totalCost=4.04,outputRows=20,outputHeapSize=20 B,partitions=1,parallelTasks=1)
                    ->  TableScan[T2(5744)](n=2,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=20 B,partitions=1,parallelTasks=1)
          ->  ProjectRestrict(n=1,totalCost=4.04,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)
            ->  TableScan[**T1**(5728)](n=0,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)

To process this, a new predicate type is introduced in ExistsSubqueryWhereVisitor: CorrelatedInequalityBronPredicate, which will match any BinaryRelationalOperatorNode which has exactly one correlated column in it at a level one up from the subquery, and any other columns in the predicate are at the subquery level.  A new CorrelatedColRefCollectingVisitor is written to perform the work of collecting and checking column references in the predicate using this rule.  GroupByUtil.addGroupByNodes is trained to process the found correlated column from a CorrelatedInequalityBronPredicate, clone the corresponding outer table and add this copy to the subquery's FROM list.  It also builds the outer_t1.a = inner_t1.a join-back predicate and places it into the correlatedSubqueryPreds list, which is also used in the CorrelatedEqualityBronPredicate case.  There is pre-existing logic which takes predicates in this list and transforms them into the ExistsFlatSubquery-0-1.A[15:4] = T1.A[15:1] and is null(ExistsFlatSubquery-0-1.A[15:4]) predicates seen in the explain above, so this functionality is re-used.  

There was a prior restriction that there may be only one table in the outer FROM clause of NOT EXISTS queries in order to flatten them because there was no logic to identify the outer table to remove from the FROM clause after internally building a left outer join to replace it.  Logic to find the table to remove is added in ExistsSubqueryFlatteningVisitor.addFromSubquery via new method getFromTableContainingOuterColumnReference.

A new database property is added to disable subquery flattening for testing purposes, or if there is a bug in the field:

> call syscs_util.syscs_set_global_database_property('splice.optimizer.disableSubqueryFlattening', true);

## How to test

Create an EXISTS or NOT EXISTS correlated subquery with at least 2 tables in the subquery, where the correlated predicate is not an equality predicate.  The EXPLAIN should not indicate a "Subquery" operation.  Use the above database property to compare query results with and without flattening.  Example case:

> create table t1 (a char(2), b char(2), c char(2));
> create table t2 (a char(2), b char(2), c char(2));
> create table t3 (a char(2), b char(2), c char(2));

> explain
> select * from t1 where not exists (select 1 from t2, t3 where t2.a = t3.a and locate(t1.a,t2.a) > 0);

Plan
//----
Cursor(n=18,rows=20,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=18,totalCost=329.335,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=17,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1)
      ->  ProjectRestrict(n=16,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1,preds=[is null(ExistsFlatSubquery-0-1.A[15:4])])
        ->  BroadcastLeftOuterJoin(n=15,totalCost=312.669,outputRows=20,outputHeapSize=137 B,partitions=1,parallelTasks=1,preds=[(ExistsFlatSubquery-0-1.A[15:4] = T1.A[15:1])])
          ->  Distinct(n=13,totalCost=279.807,outputRows=113,outputHeapSize=319 B,partitions=1,parallelTasks=1)
            ->  ProjectRestrict(n=12,totalCost=279.807,outputRows=113,outputHeapSize=321 B,partitions=1,parallelTasks=1)
              ->  NestedLoopJoin(n=10,totalCost=279.807,outputRows=113,outputHeapSize=321 B,partitions=1,parallelTasks=1)
                ->  ProjectRestrict(n=9,totalCost=4.04,outputRows=7,outputHeapSize=321 B,partitions=1,parallelTasks=1,preds=[(LOCATE(T1.A[8:1], T2.A[7:1], 1)  > 0)])
                  ->  TableScan[T1(5776)](n=8,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=321 B,partitions=1,parallelTasks=1)
                ->  BroadcastJoin(n=6,totalCost=12.296,outputRows=16,outputHeapSize=32 B,partitions=1,parallelTasks=1,preds=[(T2.A[6:1] = T3.A[6:2])])
                  ->  TableScan[T3(5808)](n=4,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=32 B,partitions=1,parallelTasks=1)
                  ->  ProjectRestrict(n=3,totalCost=4.04,outputRows=20,outputHeapSize=20 B,partitions=1,parallelTasks=1)
                    ->  TableScan[T2(5792)](n=2,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=20 B,partitions=1,parallelTasks=1)
          ->  ProjectRestrict(n=1,totalCost=4.04,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)
            ->  TableScan[T1(5776)](n=0,totalCost=4.04,scannedRows=20,outputRows=20,outputHeapSize=80 B,partitions=1,parallelTasks=1)


msirek:
IT run:
http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/2094/

msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
```diff
@@ -370,7 +378,15 @@ private FromSubquery addFromSubquery(SelectNode outerSelectNode,
             /*
              * Left join outer tables with new FromSubquery.
              */
-            HalfOuterJoinNode outerJoinNode = nf.buildOuterJoinNode(outerSelectNode.getFromList(), fromSubquery, joinClause);
+            FromTable outerFromTable =
+                getFromTableContainingOuterColumnReference(outerSelectNode.getFromList(),
+                                                           outerColumnReference);
+            if (outerFromTable == null && outerSelectNode.getFromList().size() > 1) {
+                throw StandardException.newException(LANG_INTERNAL_ERROR,
```
ipraznik-splice, line 385: Is there an example, when it can happen, or it's just defense?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
```diff
@@ -381,12 +397,60 @@ private FromSubquery addFromSubquery(SelectNode outerSelectNode,
...
+            // For cases with multiple items in the FROM clause, remove the appropriate
+            // table, otherwise just clear the FROM list.
+            if (outerFromTable == null) {
+                outerSelectNode.getFromList().getNodes().clear();
+            }
+            else {
+                int originalSize = outerSelectNode.getFromList().size();
+                outerSelectNode.getFromList().removeElement(outerFromTable);
+                if (outerSelectNode.getFromList().size() == originalSize)
+                    throw StandardException.newException(LANG_INTERNAL_ERROR,
```
ipraznik-splice, line 409: the same question as above

ipraznik-splice (changes requested):

ascend1 (commented):
I have a general question to the rewriting. For inequality predicates, why `T1` has to be joined twice? It's a valid plan for sure. But I don't understand why it's necessary. For the example `locate(T1.a, T2.a) > 0`, wouldn't it just work if we do that same, joining `T2` and `T3` with all non-correlated predicates, grouping on `T2.a`, and evaluating the correlated predicates in the final join?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/aggregate/AggregateSubqueryFlatteningVisitor.java
```diff
@@ -145,19 +145,27 @@ private void flatten(SelectNode topSelectNode,
...
          * The following lines collect correlated predicates from the subquery where clause while removing them.
          */
+        List<BinaryRelationalOperatorNode> correlatedSubqueryInequalityPreds = new ArrayList<>();
         ValueNode subqueryWhereClause = subquerySelectNode.getWhereClause();
         List<BinaryRelationalOperatorNode> correlatedSubqueryPreds = new ArrayList<>();
         subqueryWhereClause = FlatteningUtils.findCorrelatedSubqueryPredicates(
                 subqueryWhereClause,
                 correlatedSubqueryPreds,
                 new CorrelatedEqualityBronPredicate(topSelectNode.getNestingLevel()));
+        // TODO: Support aggregate subquery flattening with inequality conditions.
```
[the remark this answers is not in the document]
msirek: I've opened DB-12090 to handle this and updated the TODO to mention this Jira.  This is a very useful optimization so I'd like to keep the TODO in place so we don't forget about it.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
msirek: This is mainly just defensive.  It is the job of getFromTableContainingOuterColumnReference to locate the outer table that outerColumnReference refers to, if there is more than one outer table (if there is just one, then we don't need to specifically locate the outer table, because there is just one).  If we can't locate the outer table, we don't know which one to remove from the WHERE clause so we need to error out.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
msirek: Yes, this is defensive.   If for whatever reason we could not remove the table from the FROM list, that means perhaps that outerFromTable is pointing to an object not in the FROM list, which should never happen.  We want to catch this case if it does happen, otherwise the query would return incorrect results.

msirek:
> I have a general question to the rewriting. For inequality predicates, why `T1` has to be joined twice? It's a valid plan for sure. But I don't understand why it's necessary. For the example `locate(T1.a, T2.a) > 0`, wouldn't it just work if we do that same, joining `T2` and `T3` with all non-correlated predicates, grouping on `T2.a`, and evaluating the correlated predicates in the final join?
> 
> What I meant is that when grouping the subquery join results, we don't have to rely on equality predicates. We group on all subquery columns that are referenced in correlated predicates. Something like:
> 
> `T1 join exists subq(T2 join T3 on T2.a = T3.a and T1.b > T2.b and T1.c < T3.c)`
> 
> We can do `e = T2 join T3 on T2.a = T3.a`, then `e' = GroupBy(e, T2.b, T3.c)`, and finally `T1 join e' on T1.b > e'.b and T1.c < e'.c`.
> 
> Am I missing some cases?

Hi @ascend1, I was looking to see how to do this at one point.  This is not straightforward because we need to guarantee that at most one row from the outer table be returned.  When pushing the join with the outer table into the subquery and doing a join-back with an equality condition, since the subquery has duplicates removed, the equality condition guarantees each outer row is returned only once.  But if we apply the inequality predicate in the last join it may return a given outer table row multiple times.  I have prototyped the code for the no-push approach (same as you outlined above) in branch DB-11925_NoPushJoin if you are curious.  Running Subquery_Flattening_Exists_IT gets a bunch of failures.  If there is a way to teach the optimizer to apply the generated predicate via an inclusion join or left outer inclusion join instead of a regular inner or left outer join, then I think this approach would work.  I tried some experimental changes with no success so I think this requires extra effort.  I have opened the following Jira to investigate it:

https://splicemachine.atlassian.net/browse/DB-12092

ipraznik-splice (approved):

ascend1:
> > I have a general question to the rewriting. For inequality predicates, why `T1` has to be joined twice? It's a valid plan for sure. But I don't understand why it's necessary. For the example `locate(T1.a, T2.a) > 0`, wouldn't it just work if we do that same, joining `T2` and `T3` with all non-correlated predicates, grouping on `T2.a`, and evaluating the correlated predicates in the final join?
> > What I meant is that when grouping the subquery join results, we don't have to rely on equality predicates. We group on all subquery columns that are referenced in correlated predicates. Something like:
> > `T1 join exists subq(T2 join T3 on T2.a = T3.a and T1.b > T2.b and T1.c < T3.c)`
> > We can do `e = T2 join T3 on T2.a = T3.a`, then `e' = GroupBy(e, T2.b, T3.c)`, and finally `T1 join e' on T1.b > e'.b and T1.c < e'.c`.
> > Am I missing some cases?
> 
> Hi @ascend1, I was looking to see how to do this at one point. This is not straightforward because we need to guarantee that at most one row from the outer table be returned. When pushing the join with the outer table into the subquery and doing a join-back with an equality condition, since the subquery has duplicates removed, the equality condition guarantees each outer row is returned only once. But if we apply the inequality predicate in the last join it may return a given outer table row multiple times. I have prototyped the code for the no-push approach (same as you outlined above) in branch **DB-11925_NoPushJoin** if you are curious. Running Subquery_Flattening_Exists_IT gets a bunch of failures. If there is a way to teach the optimizer to apply the generated predicate via an inclusion join or left outer inclusion join instead of a regular inner or left outer join, then I think this approach would work. I tried some experimental changes with no success so I think this requires extra effort. I have opened the following Jira to investigate it:
> 
> https://splicemachine.atlassian.net/browse/DB-12092
> 
> One limitation of the no-push approach is that the inequality predicate must have at most one column from the outer table and one column from the inner table. With the push join approach, this restriction only applies to the outer table and there can be multiple inner table column references..

That makes sense. I think we are on the same page. Since it's for (NOT) EXISTS subqueries, I automatically think the final join between `T1` and the subquery is an inclusion (semi/anti) join. So I thought the problem of having duplicate rows from `T1` is not a thing here. I thought this can be done in flattening an EXISTS subquery through setting `existsTable` flag or maybe directly set join type. Maybe I understand the flag wrongly?

Regarding the approach of joining `T1` twice, maybe we can add simple checks on `T1` for row uniqueness. E.g., if there is a PK defined on `T1`, or some column has a unique constraint. In those cases, we know rows in `T1` are unique, so we don't have to join `T1` twice.

A more aggressive approach is to use `T1`'s ROWID since it's always unique for rows. And we include `T1.ROWID` in grouping columns. Of course, this might require much more effort.

But I think my main question is answered. I'll look again into details of the code.

msirek:
> > > I have a general question to the rewriting. For inequality predicates, why `T1` has to be joined twice? It's a valid plan for sure. But I don't understand why it's necessary. For the example `locate(T1.a, T2.a) > 0`, wouldn't it just work if we do that same, joining `T2` and `T3` with all non-correlated predicates, grouping on `T2.a`, and evaluating the correlated predicates in the final join?
> > > What I meant is that when grouping the subquery join results, we don't have to rely on equality predicates. We group on all subquery columns that are referenced in correlated predicates. Something like:
> > > `T1 join exists subq(T2 join T3 on T2.a = T3.a and T1.b > T2.b and T1.c < T3.c)`
> > > We can do `e = T2 join T3 on T2.a = T3.a`, then `e' = GroupBy(e, T2.b, T3.c)`, and finally `T1 join e' on T1.b > e'.b and T1.c < e'.c`.
> > > Am I missing some cases?
> > 
> > 
> > Hi @ascend1, I was looking to see how to do this at one point. This is not straightforward because we need to guarantee that at most one row from the outer table be returned. When pushing the join with the outer table into the subquery and doing a join-back with an equality condition, since the subquery has duplicates removed, the equality condition guarantees each outer row is returned only once. But if we apply the inequality predicate in the last join it may return a given outer table row multiple times. I have prototyped the code for the no-push approach (same as you outlined above) in branch **DB-11925_NoPushJoin** if you are curious. Running Subquery_Flattening_Exists_IT gets a bunch of failures. If there is a way to teach the optimizer to apply the generated predicate via an inclusion join or left outer inclusion join instead of a regular inner or left outer join, then I think this approach would work. I tried some experimental changes with no success so I think this requires extra effort. I have opened the following Jira to investigate it:
> > https://splicemachine.atlassian.net/browse/DB-12092
> > One limitation of the no-push approach is that the inequality predicate must have at most one column from the outer table and one column from the inner table. With the push join approach, this restriction only applies to the outer table and there can be multiple inner table column references..
> 
> That makes sense. I think we are on the same page. Since it's for (NOT) EXISTS subqueries, I automatically think the final join between `T1` and the subquery is an inclusion (semi/anti) join. So I thought the problem of having duplicate rows from `T1` is not a thing here. I thought this can be done in flattening an EXISTS subquery through setting `existsTable` flag or maybe directly set join type. Maybe I understand the flag wrongly?
> 
> Regarding the approach of joining `T1` twice, maybe we can add simple checks on `T1` for row uniqueness. E.g., if there is a PK defined on `T1`, or some column has a unique constraint. In those cases, we know rows in `T1` are unique, so we don't have to join `T1` twice.
> 
> A more aggressive approach is to use `T1`'s ROWID since it's always unique for rows. And we include `T1.ROWID` in grouping columns. Of course, this might require much more effort.
> 
> But I think my main question is answered. I'll look again into details of the code.

Regarding the existsTable flag, I tried setting this flag and it was not sufficient to trigger inclusion join.  I feel like this needs to be researched more thoroughly so that we generate the inclusion join query as a higher level input to the optimizer in the same manner as a normal inclusion join is produced to ensure it is executed properly, and so that we're guaranteed to always pick the outer table as the left table of the join.  

Regarding avoiding the last join using uniqueness checks, yes this sounds like a good enhancement for EXISTS joins.  NOT EXISTS queries would still need the final join.  

For using ROWID as a grouping column and join-back column.  Yes, other databases do this, and we can too.  With DB-11449 I made it so a BASEROWID = BASEROWID join predicate can be used as a startKey/stopKey in nested loop join and skip over rows, which should make it faster.  I'm not sure how efficient left outer nested loop join is though.  If the outer table of the subquery must be on the left, this may not help because we can only do the primary key lookups on the right table in nested loop join.  I think we may want to right some right outer join versions of join algorithms where the outer table's primary key or index can be probed.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/GroupByUtil.java
```diff
@@ -62,23 +67,94 @@
      * (actually the correlated subquery predicate(s) are removed from the subquery tree before we get here but I left
      * one in the above example for clarity)
      */
-    public static void addGroupByNodes(SelectNode subquerySelectNode,
-                                       List<BinaryRelationalOperatorNode> correlatedSubqueryPreds) throws StandardException {
+    public static ColumnReference addGroupByNodes(SelectNode subquerySelectNode,
```
ascend1, line 70: Bookmark for myself: resume from here.

msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/SubqueryNodeFactory.java
```diff
@@ -105,10 +105,10 @@ public FromSubquery buildFromSubqueryNode(SelectNode outerSelectNode,
      */
     public HalfOuterJoinNode buildOuterJoinNode(FromList outerFromList,
                                                 FromSubquery fromSubquery,
-                                                ValueNode joinClause) throws StandardException {
+                                                ValueNode joinClause,
+                                                FromTable outerFromTable) throws StandardException {
         /* Currently we only attempt not-exist flattening if the outer table has one table. */
-        assert outerFromList.size() == 1 : "expected only one FromList element at this point";
-        QueryTreeNode outerTable = outerFromList.getNodes().get(0);
+        FromTable outerTable = outerFromTable == null ? (FromTable)outerFromList.getNodes().get(0) : outerFromTable;
```
ascend1, line 111: Is the comment still valid after the assertion is removed?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
ascend1: Hmm, now I see that we still assume `outerSelectNode.getFromList().size == 1`. Maybe we can duplicate or move the comment above here because now the check is actually outside of `buildOuterJoinNode`.

ascend1 (approved):
Great change! I made only a few minor comments.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/exists/ExistsSubqueryFlatteningVisitor.java
msirek: Not sure which comment you're referring to.  I saw the following in buildOuterJoinNode:
   
> Currently we only attempt not-exist flattening if the outer table has one table.

Changed the comment to:

>         /* If outerFromTable is null, that means there is only one outer table.
>          * If outerFromTable is not null, it specifies the outer table we are flattening. 
>          */

db-engine/src/main/java/com/splicemachine/db/impl/ast/CorrelatedColRefCollectingVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+        correlatedColumns = new ArrayList<>();
+    }
+
+    @Override
+    public Visitable visit(Visitable node, QueryTreeNode parent) {
+        super.visit(node, parent);
+        List<T> foundColumns = super.getCollected();
+        if (!foundColumns.isEmpty()) {
+            for (Object qtNode:foundColumns) {
+                ColumnReference colRef = (ColumnReference)qtNode;
```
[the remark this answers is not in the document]
msirek: The predicates are processed one-at-a-time, so should not conflict.  If there are duplicate outer columns in the same predicate, then we don't pick flattening.  Your test case did find a bug though, where we may push the outer table more than once, so I added code to check if it's already been pushed and not re-add it.  I added a couple tests to the ITs to check this case.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/GroupByUtil.java
```diff
@@ -62,23 +67,94 @@
...
+            }
+        }
+        else if (!correlatedSubqueryInequalityPreds.isEmpty()) {
+            for (BinaryRelationalOperatorNode bro : correlatedSubqueryInequalityPreds) {
+                CorrelatedColRefCollectingVisitor correlatedColRefCollectingVisitor
+                    = new CorrelatedColRefCollectingVisitor<>(1, outerNestingLevel);
+                bro.accept(correlatedColRefCollectingVisitor);
+                if (correlatedColRefCollectingVisitor.getCollected().isEmpty())
                     throw new IllegalArgumentException("Did not find correlated column ref on either side of BRON");
+                else {
```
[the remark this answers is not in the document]
msirek: Great point.  This would make it easier to see all the code.  I've made the change.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/GroupByUtil.java
```diff
@@ -62,23 +67,94 @@
...
+                        if (tableName == null || schemaName == null) {
+                            fromTable = outerSelectNode.getFromList().getFromTableByTableNumber(correlatedColumnReference.getTableNumber());
+                            correlatedColumnReference.setTableNameNode(fromTable.getTableName());
+                        }
+                        else
+                            fromTable = outerSelectNode.getFromList().
+                                             getFromTableByName(correlatedColumnReference.getTableName(),
+                                                                correlatedColumnReference.getSchemaName(), true);
+                        leftColRef = (ColumnReference)correlatedColumnReference.getClone();
+                        rightColRef = (ColumnReference)correlatedColumnReference.getClone();
```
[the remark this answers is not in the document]
msirek: Change the names to subqueryColRef and outerColumnReference and added this comment:

                // subqueryColRef is the version of the correlatedColumnReference pushed into the join
                //                in the new generated outer_table.col = inner_table.col predicate.
                // outerColumnReference is the version of the correlatedColumnReference
                //                      in the new generated outer_table.col = inner_table.col predicate,
                //                      but cloned so that it has memory separate from the original
                //                      correlated inequality predicate.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/SubqueryNodeFactory.java
msirek: Updated the comment to:

>         /* If outerFromTable is null, that means there is only one outer table.
>          * If outerFromTable is not null, it specifies the outer table we are flattening.
>          */

msirek:

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/SubqueryNodeFactory.java
ascend1: Is it possible to have a case where `outerFromList.size() > 1`?

msirek:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3001/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/subquery/SubqueryNodeFactory.java
msirek: Yes.  This was a severe limitation to flattening before, where we only supported outerFromList.size() == 1.  With DB-11925, this restriction is lifted.

msirek:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5452, opened 2021-05-06
DB-10621 Batch DDL Changes in drop schema cascade

arnaud-splice:
## Short Description
Improve drop schema cascade performance

## Long Description
Drop schema cascade was slow for two reasons:
1. DB-9553 introduced a savepoint in TabInfoImpl.deleteRows, so we would effectively create a child transaction for every single delete that is caused by schema cascade, and we ended up with a transaction nesting level as large as the number of items to drop. We can now remove that savepoint as SICompactionStateMutate prevents us from purging the antitombstone if there is a tombstone with the same timestamp. Removing that savepoint brought `drop schema foo cascade` where foo contains 1000 tables down to 150-200 seconds on standalone. It used to crash before that point.
2. Each time we drop an object, we issue an RPC call to other region servers to notify them of the metadata change. This is especially important for them to invalidate data dictionary cache. Instead, we now group DDLChange into independent messages that are sent to other region servers. This brought `drop schema foo cascade` where foo contains 1000 tables down to  20-30 seconds on standalone.

Example of dropping objects buckets by bucket:
```
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
x                 x                           Bucket1  x
x Bucket2 ┌──────┐x                 ┌──────┐           x
x         │Table │x                 │Table │           x
x         └──────┘x                 └──────┘◄────┐     x
x          ▲    ▲ x                  ▲           │     x
x          │ xxx│xx        xxxxxxxxxx│xxxxxxxxxxx│xxxxxxxxxx
xxxxxxxxxxx│xx  │          x         │           │         x
x  ┌───────┴┐x ┌┴────┐     x      ┌──┴────┐     ┌┴──────┐  x
x  │Trigger │x │View │     x      │Trigger│     │Trigger│  x
x  └────────┘x └─────┘     x      └───────┘     └───────┘  x
x            x  ▲          x   xxxxxxxxxxxxxxxxxxxxxxxxxxxxx
x            x  │          x   x
x            xxx│xxxxxxxxxxx   x
x              ┌┴────┐         x
x              │View │         x
x  Bucket0     └─────┘         x
x                              x
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

```

## How to test
```
create schema s;
create table s.foo0 (a int);
create table s.foo1 (a int);
create table s.foo2 (a int);
create table s.foo3 (a int);
create table s.foo4 (a int);
create table s.foo5 (a int);
create table s.foo6 (a int);
create table s.foo7 (a int);
create table s.foo8 (a int);
...
create table s.foo991 (a int);
create table s.foo992 (a int);
create table s.foo993 (a int);
create table s.foo994 (a int);
create table s.foo995 (a int);
create table s.foo996 (a int);
create table s.foo997 (a int);
create table s.foo998 (a int);
create table s.foo999 (a int);
drop schema s cascade;
```
should now return promptly.

## Follow-up work
DB-12040 tracks future work:
Currently, drop table still drops underlying dependencies automatically, so those dependencies are not included in the grouped DDLChange, so each dependency issues a separate RPC call to other region servers. `drop schema cascade` should explicitly drop all table dependencies before dropping the table


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2877/

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2879/

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2882/

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2885/

arnaud-splice:

arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2889/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DependencyBucketing.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        return buckets;
+    }
+
+    private void propagateLevels(T node) {
+        for (T child : dependencies.get(node)) {
+            if (nodeLevels.get(child) <= nodeLevels.get(node)) {
+                nodeLevels.put(child, nodeLevels.get(node) + 1);
+                maxLevel = Integer.max(nodeLevels.get(node) + 1, maxLevel);
+                propagateLevels(child);
```
dgomezferro, line 77: Is it possible to get a cyclic dependency here? It looks we'd enter infinite recursion if so.

arnaud-splice: We check for cyclic dependencies in dropAllSchemaObjects, so it isn't possible.

dgomezferro (approved):
Great change, nice refactors!

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLConstantOperation.java
```diff
@@ -58,549 +61,568 @@
...
+        } else if (changes.size() > 1) {
+            notifyMetadataChanges(tc, ProtoUtil.createMultiChange(changes.get(0).getTxnId(), changes));
+        }
+    }
+
+    /**
+     * Notify other region servers of potential metadata changes
+     * @param tc
+     */
+    public void notifyMetadataChanges(TransactionController tc, DDLMessage.DDLChange change) throws StandardException {
```
ascend1, line 341: Maybe use singular form in function name like `notifyMetadataChange` since this function deals with one change only. But it's a minor thing.

ascend1 (approved):

ascend1 (approved):
Awesome change!

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DependencyBucketing.java
```diff
@@ -0,0 +1,81 @@
...
+        propagateLevels(parent);
+    }
+
+    List<List<T>> getBuckets() {
+        List<List<T>> buckets = new ArrayList<>(maxLevel + 1);
+        for (int i = 0; i <= maxLevel; ++i) {
+            buckets.add(new ArrayList<>());
+        }
+        for (Map.Entry<T, Integer> level: nodeLevels.entrySet()) {
+            buckets.get(level.getValue()).add(level.getKey());
```
hatyo, line 67: Adding too many nodes increases the size of the PB message stored in ZK, ZooKeeper has [limits on the size](https://zookeeper.apache.org/doc/r3.1.2/api/org/apache/zookeeper/ZooKeeper.html) of the nodes:

>  The maximum allowable size of the data array is 1 MB (1,048,576 bytes). Arrays larger than this will cause a KeeperExecption to be thrown. 

If each table consumes 50 bytes then we would hit this limit with around 20K tables per schema which I don't think is unrealistic in production.

Please test this with larger tables, and think about chopping off the buckets into smaller sizes.

hatyo (approved):
Good work.

arnaud-splice:

carolp-503 (approved):
+1

PR #5454, opened 2021-05-07
DB-11856: upgrade SPLICE_SEQUENCE table

jyuanca:
## Short Description
Identity column value reset to sequence start value after upgrading to a build later than 2003.

## Long Description
Splice stores identity column value as a sequence in SPLICE_SEQUENCES table. The row address of the sequence is the same as the address of the column descriptor stored in sys.syscolumns. DB-10410 rewrites sys.syscolumns table, which changes row address of all column descriptors. However, it does not update the address in SPLICE_SEQUENCES. If a build upgrades to 2003 or later, the relationship between sys.syscolumns and SPLICE_SEQUENCES table is lost, and all identity column's next value are reset to the sequence start value.

## How to test
To test, create a table with identity column and insert rows into it. Upgrade to a build later than 2003. Insert a couple of rows and inspect identity column value. The values should not go back to sequence start value.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
@@ -57,16 +57,20 @@ import com.splicemachine.management.Manager;
 import com.splicemachine.pipeline.Exceptions;
 import com.splicemachine.primitives.Bytes;
 import com.splicemachine.si.api.data.TxnOperationFactory;
+import com.splicemachine.si.constants.SIConstants;
 import com.splicemachine.si.impl.driver.SIDriver;
+import com.splicemachine.storage.*;
 import com.splicemachine.tools.version.ManifestReader;
 import com.splicemachine.utils.SpliceLogUtils;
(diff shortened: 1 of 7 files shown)


jyuanca:

spliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1363,13 +1367,81 @@ public void rewriteDescriptors(int catalogNum, long cloned_conglomerate) throws
         }
     }
 
+    public void rewriteColumnDescriptors(int catalogNum, long cloned_conglomerate) throws StandardException {
```
martinrupp, line 1370: we might want to think about moving these upgrade helper functions e.g. create class SpliceDataDictionaryUpgrade . Having those functions that only do upgrade in between other things of SpliceDataDictionary makes the class hard to understand

martinrupp (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5458, opened 2021-05-10
DB-11979 Use one row to estimate scan cost when row count is zero

ascend1:
## Short Description
This change fixes unfair comparison between PK access path and index access path when estimated rows to be scanned is zero.

## Long Description
The problem is a combination of two:

- gf.gf#='ccce30bb-530c-11ea-a885-005056b28b7b' gives 0 rows according to statistics
- PK access goes to estimateOneRowCost() without calculating scan selectivity

In the end, the cost of index access path using XPK_INCL_DUP_GFLL is estimated using 0 rows, while the cost of primary key access path is estimated using 1 row. For this reason, PK access path has higher cost.

The fix is to use 1 row to estimate scan cost in generateCost() when estimated row count is smaller than 1. Important thing is that we use 1 row only in estimating cost, but not really overwriting 1 row in all cases. Specifically, raw row count is critical in estimating NLJ output row count. The following is a comment left in code but for completeness also copied here:

```
// Set raw row count for join output row count estimation later.
// For issues like DB-11979, we need to bump row count to 1 if estimated row count is 0. Otherwise we may
// select an index that potentially needs index lookup on top because equivalent PK access goes to
// generateOneRowCost(). There we simply use one row everywhere without even looking at selectivities.
// As a result, PK access path (1 row) may have higher cost than index + lookup (0 row).
// However, we cannot simply bump output row count to 1. Consider the following example:
//     T1 (1000 rows) join T2 (1 row) on T1.a = T2.a using NLJ.
// Join predicate is pushed down to T2. Suppose join selectivity is estimated to be 0.01. Since T2 has 1
// row only, estimated row count is then 0.01. If we bump the row count to 1 at this point, NLJ output
// row count estimation later would simply be 1000 x 1 = 1000 rows. The join predicate effectively has
// no impact! Moreover, on a different join order T2 join T1, since predicate is pushed to T1 now, we
// would have 1 x 1000 x 0.01 = 10 rows! In the end, different join orders give different estimates.
// Raw row count is used to fix this issue. By recording 0.01 for T2 (T1 join T2) or 10 for T1 (T2 join
// T1), NLJ output row count is always 10.
scanCost.setRawRowCount(outputRowCount);
```

## How to test
A new IT `testZeroRow` is added. Copying this IT to current code without the fix, running it and it should because it selects index scan + index lookup. With the fix, PK access path should be selected.

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/v1/V1ScanCostEstimator.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/v1/V1ScanCostEstimator.java
@@ -118,7 +118,9 @@ public class V1ScanCostEstimator extends AbstractScanCostEstimator {
         double totalRowCount = scc.baseRowCount();
         assert totalRowCount >= 0 : "totalRowCount cannot be negative -> " + totalRowCount;
         // Rows Returned is always the totalSelectivity (Conglomerate Independent)
-        scanCost.setEstimatedRowCount(Math.round(totalRowCount*totalSelectivity));
+        double outputRowCount = totalRowCount * totalSelectivity;
+        scanCost.setEstimatedRowCount(Math.round(outputRowCount));
+        outputRowCount = scanCost.rowCount();  // >= 1
 
         int numCols = getTotalNumberOfBaseColumnsInvolved();
         if (isIndexOnExpression && numCols == 0) {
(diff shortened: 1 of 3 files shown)


ascend1:

arnaud-splice (approved):

martinrupp (approved):

carolp-503 (approved):
+1

PR #5459, opened 2021-05-11
DB-11969 Throw properly for invalid update syntax set(a,b)=(c,d).

ascend1:
## Short Description
This change improves the error message when an invalid update syntax `set (a,b)=(c,d)` is used.

## Long Description
We currently have three update statement syntax:
1. `update T set a = c, b = d where ...`
2. `update T set (a, b) = (select c, d from ...)`
3. `update T set a = c, b = d from ... where ...`

When the left-hand side of a set clause is surrounded with a pair of parenthesis, right-hand side must be a subquery. This is syntax 2. Before syntax 3 is implemented, it's splice's way of allowing multi-column update from a subquery.

Currently, we don't generally have tuple assignment except for a few cases. In update statement, `set (a, b) = (c, d)` is equivalent to `set a = c, b = d`. For this reason, there is no real need at the moment to implement tuple assignment in update.

A proper error message is needed though.

## How to test
```
create table t (c1 int, c2 int);
insert into t values (1, 2);
update t set (c1, c2) = (c1 + 1, c2 + 1);
```

Before this fix, the update statement would cause the following exception:
```
ERROR XJ001: Java exception: 'java.lang.ClassCastException: com.splicemachine.db.impl.sql.compile.ValueTupleNode cannot be cast to com.splicemachine.db.impl.sql.compile.SubqueryNode'.
ERROR XJ001: Java exception: 'com.splicemachine.db.impl.sql.compile.ValueTupleNode cannot be cast to com.splicemachine.db.impl.sql.compile.SubqueryNode: java.lang.ClassCastException'.
```

After the fix, we have a proper error message:
```
ERROR 42X67: Invalid UPDATE statement syntax.
```

A new test `testUpdateMultipleColumnsInvalidSyntax` is added to `UpdateOperationIT`.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SQLGrammarImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SQLGrammarImpl.java
@@ -600,6 +600,10 @@ class SQLGrammarImpl {
         // even more hacky code.
         // The derived table flattening logic in preprocess will be triggered to
         // flatten the subquery.
+        if (!(subQuery instanceof SubqueryNode)) {
+            throw StandardException.newException(SQLState.LANG_INVALID_UPDATE_STATEMENT,
+                                                 "Expecting RHS to be a subquery when LHS of set clause is a tuple");
+        }
         SubqueryNode subq = (SubqueryNode) subQuery;
         FromTable fromSubq = (FromTable) nodeFactory.getNode(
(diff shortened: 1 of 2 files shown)


ascend1:

spliceci:
TEST SUCCEEDED +1

jyuanca (approved):

carolp-503 (approved):
+1

PR #5462, opened 2021-05-11
DB-11960 Check constant in-list condition in predicates

arnaud-splice:
## Short Description
Evaluate in-list condition at compile time if possible

## Long Description
When the in-list condition is constant, we should evaluate it at compile time to avoid doing a table scan for a predicate that we know to be false anyway

## How to test
```
splice> explain select * from a where 'abc' in ('abcd', 'def', 'ghi');
Plan
Cursor(n=3,rows=1,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=3,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=2,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
      ->  Values(n=0,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
```
Note that the absence of a table scan and the totalCost=0


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2900/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

martinrupp (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/InListOperatorNode.java
```diff
@@ -917,52 +917,81 @@ public void generateStartStopKey(boolean isAsc, boolean isStartKey,
...
+        if (!(getLeftOperand() instanceof ConstantNode)) {
+            return this;
+        }
+        if (!getRightOperandList().containsAllConstantNodes()) {
+            return this;
+        }
+
+        DataValueDescriptor lhs = ((ConstantNode) getLeftOperand()).getValue();
+        if (lhs == null || lhs.isNull()) {
+            return this;
```
hatyo, line 983: I think we can return `false` here, right?
i.e.
`return (ValueNode) getNodeFactory().getNode(C_NodeTypes.BOOLEAN_CONSTANT_NODE, false, getContextManager());`

hatyo (changes requested):
If the LHS is `null` I am guessing you can also return `false` immediately, e.g. `null in (1,2,3) => false`. Could you please check this case?

arnaud-splice:
> If the LHS is `null` I am guessing you can also return `false` immediately, e.g. `null in (1,2,3) => false`. Could you please check this case?

That is not the case. 
If this is the top node of a predicate, then `null in (1,2,3)` will indeed evaluate to false, because a null predicate is a false predicate.
However, if not the top node of a predicate, `null in (1,2,3)` should evaluate to null.
We could return null directly here, but it leads to complications that are addressed by [DB-12077](https://github.com/splicemachine/spliceengine/pull/5474)

hatyo (approved):

carolp-503 (approved):
+1

PR #5467, opened 2021-05-13
DB-11960 Check constant in-list condition in predicates (3.1)

arnaud-splice:
## Short Description
Evaluate in-list condition at compile time if possible

## Long Description
When the in-list condition is constant, we should evaluate it at compile time to avoid doing a table scan for a predicate that we know to be false anyway

## How to test
```
splice> explain select * from a where 'abc' in ('abcd', 'def', 'ghi');
Plan
Cursor(n=3,rows=1,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=3,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
    ->  ProjectRestrict(n=2,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
      ->  Values(n=0,totalCost=0,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=1)
```
Note that the absence of a table scan and the totalCost=0

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BooleanConstantNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BooleanConstantNode.java
@@ -130,6 +130,9 @@ public final class BooleanConstantNode extends ConstantNode
      */
     Object getConstantValueAsObject()
     {
+        if (unknownValue) {
+            return null;
+        }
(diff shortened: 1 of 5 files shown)


carolp-503 (approved):
Approved for 3.1

PR #5471, opened 2021-05-14
DB-12064 Fix NPE in code generation for index prefix iteration mode

ascend1:
## Short Description
This change fixes a NPE in code generation related to index prefix iteration mode.

## Long Description
This is a bug happens in generating code in index prefix iteration mode when an equality predicate and a ranged predicate are both used as scan keys.

In code generation of a `FromBaseTable`, `getScanArguments()` is called with an assumption that the predicates pushed down to this table must be sorted in index column order (for PK or index conglomerates). If this is not the case, code generation is simply wrong.

However, `getScanArguments()` eventually calls `PredicateList.generateProbeQualifiers()`. This call sorts predicates according to a heuristic on selectivity:
```
if(numberOfQualifiers>0){
    orderQualifiers();
}
```
This step is fine previously because when we are here, the logic relying on the order of predicates is already finished. We are free to sort them.

With index prefix iteration mode, a new call is added after `getScanArguments()`:
```
...
int nargs=getScanArguments(acb,mb);

mb.callMethod(VMOpcode.INVOKEINTERFACE,null,
              trulyTheBestJoinStrategy.resultSetMethodName(multiProbing),
              ClassName.NoPutResultSet,nargs);
generateIndexPrefixIteratorOperation(acb, mb);
```
`generateIndexPrefixIteratorOperation()` internally calls `getScanArguments()` again. Now since predicates are sorted by the last call of `getScanArguments()`, this code generation process could be wrong.

This problem is not observed when there are only equality predicates. In that case, `orderQualifiers()` has no effect on the order because the heuristic is only that:

- = and IS NULL are most selective
- <> and IS NOT NULL are least selective
- all others are in between

As long as we have predicates from more than one category and they are all used in index prefix iteration mode, this problem occurs.

## How to test
A new test `testCodeGenerationNoNPE` is added to `IndexPrefixIterationIT`. Running the query in this test under the IT setup without the fix would throw an NPE. With the fix, the query should run through.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/PredicateList.java
@@ -3781,7 +3781,10 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
          * by (column #, selectivity) once the store does just in time
          * instantiation.
          */
+        PredicateList cloneMe = null;
         if(numberOfQualifiers>0){
+            cloneMe = (PredicateList)getNodeFactory().getNode(C_NodeTypes.PREDICATE_LIST, getContextManager());
+            this.copyPredicatesToOtherList(cloneMe);
             orderQualifiers();
         }
 
@@ -3976,6 +3979,11 @@ public class PredicateList extends QueryTreeNodeVector<Predicate> implements Opt
                 qualNum++;
             }
 
+        }
+
+        if (cloneMe != null) {
+            this.removeAllPredicates();
+            cloneMe.copyPredicatesToOtherList(this);
         }
 
         //assert qualNum==numberOfQualifiers: qualNum+" Qualifiers found, "+ numberOfQualifiers+" expected.";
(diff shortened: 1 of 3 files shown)


ascend1:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2910/

ascend1:

spliceci:
TEST SUCCEEDED +1

msirek (approved):
Great fix!  LGTM

arnaud-splice (approved):

martinrupp (commented):
this is a nit-pick, but in general i would say the "How to test" section should be self-contained as much as possible. Refering to an IT is nice - i hope every PR has a test. However, testing this on the cluster for the QA team means additional work: It means going to the java code of the IT, getting the CREATE TABLE / INSERTs and checking out the (sometimes generated SQL). If you have some easy 5-6 SQL lines to reproduce the behavior (not always given i know), it would be cool to add.

carolp-503 (approved):
+1

PR #5474, opened 2021-05-14
DB-12077 Evaluate null constant restriction to false during execution 

arnaud-splice:
## Short description
Fix issue with null constant restriction during execution

## Long description
We try to remove nulls during constant folding & predicate simplification phase. However, if a folding or a simplification is missing, we would end up with a null constant restriction, which was evaluated to true during the execution. 
This PR fixes the execution problem, and introduces tests that enable/disable constant folding and predicate simplification, so that they do not hide null predicate problems in the future.
We also introduce a new parameter: `splice.database.disableConstantFolding` that echoes `derby.database.disablePredicateSimplification`. Both those parameters are database level parameters for now, but they we should introduce them as session parameters as well.
We bumped junit's version to 4.13.1 in in order to use [BeforeParameter/AfterParameter](https://github.com/junit-team/junit4/pull/1435/files#diff-0f0be60a1387c68e5f84d98b8abd26814e66286d24ce2180624724128174d391) annotations.

## How to test

```
create table t (a int);
insert into t values 1,3,5;
select * from t where upper(cast(null as varchar(10))) = 'a';
```
should not return any rows.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/CompilerContext.java
@@ -183,6 +183,7 @@ public interface CompilerContext extends Context
     boolean DEFAULT_MULTICOLUMN_INLIST_PROBE_ON_SPARK_ENABLED = true;
     boolean DEFAULT_CONVERT_MULTICOLUMN_DNF_PREDICATES_TO_INLIST = true;
     boolean DEFAULT_DISABLE_PREDICATE_SIMPLIFICATION = false;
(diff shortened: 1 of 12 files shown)


arnaud-splice:

spliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2913/

arnaud-splice:

spliceci:
TEST SUCCEEDED +1

msirek (approved):

ipraznik-splice (approved):

carolp-503 (approved):
+1

PR #5478, opened 2021-05-14
DB-12075: null check for namespace

jyuanca:
## Short Description
BackupEndpointObserver.postFlush crashes region server if the namespace of a table is null

## How to test
From hbase shell, create an hbase table with default namespace, insert data to it and flush the table
create 'guru99', {NAME=>'Edu', VERSIONS=>213423443}
put 'guru99', 'r1', 'Edu:c1', 'value', 10
put 'guru99', 'r1', 'Edu:c1', 'value', 15
put 'guru99', 'r1', 'Edu:c1', 'value', 30
flush 'guru99'


arnaud-splice (approved):

arnaud-splice (approved):

bklo94 (approved):

carolp-503 (approved):
+1

PR #5479, opened 2021-05-14
DB-11580 Use Spark Dataset for trigger rows

msirek:
## Short Description
For statement triggers referencing NEW TABLE or OLD TABLE, when executing on Spark, store the temporary trigger rows solely in an in-memory Spark Dataset instead of using temporary conglomerate.

## Long Description
DB-11302 https://github.com/splicemachine/spliceengine/pull/5190 improved performance for statement triggers on Spark.  One of the ways it did this was by remembering the source dataset of the INSERT, UPDATE or DELETE statement that fired the trigger, then using that instead of a temporary congomerate (written to HBase) for accessing NEW TABLE or OLD TABLE rows.
Example trigger:

> CREATE TRIGGER FeatureStoreTest.FeatureTable_History_UPDATE_OOO
> AFTER UPDATE ON FeatureStoreTest.FeatureTable
> REFERENCING NEW_TABLE AS NEWW OLD_TABLE AS OLDW
> FOR EACH STATEMENT
>     INSERT INTO FeatureStoreTest.FeatureTable (entity_key, last_update_ts, feature1, feature2) 
>     SELECT o.entity_key, o.last_update_ts, o.feature1, o.feature2
>     FROM OLDW o INNER JOIN NEWW n ON o.ENTITY_KEY=n.ENTITY_KEY AND 
>                             o.LAST_UPDATE_TS > n.LAST_UPDATE_TS
>     ;

However, it could not detect ahead of time special cases, such as those using nested loop join, or nested triggers, where the conglomerate approach is required for correctness due to execution details, so we still had to always create and write to the temporary conglomerate.  This PR adds detection for cases which can avoid writing the conglomerate, and disallows use of nested loop join for statement triggers with a referencing clause on spark.

Also as part of this PR, the default control execution cutoff where we switch to spark execution is increased from 1 million to 10 million rows.  It is costly to switch to Spark mid-stream because all the previous work has to be thrown away, so increasing the default will help avoid these cases, yet not be so high that we don't over-commit to control execution.

For the DB-11580 test case (available in https://splicemachine.atlassian.net/browse/DB-12032) the average spark runtime I'm seeing on a k8s cluster with 8 spark executors for an UPDATE with the 3 defined triggers is 34 seconds without DB-11580 and 21 seconds with DB-11580, a better than 35% improvement.  On control the average runtime is closer to 80 seconds.

## How to test

> create table t1 (a int, b int);
> create table t2 (a int, b int);
> // The following trigger should be disallowed:
> CREATE TRIGGER mytrig
> AFTER UPDATE OF a,b
> ON t1
> REFERENCING OLD TABLE AS OLD NEW TABLE AS NEW
> FOR EACH STATEMENT
> INSERT INTO T2 SELECT NEW.A, NEW.B from NEW --splice-properties joinStrategy=nestedloop
> , T1 --splice-properties joinStrategy=nestedloop, useSpark=true;


msirek:

msirek:
ITs:
http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/2101/default/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
```diff
@@ -232,6 +232,17 @@ else if (! rightTypeId.isStringTypeId() && leftTypeId.isStringTypeId())
             rightPadBitDataConstantNode((BitConstantNode) getLeftOperand(), (ColumnReference) getRightOperand());
         }
 
+        // TODO: Enable this code to allow native spark joins involving REALs.
```
ipraznik-splice, line 235: What should be done with this TODO? Should it be activated or it's not needed anymore and can be removed?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2790,4 +2794,25 @@ public int getJoinPosition() {
     public CostModel getCostModel() {
         return costModel;
     }
+
+    @Override
+    public void setOuterTableOfJoin(ResultSetNode outerTableOfJoin) {
+        if (outerTableOfJoin == null)
+            this.outerTableOfJoin = null;
+        if (outerTableOfJoin instanceof Optimizable)
+            this.outerTableOfJoin = (Optimizable)outerTableOfJoin;
```
ipraznik-splice, line 2803: If outerTableOfJoin is not null and not instanceof Optimizable, is it ok or should be an Exception?

ipraznik-splice (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
```diff
@@ -2790,4 +2794,25 @@ public int getJoinPosition() {
...
+    @Override
+    public Optimizable getOuterTable() {
+        if (joinPosition == 0) {
+            if (getJoinType() >= INNERJOIN)
+                return outerTableOfJoin;
+            return null;
+        }
+        int propJoinOrder = proposedJoinOrder[joinPosition-1];
+        Optimizable thisOpt = optimizableList.getOptimizable(propJoinOrder);
+        return thisOpt;
```
hatyo, line 2815: So, for join position 0, we only return the outer table for OUTER join types, but if position is larger than 0, we return the previous optimizable in the join order, regardless of the join type, is this intentional?

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
```diff
@@ -788,13 +788,14 @@ public boolean hasGeneratedColumn() {
...
 
-    public boolean needsTemporaryConglomerate() {
-        // TODO:  Can we detect when a DML statement and associated
-        //        triggers needs no temporary conglomerate?
-        //        The persisted Dataset may not always be accessible,
-        //        for example in a nested loop join iterator,
-        //        so currently we still need it.
-        return true;
-        //return hasGeneratedColumn || !fromSparkExecution;
+    public boolean needsTemporaryConglomerate(boolean nestedTrigger) {
```
hatyo, line 791: This comment is not related to CR, but just to get better understanding of triggers logic.

Could you please explain why do we need temporary conglomerate for triggers, and what is the life-cycle of the temporary conglomerate? Are there other cases besides triggers where temp. conglomerates are required?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
```diff
@@ -61,6 +61,14 @@ public boolean feasible(Optimizable innerTable,
         if (outerCost != null && outerCost.getJoinType() == JoinNode.FULLOUTERJOIN)
             return false;
 
+        if (!isSingleTableScan(optimizer)) {
+            if (innerTable.isTriggerVTI())
+                return false;
+            Optimizable outerTable = optimizer.getOuterTable();
+            if (outerTable != null && outerTable.isTriggerVTI())
+                return false;
+        }
```
hatyo, line 70: From the PR commit message:

> and disallows use of nested loop join for statement triggers with a referencing clause on spark.

Do you want to strengthen the condition by adding `optimizer.isForSpark`?

In general, it would be nice to have a comment on why we don't want to use the NLJ in case of triggers (just to quickly read it without having to back to the PR itself :) ).

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/V1ScanCostEstimator.java
```diff
@@ -127,8 +127,9 @@ public void generateCost(long numFirstIndexColumnProbes) throws StandardExceptio
         double closeLatency = scc.getCloseLatency();
         double localLatency = scc.getLocalLatency();
         double remoteLatency = scc.getRemoteLatency();
+        // IndexPrefixIteration is not currently parallel, so penalize by a factor of the parallelism.
         double remoteCost = (openLatency + closeLatency) +
-                (numFirstIndexColumnProbes*2)*remoteLatency*(1+colSizeFactor/1024d) +
+                (numFirstIndexColumnProbes*2*scc.getParallelism())*remoteLatency*(1+colSizeFactor/1024d) +
```
hatyo, line 132: are these changes to scan cost relevant to this PR, if not, could you please revert them and create a special PR for them? (this would be nice so we can adapt to these changes in V2 cost model when we bring DB-11931 in.

hatyo (commented):
Nice work @msirek! few comments about triggers, I also noticed some changes w.r.t. IndexPrefixIterator which doesn't seem relevant to this PR, but maybe I am wrong.

msirek:
> Nice work @msirek! few comments about triggers, I also noticed some changes w.r.t. IndexPrefixIterator which doesn't seem relevant to this PR, but maybe I am wrong.

Thanks!  The IndexPrefixIterator isn't specific to triggers, but is specific to the DB-11580 test case.  SPLICE-2399 is slow when there are a large number of values in the first column of the index, and it is affecting the useability of feature store.  I will address the root cause, maybe by writing a custom HBase filter, through DB-11930.  But in the meantime I need costing to reflect the slow performance so they are not impacted.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryComparisonOperatorNode.java
msirek: @ipraznik-splice I just added this comment and example code as a reminder to do something like this in the future so we trigger more native spark joins.  Non-native spark can be very slow, but sometimes spark returns different results than splice, maybe due to storing Floats with a different amount of precision than us, or hashing Floats different from Doubles.  Adding this cast would cause the containsUnsafeSQLRealComparison method to return false in join code (such as BroadcastJoinOperation), and we'd pick up native spark join.  The downside is, the cast can remove our ability to use a predicate for index access, so I don't want to add it without seeing if there is a better way to do this.  But I want to leave it in as a reminder of work to be done.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: This is a good question.  I have modified setOuterTableOfJoin to save the outer table, even if it is not an Optimizable.  A trigger VTI will always be an Optimizable.  But this general-purpose function should always record the outer table, even if it is not an Optimizable.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
msirek: This is intentional.  There are two flows to costing a join.  If the syntax 'JOIN ON' is used to join tables and the join cannot be flattened (JoinNode.flatten), then a JoinNode is generated and JoinNode.optimizeSource is called on the right (inner) table.  In this case, the joinPosition will always be zero and joinType reflects if it is an inner join or a type of outer join.  If the join is specified in the FROM clause via comma syntax: FROM t1,t2,t3  ; or if the join is flattened, then SelectNode.findBestPlan is called which iterates through all of the Optimizables in the FROM clause, trying all different combinations.  When joinPosition is zero, we are not planning a join, but a single-table scan.  When joinPosition is greater than 0, then the Optimizable at joinPosition - 1  is the outer table and the Optimizable at joinPosition is the inner table of the join.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
msirek: We have to disable nested loop join even if not for spark.  I've added the following comment explaining why:

>     // Nested loop join on Spark does not work correctly when using a common
>     // Dataset to access the trigger REFERENCING NEW/OLD TABLE rows:
>     //         see useCommonDataSet in TriggerNewTransitionRows.
>     // The compilation of a trigger is saved as a stored prepared statement in
>     // the data dictionary, and reloaded/reused by each new triggering statement.
>     // Even if the trigger is compiled to run in OLTP mode, if the triggering
>     // statement runs in OLAP, the trigger must run in OLAP too.  Since we
>     // cannot tell from the SPSDescriptor whether the trigger was compiled
>     // to run on OLTP or OLAP, we would not be able to detect when an
>     // OLTP-compiled trigger which uses nested loop join would need to be
>     // recompiled as forced-OLAP, and avoid choosing nested loop join.
>     // Therefore we must always avoid nested loop join for statement triggers
>     // with a REFERENCING clause, even if compiled for OLTP execution.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/V1ScanCostEstimator.java
msirek: This is relevant because the test case runs slowly without it.  It is meant only as a temporary change, and I would revert it as soon as I'm done with DB-11930 (which I'm currently working on).  If you don't mind, I'd rather spend the time working on the root cause, DB-11930, instead of spending extra time to separate this temporary change out when it would be reverted soon anyway.

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
msirek: When running on control, the temporary trigger rows can simply be cached in memory (see rowArray in TriggerRowHolderImpl) because the entire execution runs in a single process on the same node (even if using multiple threads).  But on Spark, the triggering statement (insert, update, delete) runs in different spark executors on different nodes.  There really is no good way to collect all of these rows together in a cache managed by Splice (maybe you could do it via Zookeeper, but perhaps this would not perform well).  Using a spark Dataset makes sense, but requires that the trigger actions be invoked from the OlapServer.  For trigger actions marked as OLTP, they may cause trigger statements to be invoked from the executors, which do not have access to the Dataset collected from the OlapServer.  So, the original implementation creates a temporary conglomerate to store the rows.  Currently only spark uses the temporary conglomerate, though code-wise OLTP execution could also use it, though for performance reasons and to avoid potential bugs due to extra complexity, we only uses the in-memory cache on OLTP.  If we go above the OLTP execution threshold (1 million rows, bumped up to 10 million rows by this PR), the trigger execution is rolled back by throwing a ResubmitDistributedException and we re-run it in spark.  The temporary trigger conglomerate is always removed via TriggerRowHolderImpl.close -> dropConglomerate.  I don't know if any other features use a temporary conglomerate.

ipraznik-splice (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
carolp-503: Would this be a good comment in the code?

msirek: @carolp-503 I've added this comment, slightly modified, to the top of TriggerRowHolderImpl.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/TriggerRowHolderImpl.java
```diff
@@ -265,6 +292,8 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
         CID = in.readLong();
         conglomCreated = in.readBoolean();
         resultDescription = (ResultDescription) in.readObject();
+        needsTemporaryConglomerate = in.readBoolean();
+        needsTemporaryConglomerateSet = in.readBoolean();
```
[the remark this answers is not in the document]
msirek: This is just to make sure any decision made in the OlapServer about needing the temporary conglomerate would get reflecting in any execution happening in the executors.  It may not be needed at all, but is there for consistency.

carolp-503 (approved):
+1

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/TriggerRowHolderImpl.java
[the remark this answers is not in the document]
ascend1: I see. Thanks for explaining!

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OptimizerImpl.java
hatyo: Great explanation, thank you

db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
hatyo: Thanks Mark for taking the time explaining this.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
hatyo: Great!

PR #5481, opened 2021-05-16
DB-11523 sort merge join infeasible for correlated subq. join condition

hatyo:
## Short Description
 In this PR the sort merge join strategy is restricted from handling left and full outer joins containing a non-flattened correlated sub-query in the join condition for OLAP.

### 1. The bug

To understand the root cause of this issue, we should have a look at the generated code logic and the actual execution model of sort merge join in Spark.

#### Generated code
The generated code for evaluating a non-flattened correlated subquery constructs a `ScalarAggregateOperation` which performs a `ScalarAggregateFlatMapFunction` on `TableScannerIterator`, the table scanner itself defines a set of row qualifiers that filters in the rows the satisfy the correlation. For example, if the query is:

```
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;
```

Then the row qualifier in the scanner will filter in rows that satisfy the condition `SUBQ.b = ALPHA.b`. The generated code assumes that the row from `ALPHA` is cached in the `Activation` itself (from a previous scan of the outer table).

#### Spark execution model for sort merge join 
Sort merge join uses internally Spark's `cogroup` operation to execute the join in OLAP mode. If we have a correlated non-flattened sub-query in the join condition, then the operation will be running into steps:

1. `cogroup` left and right tables
2. apply `flatMap` using the `CogroupFullOuterJoinRestrictionFlatMapFunction` to qualify joined rows.

Note that `cogroup` is a shuffle-inducing operation, during which, new `Activation`(s) will be created for each task doing the cogrouping and writing into their `Activation`s the result of scanning both tables which is necessary for the next step.

In `CogroupFullOuterJoinRestrictionFlatMapFunction` each worker thread will attempt to qualify the joined rows, in case of correlated non-flattened subquery, the thread will execute the sub-query, while doing so, it will qualify the rows it scanned with the outer table's current row (the correlation).

This is where the logic fails, because at step 1, due to the `cogroup` operation (and its shuffling), the worker threads have written their data into a different `Activation` instance, so in the next step, when the subquery is evaluated the worker thread will not find the outer table rows in its `Activation` causing an NPE.

### 2. The fix

Since the sort merge join relies on Spark's internal `cogroup` function to perform the actual join (in OLAP), I limited the scope of sort merge join applicability for that particular case of having correlated non-flattened subquery in join condition. To do this, I collect all the join condition predicates (pushable and non-pushable, see [1]) and check if any of these predicates has a correlated subquery, if so, we mark the sort merge join strategy as infeasible making the optimizer skip over it.

[1] I noticed that in when costing a join we don't include any non-pushable join predicate in the costing process, this seems like a bug since it might lead to underestimation of the join's true cost and it needs further investigation.

## How to test

Here is an example:

```sql
MAXIMUMDISPLAYWIDTH 0;
drop table t1 if exists;
drop table t2 if exists;

create table t1 (a int, b int);
create table t2 (a int, b int, primary key(a));
 
insert into t1 values (155,540);
insert into t2 select a, b from t1;

analyze table t1;
analyze table t2;

-- previously we got NPE when executing this statement:
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;

-- with this fix, we prevent the optimizer from choosing the sort merge join strategy for this type of query:
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;
```


hatyo:

spliceci:
TEST SUCCEEDED +1

martinrupp (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
```diff
@@ -1628,4 +1628,17 @@ public boolean isVarcharBinaryRelationalOperator() {
         return bron.getLeftOperand().getTypeServices().
                                      getTypeName().equals(TypeId.VARCHAR_NAME);
     }
+
+    @Override
+    public boolean hasCorrelatedSubquery() throws StandardException {
+        CollectNodesVisitor visitor= new CollectNodesVisitor(SubqueryNode.class);
```
msirek, line 1634: It may be expensive to scan the Predicate multiple times throughout join planning every time we check if a join is feasible.  I suggest we cache the result of this check in the Predicate object and re-use that result on subsequent calls to hasCorrelatedSubquery.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeSortJoinStrategy.java
```diff
@@ -72,4 +74,24 @@ public JoinStrategyType getJoinStrategyType() {
     public int maxCapacity(int userSpecifiedCapacity,int maxMemoryPerTable,double perRowUsage){
         return Integer.MAX_VALUE;
     }
+
+    private boolean correlatedSubqueryRestriction(Optimizer optimizer,
+                                                  OptimizablePredicateList predList,
+                                                  int joinType) throws StandardException {
+        if (optimizer.isForSpark()
+            && (joinType == LEFTOUTERJOIN || joinType == FULLOUTERJOIN)
```
msirek, line 82: Why is the fix limited to outer joins?  Inner join also does a shuffle.  Do we need to hint non-native spark to see the problem for inner joins?

msirek (changes requested):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
hatyo: Good point, I added caching logic.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/MergeSortJoinStrategy.java
hatyo: In case of inner join, the subquery is not part of the join predicate itself and it is pulled as a restriction of a PRN on top of the join, as shown in the plan below, that's why it is OK, but I think your point is nevertheless valid, if, for some reason, the subquery becomes part of the join condition, then we will definitely run into this problem if we choose sort-merge (merge-sort?) join strategy. I apply this restriction for all types of joins.

```
explain SELECT * FROM 
t1
where a in (
select ALPHA.a from
t1 ALPHA  
INNER JOIN 
t2 BETA --splice-properties joinstrategy=sortmerge, useSpark=True
ON BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) --spark-properties doNotFlatten=true
AND ALPHA.b=BETA.b
);

Cursor(n=16,rows=1,updateMode=READ_ONLY (1),engine=OLAP (query hint))
  ->  ScrollInsensitive(n=16,totalCost=8.011,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3)
    ->  ProjectRestrict(n=15,totalCost=4.001,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3,preds=[is not null(subq=14)])
      ->  Subquery(n=14,totalCost=10.682,outputRows=1,outputHeapSize=16 B,partitions=1,parallelTasks=3,correlated=false,expression=false,invariant=true)
        ->  ProjectRestrict(n=14,totalCost=8.008,outputRows=1,outputHeapSize=16 B,partitions=1,parallelTasks=3,preds=[(A[1:1] = ALPHA.A[13:1])])
          ->  ProjectRestrict(n=13,totalCost=8.008,outputRows=1,outputHeapSize=16 B,partitions=1,parallelTasks=3,preds=[(BETA.A[7:3] = subq=12)])
            ->  Subquery(n=12,totalCost=17.347,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=3,correlated=true,expression=true,invariant=true)
              ->  ProjectRestrict(n=12,totalCost=13.337,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=3)
                ->  GroupBy(n=11,totalCost=13.337,outputRows=1,outputHeapSize=0 B,partitions=1,parallelTasks=3)
                  ->  ProjectRestrict(n=10,totalCost=4.001,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3)
                    ->  TableScan[T2(1664)](n=8,totalCost=4.001,scannedRows=1,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3,preds=[(SUBQ.B[8:2] = ALPHA.B[7:2])])
            ->  MergeSortJoin(n=6,totalCost=8.008,outputRows=1,outputHeapSize=16 B,partitions=1,parallelTasks=3,preds=[(ALPHA.B[6:2] = BETA.B[6:4])])
              ->  TableScan[T2(1664)](n=4,totalCost=4.001,scannedRows=1,outputRows=1,outputHeapSize=16 B,partitions=1,parallelTasks=3)
              ->  TableScan[T1(1648)](n=2,totalCost=4.001,scannedRows=1,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3)
      ->  ProjectRestrict(n=1,totalCost=4.001,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3)
        ->  TableScan[T1(1648)](n=0,totalCost=4.001,scannedRows=1,outputRows=1,outputHeapSize=8 B,partitions=1,parallelTasks=3)

```

hatyo:

msirek (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/Predicate.java
```diff
@@ -1628,4 +1633,23 @@ public boolean isVarcharBinaryRelationalOperator() {
         return bron.getLeftOperand().getTypeServices().
                                      getTypeName().equals(TypeId.VARCHAR_NAME);
     }
+
+    @Override
+    public boolean hasCorrelatedSubquery() throws StandardException {
+        if(hasCorrelatedSubqerySet) {
+            return hasCorrelatedSubquery;
+        }
+        hasCorrelatedSubqerySet = true;
```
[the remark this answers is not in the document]
hatyo: resolved.

hatyo:

carolp-503 (approved):
+1

hatyo:

PR #5492, opened 2021-05-21
DB-11523 sort merge join infeasible for correlated subq. join condition (redux + fix)

hatyo:
## Short Description
 In this PR the sort merge join strategy is restricted from handling left and full outer joins containing a non-flattened correlated sub-query in the join condition for OLAP.

### 1. The bug

To understand the root cause of this issue, we should have a look at the generated code logic and the actual execution model of sort merge join in Spark.

#### Generated code
The generated code for evaluating a non-flattened correlated subquery constructs a `ScalarAggregateOperation` which performs a `ScalarAggregateFlatMapFunction` on `TableScannerIterator`, the table scanner itself defines a set of row qualifiers that filters in the rows the satisfy the correlation. For example, if the query is:

```
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;
```

Then the row qualifier in the scanner will filter in rows that satisfy the condition `SUBQ.b = ALPHA.b`. The generated code assumes that the row from `ALPHA` is cached in the `Activation` itself (from a previous scan of the outer table).

#### Spark execution model for sort merge join 
Sort merge join uses internally Spark's `cogroup` operation to execute the join in OLAP mode. If we have a correlated non-flattened sub-query in the join condition, then the operation will be running into steps:

1. `cogroup` left and right tables
2. apply `flatMap` using the `CogroupFullOuterJoinRestrictionFlatMapFunction` to qualify joined rows.

Note that `cogroup` is a shuffle-inducing operation, during which, new `Activation`(s) will be created for each task doing the cogrouping and writing into their `Activation`s the result of scanning both tables which is necessary for the next step.

In `CogroupFullOuterJoinRestrictionFlatMapFunction` each worker thread will attempt to qualify the joined rows, in case of correlated non-flattened subquery, the thread will execute the sub-query, while doing so, it will qualify the rows it scanned with the outer table's current row (the correlation).

This is where the logic fails, because at step 1, due to the `cogroup` operation (and its shuffling), the worker threads have written their data into a different `Activation` instance, so in the next step, when the subquery is evaluated the worker thread will not find the outer table rows in its `Activation` causing an NPE.

### 2. The fix

Since the sort merge join relies on Spark's internal `cogroup` function to perform the actual join (in OLAP), I limited the scope of sort merge join applicability for that particular case of having correlated non-flattened subquery in join condition. To do this, I collect all the join condition predicates (pushable and non-pushable, see [1]) and check if any of these predicates has a correlated subquery, if so, we mark the sort merge join strategy as infeasible making the optimizer skip over it.

[1] I noticed that in when costing a join we don't include any non-pushable join predicate in the costing process, this seems like a bug since it might lead to underestimation of the join's true cost and it needs further investigation.

## How to test

Here is an example:

```sql
MAXIMUMDISPLAYWIDTH 0;
drop table t1 if exists;
drop table t2 if exists;

create table t1 (a int, b int);
create table t2 (a int, b int, primary key(a));
 
insert into t1 values (155,540);
insert into t2 select a, b from t1;

analyze table t1;
analyze table t2;

-- previously we got NPE when executing this statement:
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;

-- with this fix, we prevent the optimizer from choosing the sort merge join strategy for this type of query:
SELECT * FROM t1 ALPHA LEFT OUTER JOIN 
t2 BETA --splice-properties useSpark=True, joinStrategy=sortmerge
ON ALPHA.b=BETA.b AND BETA.a = (SELECT MAX(A) FROM t2 SUBQ WHERE SUBQ.b=ALPHA.b) ;
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/OptimizablePredicate.java
@@ -204,4 +204,6 @@ public interface OptimizablePredicate
 	// of one of the OrNodes (OrNode chains are right-deep, with each left child being
 	// something other than an OrNode, if normalized properly).
 	List<OptimizablePredicateList> separateOredPredicates() throws StandardException;
+
+	boolean hasCorrelatedSubquery() throws StandardException;
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizer.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/compile/Optimizer.java
@@ -43,6 +43,7 @@ import com.splicemachine.db.impl.sql.compile.FromBaseTable;
 import com.splicemachine.db.impl.sql.compile.GroupByList;
 import com.splicemachine.db.impl.sql.compile.OrderByList;
 import com.splicemachine.db.impl.sql.compile.ResultSetNode;
+import com.splicemachine.db.impl.sql.compile.PredicateList;
 
 import java.util.List;
 
@@ -439,4 +440,12 @@ public interface Optimizer{
 
     FromBaseTable getOuterBaseTable() throws StandardException;
 
+
+    void setNonPushablePredicates(PredicateList predicateList);
+
+    /**
+     * Postcondition @return != null.
+     * @return the non-pushable predicate list
+     */
+    PredicateList getNonPushablePredicates();
 }
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FullOuterJoinNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FullOuterJoinNode.java
@@ -139,22 +139,7 @@ public class FullOuterJoinNode extends JoinNode {
 		 * conditions only get pushed down 1 level.
 		 * We use the optimizer's logic for pushing down join clause here.
 		 */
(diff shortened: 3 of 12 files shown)


hatyo:

dgomezferro (approved):

hatyo:

msirek (approved):

carolp-503 (approved):
+1

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2982/

hatyo:

hatyo:
Test failed due to timeout in `com.splicemachine.triggers.Trigger_Referencing_Clause_IT`. Not related to this PR, I also tried it locally without being able to reproduce it. Running the suite again.

cloudspliceci:
TEST SUCCEEDED +1

PR #5493, opened 2021-05-21
DB-10980 Wait for real exception when an OLAP job fails

ascend1:
## Short Description
This change fixes an exception propagation issue that causes test `ProjectRestrictSparkExpressionIT.testBugs` to be sporadic.

### Problem description
Submitting a job to OLAP is asynchronous in our system. On the very top level, this is done in `stmt.execute()` call in `DRDAConnThread.java`, around line 947. When submitting an OLAP job, we also register an handler in case the job's status is changed. This happens at `RemoteQueryClientImpl.java`, around line 132.

After submitting the OLAP job, execution in DRDA thread continues. Eventually, it blocks inside `writeQRYDTA(stmt)` (`DRDAConnThread.java`, around line 986). At this point, we need to block and wait because we are about to read the result set through the iterator. This happens at `getNextRowCore()` in `SpliceBaseOperation.java`.

When we are blocking at `execRowIterator.hasNext()`, the OLAP job could be 1) still running; 2) failed; or 3) finished. Cases 1 and 3 are not interesting in the context of DB-10980.

In case the OLAP job has failed, real exception need to be propagated from Spark to the DRDA thread. This is done through the handler we registered when submitting the job. Eventually, the exception should be stored in `failure` field of `streamListener`, a member inside `remoteQueryClientImpl`.

`execRowIterator.hasNext()` internally checks the `failure` field. If it's set, then it throws the exception directly.

Now comes the problem. Since the OLAP job has failed, the OLAP server could send a "Connection reset by peer" exception directly to the remote query client. There is a race. Whichever exception comes first is reported by client. The other is lost. However, "connection reset by peer" is too general to be useful for understanding what's going wrong. In this case, we need the actual exception thrown from Spark side.

### Solution
When an exception thrown from `execRowIterator.hasNext()` is caught and being processed, we wait until the OLAP job handler is called with a timeout (currently 5 seconds). Once we know that the OLAP job handler is already called, we replace the current caught exception with the one propagated from Spark. In case there is no exception from Spark or the wait has timed out, we continue use the current exception.

### Further question
The OLAP job handler calls `olapFuture.get()` and then sets `streamListener.completed(olapResult)`. In debugging this problem, we see that this handler can be called way later than `execRowIterator.hasNext()` has returned (and the client received something, thus also returned from this query) no matter if the job was a failure or success. In case the job failed, we have the problem described above. If the job succeeded, by the time this handler is invoked, the client has already showed the result to user and possibly already accepted another query. Between this handler and `writeQRYDTA(stmt)`, there is no synchronization at all.

## How to test
- Setup:
  ```
  create table ts_double (a double, b double, c int);
  insert into ts_double values (1.79769E+308, 1.49769E+308, 1);
  ```
- Query:
  ```
  -- query 1
  select nullif(a+a, b+b) from ts_double --splice-properties useSpark=true
  ;
  -- query 2
  select case when a is not null and a > 1 then a else a+a end from ts_double --splice-properties useSpark=true
  ;
  ```
- Before this fix, query 1 should give one of the following exceptions sporadically:
  - ERROR XJ001: Java exception: 'Connection reset by peer: java.io.IOException'.
  - ERROR 22003: The resulting value is outside the range for the data type DOUBLE.
  Query 2 should be always successful and finishes in a short amount of time.
- After this fix, query 1 should return the following exception deterministically:
  - ERROR 22003: The resulting value is outside the range for the data type DOUBLE.
  Query 2 should still be successful and finishes in a short amount of time. It should never get stuck.

### DIFF
--- a/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
+++ b/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
@@ -42,7 +42,6 @@ import java.util.*;
 import java.util.concurrent.*;
 
 
-
 /**
  * Created by dgomezferro on 5/20/16.
  */
@@ -58,6 +57,7 @@ public class RemoteQueryClientImpl implements RemoteQueryClient {
     private StreamListener streamListener;
     private long offset = 0;
     private long limit = -1;
+    private CountDownLatch olapFutureCallbackInvoked = new CountDownLatch(1);
 
     public RemoteQueryClientImpl(SpliceBaseOperation root, String hostname) {
         this.root = root;
@@ -148,6 +148,7 @@ public class RemoteQueryClientImpl implements RemoteQueryClient {
                         LOG.error("Unexpected exception, shouldn't happen", e);
                         streamListener.failed(e);
                     }
+                    olapFutureCallbackInvoked.countDown();
                 }
             }, MoreExecutors.sameThreadExecutor());
         } catch (IOException e) {
@@ -275,5 +276,10 @@ public class RemoteQueryClientImpl implements RemoteQueryClient {
         streamListener.stopAllStreams();
         if (olapFuture != null)
             olapFuture.cancel(false);
+    }
+
+    public Exception getException() throws InterruptedException {
+        olapFutureCallbackInvoked.await(5, TimeUnit.SECONDS);
+        return (Exception) streamListener.getFailure();
     }
 }
--- a/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
+++ b/hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
@@ -16,6 +16,7 @@ package com.splicemachine.stream;
 
 import com.splicemachine.db.iapi.reference.GlobalDBProperties;
 import com.splicemachine.db.iapi.services.property.PropertyUtil;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import splice.com.google.common.net.HostAndPort;
 import com.splicemachine.EngineDriver;
 import com.splicemachine.access.HConfiguration;
@@ -278,6 +279,7 @@ public class RemoteQueryClientImpl implements RemoteQueryClient {
             olapFuture.cancel(false);
     }
 
+    @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED", justification = "intended")
     public Exception getException() throws InterruptedException {
         olapFutureCallbackInvoked.await(5, TimeUnit.SECONDS);
         return (Exception) streamListener.getFailure();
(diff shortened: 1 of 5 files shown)


ascend1:

ascend1:

ascend1:

martinrupp (approved):
great catch!

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
```diff
@@ -276,4 +278,10 @@ public void close() throws Exception {
         if (olapFuture != null)
             olapFuture.cancel(false);
     }
+
+    @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED", justification = "intended")
+    public Exception getException() throws InterruptedException {
+        olapFutureCallbackInvoked.await(5, TimeUnit.SECONDS);
```
arnaud-splice, line 284: Should we handle the timeout? If `await` returns false, we do not have a guarantee that streamListener.getFailure() is going to work as we expect.

arnaud-splice (changes requested):

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
ipraznik-splice, line 284: another way, where failure is updated, is exceptionCaught, which handles channel exceptions. Not sure if it can happen, but failure can be set already from the channel before olapFutureCallbackInvoked downs. To avoid additional waiting, maybe it's possible to check streamListener.getFailure() != null before olapFutureCallbackInvoked.await and return immediately or wait if still null.

ipraznik-splice (changes requested):
please take a look at my comment, if it makes sense

hbase_sql/src/main/java/com/splicemachine/stream/RemoteQueryClientImpl.java
ascend1: And that's exactly the point of this fix. If some other thread set the `failure` field before `olapFutureCallbackInvoked` count down to 0, that means the exception is not the real exception from olap side and we should wait for olapFuture callback to be invoked.

Point is, as long as `olapFutureCallbackInvoked` is not 0, we wait and don't trust the exception set to `failure`. We use that exception only when 1) olap callback is executed but no exception, or 2) olap callback timed out in 5 seconds.

ascend1: `streamListener.getFailure()` simply returns the `failure` field. We wait for maximum 5 seconds, hoping that the olapFuture callback is invoked and the `failure` field has the correct exception. But if that's not the case, it's also fine to return whatever exception that has been set to `failure` (the same behavior as before) or `null` if it's not set. We check for `null` on caller side.

ipraznik-splice: thank you, got it

dgomezferro (approved):

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

carolp-503 (commented):
+1

PR #5495, opened 2021-05-21
DB-10980 Wait for real exception when an OLAP job fails (3.0)

ascend1:
See https://github.com/splicemachine/spliceengine/pull/5493.


carolp-503 (commented):
Approved for 3.0

PR #5497, opened 2021-05-21
DB-12108 Back out the increase of default value of splice.controlExecution.rowsLimit (3.1)

msirek:


carolp-503 (approved):
Approved for 3.1

PR #5498, opened 2021-05-21
DB-12032 a micro-benchmark for Feature Store trigger updates

OlegMazurov:
## Short Description
This is a micro-benchmark for measuring Feature Store trigger update performance.

## Long Description
The micro-benchmark is based on SQL scripts provided in JIRA. Two modes of updates are implemented: batched updates from a staging table, concurrent single updates from multiple connections. 
Two approaches in how updates are applied and propagated:
- insert into the history table triggers update of the feature table
- insert/update of the feature table triggers insert into the history table

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
+++ b/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
@@ -33,7 +33,7 @@ public class FeatureStoreTriggerBenchmark extends Benchmark {
 
     private static final Logger LOG = Logger.getLogger(FeatureStoreTriggerBenchmark.class);
     private static final int DEFAULT_CONNECTIONS = 10;
-    private static final int DEFAULT_ENTITIES = 1 << 17;
+    private static final int DEFAULT_ENTITIES = 1 << 10;
     private static final int DEFAULT_OPS = 16;
 
     @ClassRule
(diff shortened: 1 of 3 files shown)


martinrupp (approved):

ascend1 (approved):

splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
```diff
@@ -0,0 +1,498 @@
...
+                        "SELECT NEWW.entity_key, NEWW.last_update_ts, CURRENT_TIMESTAMP, NEWW.feature1, NEWW.feature2 FROM NEWW"
+        );
+        testStatement.execute(
+                "CREATE TRIGGER FeatureTableHistory_UPDATE_OOO " +
+                        "AFTER UPDATE ON FeatureTable " +
+                        "REFERENCING NEW_TABLE AS NEWW OLD_TABLE AS OLDW " +
+                        "FOR EACH STATEMENT " +
+                        "INSERT INTO FeatureTable (entity_key, last_update_ts, feature1, feature2) --splice-properties insertMode=UPSERT\n" +
+                        "SELECT o.entity_key, o.last_update_ts, o.feature1, o.feature2 " +
+                        "FROM OLDW o INNER JOIN NEWW n ON o.ENTITY_KEY=n.ENTITY_KEY AND o.LAST_UPDATE_TS > n.LAST_UPDATE_TS"
```
hatyo, line 255: We're upserting here again all the rows which were newer than the newly updated row w.r.t. `LAST_UPDATE_TS`, this triggers a corresponding update of history table rows, is this intentional? because I think we can directly update the history table rows in this trigger, right?

OlegMazurov: This set of triggers was originally proposed in DB-12032 and serves as a baseline for other implementations.
The idea is that all new data goes to the feature table first and the history table is updated using the first two triggers upon every insert/update. The last trigger is needed to maintain the most recent data in the feature table.
Alternatively, as implemented by the last two tests, _updateHistoryTable_ and _updateHistoryTableConcurrently_, we may send all new data, not necessarily ordered, to the history table first and maintain the feature table so that it always contains the most recent data for each entity.
The goal is to compare the two approaches (in two modes: batch and concurrent) and hopefully show that the cleaner approach based on history table updates is also more performant. Currently, both approaches are not functioning as expected due to bugs revealed during this micro-benchmark development.

hatyo: Thanks for the thorough explanation, I understand it better now.

carolp-503 (approved):
+1

PR #5505, opened 2021-05-25
DB-12085 Fix isQualifier() for index conglomerates

ascend1:
## Short Description
This change prevents predicates that cannot be used to limit a table scan from being pushed down to base table wrongly and used a scan qualifier.

## Long Description
Before this fix, `isQualifier()` (`RelationalOperator` interface method) implementations simply checks if the column reference operand of this relational operator refers to a column of `optTable`. Depending on the type of relational operator, there might be other checks. But this is a necessary part of all implementations.

However, this check is not sufficient in case we want to push this predicate down to scan. For an index conglomerate, the column reference must refer to one of the index columns, too. This check is currently missing.

Note that `isQualifier()` is used in two cases, which have slightly different semantics:
1. check if a predicate can be used as a hash key
2. check if a predicate can be used as a scan qualifier

In case 1, the conglomerate columns check is not needed because we assume a predicate is on the level of a join. Down to the subtree, it can be an index scan + index lookup. But eventually, we have access to all columns referenced by predicates. Point is, we are not going to push the predicate down.

In case 2, we care about if it's possible to actually push this predicate down to the scan. In this case, the conglomerate columns check is necessary.

The two cases above are distinguished by the `forPush` argument.

One may wonder in case 2, how simple queries would even work without this check because it's very fundamental. Answer is, in `orderUsefulPredicates()`, `isQualifier()` is mostly called under a guard that the current conglomerate is the table conglomerate. If it's an index conglomerate, the code path calls `isIndexUseful()`.

But there is a missing piece. In case of a disjunction of predicates, e.g., `col1 = 2 or col2= 3`, we call `Predicate.isPushableOrClause()`, which in turns calls `isQualifier()` on each term of the disjunction without considering the current conglomerate. Consequently, the whole or-list can be pushed down to an index scan as a qualifier. However, some columns used in the predicate are not accessible during the scan, causing an ArrayIndexOutOfBound exception at run time.

The fix is to add conglomerate columns check for case 2 for index conglomerates. 

## How to test
Setup:
```
create table TA (c1 char(36));
create table TB (c2 char(36), s char(1), z char(1), hsp char(2), f char(1), pfnr char(10));
create index XPFHSPF on TB (hsp, f, s, z, c2);
```

Test query:
```
select * from --splice-properties joinOrder=fixed
 TA, TB --splice-properties index=XPFHSPF, joinStrategy=sortmerge
 where TA.c1 = TB.c2
   AND TB.s = '3' AND TB.z <> 'S' AND TB.hsp = 'LE'
   AND NOT (TB.f = 'L' AND TB.pfnr = '12345678  ')
```

Before this fix, running the test query would throw `ArrayIndexOutOfBound` exception. Note that explain this query returns a plan without any error. However, this plan is wrong because it takes `TB.f <> 'L' OR TB.pfnr <> '12345678  '` as a qualifier of index scan on `XPFHSPF`.

After this fix, the query runs correctly. Also, the explained plan should put `TB.f <> 'L' OR TB.pfnr <> '12345678  '` at the `ProjectRestrict` step on top of `IndexLookup`.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
@@ -426,7 +426,7 @@ public class IndexRowGenerator implements IndexDescriptor, Formatable
 
 	/** @see IndexDescriptor#baseColumnPositions */
 	public int[] baseColumnPositions() {
-		return id.baseColumnPositions();
+		return id == null ? null : id.baseColumnPositions();
 	}
 
 	/** @see IndexDescriptor#getKeyColumnPosition */
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryComparisonOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryComparisonOperatorNode.java
@@ -36,6 +36,7 @@ import com.splicemachine.db.iapi.services.compiler.MethodBuilder;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.sql.compile.Optimizable;
 import com.splicemachine.db.iapi.sql.dictionary.ConglomerateDescriptor;
+import com.splicemachine.db.iapi.sql.dictionary.IndexRowGenerator;
 import com.splicemachine.db.iapi.store.access.ScanController;
 import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.types.TypeId;
@@ -379,7 +380,7 @@ public abstract class UnaryComparisonOperatorNode extends UnaryOperatorNode impl
     public boolean orderedNulls() { return true; }
 
     @Override
-    public boolean isQualifier(Optimizable optTable, boolean forPush) {
+    public boolean isQualifier(Optimizable optTable, ConglomerateDescriptor cd, boolean forPush) {
         /*
         ** It's a Qualifier if the operand is a ColumnReference referring
         ** to a column in the given Optimizable table.
@@ -390,7 +391,15 @@ public abstract class UnaryComparisonOperatorNode extends UnaryOperatorNode impl
         ColumnReference cr = (ColumnReference) getOperand();
         FromTable ft = (FromTable) optTable;
 
-        return cr.getTableNumber()==ft.getTableNumber();
+        if (cr.getTableNumber() != ft.getTableNumber()) {
+            return false;
+        }
+
+        if (forPush && cd != null && cd.isIndex()) {
+            return isIndexColumn(cr, cd);
+        }
+
+        return true;
     }
 
     @Override
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryRelationalOperatorNode.java
@@ -884,7 +884,8 @@ public class BinaryRelationalOperatorNode
     }
 
     @Override
-    public boolean isQualifier(Optimizable optTable,boolean forPush) throws StandardException{
+    public boolean isQualifier(Optimizable optTable, ConglomerateDescriptor cd, boolean forPush)
+            throws StandardException {
         /* If this rel op is for an IN-list probe predicate then we never
          * treat it as a qualifer.  The reason is that if we treat it as
          * a qualifier then we could end up generating it as a qualifier,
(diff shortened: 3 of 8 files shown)


ascend1:

bklo94:

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
```diff
@@ -426,7 +426,7 @@ public boolean isUnique()
 
 	/** @see IndexDescriptor#baseColumnPositions */
 	public int[] baseColumnPositions() {
-		return id.baseColumnPositions();
+		return id == null ? null : id.baseColumnPositions();
```
msirek, line 429: Wouldn't an index descriptor always have id set?  When can it happen that this is null?  It seems like an error condition.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryComparisonOperatorNode.java
```diff
@@ -390,7 +391,15 @@ public boolean isQualifier(Optimizable optTable, boolean forPush) {
         ColumnReference cr = (ColumnReference) getOperand();
         FromTable ft = (FromTable) optTable;
 
-        return cr.getTableNumber()==ft.getTableNumber();
+        if (cr.getTableNumber() != ft.getTableNumber()) {
+            return false;
+        }
+
+        if (forPush && cd != null && cd.isIndex()) {
```
msirek, line 398: It seems like we'd want to do this check even when forPush is false so that higher level calls like isIndexUseful are correct and we can make the correct costing and join planning decisions.

msirek (changes requested):

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
ascend1: Right, this is super weird. Actually, no, `id` is not always set for `IndexRowGenerator`. I didn't debug into details about it, though.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryComparisonOperatorNode.java
ascend1: This is explained in the comment of `isQualifier()` in `RelationalOperator.java`. When `forPush` is `false`, this method is used to check if this predicate can be used as a hash key. In that case, we should not check for index columns because the predicate will be on top of index lookup (if the index is not covering). If we check for index columns, we may miss the chance of using this predicate as a hash key, thus disabling hashable join.

`isIndexUseful()` logic is correct even before this fix because it uses `isQualifier()` as only one of its preconditions. If a predicate is not a qualifier, it rejects the predicate. If it is, `isIndexUseful()` further checks for index columns for both classic index and index on expressions cases. So it should be fine.

The strange part, though, is that we have too many methods essentially check for the same thing: `isIndexUseful()`, `isQualifier()`, `isHashJoinQualifier()`, `isPushableOrPredicate()`, etc. ...IMHO this part should be carefully refactored.

cloudspliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/UnaryComparisonOperatorNode.java
msirek: OK, I see, thanks.  Yes, as part of DB-11449 I needed to add a special pushOnlyHashableJoinPreds flag to orderUsefulPredicates.  I found it strange that we have to push join predicates to the base table in order to find them for hashable joins, even if they aren't qualifiers.  It seems like we should allow the predicates to be pushed just to the ProjectRestrict above the FromBaseTable, and also look for hashable join predicates there.  I agree some it would be good to refactor this and maybe modify the current predicate pushing behavior.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/IndexRowGenerator.java
msirek: OK, well in any case, if we try to use a null baseColumnPositions later on it would end up with an NPE anyway, so adding this check probably doesn't hurt.

msirek (approved):

dgomezferro (approved):

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2960/

carolp-503 (approved):
+1

ascend1:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2978/

ascend1:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2985/

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

PR #5511, opened 2021-05-26
DB-12120 Database property and hint for cost model

ascend1:
## Short Description
This change adds a database property and a query hint for setting cost model.

## Long Description
- Database property:
  ```
  CALL SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY('costModel', 'v2');
  CALL SYSCS_UTIL.SYSCS_GET_GLOBAL_DATABASE_PROPERTY('costModel');
  ```
- Session property:
  ```
  set session_property costModel='v1';
  values current session_property;
  ```
- Query hint:
  ```
  explain select * from --splice-properties costModel='v2'
    sys.systables;
  ```

## How to test
```
CALL SYSCS_UTIL.SYSCS_SET_GLOBAL_DATABASE_PROPERTY('costModel', 'v2');
name                |value
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PROPERTY_NAME       |costModel
NEW VALUE           |v2
PREVIOUS VALUE      |NULL
INFO                |

4 rows selected
ELAPSED TIME = 330 milliseconds
CALL SYSCS_UTIL.SYSCS_GET_GLOBAL_DATABASE_PROPERTY('costModel');
HOST_NAME                                                                                                               |PROPERTY_VALUE
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
MacBook-Pro:1527                                                                                                        |v2
```
After setting the `costModel` database property to `v2`, explain any query should get `costModel='v2'` in plan.

Session property should overwrite database property value.

Query hint should overwrite both session property and database property values.

If none of the properties are set and no query hint is given, default cost model should be `v1`.

Any value other than `v1` and `v2` should be invalid for `costModel` property.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
@@ -292,6 +292,9 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
     // default to 6
     private int tableLimitForExhaustiveSearch = 6;
 
+    // default to v1
+    private String costModelName = "v1";
+
     // this used to be computed in OptimizerFactoryContextImpl; i.e everytime a
     // connection was made. To keep the semantics same I'm putting it out here
     // instead of in the OptimizerFactory which is only initialized when the
@@ -459,6 +462,11 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
             throw e;
         }  catch (Exception e) {
             // no op, use default value 6
+        }
+
+        String costModelString = PropertyUtil.getCachedDatabaseProperty(this, Property.COST_MODEL);
+        if (costModelString != null && CostModelRegistry.exists(costModelString)) {
+            costModelName = costModelString;
         }
 
         try {
@@ -715,6 +723,15 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
         if (tableLimit != null)
             return tableLimit;
         return tableLimitForExhaustiveSearch;
+    }
+
+    @Override
+    public String getCostModelName() {
+        String cmName = (String) sessionProperties.getProperty(
+                SessionProperties.PROPERTYNAME.COSTMODEL);
+        if (cmName != null)
+            return cmName;
+        return costModelName;
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


ascend1:

martinrupp:
you might want to use the GlobalDBProperties and add a validator for the set costModel

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -461,6 +464,11 @@ public GenericLanguageConnectionContext(
             // no op, use default value 6
         }
 
+        String costModelString = PropertyUtil.getCachedDatabaseProperty(this, Property.COST_MODEL);
```
arnaud-splice, line 467: Why set it here and not in CompilerContext like the other global database properties?

arnaud-splice:
Can you add a test for SET_GLOBAL_DATABASE_PROPERTY?

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
ascend1: Logic is moved to `GenericStatement` and `CompilerContext`.

ascend1:
> you might want to use the GlobalDBProperties and add a validator for the set costModel (for SYSCS_SET_GLOBAL_DATABASE_PROPERTY)

That makes sense. I'll do it once we have the latest master branch merge back to optimizer branch.

ascend1:

ascend1:
> Can you add a test for SET_GLOBAL_DATABASE_PROPERTY?

It would be nice to have DB-11538 PR on optimizer branch for adding a test. Is it OK to do this after merge master branch back?

cloudspliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

arnaud-splice:
> > Can you add a test for SET_GLOBAL_DATABASE_PROPERTY?
> 
> It would be nice to have DB-11538 PR on optimizer branch for adding a test. Is it OK to do this after merge master branch back?

Sure!
Can you create a jira for that?

ascend1:
> > > Can you add a test for SET_GLOBAL_DATABASE_PROPERTY?
> > 
> > 
> > It would be nice to have DB-11538 PR on optimizer branch for adding a test. Is it OK to do this after merge master branch back?
> 
> Sure!
> Can you create a jira for that?

Yes, it's here: https://splicemachine.atlassian.net/browse/DB-12132.

msirek (approved):

martinrupp (approved):

PR #5518, opened 2021-05-27
DB-12116: batch write to SPLICE_CONGLOMERATE_SI during upgrade

jyuanca:
## Short Description
Copying all conglomerates from SPLICE_CONGLOMERATE to SPLICE_CONGLOMERATE_SI is slow if there are a large amount of conglomerates.

## Long Description
It is slow because conglomerates are copied one at a time, each invoking an RPC. To speed up, conglomerates are batched to write to SPLICE_CONGLOMERATE_SI.

## How to test

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeUtils.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeUtils.java
@@ -49,8 +49,7 @@ public class UpgradeUtils {
     public static void initializeConglomerateSITable(TransactionController tc) throws IOException {
         SIDriver driver = SIDriver.driver();
         int rowsRewritten = 0;
-
-        int batchSize = 100;
+        int batchSize = driver.getConfiguration().getMaxBufferEntries();
         List<DataPut> mutations = Lists.newArrayList();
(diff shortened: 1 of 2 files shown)


jyuanca:

cloudspliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeUtils.java
```diff
@@ -50,6 +50,8 @@ public static void initializeConglomerateSITable(TransactionController tc) throw
         SIDriver driver = SIDriver.driver();
         int rowsRewritten = 0;
 
+        int batchSize = 100;
```
dgomezferro, line 53: Maybe we could use `conf.getMaxBufferEntries()`, what we use for the write pipeline. This defaults to 1000 rows.

dgomezferro (approved):
We could use a configurable batch size, but I'm ok with merging as it is.

jyuanca:

carolp-503 (approved):
+1

cloudspliceci:
TEST SUCCEEDED +1

PR #5520, opened 2021-05-27
DB-12102 Properly escape ' and \ when sending a query to spark 

arnaud-splice:
## Description

When sending predicates to spark, we need to escape ' and \ properly. Predicates such as `World's` would cause a ParserException on spark's side.

## How to test

```
set session_property useOlap=true;
drop schema foo cascade;
create schema foo;
set schema foo;

call SYSCS_UTIL.SYSCS_EMPTY_GLOBAL_STATEMENT_CACHE();
CREATE TABLE ITEM_CODES ( ITEMTEXT VARCHAR (50) NOT NULL);

CREATE VIEW VERBOSE_INV AS SELECT * FROM item_codes UNION SELECT * FROM item_codes;
INSERT INTO ITEM_CODES VALUES ('a''b');
INSERT INTO ITEM_CODES VALUES ('a\b');
INSERT INTO ITEM_CODES VALUES ('a"b');


SELECT COUNT(*) FROM VERBOSE_INV WHERE ITEMTEXT = 'a''b';
SELECT COUNT(*) FROM VERBOSE_INV WHERE ITEMTEXT = 'a\b';
SELECT COUNT(*) FROM VERBOSE_INV WHERE ITEMTEXT = 'a"b';
```

All three select should return 1

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/OperatorToString.java
@@ -856,9 +856,11 @@ public class OperatorToString {
                         dvd instanceof SQLLongvarchar ||
                         dvd instanceof SQLClob) {
                         if (vars.buildExpressionTree)
-                            str = format("%s", cn.getValue().getString());
+                            str = cn.getValue().getString();
                         else
-                            str = format("\'%s\' ", cn.getValue().getString());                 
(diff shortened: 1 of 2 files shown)


arnaud-splice:

cloudspliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

martinrupp (approved):

carolp-503 (approved):
+1

PR #5523, opened 2021-05-28
DB-11095 Fix shouldRunSpark() for DT as RHS of a join

ascend1:
## Short Description
This change fixes execution engine decision when RHS of a join is not a base table.

## Long Description
In `shouldRunSpark()`, `determineSpark()` is called under a guard that `obj instanceof FromBaseTable`. However, for the following query, join cost is not stored in any `FromBaseTable`:

```
select * from --splice-properties joinOrder=fixed
    t1, (select distinct c2 from t2);
```

This is because the RHS of the join is a derived table. The join cost is stored in the `ProjectRestrictNode` on top of the `SelectNode` of the derived table. Since we check `obj instanceof FromBaseTable`, effectively join cost is not considered in deciding execution engine of this query.

The fix is to move `determineSpark()` to `FromTable` and remove the `FromBaseTable` check. Now we call `determineSpark()` on all `FromTable`s collected from the plan, including both PRNs and `FromBaseTable`s.

## How to test
Setup:
```
create table if not exists t1(c1 int);
create table if not exists t2(c2 int);

call syscs_util.fake_table_statistics('SPLICE', 'T1', 5000, 13, 1);
call syscs_util.fake_table_statistics('SPLICE', 'T2', 1000, 4, 1);
```
Query:
```
explain select * from --splice-properties joinOrder=fixed
    t1, (select distinct c2 from t2);
```
Before the fix, the query goes to OLTP despite of 5000000 output rows. After the fix, the query goes to OLAP.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLStatementNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/DMLStatementNode.java
@@ -278,9 +278,7 @@ public abstract class DMLStatementNode extends StatementNode {
         CollectNodesVisitor cnv = new CollectNodesVisitor(FromTable.class);
         resultSet.accept(cnv);
         for (Object obj : cnv.getList()) {
-            if (obj instanceof FromBaseTable) {
-                ((FromBaseTable) obj).determineSpark();
-            }
+            ((FromTable) obj).determineSpark();
             type = type.combine(((FromTable) obj).getDataSetProcessorType());
             sparkExecType = sparkExecType.combine(((FromTable) obj).getSparkExecutionType());
         }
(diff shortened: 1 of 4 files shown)


ascend1:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2992/

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/2998/

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5525, opened 2021-05-29
DB-9898: alter table drop column

jyuanca:
## Short Description
Support alter table drop column

## Long Description
We have two representations of a column's position, logical position and storage position. However, when we implement a feature that access a column, we always overlook the possibility that a column may be dropped from the table, and use logical position. This PR is to find and fix the case where storage column position should be used instead of logical position.

## How to test
Test various cases where a column is dropped, read/write to table and index works. Below are some test cases

----------------------------------
create table t(i int, j bigint, k varchar(10), l real);
insert into t values (1,2,'3', 4.0);
alter table t drop j;
create index ti on t(k, i);

select * from t;
select * from t --splice-properties index=ti
;
select * from t --splice-properties index=null
;
select i, k from t;

-----------------------------------
create table t1(a int, b int, c int, d int, e int, f int, g int, h int, z int); 
alter table t1 drop column b;
alter table t1 drop column c;
alter table t1 drop column d;
create index it1 on t1(a, z);
insert into t1 values(1,5,6,7,8,9);
alter table t1 drop column f;
alter table t1 drop column g;
alter table t1 drop column h;

insert into t1 values (11,55,99);
select * from t1 --splice-properties index=it1
;
select * from t1 --splice-properties index=null
;
select a,z from t1;

-----------------------------------
create table TEST_TABLE (a int, b int, c int, d int, e int, f int, g int, primary key(c));
insert into TEST_TABLE values (1,2,3,4,5,6,7);
create index TEST_IDEXN on TEST_TABLE(c,f);
alter table test_table drop a;
select c, f from test_table;

create table TABLE_WITH_CONSTRAINTS (a int, b int, c int, d int, primary key(c), unique(b));
insert into TABLE_WITH_CONSTRAINTS values (1,2,3,4);

create table TABLE_WITH_CONSTRAINTS2 (a int, b int, c int references TEST_TABLE(c), d int);
insert into TABLE_WITH_CONSTRAINTS2 values (1,2,3,4);

alter table test_table drop d;
select c, f from test_table;

alter table test_table drop e;
select c, f from test_table;

alter table test_table drop g;
select c, f from test_table;

explain select c, f from test_table;

alter table test_table drop b;
select c, f from test_table;
create table a(i int, j int, k int);
create index ak on a(k);
alter table a drop i;
insert into a values (1,2);
explain select k from a;


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLSingleTableConstantOperation.java
```diff
@@ -135,6 +135,21 @@ void dropConstraint(ConstraintDescriptor consDesc,TableDescriptor skipCreate, Li
...
             dropIndex(td, indexConglom, (SpliceTransactionManager)lcc.getTransactionExecute(),lcc, schemaName, indexName);
         }
+
+        if (skipCreate != null && consDesc.getConstraintType() == DataDictionary.PRIMARYKEY_CONSTRAINT) {
+            ConglomerateDescriptorList cdl = skipCreate.getConglomerateDescriptorList();
+            DataDictionary dd = lcc.getDataDictionary();
+            TransactionController tc = lcc.getTransactionExecute();
+            for (ConglomerateDescriptor cd : cdl) {
+                if (cd.isPrimaryKey()) {
+                    cd.setIndexRowGenerator(new IndexRowGenerator());
```
ascend1, line 145: Is it necessary to set an empty `IndexRowGenerator` instead of just `null`?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/IndexConstantOperation.java
```diff
@@ -204,28 +206,29 @@ public String getScopeName() {
         List<Integer> indexCols = tentativeIndex.getIndex().getIndexColsToMainColMapList();
         List<Integer> allFormatIds = tentativeIndex.getTable().getFormatIdsList();
         List<Integer> columnOrdering = tentativeIndex.getTable().getColumnOrderingList();
-        td.getColumnDescriptorList().elementAt(0).getType().getNull();
-        int[] rowDecodingMap = new int[allFormatIds.size()];
-        int[] baseColumnMap = new int[allFormatIds.size()];
+        List<Integer> storagePositions = tentativeIndex.getTable().getStoragePositionsList();
+        int maxStoragePosition = Collections.max(indexCols);
```
ascend1, line 210: It would be more intuitive to me if `maxStoragePosition = Collections.max(storagePositions)`. Not sure if `indexCols` is the same as `storagePositions` at this point.

martinrupp (changes requested):
please add short/long/how to test to description. title should maybe be "fix showprimarykeys after alter table drop column"

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/ModifyColumnConstantOperation.java
```diff
@@ -889,6 +889,25 @@ private void dropColumnFromTable(Activation activation,
...
+                for (int i = 0; i < positions.length; ++i) {
+                    if (positions[i] > pos) {
+                        positions[i] = positions[i] - 1;
+                        modified = true;
+                    }
+                }
+                if (modified) {
+                    dd.dropConglomerateDescriptor(cd, tc);
+                    dd.addDescriptor(cd, sd,
+                            DataDictionary.SYSCONGLOMERATES_CATALOG_NUM, false, tc, false);
```
ascend1, line 907: It's a bit confusing here. Since we have modified the logical column positions in this `cd`, why would we drop it and add it back again? I can only image we do this to refresh data dictionary cache. But then it's kind of using its side effect rather than the actual effect.

If the purpose is to refresh cache only, can we achieve it without dropping and then adding it back? Or there is some other reason here?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/TruncateTableConstantOperation.java
```diff
@@ -160,10 +161,15 @@ private void truncateTable(TableDescriptor td, Activation activation) throws Sta
...
             int[] columnOrder = conglomerate.getColumnOrdering();
+            keyFormatIds = conglomerate.getKeyFormatIds();
+            //ColumnDescriptorList cds = td.getColumnDescriptorList();
             if (columnOrder != null) {
                 columnOrdering = new ColumnOrdering[columnOrder.length];
                 for (int i = 0; i < columnOrder.length; i++) {
                     columnOrdering[i] = new IndexColumnOrder(columnOrder[i]);
+//                    int storagePosition = columnOrder[i]+1;
+//                    int position = cds.getColumnDescriptorByStoragePosition(storagePosition).getPosition();
+//                    columnOrdering[i] = new IndexColumnOrder(position-1);
```
ascend1, line 172: Remove commented code?

splice_machine/src/main/java/com/splicemachine/derby/impl/store/access/base/SpliceConglomerate.java
```diff
@@ -104,7 +106,12 @@ protected void create(
         collation_ids=ConglomerateUtil.createCollationIds(format_ids.length,collationIds);
         hasCollatedTypes=hasCollatedColumns(collation_ids);
         this.tmpFlag=tmpFlag;
+//        keyFormatIds = new int[columnOrdering.length];
+//        for (int i = 0; i < keyFormatIds.length; ++i) {
+//            keyFormatIds[i] = format_ids[columnOrdering[i]];
+//        }
```
ascend1, line 112: Remove?

ascend1 (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/IndexConstantOperation.java
jyuanca: IndexCols stores storage positions

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/ModifyColumnConstantOperation.java
jyuanca: To update a conglomerate descriptor, we need to drop the old one from sysconglomerates first. Though this cd is new, only its schemaid and conglomerte name will be used to drop old conglomerates. The add this new conglomerate descriptor to dictionary

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/DDLSingleTableConstantOperation.java
jyuanca: it can be null

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/AlterTableConstantOperation.java
```diff
@@ -939,7 +949,7 @@ protected ConglomerateDescriptor updateTableConglomerateDescriptor(TableDescript
                 ascending[i] = true;
             }
             IndexDescriptor indexDescriptor =
-                new IndexDescriptorImpl("PRIMARYKEY",true,false,pkColumns,ascending,pkColumns.length,false,false);
+                new IndexDescriptorImpl("PRIMARYKEY",true,false,pkColumns,pkColumns,ascending,pkColumns.length,false,false);
```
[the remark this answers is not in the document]
jyuanca: Good point. I changed to pass in storage positions too

jyuanca:

cloudspliceci:
TEST SUCCEEDED +1

ascend1:
DB-11062 is related to this and on top of this PR, the problematic query using an index on expressions gives NPE. For this reason, I think we need more commits to this PR and I'm currently looking into this.

jyuanca:

cloudspliceci:
TEST SUCCEEDED +1

ascend1:
A minimal NPE test case:
```
create table t (c int, d int);
ELAPSED TIME = 7 milliseconds
insert into t values (1,1), (2,2);
2 rows inserted/updated/deleted
ELAPSED TIME = 12 milliseconds
create index tidx on t (c, d);
ELAPSED TIME = 14 milliseconds
alter table t drop column d;
ELAPSED TIME = 10 milliseconds
select * from t where d = 2;
ERROR 42X04: Column 'D' is either not in any table in the FROM list or appears within a join specification and is outside the scope of the join specification or appears in a HAVING clause and is not in the GROUP BY list. If this is a CREATE or ALTER TABLE  statement then 'D' is not a column in the target table.
alter table t add column d integer;
ELAPSED TIME = 8 milliseconds
select * from t where d = 2;
ERROR XJ001: Java exception: 'java.lang.NullPointerException: '.
ERROR XJ001: Java exception: ': java.lang.NullPointerException'.
```

ascend1 (changes requested):
It looks like we need to define the behavior of dropping a column that is used in an index. Should we drop the index as well? Or we prevent the user from dropping the column? Currently we do nothing, meaning the index still has old storage position in `baseColumnStoragePositions`, and this causes NPE in a few places (cost estimation, code generation, ...).

SQL standard 2015 defines a "drop behavior" clause for alter table drop column statement. However, it doesn't have explicit definition on columns referenced in an index. For this reason, current behavior is vendor implementation dependent. Daniel and Arnaud tried a few other DBs:
- SQL Server: cannot drop the column
- DB2: drop the column and index
- PostgreSQL: drop the column and index
- Snowflake: "cascade" drops index, "restrict" prevent from dropping the column

arnaud-splice (changes requested):
Fantastic PR!
A couple of minor comment, on top of Zhen's comment about dropping columns with indexes attached to them.

arnaud-splice:
> A minimal NPE test case:
> 
> ```
> create table t (c int, d int);
> 0 rows inserted/updated/deleted
> ELAPSED TIME = 7 milliseconds
> insert into t values (1,1), (2,2);
> 2 rows inserted/updated/deleted
> ELAPSED TIME = 12 milliseconds
> create index tidx on t (c, d);
> 0 rows inserted/updated/deleted
> ELAPSED TIME = 14 milliseconds
> alter table t drop column d;
> 0 rows inserted/updated/deleted
> ELAPSED TIME = 10 milliseconds
> select * from t where d = 2;
> ERROR 42X04: Column 'D' is either not in any table in the FROM list or appears within a join specification and is outside the scope of the join specification or appears in a HAVING clause and is not in the GROUP BY list. If this is a CREATE or ALTER TABLE  statement then 'D' is not a column in the target table.
> alter table t add column d integer;
> 0 rows inserted/updated/deleted
> ELAPSED TIME = 8 milliseconds
> select * from t where d = 2;
> ERROR XJ001: Java exception: 'java.lang.NullPointerException: '.
> ERROR XJ001: Java exception: ': java.lang.NullPointerException'.
> ```

This is a regression from the current version. On the main branch, the last select does not return anything.
However, 
```
select * from t --splice-properties index=tidx
where d = 2
```
returns wrongly:
```
C          |D
-----------------------
2          |2
```
on the main branch.

> It looks like we need to define the behavior of dropping a column that is used in an index. Should we drop the index as well? > Or we prevent the user from dropping the column? Currently we do nothing, meaning the index still has old storage position in baseColumnStoragePositions, and this causes NPE in a few places (cost estimation, code generation, ...).
>
> SQL standard 2015 defines a "drop behavior" clause for alter table drop column statement. However, it doesn't have explicit definition on columns referenced in an index. For this reason, current behavior is vendor implementation dependent. Daniel and Arnaud tried a few other DBs:
> 
>    SQL Server: cannot drop the column
>    DB2: drop the column and index
>    PostgreSQL: drop the column and index
>    Snowflake: "cascade" drops index, "restrict" prevent from dropping the column

This PR should probably emulate the SQL Server behavior. We can then extend it and support both restrict and cascade in a future PR

arnaud-splice (approved):

ascend1 (approved):
Maybe we can clean up the code being commented out?

martinrupp (approved):

carolp-503 (approved):
+1

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

PR #5533, opened 2021-05-31
DB-12039 Calibrate join selectivity

hatyo:
## Short Description
The PR fixes the join selectivity and gives a much better estimate of join cardinality.

## Long Description
Before this PR, the inner join selectivity is not exactly correct, it is given in the following formula:

((1 - NullSelectivity(L))*(1 - NullSelectivity(R)) / Min(Cardinality(L), Cardinality(R))

This could cause the join cardinality to give the wrong result, e.g. assume we have two tables `A` and `B` with join condition `A.a = B.b`. a has a cardinality of 800 with RPV=1 and `b` is a key has a cardinality of 10 also with RPV=1, both have no null values.

In this case, the join selectivity will be: 1 / Min(800, 10) = 0.1 which is incorrect, to see why, let’s calculate the join cardinality:

jc = 0.1 * 800 * 10 = 800

This is wrong, if we choose Max instead of Min we will have a much more conservative join selectivity: 1 / Max(800, 10) = 0.00125.

jc = 0.00000125 * 800 * 10 = 10

This is addressed in the v2 cost estimation model.

A new microbenchmark is added to test selectivity calibration, it will define the following dimensions:

    number of rows of LHS
    number of rows on RHS
    OLAP / OLTP
    RPV on LHS
    RPV on RHS

RPV is either fixed per value or is following [Zipfian distribution](https://en.wikipedia.org/wiki/Zipf%27s_law) to simulate a realistic workload.

## How to test
Run the tests in `JoinSelectivityBenchmark`, preferably against a cluster running in cloud environment, similar to what's explained [here](https://github.com/splicemachine/spliceengine/pull/5360#issue-605614457), basically:

1. Prepare the necessary jar files as a tarball.
```
mvn -Pcore,cdh6.3.0 clean install -DskipTests
mkdir benchmark
cp ./platform_it/src/main/resources/info-log4j.properties /Users/<your username>/.m2/repository/junit/junit/4.12/junit-4.12.jar /Users/<your username>/.m2/repository/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar ./splice_machine/target/splice_machine-<VERSION>-tests.jar ./platform_it/target/dependency/log4j-1.2.17.jar ./platform_it/target/dependency/db-client-<VERSION>.jar benchmark
tar czvf benchmark.tar.gz ./benchmark
```
2. Upload the tarball to your cluster. You need to be authorized to the k8s environment before doing this step.
```
kubectl cp benchmark.tar.gz splicedb-hregion-0:/home/hbase/ -n <your cluster namespace>
```
3. Log into the `splicedb-hregion-0` pod of your cluster:
```
kubectl exec -it splicedb-hregion-0 -n <your cluster namespace> -- bash
```
4. Once in pod, run the benchmark:
```
cd home/hbase
tar zxvf benchmark.tar.gz
java -Dlog4j.configuration=file:./benchmark/info-log4j.properties -cp 'benchmark/*' org.junit.runner.JUnitCore com.splicemachine.benchmark.<benchmark name>

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/v2/V2SelectivityEstimator.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/costing/v2/V2SelectivityEstimator.java
@@ -45,7 +45,7 @@ public class V2SelectivityEstimator implements SelectivityEstimator {
         double rightNonZeroCardinality = Math.max(1.0, rightColumnIndexSelectivity * rightRowCount);
 
         selectivity = ((1.0d - leftColumn.nullSelectivity()) * (1.0d - rightColumn.nullSelectivity())) /
-                Math.min(leftNonZeroCardinality, rightNonZeroCardinality);
+                Math.max(leftNonZeroCardinality, rightNonZeroCardinality);
         selectivity = selectivityJoinType.equals(SelectivityEstimator.SelectivityJoinType.INNER) ?
                 selectivity : 1.0d - selectivity;
         if (rightTable instanceof FromTable && ((FromTable) rightTable).getExistsTable()) {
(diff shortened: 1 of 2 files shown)


arnaud-splice (approved):
Great change!

ascend1 (approved):
Awesome fix!

carolp-503 (approved):
+1

Great fix and great instructions on how to build and test

PR #5535, opened 2021-05-31
DB-12140 Fix index on expressions logic for UnaryDateTimestampOperatorNode

ascend1:
## Short Description
This change fixes two problems:
1. An index created based on expressions using `UnaryDateTimestampOperatorNode` is not treated as an covering index even when an query refers to exactly the same expression.
2. An index on expression is not chosen based on cost if index expressions do not occur in any predicates.

## Long Description
For the first problem, root cause is that `bindExpression` of `UnaryDateTimestampOperatorNode` modifies the subtree in some cases. If this happens, the stored AST for index expression at index creation time is different than the AST parsed in query time. Solution is to bind `UnaryDateTimestampOperatorNode` nodes in index expression ASTs in query time, too.

Since we currently do this to `JavaToSQLValueNode` already, I refactor the code to have a new visitor doing this binding job for index expressions.

For the second problem, root cause is that in current optimizer architecture, best plan is determined based on `FromList` cost, not including other parts of the query (e.g., grouping, ordering, window function, etc.). In the problematic query, index expression is used only in select list, which is currently not costed at all.

Current solution is to prefer an index on expression if we know it's covering. Concretely, base cost of such an access path is multiplied by 0.999.

In theory, this problem should be resolved once we move to our next optimizer.

## How to test
Setup:
```
CREATE TABLE T1 (
    "T" TIMESTAMP NOT NULL
    , CONSTRAINT SQLBF0B9DCD0179B01B36930004E01C4FE8 PRIMARY KEY("T")) ;

create index tidx on t1 (timestampdiff(SQL_TSI_FRAC_SECOND, timestamp('1970-01-01 00:00:00'), t));
```

Query:
```
1. explain select timestampdiff(SQL_TSI_FRAC_SECOND, timestamp('1970-01-01 00:00:00'), t) from T1 --splice-properties index=TIDX
    ;
2. explain select timestampdiff(SQL_TSI_FRAC_SECOND, timestamp('1970-01-01 00:00:00'), t) from T1;
```
Before this fix:
- Query 1 plan contains an index lookup step
- Query 2 plan contains a table scan

After this fix:
- Query 1 plan doesn't contain the index lookup step
- Query 2 plan contains an index scan

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
@@ -42,9 +42,11 @@ import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
 import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.types.ProtobufUtils;
 import com.splicemachine.db.iapi.util.ByteArray;
+import com.splicemachine.db.impl.ast.CollectingVisitor;
 import com.splicemachine.db.impl.sql.CatalogMessage;
 import com.splicemachine.db.impl.sql.compile.*;
 import com.splicemachine.db.impl.sql.execute.BaseExecutableIndexExpression;
+import splice.com.google.common.base.Predicates;
 
 import java.io.IOException;
 import java.io.ObjectInput;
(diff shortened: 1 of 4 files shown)


ascend1:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3014/

martinrupp (approved):

ipraznik-splice (approved):

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5539, opened 2021-05-31
DB-12051 Add support for additional constant folding

arnaud-splice:
## Short Description
Support NOT, unary minus, conditional and binary relational operator constant folding

## Long Description
- **DB-12145** NOT: Support for constant folding added for null and non null operands

- **DB-12148** Conditional node: Support for constant folding added for null and non null operands. If the condition is constant but then or else isn't, remove the condition and return the correct underlying operand

- **DB-12146** Unary plus / minus: Support for constant folding added for nulls and non null number operand.

- **DB-12147** BinaryRelationalOperatorNode: Support added for null operands. Non null operand constant folding was already implemented.

## How to test
This can be tested via an explain on a simple select: 
```
explain select * from sysibm.sysdummy1 where 1 = -(1)
```
will return a plan without a table scan because `1 = -(1)` is evaluated to be false at compile time
The same principle will apply to `true = not false`, `true = case when true then true else false end`, etc.
A new IT was created and should be referred to for more details.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/types/NumberDataValue.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/types/NumberDataValue.java
@@ -39,134 +39,134 @@ import static com.splicemachine.db.iapi.reference.Limits.DB2_MAX_DECIMAL_PRECISI
 
 public interface NumberDataValue extends DataValueDescriptor
 {
-	/**
-	 * The minimum scale when dividing Decimals
-	 */
-	int MIN_DECIMAL_DIVIDE_SCALE = 4;
-	int MIN_DECIMAL_MULTIPLICATION_SCALE = 6;
-	int MAX_DECIMAL_PRECISION_SCALE = DB2_MAX_DECIMAL_PRECISION_SCALE;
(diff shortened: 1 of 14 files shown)


arnaud-splice:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3018/

martinrupp (approved):
This is a nitpick but i've seen it a couple of times now people just write "see IT" in the "how to test" section, without any nice example. "see IT for details" should - if at all - come second (not first) after a nice example that i can run in a sqlshell on a cluster (if possible).

ipraznik-splice (approved):

arnaud-splice:
> This is a nitpick but i've seen it a couple of times now people just write "see IT" in the "how to test" section, without any nice example. "see IT for details" should - if at all - come second (not first) after a nice example that i can run in a sqlshell on a cluster (if possible).

I will switch that around :+1:

ascend1 (approved):

arnaud-splice:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5541, opened 2021-06-01
DB-11905/12122 Write Streaming result to HDFS if a client is not consuming

ipraznik-splice:
WARNING! DON'T MERGE before DB-11905/12121

## Short Description
After a long pause in client consumption, it starts to write massages from Spark to the file system to prepare for resuming.

## Long Description
The idea comes from the discussion in SPLICE-1891. If the client is not consuming a long time, but the connection is still alive, the first defence against overloading is to start the throttling: DB-12121. After 5 minutes(hardcoded now), if a connection is still alive, it starts a second phase - enables a spark submitting again, but stores all messages in the file system, to optimize when a client starts to resume to consume again the messages are prepared alrerady without a spark latency.

The backup functionality can be disabled by a parameter **spark.result.streaming.backup.enabled**, by default true.

The messages are stored in [root]/stream_result/[StreamListener.uuid]/[partitionId] and deleted if are consumed or on close of StreamListener.

## How to test
```
create table a ( i int);
insert into a values 1,2,3;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
insert into a select i + (select count(*) from a) from a;
```

```
get cursor c as 'select * from a --splice-properties useSpark=true
order by i';
next c;

After 5 mins the messages will be stored in platform_it/target/hbase/hbase/stream_result

Or with StreamListenerIT adding

//                        s.setQueryTimeout(240);

                        while (rs.next()) {
                            rs.getString(1);
                            counter++;
                            if (counter == 40000) {
                              Thread.sleep(360000);
                            }
                        } 
The client is pausing the consumption, it will start the throttling, after 5 minutes it starts to backup messages and the spark jobs are running again but without reading on the client. After 6 minutes the client will continue, the messages will be read from files.
```


hbase_sql/src/main/java/com/splicemachine/stream/MessageBackupHandler.java
```diff
@@ -0,0 +1,116 @@
...
+public class MessageBackupHandler {
+    private static final Logger LOG = Logger.getLogger(MessageBackupHandler.class);
+
+    static final int CONSUMPTION_TIMEOUT = 300000;
+
+    private UUID uuid;
+    private HdfsStreamWriter hdfsStreamWriter;
+    private ScheduledExecutorService timeoutMonitorExecutor;
+    private ExecutorService backupExecutor;
+
```
jyuanca, line 38: try to use general purpose thread pool from SIDriver if possible.

hbase_sql/src/main/java/com/splicemachine/stream/StreamListener.java
```diff
@@ -73,21 +75,25 @@
     }
 
     public StreamListener(long limit, long offset, int batches, int batchSize) {
-        this(limit, offset, batches, batchSize, StreamableRDD.DEFAULT_PARALLEL_PARTITIONS, true);
+        this(limit, offset, batches, batchSize, StreamableRDD.DEFAULT_PARALLEL_PARTITIONS, true, true);
```
jyuanca, line 78: Should the last two parameter be always true or configurable?

jyuanca (changes requested):

hbase_sql/src/main/java/com/splicemachine/stream/StreamListener.java
ipraznik-splice: The first and second constructors are used only in ITs with default parameters, I've added comments. Only the last one is used in the standard cases and gets parameters configured:

`public StreamListener(long limit, long offset, int batches, int batchSize, int parallelPartitions,
                          boolean throttleEnabled, boolean fileCacheEnabled)`

hbase_sql/src/main/java/com/splicemachine/stream/MessageBackupHandler.java
ipraznik-splice: backupExecutor is replaced
timeoutMonitorExecutor is still initialized here, as there is no general ScheduledExecutorService

hbase_sql/src/main/java/com/splicemachine/stream/HdfsStreamWriter.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.UUID;
+import java.util.concurrent.ArrayBlockingQueue;
+
+/**
+ * Handles writing and loading operations to the file system for streaming message queues.
+ */
+public class HdfsStreamWriter {
+    private static final Logger LOG = Logger.getLogger(HdfsStreamWriter.class);
+
+    private static final String BACKUP_DIR = "stream_result";
```
hatyo, line 39: I think it makes sense to make this configurable (in a separate ticket).

hatyo (approved):
LGTM

PR #5542, opened 2021-06-02
DB-12089 fix equijoin condition containing SECOND function

hatyo:
## Short Description
In this PR we fix a NPE caused by having a `SECOND` function in an equijoin condition.

## Long Description
When the optimizer analyses join clauses, it attempts to push some of these clauses after normalizing them as deep as possible down the tree, this is very useful to prune as many rows as possible early on in the execution pipeline.

If a join condition contains a `SECOND` function referencing a column from the RHS of the join, we might end up in certain cases in translating the join condition into a where clause and incorrectly pushing it to the LHS of the join thinking this is a normal search clause. This goes unnoticed until a NLJ strategy is chosen and the actual execution begins, at that point, the NLJ operation will use the (wrong) search clause to filter out the rows from the LHS, however since the search clause contains a reference to some columns from the RHS, it attempts to read these column values from the activation's list of rows and we end up with NPE, because we haven't even started looking into the RHS (yet).

The root cause of this problem, is that the logic plan node of `SECOND` function is not collecting the referenced tables / columns correctly when visited by calling `categorize`, for example, if a predicate is something like `T1.col1 = SECOND(T2.col2)`, during predicate categorization, each term in the predicate is visited to determine  column references and table references, however `SECOND` is not passing control to its own operands and simply returns `true`, therefore, we end up thinking the above predicate contains only references to table `T1` and its column `col1`, instead of `T1` and `T2` and columns `col1` resp. `col2`.

To fix this issue, we extend `SECOND` function's `categorize` hook and allow it to recursively process its own operands. By doing this, the optimizer now recognizes correctly an equijoin predicate even if it contains `SECOND` function and does not incorrectly pushes it down to the LHS, instead it pushes it to the RHS allowing NLJ to leverage it later on during probing the RHS.

## How to test

```sql

CREATE TABLE DTFS1(col1 TIMESTAMP);
CREATE TABLE DTFS2(col2 DEC);
insert into DTFS1 values('2015-11-19 11:03:45.095'),('2015-01-18 11:03:44.095');
insert into DTFS2 values(45.095),(44.095);

-- previously, this gave NPE
SELECT * from --splice-properties joinOrder=fixed
DTFS2
join DTFS1
on INTEGER(SECOND(col1)) = INTEGER(col2);
ERROR XJ001: Java exception: 'java.lang.NullPointerException: java.io.IOException'.
ERROR XJ001: Java exception: ': java.lang.NullPointerException'.

-- with this PR, it works as expected:
SELECT * from --splice-properties joinOrder=fixed
DTFS2
join DTFS1
on INTEGER(SECOND(col1)) = INTEGER(col2);
COL2|COL1
---------
45|2015-11-19 11:03:45.095
44|2015-01-18 11:03:44.095
```

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>spliceengine-parent</artifactId>
         <groupId>com.splicemachine</groupId>
-        <version>3.2.0.2017-SNAPSHOT</version>
+        <version>3.2.0.2018-SNAPSHOT</version>
     </parent>
     <dependencies>
         <dependency>
--- a/coverage/pom.xml
+++ b/coverage/pom.xml
@@ -9,7 +9,7 @@
     <parent>
         <artifactId>spliceengine-parent</artifactId>
         <groupId>com.splicemachine</groupId>
-    	  <version>3.2.0.2017-SNAPSHOT</version>
+    	  <version>3.2.0.2018-SNAPSHOT</version>
     </parent>
     <dependencies>
         <dependency>
--- a/db-build/pom.xml
+++ b/db-build/pom.xml
@@ -33,7 +33,7 @@
     <parent>
         <artifactId>spliceengine-parent</artifactId>
         <groupId>com.splicemachine</groupId>
-        <version>3.2.0.2017-SNAPSHOT</version>
+        <version>3.2.0.2018-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>db-build</artifactId>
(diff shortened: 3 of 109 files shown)


hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3043/

msirek (approved):

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3050/

hatyo:
The only failed test is `testKillOlapServer`, it is a known sporadic that is tracked in DB-9942.

dgomezferro (approved):

dgomezferro:
Thanks for the great explanation @hatyo !

carolp-503 (approved):
+1

PR #5543, opened 2021-06-02
DB-11905/12154 Introduce parameter to configure amount of threads to execute partitions in StreamableRDD

ipraznik-splice:
WARNING! DON'T MERGE before DB-11905/12121

## Short Description
Add a parameter to configure the number of threads to submit partition jobs in StreamableRDD.

## Long Description
Currently, the thread executor, which is submitting partition jobs in StreamableRDD, is initialized only with 2 threads. To make it configurable a new parameter is added: spark.result.streaming.threads, by default 2.

Also it's adjusted by parallelPartitions: parallelPartitions / 2, if it's greater than a default value, but cannot be larger than 8. 

spark.result.streaming.threads can set any value, even greater than 8.

## How to test
Regression testing with different values of spark.result.streaming.threads, that corresponds to the number of executor.cores on the testing platform.


hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
```diff
@@ -60,25 +63,30 @@
...
 
 
     StreamableRDD(JavaRDD<T> rdd, UUID uuid, String clientHost, int clientPort) {
         this(rdd, null, uuid, clientHost, clientPort, 2, 512);
     }
 
     public StreamableRDD(JavaRDD<T> rdd, OperationContext<?> context, UUID uuid, String clientHost, int clientPort, int batches, int batchSize) {
-        this(rdd, context, uuid, clientHost, clientPort, batches, batchSize, DEFAULT_PARALLEL_PARTITIONS);
+        this(rdd, context, uuid, clientHost, clientPort, batches, batchSize, DEFAULT_PARALLEL_PARTITIONS, OlapConfigurations.DEFAULT_SPARK_RESULT_STREAMING_THREADS);
     }
```
jyuanca, line 75: Should always use the default value, or the value configured by user?

ipraznik-splice: The first and second constructors are used only in ITs with default parameters, I've added comments. Only the last one is used in the standard cases and gets parameters configured:

`public StreamableRDD(JavaRDD<T> rdd, OperationContext<?> context, UUID uuid, String clientHost, int clientPort,
                         int batches, int batchSize, int parallelPartitions, int executorThreads) {`

hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
```diff
@@ -60,25 +63,36 @@
...
         this.uuid = uuid;
         this.host = clientHost;
         this.port = clientPort;
-        this.parallelPartitions = parallelPartitions % 2 == 0 ? parallelPartitions : parallelPartitions + 1;
-        this.executor = Executors.newFixedThreadPool(DEFAULT_PARALLEL_THREADS);
+        int defaultExecutorThreads = (int) max(min(ceil((float)parallelPartitions / 2), StreamableRDD.MAX_PARALLEL_THREADS),
+                OlapConfigurations.DEFAULT_SPARK_RESULT_STREAMING_THREADS);
+        this.executorThreads = max(defaultExecutorThreads, executorThreads);
+        this.parallelPartitions = this.executorThreads < parallelPartitions ? parallelPartitions : this.executorThreads;
+        this.executor = Executors.newFixedThreadPool(this.executorThreads);
```
hatyo, line 95: Could you provide a descriptive name of the threads in the thread pool to make it easy for debugging? something like `streamablerdd-streaming-thread-pool`

hatyo (approved):
LGTM, I noticed that you used the terms "ExecutorThread" and "StreamerThread" interchangeably, does it make sense to use "StreamerThread" (even in `StreamableRDD`)? it is up to you, but personally I think it makes sense.

jyuanca (approved):

hbase_sql/src/main/java/com/splicemachine/stream/StreamableRDD.java
ipraznik-splice: @hatyo , I've added the name for pool and rename some variables to use 'stream' in names

hatyo (approved):

carolp-503 (approved):
+1

PR #5544, opened 2021-06-02
DB-12094 Fix TriggerHandler cleanup

msirek:
## Short Description
TriggerHandler does not clean up the trigger rows result sets properly for prepared statements.

## Long Description
TriggerHandler -> triggerActivator -> tec (TriggerExecutionContext) may accumulate ResultSets in the resultSetVector of TriggerExecutionContext because TriggerHandler.cleanup() skips calling triggerActivator.cleanup when cleanup1Done is true. The specialized cleanup logic was originally added to avoid calling cleanup code repetitively in case it's called through different code flows.  When prepared statements are used, the same TriggerHandler may be used over and over again.  cleanup1Done doesn't get reset in between uses of the TriggerHandler, so any subsequent executions of the prepared statement may skip cleanup of the ResultSet which holds the trigger rows that were inserted/updated/deleted.

The fix is to restore the old TriggerHandler cleanup code and always call the cleanup routines.

Also, DB-11580 disallowed nested loop join for statement triggers.  Really, only the inner table of a nested loop join may not be the trigger VTI because the right side is executed in a separate thread, but not the left side (left == outer, right == inner).  So, the trigger VTI restriction on the outer table is removed.  This restriction was causing the updateHistoryTableConcurrently test in the FeatureStoreTriggerBenchmark not to run.

A canBeReused method is also added to TriggerRowHolderImpl.  Now, TriggerHandler will attempt to reuse a previously allocated trigger row holder on subsequent statement executions to avoid allocating the row buffer over and over again, which should help avoid unnecessary memory pressure.

## How to test
Run FeatureStoreTriggerBenchmark#updateHistoryTableConcurrently multiple times with a memory profiler or debugger attached and verify there is no leak of ResultSets.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
@@ -761,6 +761,7 @@ public class TriggerExecutionContext implements ExecutionStmtValidator, External
         if (triggeringResultSet != null)
         {
             triggeringResultSet.close();
+            triggeringResultSet = null;
         }
     }
 
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/compile/NestedLoopJoinStrategy.java
@@ -57,14 +57,15 @@ public class NestedLoopJoinStrategy extends BaseJoinStrategy{
         if (outerCost != null && outerCost.getJoinType() == JoinNode.FULLOUTERJOIN)
             return false;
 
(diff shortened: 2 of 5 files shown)


msirek:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3052/

msirek:

dgomezferro (approved):

hatyo (approved):

jyuanca (approved):

carolp-503 (approved):
+1

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3057/

PR #5549, opened 2021-06-03
DB-12109 Favor NLJ for PK joins with the trigger VTI

msirek:
## Short Description
Improve performance of joins between NEW TABLE or OLD TABLE and a splice internal table in a statement trigger.

## Long Description
A join between a single-row NEW TABLE and a base table on the primary key chose broadcast join instead of nested loop join, resulting in poor throughput.  Triggers are precompiled, so we have no way of knowing how many rows will be updated/inserted/deleted in the triggering statement, so no way to cost joins accurately.  The feature store use cases may involve single row inserts/updates or batch updates of thousands of rows.  We need to handle both efficiently and hopefully also avoid trigger recompilation.  The solution in this case is to adjust the DUMMY_ROWCOUNT_ESTIMATE in TriggerNewTransitionRows from 1000 down to 1 and also scale the join cost of nested loop join by a factor of 0.1 when the outer table is the trigger VTI and there are join predicates that can be used as start/stop keys.  Nested loop join on the base table PK will be fast for both small updates and large updates.  In the updateHistoryTableConcurrently test of FeatureStoreTriggerBenchmark the joinStrategy hint in the trigger of test FeatureStoreTriggerBenchmark is removed and no longer required.  Nested loop join is now picked without the hint.

In the future maybe there is something we can do to dynamically detect the number of rows in NEW/OLD TABLE, and force a trigger recompilation when this value differs significantly from the row count estimate used in the last trigger compilation.  This would be a large change though and would require tuning to avoid recompiling triggers too often.

## How to test
Run the updateHistoryTableConcurrently test of FeatureStoreTriggerBenchmark and observe that the ops/second throughput is high (above 2000) and matches the result when hinting nested loop join.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/execute/TriggerExecutionContext.java
@@ -761,6 +761,7 @@ public class TriggerExecutionContext implements ExecutionStmtValidator, External
         if (triggeringResultSet != null)
         {
             triggeringResultSet.close();
+            triggeringResultSet = null;
         }
     }
 
--- a/splice_machine/src/main/java/com/splicemachine/derby/catalog/TriggerNewTransitionRows.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/catalog/TriggerNewTransitionRows.java
@@ -88,7 +88,7 @@ import static com.splicemachine.derby.impl.sql.execute.operations.ScanOperation.
 public class TriggerNewTransitionRows
                    implements DatasetProvider, VTICosting, AutoCloseable {
 
-        private static final double DUMMY_ROWCOUNT_ESTIMATE = 1000;
+        private static final double DUMMY_ROWCOUNT_ESTIMATE = 1;
         private static final double DUMMY_COST_ESTIMATE = 1000;
 	private ResultSet resultSet;
 	private DataSet<ExecRow> sourceSet;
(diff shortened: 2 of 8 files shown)


msirek:

dgomezferro (approved):
I'm OK with the change but this reminds me of the index access vs full table scan for empty tables issue. I think we should have a more general solution that favors the safer approach (using lookups) when the table might grow beyond the current statistics.

msirek:
> I'm OK with the change but this reminds me of the index access vs full table scan for empty tables issue. I think we should have a more general solution that favors the safer approach (using lookups) when the table might grow beyond the current statistics.

Yes, but it's even more basic than that.  We don't have any information as to the size of the trigger rows at query planning time, and the framework for checking this at runtime and making this number available to the optimizer isn't there.  One approach could be to make dictionary changes to allow more that one SPSDescriptor (stored statements) for a trigger to be stored in sysstatements for different potential sizes of trigger rows or combinations of OLTP vs. OLAP execution.  Then at runtime we could check the available SPSDescriptors and pick the one that was optimized for our current situation, and if the trigger rows size is too far away from what we've compiled for, create a new trigger compilation and add it to sysstatements for future use.  It's not a small change at all, but may be worth the effort.  This approach would prevent excessive recompilations of triggers for simultaneously-running DMLs of different sizes.

jyuanca (approved):

msirek:
IT run:
http://nexus.splicemachine.com:8080/job/spliceengine-branch/job/spliceengine-branch/2120/default/

carolp-503 (approved):
+1

PR #5551, opened 2021-06-03
DB-11809 FeatureStoreBenchmark: implemented queries based on UNION

OlegMazurov:
## Short Description
A different variant of FeatureStoreBenchmark queries was proposed to ensure more parallelism in multi-table PK lookups. This PR adds that variant to the benchmark.  

## Long Description
In a table with 10 columns, select columns c2, c4, c6, c8, c10, with 100 columns: c20, c40, c60, c80, c100 and with 1000: c200, c400, c600, c800, c1000.
```
SELECT max(t1_c2), ... max(t2_c20), ... max(t3_c200), ... FROM (
    SELECT T1.c1 t1_c1, ... cast(NULL as int) t2_c20, ..., cast(NULL as int) t3_c200, ... FROM T1 WHERE (T1.<primaryKey>, ...) = (<random value>, ...)
    UNION ALL
    SELECT cast(NULL as int) t1_c1, ... T2.c20 t2_c20, ... cast(NULL as int) t3_c200, ... FROM T2 WHERE (T2.<primaryKey>, ...) = (<random value>, ...)
 ... )
```
## How to test
Run FeatureStoreBenchmark as an integration test either in standalone or cluster environment.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreBenchmark.java
+++ b/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreBenchmark.java
@@ -21,13 +21,18 @@ import com.splicemachine.test.Benchmark;
 import org.apache.log4j.Logger;
 import org.junit.*;
 import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
 import java.sql.*;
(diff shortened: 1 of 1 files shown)


dgomezferro (approved):

carolp-503 (approved):
+1

PR #5556, opened 2021-06-09
DB-12032 FeatureStoreTriggerBenchmark: improve performance of batch updates 

OlegMazurov:
## Short Description
A pre-populated staging table is used in batch updates of FeatureTable or FeatureHistoryTable. How the table is generated is crucial for batch update performance.

## Long Description
The staging table is populated by batched inserts from multiple concurrent connections. It's important to turn autocommit off in those connections in order to avoid generating a large number of small transactions which will need to be resolved when the table is queried.
TRUNCATE does not preserve pre-split points specified at table creation time. Had to replace with DROP and re-CREATE.
ANALYZE the staging table to get better load balance for Spark tasks scanning the table. 

## How to test
Run FeatureStoreTriggerBenchmark as an integration test either in standalone or cluster environment.

### DIFF
--- a/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
+++ b/splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
@@ -158,6 +158,7 @@ public class FeatureStoreTriggerBenchmark extends Benchmark {
     // Generate timestamps between ts0 inclusive and ts1 exclusive
     private static void populateStagingTable(long ts0, long ts1) {
         try (Connection conn = makeConnection()) {
+            conn.setAutoCommit(false);
             PreparedStatement insert = conn.prepareStatement("INSERT INTO FeatureStaging VALUES (?, ?, ?, ?, ?)");
             Random rnd = ThreadLocalRandom.current();
             int maxDay = (int)((ts1 - ts0 + DAYMS - 1) / DAYMS);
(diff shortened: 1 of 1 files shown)


dgomezferro (approved):

splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
```diff
@@ -200,23 +184,45 @@ private static void populateStagingTable(long ts0, long ts1) {
...
+                        "PRIMARY KEY (entity_key, update_order))" +
+                        "LOGICAL SPLITKEYS LOCATION '%s'", fileName)
+        );
+
+        long nextTimestamp = curTimestamp + updates * 2 * DAYMS;
         taskId.set(0);
         runBenchmark(connections, () -> populateStagingTable(curTimestamp, nextTimestamp));
         curTimestamp = nextTimestamp;
+
+        testStatement.execute("ANALYZE TABLE FeatureStaging");
```
dgomezferro, line 213: I don't think this would actually help Spark balance tasks better, if anything just because we are delaying the execution and HBase might complete flushes/compactions/splits, but the analyze itself wouldn't play a role I think.

If you wanted that you could force a flush and major_compaction, I think that would help more.

arnaud-splice (approved):

carolp-503 (approved):
+1 after Daniel's question is resolved.

splice_machine/src/test/java/com/splicemachine/benchmark/FeatureStoreTriggerBenchmark.java
OlegMazurov: I think you are right. I was able to get comparable performance with a simple delay instead of ANALYZE.
FLUSH + MAJOR COMPACTION improve batch performance from 35K op/sec to 50K ops/sec (but only in one out of the two tests).

PR #5558, opened 2021-06-10
DB-11855 Fix index corruption caused by foreign key on delete set null action.

hatyo:
In this PR an index corruption caused by foreign key action on the child table is fixed.

## Long Description
Let us say we have a parent table `p` and child table `c` have a foreign key relationship with `on delete set null action`. If we delete a row from the parent the is referenced by row(s) in the child, then we generate an `UPDATE` mutation on the child table and its indices to set all referenced columns to `null`.

However, while doing so, we incorrectly set all other columns in the child indices to null as well! this is because the generated `update` mutation does not have the values of the non-fk index columns and we simply set them to `null`.

To fix this problem, a new mutation type is added `0x09` under the name `BLIND_UPDATE`. This mutation type is leveraged in situations where we don't have the full overview of the update's scope, as for example in the case where we update some descendant table deep down in a foreign key hierarchy; unlike normal updates where we know about both old and new columns at the moment of update allowing us to eliminate an extra base table lookup to reconstruct the not-updated index columns (see #5391 for more information).

When we receive a `BLIND_UPDATE` mutation in `IndexWriteHandler`, we perform a base table lookup on the affected row, meld the column values of the mutation with the base table row, and continue to perform the index update (delete + insert).

I implemented a new method `Utils.meld` that allows melding two `EntryDecoder` columns into one. This method is used to above to reconstruct the "blindly" updated row.

## How to test

Here is a simple test:

```sql
create table p_a(a int, d int primary key, e int);
create table c_a (x int primary key, y int references p_a(d) on delete set null, z int);
create index c_ix on c_a(y, z);
insert into p_a values (22, 98, 66);
insert into c_a values (4 , 98, 4);

-- delete from the parent table to trigger child table update via the foreign key action
delete from p_a;

-- previously, querying c_a columns via the index was returning wrong results
-- because the index is corrupted
select z from c_a --splice-properties index=c_ix
;
Z
-
NULL

-- with this PR, the index is not corrupted any more , it is working as expected
select z from c_a --splice-properties index=c_ix
;
Z
-
4
```

I added an extensive randomized test suite for `Utils.meld` and a number of ITs in `ForeignKeyActionIT` for testing the index update patch fix.


hatyo:

cloudspliceci:
TEST SUCCEEDED +1

splice_machine/src/main/java/com/splicemachine/pipeline/IndexWriteHandler.java
```diff
@@ -109,7 +102,13 @@ public boolean route(KVPair mutation,WriteContext ctx) {
                     mutation = getBaseUpdateMutation(mutation);
                     return createIndexRecord(mutation, ctx, delete);
                 }
-                return true; // No index columns modifies ignore...
+                return true; // No index columns modified, ignore...
+            case BLIND_UPDATE:
+                if(transformer.areIndexKeysModified(mutation)) {
```
dgomezferro, line 107: I think this function has to be updated, or a different function is needed for BLIND_UPDATES. The logic here has to be: if any indexed column is modified, we process the update. 
Instead of: only when all indexed columns are modified, we process the update.

dgomezferro: I think a test case for this would be something like:
```
create table a (i int primary key);
create table b (i int references a(i) on delete set null, j int);
create index bi on b (i,j);
insert into a values 1;
insert into b values (1,1);
delete from a;
select * from b --splice-properties index=bi
;
```

dgomezferro (changes requested):
This is great @hatyo ! Nice test cases too.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
```diff
@@ -223,9 +223,8 @@ public KVPair encodeSystemTableIndex(ExecRow execRow) throws StandardException,
         byte[] key = DerbyBytesUtil.generateIndexKey(dvds, order,tableVersion,false);
 
         if(entryEncoder==null){
-            int[] validCols= EngineUtils.bitSetToMap(null);
             DescriptorSerializer[] serializers=VersionedSerializers.forVersion(tableVersion,true).getSerializers(execRow);
-            entryEncoder=new EntryDataHash(validCols,null,serializers);
```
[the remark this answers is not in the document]
hatyo: Oh it is not needed, it is a small refactoring, because if you pass `null` to `EngineUtils.bitSetToMap` you get back `null`. I thought I make the intended usage clear here by directly passing the `null`. I can revert it if you think it irrelevant.

splice_machine/src/main/java/com/splicemachine/pipeline/IndexWriteHandler.java
hatyo: Thanks, I had that in my notes to address but I forgot to do it. I will push a fix for it plus a test.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/index/IndexTransformer.java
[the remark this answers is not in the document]
dgomezferro: That make sense, I also think it's clearer this way, thanks!

dgomezferro (approved):
Looks good, thanks!

arnaud-splice (approved):

ascend1 (approved):

hatyo:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5567, opened 2021-06-14
DB-12132 Put cost model property into GlobalDBProperties, add an IT

ascend1:
## Short Description
This PR refactors `costModel` database property to be part of `GlobalDBProperties` class, using its capability of having a validator. Also, an IT `testCostModel()` is added to `GlobalDatabasePropertyIT`.

## Long Description
No long description needed for this PR.

## How to test
```
call syscs_util.syscs_set_global_database_property('costModel', 'v2');
call syscs_util.syscs_get_global_database_property('costModel');  -- should see v2 as the value
explain select * from sys.systables;  -- should see costModel=v2 in plan, no need to drop current session

call syscs_util.syscs_set_global_database_property('costModel', 'NULL');
call syscs_util.syscs_get_global_database_property('costModel');  -- should see NULL as the value
explain select * from sys.systables;  -- should see costModel=v1 in plan since v1 is the default value, no need to drop current session
```

### DIFF
--- a/db-shared/src/main/java/com/splicemachine/db/iapi/reference/GlobalDBProperties.java
+++ b/db-shared/src/main/java/com/splicemachine/db/iapi/reference/GlobalDBProperties.java
@@ -136,4 +136,9 @@ public class GlobalDBProperties {
             new PropertyType("splice.execution.newMergeJoin",
                     "use new merge join. possible values: on, off, forced",
                     new Validator.MultipleOptions(new String[]{"on", "off", "forced"}));
+
+    public static PropertyType COST_MODEL =
+            new PropertyType("costModel",
+                             "cost model version, usable versions are v1 and v2",
+                             new Validator.MultipleOptions(new String[]{"v1", "v2"}));
 }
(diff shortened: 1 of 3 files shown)


ascend1:

cloudspliceci:
TEST SUCCEEDED +1

db-shared/src/main/java/com/splicemachine/db/iapi/reference/GlobalDBProperties.java
```diff
@@ -136,4 +136,9 @@ public String getInformation() {
             new PropertyType("splice.execution.newMergeJoin",
                     "use new merge join. possible values: on, off, forced",
                     new Validator.MultipleOptions(new String[]{"on", "off", "forced"}));
+
+    public static PropertyType COST_MODEL =
+            new PropertyType("costModel",
+                             "cost model version, usable versions are v1 and v2",
```
hatyo, line 142: instead of "usable" I would write "available"

hatyo (approved):

martinrupp (approved):

db-shared/src/main/java/com/splicemachine/db/iapi/reference/GlobalDBProperties.java
ascend1: Right, that's better. Changed to "available" in the latest commit.

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5571, opened 2021-06-15
DB-11857 Short-circuit HBaseConglomerate.readExternal once the engine has started (3.1)

arnaud-splice:


carolp-503 (approved):
Approved for 3.1

PR #5573, opened 2021-06-15
DB-11855 Fix index corruption caused by foreign key on delete set null action (3.0)

hatyo:
In this PR an index corruption caused by foreign key action on the child table is fixed.

## Long Description
Let us say we have a parent table `p` and child table `c` have a foreign key relationship with `on delete set null action`. If we delete a row from the parent the is referenced by row(s) in the child, then we generate an `UPDATE` mutation on the child table and its indices to set all referenced columns to `null`.

However, while doing so, we incorrectly set all other columns in the child indices to null as well! this is because the generated `update` mutation does not have the values of the non-fk index columns and we simply set them to `null`.

To fix this problem, a new mutation type is added `0x09` under the name `BLIND_UPDATE`. This mutation type is leveraged in situations where we don't have the full overview of the update's scope, as for example in the case where we update some descendant table deep down in a foreign key hierarchy; unlike normal updates where we know about both old and new columns at the moment of update allowing us to eliminate an extra base table lookup to reconstruct the not-updated index columns (see #5391 for more information).

When we receive a `BLIND_UPDATE` mutation in `IndexWriteHandler`, we perform a base table lookup on the affected row, meld the column values of the mutation with the base table row, and continue to perform the index update (delete + insert).

I implemented a new method `Utils.meld` that allows melding two `EntryDecoder` columns into one. This method is used to above to reconstruct the "blindly" updated row.

## How to test

Here is a simple test:

```sql
create table p_a(a int, d int primary key, e int);
create table c_a (x int primary key, y int references p_a(d) on delete set null, z int);
create index c_ix on c_a(y, z);
insert into p_a values (22, 98, 66);
insert into c_a values (4 , 98, 4);

-- delete from the parent table to trigger child table update via the foreign key action
delete from p_a;

-- previously, querying c_a columns via the index was returning wrong results
-- because the index is corrupted
select z from c_a --splice-properties index=c_ix
;
Z
-
NULL

-- with this PR, the index is not corrupted any more , it is working as expected
select z from c_a --splice-properties index=c_ix
;
Z
-
4
```

I added an extensive randomized test suite for `Utils.meld` and a number of ITs in `ForeignKeyActionIT` for testing the index update patch fix.

### DIFF
--- a/pipeline_api/src/main/java/com/splicemachine/pipeline/PipelineWriter.java
+++ b/pipeline_api/src/main/java/com/splicemachine/pipeline/PipelineWriter.java
@@ -279,7 +279,8 @@ public class PipelineWriter{
                 case INSERT:
                     privileges.add(Authorizer.INSERT_PRIV);
                     break;
-                case UPDATE:
+                case UPDATE: // fallthrough
+                case BLIND_UPDATE:
                     privileges.add(Authorizer.UPDATE_PRIV);
                     break;
                 case UPSERT:
--- a/splice_encoding/src/main/java/com/splicemachine/kvpair/KVPair.java
+++ b/splice_encoding/src/main/java/com/splicemachine/kvpair/KVPair.java
@@ -31,7 +31,13 @@ public class KVPair implements Comparable<KVPair> {
         EMPTY_COLUMN((byte)0x04),
         UPSERT((byte)0x05),
         /* For import process to cancel out an inserted row that violates a unique constraint */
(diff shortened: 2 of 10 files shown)


hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3132/

carolp-503 (approved):
Approved for 3.0

PR #5579, opened 2021-06-16
DB-10800 Make hashable joins infeasible for non-covering IoEs

ascend1:
## Short Description
This change makes hashable joins (broadcast, merge, sortmerge) infeasible for non-covering indexes on expressions (IoEs).

## Long Description
In previous work of implementing indexes on expressions, one extension to `findHashKeys()` is to let an expression on one side of an equi-join predicate that matches an index expression be recognized as a hash key. This is necessary to make hashable joins feasible for IoE access paths as well as costing them correctly. Executing such a plan is possible because the matching expression is replaced with a column reference pointing to the corresponding index column right before code generation phase.

For classic access paths like table scan or classic index scan, a hash key must be a column reference. An expression cannot be used as a hash key for now.

However, this process should happen only for covering IoEs. If the index is not covering, index lookup operation will retrieve base table rows and expressions (even they match an index expression) are not replaced because we are not directly accessing any index column in this case. If `findHashKeys()` still treat matching expressions as hash keys, we would generate code in which hash keys are the evaluation results of the expressions.

This is generally fine. But our current execution has a step of asking for an output row template of `ProjectRestrictOperation` in `fillResultSet()`, i.e., execution preparation phase. This is done by invoking the generated projection method. If this method tries to evaluate an expression like `a + b`, it will cause a NPE because data is not ready yet. We can only evaluate such an expression later in real execution, i.e., in `getResultSet()` call tree.

For this reason, this change makes hashable joins feasible only for covering IoEs. It should be generally possible to support expression as a hash key, but that's a different and bigger topic. 

## How to test
Setup:
```
drop table if exists T1;
create table T1 (a1 int, b1 int, c1 int, d1 int);
insert into T1 values (1,1,1,1), (2,2,2,2), (3,3,3,3), (4,4,4,4), (5,5,5,5);
create index T1IDX on T1(b1 + c1);
create index T1IDX2 on T1(b1);

drop table if exists T2;
create table T2 (a1 int, b1 int, c1 int, d1 int);
insert into T2 values (1,1,1,1), (2,2,2,2);
create index T2IDX on T2(b1 + c1);
create index T2IDX2 on T2(b1);
```

Query:
```
select *
from T1 --splice-properties index=T1IDX
         join T2 --splice-properties index=T2IDX
              on T1.b1 + T1.c1 = T2.b1 + T2.c1;
```

Before this fix, the query fails due to an NPE (assertion disabled) or assertion failure (assertion enabled). After this fix, the query should return the following result without any error:
```
A1|B1|C1|D1|A1|B1|C1|D1
-----------------------
1|1|1|1|1|1|1|1
2|2|2|2|2|2|2|2
```


ascend1:

cloudspliceci:
TEST SUCCEEDED +1

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

jyuanca (approved):

msirek:
@ascend1 This seems like a limitation we'd want to remove as soon as we can.  Could you please open a Jira for it if one is not already open?

msirek (approved):

carolp-503 (approved):
+1

ascend1:
> @ascend1 This seems like a limitation we'd want to remove as soon as we can. Could you please open a Jira for it if one is not already open?

Sure, I just created it: https://splicemachine.atlassian.net/browse/DB-12265.

PR #5582, opened 2021-06-16
DB-11826 DB-12226 Fix time-travel returning incorrect results due to using temporal-inconsistent index

hatyo:
## Short Description
In this PR, we target a subtle inconsistency caused by time-travel allowing the optimizer to choose a temporal-inconsistent index, causing wrong results to be returned.

## Long Description

With time-travel, the user view the historical data of a table within a min. retention period. It could happen that the optimizer prefers an AccessPath involving an index conglomerate that is created _after_ the past transaction ID which is dangerous, because the base table could have changed between the past transaction ID and the index creation time (e.g. some rows are deleted before the index was created), if the optimizer chooses that index conglomerate we will get empty results because the index does not have any data at the past transaction ID since it was created after it.

To fix this issue we make sure, during optimization and when deciding whether an index access path is eligible or not, that the index's creation time precedes the past transaction ID, and only accept it in that case.

To achieve this, I implemented the following changes in this PR:

1. Add API extensions to `DataDictionary` that allows, for a given conglomerate, to return its creation transaction ID from the conglomerate's metadata.
2. Move `AS OF` expression evaluation from execution time to binding phase, since it is crucial to know upfront the past transaction ID during optimization to figure out whether an index is eligible or not by comparing it to the index's creation transaction ID.
3. Limit `AS OF` to expressions that can be evaluated at binding phase, for now, we limit to the subset of constant expressions, but in the future we can extend it to more expressions.
4. Implement constant evaluation for expressions that are used quite often with `AS OF`, i.e. the `TIMESTAMP` expression (in `UnaryDateTimestampOperatorNode`) and `TIMESTAMPADD` expression (in `TernaryOperatorNode`).
5. Throw a proper error message if the user attempts to force a temporal-inconsistent index.

With this fix, the optimizer is able now to choose valid indices that are temporal-consistent if an `AS OF` expression is present.

## How to test

Here is a quick example:

```sql


create table t1(col1 int);
insert into t1 values 42, 42;
CALL SYSCS_UTIL.SYSCS_GET_CURRENT_TRANSACTION();
txnId
----

19200

delete from t1;
alter table t1 add constraint c1  unique(col1);

select * from t1 as of 19200;
COL1
-----------
42
42

select * from t1 as of 19200 where col1 >= 42; –- wrong result
COL1
-----------

-– because we are using the newly introduced index, which is empty
explain select * from t1 as of 19200> where col1 >= 42;
Plan
-------------------------------------------------------------------
Cursor(n=3,rows=18,updateMode=READ_ONLY (1),engine=OLTP (default))
   ->  ScrollInsensitive(n=3,totalCost=8.216,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
    ->  IndexScan[SQLA35EC2CE0178AC35AAF500000B960C28(1729)] (n=0,totalCost=4.036,scannedRows=18,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,baseTable=T1(1712),preds=[(COL1[0:1] >= 42)]) 
    
-- with the fix we get the correct results, since the index was not chosen!
select * from t1 as of 19200 where col1 >= 42;
COL1
-----------
42
42

explain select * from t1 as of 19200 where col1 >= 42;

Cursor(n=3,rows=18,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=3,totalCost=8.22,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
    ->  TableScan[T1(2512)](n=0,totalCost=4.04,scannedRows=20,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,preds=[(COL1[0:1] >= 42)])
Table statistics are missing or skipped for the following tables:
SPLICE.T1

-- moreover, force-using the index throws an error now:

select * from t1 as of 19200 --splice-properties index=i1
where col1 >= 42;
ERROR 42Z03: Specified index 'I1' is created at tx '21042' which is after time-travel mapped past transaction id '19200'.
```


hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3164/

hatyo:

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3177/

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -344,6 +346,10 @@ else if (currentConglomerateDescriptor.getIndexDescriptor().excludeDefaults()
                     && (predList == null || !predList.canSupportIndexExcludedDefaults(tableNumber,currentConglomerateDescriptor, tableDescriptor))) {
                 return false;
             }
+            else if (pastTxnId != -2) {
```
jyuanca, line 349: What is magic number -2?

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
```diff
@@ -2319,4 +2320,18 @@ PermDescriptor getGenericPermissions(UUID objectUUID,
...
+     * @return The ID of the transaction used to create it.
+     * @throws StandardException In case of e.g. connection error to HBase admin.
+     */
+    long getConglomerateCreationTxId(long conglom) throws StandardException;
+
+    long getTxnAt(long ts) throws StandardException;
+
+    boolean txnWithin(long period, long pastTx) throws StandardException;
+
+    boolean txnWithin(long period, Timestamp pastTx) throws StandardException;
```
jyuanca, line 2336: getTxnAt and txnWithin are not related to database data dictionary, so you may want to refactor it to somewhere else.

jyuanca (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
```diff
@@ -1404,6 +1406,66 @@ public boolean useTxnAwareCache() {
...
+            }
+            PartitionAdmin admin = tableFactory.getAdmin();
+            if(SanityManager.DEBUG) {
+                SanityManager.ASSERT(admin != null);
+            }
+            com.splicemachine.access.api.TableDescriptor descriptor = admin.getTableDescriptor(Utils.constructHbaseName(partition.getTableName()));
+            if(SanityManager.DEBUG) {
+                SanityManager.ASSERT(descriptor != null);
+            }
+            return Long.parseLong(descriptor.getTransactionId());
```
dgomezferro, line 1432: There's old tables/indexes that don't have a transaction id, this would fail for them and make them inaccessible. I think we can assume they are always older than the requested TXN

dgomezferro (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/SpliceDataDictionary.java
hatyo: Thanks, I didn't know that, I fixed it assuming we get `null` from `TableDescriptor.getTransactionId` for these old indexes / tables, and return 0 instead allowing the optimizer to consider them for time-travel queries.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
hatyo: -2 indicates that time-travel is not set at all, i.e. we should go ahead and use current transaction for the scan.
-1 indicates that time-travel is beyond the oldest transaction, therefore, when doing the scan, we choose `SIConstants.OLDEST_TIME_TRAVEL_TX` as past txn id.
I use now proper constants instead of these magic numbers for better readability, thanks for hinting it out.

db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
hatyo: I changed it, and moved these methods to `TransactionController` instead, please let me know if you think there is a better location for them.

dgomezferro (changes requested):
Changes look good to me, I just have some doubts regarding `TransactionController.TIME_TRAVEL_OLDEST`.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromBaseTable.java
```diff
@@ -240,7 +239,7 @@
 
     private ValueNode pastTxIdExpression = null;
 
-    private long pastTxnId = -1;
+    private long pastTxnId = TransactionController.TIME_TRAVEL_UNSET;
```
dgomezferro, line 242: Is this correct? We were setting it to `-1` (`OLDEST`) and now it's being set to `-2` (`UNSET`)

hatyo: Yes I think it is, `pastTxnId` is set accordingly later on in `bindNonVTITables`.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNode.java
```diff
@@ -71,6 +71,7 @@
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.spark.sql.execution.command.DDLUtils$;
```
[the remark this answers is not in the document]
hatyo: done.

db-engine/src/main/java/com/splicemachine/db/iapi/store/access/TransactionController.java
```diff
@@ -264,6 +265,8 @@ gaining access to resources (conglomerates, scans, etc.) in the transaction
 	byte IS_TEMPORARY	=	(byte) 0x01; // conglom is temporary
 	byte IS_KEPT		=	(byte) 0x02; // no auto remove
 
+	int TIME_TRAVEL_OLDEST = -1;
```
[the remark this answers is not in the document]
hatyo: We do return -1 in `com.splicemachine.si.impl.SkeletonTxnNetworkLayer#getTxnAt`, -1 indicates that the past transaction id is older than the oldest transaction in time-travel.

We use the above constant `TIME_TRAVEL_OLDEST` in `com.splicemachine.derby.impl.sql.execute.operations.ScanOperation#ScanOperation` for checking for that case.

dgomezferro (approved):

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3292/

hatyo:

arnaud-splice (approved):

carolp-503 (approved):
+1

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3297/

ipraznik-splice:

hatyo:
> TEST FAILED
> http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3297/

Failure is a sporadic that is not related to the PR itself; creation of test tables (in @BeforeClass) is failing for some reason:

```java
@BeforeClass
    public static void createSharedTables() throws Exception {
        TestConnection connection=classWatcher.getOrCreateConnection();
        new TableCreator(connection)
                .withCreate("create table t1 (a1 int, b1 int, c1 int, d1 int, primary key(d1))")
                .withInsert("insert into t1 values(?,?,?,?)")
                .withRows(rows(row(1, 1, 1, 1),
                        row(1, 1, 1, 11),
                        row(2, 2, 2, 2),
                        row(3, 3, 3, 3),
                        row(4, 4, 4, 4))).create(); <----- here
```

It looks like we have a similar occurrence of it but in a different test suite reported in [DB-9892](https://splicemachine.atlassian.net/browse/DB-9892).

cloudspliceci:
TEST SUCCEEDED +1

PR #5596, opened 2021-06-21
DB-12232 Add Triggering Query Identifier to Log Output

mush-zhang:
## Short Description
Add Triggering Query Identifier to Log Output

## Long Description
Add an addtional field to Execution Start log record about a query’s tringerring query if exists. 

Sample Log record:

'''
2021-06-16 11:13:56,501 Thread[DRDAConnThread_1402] Start executing query. XID=SpliceTransaction[ACTIVE,WritableTxn(14848,ACTIVE)], SessionID=4, Database=splicedb, DRDAID=��������.����-738307240056184952{2}, UserID=SPLICE, uuid=f4cf0d28-8f9a-92d9-8ab2-fb9ed7dec384, engine=CONTROL, sqlHash=2bd3289e34490a73cc1feab46f664956, statement=[ insert into tableb values (CAST (com.splicemachine.db.iapi.db.Factory::getTriggerExecutionContext().getNewRow().cloneColumn(1).getObject() AS INTEGER) *40) ], paramsCount=0, params=[  ], sessionProperties=[ null ], triggeredBy=[ 84a8a5db-9b8c-709b-ffec-9293f4a29ac5 ]
'''

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
@@ -617,6 +617,12 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
             triggerInfos = new ArrayList<>(queryIdToTriggerInfoMap.values());
         }
     }
+
+    @Override
+    public java.util.UUID getLatestQueryId() {
+        return null;
+    }
+
     @Override
     public void initialize() throws StandardException{
         interruptedException = null;
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
@@ -4108,12 +4108,12 @@ public class GenericLanguageConnectionContext extends ContextImpl implements Lan
 
     @Override
     public void logStartExecuting(String uuid, String engine, String stmt, ExecPreparedStatement ps,
-                                  ParameterValueSet pvs) {
+                                  ParameterValueSet pvs, String parentId) {
         if (stmtLogger.isInfoEnabled()) {
             stmtLogger.info(String.format(
-                "Start executing query. %s, uuid=%s, engine=%s, %s, paramsCount=%d, params=[ %s ], sessionProperties=[ %s ]",
+                "Start executing query. %s, uuid=%s, engine=%s, %s, paramsCount=%d, params=[ %s ], sessionProperties=[ %s ], triggeredBy=[ %s ]",
                 getLogHeader(), uuid, engine, formatLogStmt(stmt),
-                pvs.getParameterCount(), pvs.toString(), ps.getSessionPropertyValues()));
+                pvs.getParameterCount(), pvs.toString(), ps.getSessionPropertyValues(), parentId));
         }
     }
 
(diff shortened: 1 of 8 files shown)


splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
```diff
@@ -608,11 +609,19 @@ a triggered statement(GenericTriggerExecutor.compile(), in those cases, we want
                 stmtForLogging = ps.getSourceTxt();
         }
 
+        UUID parentId = null;
+
+        try {
+            parentId = getCurrentTransaction().getParentQueryIdForTrigger();
+        } catch (Exception ignored) {}
+
```
jyuanca, line 617: Log a warning message for the error

splice_si_api/src/main/java/com/splicemachine/si/impl/txn/AbstractTxn.java
```diff
@@ -57,6 +68,23 @@ protected AbstractTxn(
...
+    public UUID getParentQueryIdForTrigger() {
+        AbstractTxn parent;
+        if (parentReference == null || parentReference.getCurrentQueryId() == null) {
+            try {
+                parent = (AbstractTxn) getParentTxnView();
+                if (parent.getCurrentQueryId() == null) {
+                    return null; // maybe should throw an error here
+                } else {
+                    return parent.getCurrentQueryId();
+                }
```
jyuanca, line 80: simplify to return parent.getCurrentQueryId()?

jyuanca (changes requested):

splice_si_api/src/main/java/com/splicemachine/si/impl/txn/AbstractTxn.java
```diff
@@ -57,6 +68,23 @@ protected AbstractTxn(
...
+            } catch (Exception ignore) {
+                return null;
```
ipraznik-splice, line 82: please add a warning log if possible

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -617,6 +617,12 @@ public void recordAdditionalDisplayedTriggerInfo(long elapsedTime, long modified
             triggerInfos = new ArrayList<>(queryIdToTriggerInfoMap.values());
         }
     }
+
+    @Override
+    public java.util.UUID getLatestQueryId() {
```
ipraznik-splice, line 622: is this method used anywhere?

ipraznik-splice (changes requested):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/SpliceBaseOperation.java
mush-zhang: I am logging the warning message in case getParentQueryIdForTrigger() results in one. Functino getCurrentTransaction may throw standardException, which I ignore it, as `logExecutionStart` originally did not call `getCurrentTransaction`

arnaud-splice (approved):

msirek (changes requested):
It seems like it may be a little problematic to add tracking of the triggering statement through relationships between transactions.  New transactions can be started any time, or they may be looked up via transaction id and not have the query UUID filled in. Luckily, the triggering operation can already be found via the parent activation.  If activation.getParentActivation() is not null, then activation.getParentActivation().getResultSet().getUuid() should give you the UUID of the triggering DML operation (insert, update, or delete), and there is no need to add the tracking of UUIDs in transactions.

PR #5599, opened 2021-06-22
DB-9348 Fix natural right outer join

arnaud-splice:
## Short Description
Fix natural right outer join 

## Long Description
During Parsing phase, we construct a HalfOuterJoinNode and if it is a right outer join, we set joinResultSet and rightOuterJoinUsingClause in every result column of top select nodes.
However, during optimization phase, we remove the original HalfOuterJoinNode to create a new one. The new one will get a resultSetNumber assigned to it. Since the original HalfOuterJoinNode was removed, we never assign a resultSetNumber to it.
However, the top SelectNode was never cleared of its reference to the original HalfOuterJoinNode, leading to issues in code generation.

## How to test

```
drop table staff3;
drop table staff66;

create table staff3(EMPNUM   VARCHAR(3) NOT NULL, GRADE    DECIMAL(4));
INSERT INTO STAFF3 VALUES ('E2',10);

CREATE TABLE STAFF66 (SALARY DECIMAL(14,0),EMPNUM VARCHAR(3) NOT NULL) ;

SELECT EMPNUM, SALARY FROM STAFF66 NATURAL RIGHT OUTER JOIN STAFF3 ;
```

We should increase our test coverage for Natural joins. This is covered by DB-12270

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
@@ -2054,6 +2054,12 @@ public class SelectNode extends ResultSetNode {
                         getContextManager()
                         );
             }
+            for (ResultColumn rc : this.getResultColumns()) {
+                rc.setJoinResultset(null);
+                rc.setRightOuterJoinUsingClause(false);
+                joinNode.isJoinColumnForRightOuterJoin(rc);
+            }
+
             joinNode.setCostEstimate(rightResultSet.getCostEstimate());
 
             ResultSetNode newPRNode = joinNode.genProjectRestrict();
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
@@ -2054,6 +2054,13 @@ public class SelectNode extends ResultSetNode {
                         getContextManager()
                         );
             }
+            /*
+             * During Parsing phase, we construct a HalfOuterJoinNode and if it is a right outer join,
+             * we set joinResultSet and rightOuterJoinUsingClause in every result column of top select nodes.
+             * However, here, we removed the original HalfOuterJoinNode to create a new one.
+             * The new one will get a resultSetNumber assigned to it.
+             * We need to clear the top SelectNode of its reference to the original HalfOuterJoinNode
+             */
             for (ResultColumn rc : this.getResultColumns()) {
                 rc.setJoinResultset(null);
                 rc.setRightOuterJoinUsingClause(false);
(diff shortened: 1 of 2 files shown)


arnaud-splice:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3209/

arnaud-splice:

cloudspliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2054,6 +2054,12 @@ public ResultSetNode modifyAccessPaths() throws StandardException {
                         getContextManager()
                         );
             }
+            for (ResultColumn rc : this.getResultColumns()) {
```
hatyo, line 2057: a comment here would be nice.

hatyo (commented):
Great work

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
arnaud-splice: Very good point. Done

hatyo (approved):

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
```diff
@@ -2054,6 +2054,19 @@ public ResultSetNode modifyAccessPaths() throws StandardException {
...
+             * However, here, we removed the original HalfOuterJoinNode to create a new one.
+             * The new one will get a resultSetNumber assigned to it.
+             * We need to clear the top SelectNode of its reference to the original HalfOuterJoinNode
+             */
+            for (ResultColumn rc : this.getResultColumns()) {
+                rc.setJoinResultset(null);
+                rc.setRightOuterJoinUsingClause(false);
+                joinNode.isJoinColumnForRightOuterJoin(rc);
+            }
+
```
martinrupp, line 2069: looks good to me, however i don't really understand why this would only apply to natural joins. to me it looks like natural join is a normal join without a ON clause and a generated USING clause. i think one could create a right outer join which would run into the same problem, right?

martinrupp (approved):

carolp-503 (approved):
+1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/SelectNode.java
arnaud-splice: See comments in `ResultColumn.rightOuterJoinUsingClause`. This was introduced by [DERBY-4631](https://issues.apache.org/jira/browse/DERBY-4631?attachmentSortBy%3DdateTime) and is only used for right outer join with using / natural

PR #5604, opened 2021-06-23
DB-12225 Implement constant folding for numeric binary +-*/% 

arnaud-splice:
## Description

Support plus, minus, times, divide, mod constant folding for numeric operands

## How to test

This can be tested via an explain on simple selects:

```
explain select * from sysibm.sysdummy1 where 1 = 1 + 1
```

will return a plan without a table scan because 1 = 1 + 1 is evaluated to be false at compile time
The same principle will apply to the other operators supported in that ticket.

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryArithmeticOperatorNode.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryArithmeticOperatorNode.java
@@ -363,7 +363,7 @@ public final class BinaryArithmeticOperatorNode extends BinaryOperatorNode
     /** @see ValueNode#evaluateConstantExpressions */
     @Override
     ValueNode evaluateConstantExpressions() throws StandardException {
-        if (isConstantExpression()) {
+        if (getLeftOperand() instanceof ConstantNode && getRightOperand() instanceof ConstantNode) {
             ConstantNode lhs = (ConstantNode) getLeftOperand();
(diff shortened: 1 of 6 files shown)


cloudspliceci:
TEST SUCCEEDED +1

arnaud-splice:

dgomezferro (approved):
Nice and clean!

cloudspliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/BinaryArithmeticOperatorNode.java
```diff
@@ -363,4 +359,56 @@ public double getBaseOperationCost() throws StandardException {
...
+                        operator
+                );
+                NumberDataValue result;
+                NumberDataValue leftValue = (NumberDataValue) lhs.getValue();
+                NumberDataValue rightValue = (NumberDataValue) rhs.getValue();
+                NumberDataValue receiver = leftIsReceiver() ? leftValue : rightValue;
+                switch (operator) {
+                    case TypeCompiler.TIMES_OP:
+                        result = receiver.times(leftValue, rightValue, null);
+                        break;
```
ascend1, line 389: A general question is what's the behavior if left + right overflows. If in runtime there is a SQL warning, we wouldn't be able to set the warning at this step.

But I guess that's something we can discuss later.

ascend1 (approved):

PR #5614, opened 2021-06-25
DB-12280 Stop key for Index Prefix Iteration should not override stopSearchOperator

msirek:
## Short Description
Avoid incorrect results on queries flagged as IndexPrefixIteratorMode in the EXPLAIN with predicates of the form _2nd_index_column_ < _constant_.

## Long Description
Scans.attachScanKeys was updated by [SPLICE-2399](https://github.com/splicemachine/spliceengine/pull/5013) to combine a list of key values on the first index column (firstIndexColumnKeys) with key values on the remaining index columns from predicates.  However this path always passed the stopSearchOperator parameter of buildStartAndStopKeys as hard-coded GT, which causes the stop key to be appended with an extra byte of value 1, causing the stop key to be inclusive, even if we are using a less than operator in the predicate.  The range was made wider in an attempt to never exclude necessary values with the assumption that the scan qualifier could filter out any unqualified values.  This works fine on control, but on spark, since the IndexPrefixIteratorOperation and associated TableScanOperation have the same result set number, spark SerDe ends up sharing the same DerbyScanInformation between the two operations (see SparkLeanOperationContext.readExternal):

> op=(Op)ah.getOperationsMap().get(in.readInt());

... and we pick up the qualifier of the IndexPrefixIteratorOperation, which has been set to null, since it is not desirable for the initial scanner to have a predicate (it may read to the end of the table in control mode before finding the first row!).  

The solution is:
1.  Use the passed-in startSearchOperator and stopSearchOperator parameters for building the scan keys instead of hard-coded arguments.
2. Build the IndexPrefixIteratorOperation with the same qualifier as the main table scan so the TableScanPredicateFunction of the TableScanOperation can see it and use it.  However, for performing the IndexPrefixIteratorOperation itself, temporarily set the qualifier to null when building that operation's scan to avoid applying an undesirable qualifier.

## How to test
create table t1 (a int, b timestamp, primary key(a,b));
insert into t1 values (1,timestamp('2021-06-24 18:00:00'));
insert into t1 values (2,timestamp('2021-06-24 17:00:00'));
/- the following should return 1 row:
select distinct(b) from t1 where b >= timestamp('2021-06-24 17:00:00') and b < timestamp('2021-06-24 18:00:00');
B
/-
2021-06-24 17:00:00.0

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/SpliceGenericResultSetFactory.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/SpliceGenericResultSetFactory.java
@@ -649,7 +649,7 @@ public class SpliceGenericResultSetFactory implements ResultSetFactory {
                     stopSearchOperator,
                     sameStartStopPosition,
                     rowIdKey,
-                    null, // qualifiersField
+                    qualifiersField,
                     tableName,
                     userSuppliedOptimizerOverrides,
                     indexName,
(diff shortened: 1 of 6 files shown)


msirek:

cloudspliceci:
TEST SUCCEEDED +1

dgomezferro (approved):

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5615, opened 2021-06-25
DB-12280 Stop key for Index Prefix Iteration should not override stopSearchOperator (3.1)

msirek:


carolp-503 (approved):
Approved for 3.1

PR #5620, opened 2021-06-25
DB-12210 NSDSv2 Continuous Topic Deletion

jpanko1:
## Short Description
NSDS now deletes used topics during processing instead of at shutdown.

## Long Description
NSDS v2 (eNSDS) uses a new Kafka topic for each transaction, for both transactions to and from the DB.  After the call to the DB returns (which should mean the transaction is completed), NSDS puts the used topic name on a queue to be deleted.  A separate thread now checks the deletion queue and sends the topic names to Kafka to delete the topics.  The thread sleeps for a minute between checks of the deletion queue.

The code was previously queueing the used topic names after each transaction and sending them for deletion when NSDS shut down.  For a long running process like streaming, a large number of topics accumulate and fill disk and memory in Kafka.  Now NSDS performs continual cleanup.

## How to test
Use NSDS v2 (eNSDS) normally.  It's set to do topic deletion by default.  Watch the Kafka controller.log for messages containing the phrase `Deletion of topic`.

### DIFF
--- a/splice_spark2/src/main/scala/com/splicemachine/nsds/kafka/KafkaTopics.scala
+++ b/splice_spark2/src/main/scala/com/splicemachine/nsds/kafka/KafkaTopics.scala
@@ -43,9 +43,7 @@ class KafkaTopics(
   private def deleteTopics(): Unit = {
     val toDelete = collection.mutable.Set.empty[String]
     unneeded.drainTo(toDelete.asJava)
-    println(s"Deleting topics $toDelete")  // todo remove println
-    admin.deleteTopics(toDelete, 60*1000)  // todo will admin.deleteTopics accept an empty toDelete when there are no topics?
-    println(s"Delete done")  // todo remove println
+    admin.deleteTopics(toDelete, 60*1000)
   }
   
   if(continuousCleanup) {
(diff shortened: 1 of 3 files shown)


jpanko1:

cloudspliceci:
TEST SUCCEEDED +1

arnaud-splice (approved):
+1 once the spotbugs error is addressed

splice_spark2/src/main/scala/com/splicemachine/nsds/kafka/KafkaTopics.scala
```diff
@@ -17,16 +17,47 @@ package com.splicemachine.nsds.kafka
...
+    unneeded.drainTo(toDelete.asJava)
+    admin.deleteTopics(toDelete, 60*1000)
+  }
   
+  if(continuousCleanup) {
+    new Thread {
+      override def run {
+        while (processing.get) {
+          deleteTopics
+          Thread.sleep(60*1000)
```
ipraznik-splice, line 54: to use a schedular instead of sleep here is more thread efficient and CPU friendly, but as there is only one thread, should be ok

ipraznik-splice (approved):

jpanko1:

cloudspliceci:
TEST SUCCEEDED +1

jyuanca (approved):

carolp-503 (approved):
+1

PR #5630, opened 2021-06-29
DB-12283 Use UUIDs as session IDs instead of lcc.getInstanceNumber()

ascend1:
## Short Description
This change adds a new session ID field in `GenericLanguageConnectionContext` which stores a UUID prefixed with a machine ID together as a unique identifier across region servers.

## Long Description
`lcc.getInstanceNumber()` returns a monotonically increasing number and it's currently used as session IDs in some scenarios. However, this is wrong because this number is local to a region server. Across region servers, different LCCs could have the same instance number.

This means the current implementation of local temporary table name mangling is also incorrect. When two connections have the same instance number, one could see the local temporary table created in the other session, which breaks the definition of local temporary table.

This change adds a new session ID field in `GenericLanguageConnectionContext` which stores a UUID prefixed with a machine ID together as a unique identifier across region servers.

Another option is to use a combination of host name and instance number, which is also unique across region servers. However, we don't have a limit on host name length at the moment. In `syscs_utils.syscs_get_session_info()`, we reserve 120 characters for host name column. Thus, using host name plus instance number could potentially lead to a very long table name and that might be a problem at some point.

## How to test
All existing ITs in `TempTableIT`, `CreateTableIT`, `SpliceAdmin_GetActiveSessionsIT`, and `SpliceAdminIT` should still pass as they always do.


ascend1:

jyuanca (approved):

ascend1:

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

cloudspliceci:
TEST SUCCEEDED +1

dgomezferro (changes requested):
+1 to the change, but I think we should log the session Id when logging statement information (GenericLanguageConnectionContext.logStartExecuting() ... ) We are logging the instanceNumber as the `SessionID`, so there's a bit of a conflict there naming wise.

ipraznik-splice (approved):

ascend1:

hatyo (changes requested):
IMHO this PR is not fixing the issue completely, there is still, albeit tiny, chance of a collision since region servers generate UUIDs in isolation from each other.

I think this should be addressed, one easy approach would be to use ZK for getting a globally-unique number. This should guarantee uniqueness of session IDs across multiple region servers.

cloudspliceci:
TEST SUCCEEDED +1

ascend1:

ascend1:
> +1 to the change, but I think we should log the session Id when logging statement information (GenericLanguageConnectionContext.logStartExecuting() ... ) We are logging the instanceNumber as the `SessionID`, so there's a bit of a conflict there naming wise.

That's a good point. This is addressed in commit https://github.com/splicemachine/spliceengine/pull/5630/commits/b20824b6b2cf0c0a929c06cb3f7179c26d7fff20.

ascend1:
> IMHO this PR is not fixing the issue completely, there is still, albeit tiny, chance of a collision since region servers generate UUIDs in isolation from each other.
> 
> I think this should be addressed, one easy approach would be to use ZK for getting a globally-unique number. This should guarantee uniqueness of session IDs across multiple region servers.

The session ID is improved in commit https://github.com/splicemachine/spliceengine/pull/5630/commits/6bd63b438ca6afd9be8f3a244f8c1b192fe5c5fb.

The new session ID is prefixed with a machine ID, which is a globally unique number retrieved from ZK. We don't have to implement anything in ZK because `ZkUtils` provides this functionality already.

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3317/

ascend1:

cloudspliceci:
TEST SUCCEEDED +1

hatyo (approved):

dgomezferro (approved):

carolp-503 (approved):
+1

PR #5639, opened 2021-06-30
DB-10594 Add a system procedure to clean up obsolete temp tables

ascend1:
## Short Description
This change adds a new administrative system procedure which cleans up obsolete temporary tables.

## Long Description
Currently, local temporary tables are not dropped properly when region servers crash because the happy path in disconnecting a session is not executed.

Once the cluster is up again, there is no way to remove the obsolete local temporary tables because they were created by previous sessions, thus not visible to new sessions. Over time, there could be more and more obsolete temporary tables in system, wasting space and causing troubles in fixing inconsistency issues when they occur.

The new system procedure `SYSCS_DROP_OBSOLETE_TEMPORARY_TABLES(schemaName)` cleans up all obsolete temporary tables in schema `schemaName`. It works in the following steps:
1. get a copy of table descriptors in the give schema
2. get a list of active session IDs
3. for each local temporary table that is not created by any of the active sessions, drop it cascade

The workflow above handles all of the following cases:
a. A new session appears after step 1 and create a local temporary table. This is fine because the new temp table is not in the list of step 1. thus not processed.
b. A new session appears after step 2 and create a local temporary table. This is also fine because it's completely unknown to the procedure, thus not processed.
c. An existing session is closed after step 1 and its local temporary tables are dropped. This is fine because dropping a non-existing table is a no-op.
d. An existing session is closed after step 2 and its local temporary tables are dropped. This is fine because the table is already dropped. It's OK that we skip it.

However, if we reverse step 1 and 2, case a) would be wrong.

## How to test
This is currently a question. Since making an obsolete local temporary table requires killing the cluster, an automated test is not possible in current platform it.

Manual testing:
- create a local temporary table `T`
- create an index `T_IDX` on `T`
- verify that the index exists by looking up `sys.sysconglomerates`
- kill the cluster
- restart without build, i.e., `-b`
- verify that the index still exists
- call the new system procedure
- the index `T_IDX` should be gone because `T` is dropped cascade

### DIFF
--- a/splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
+++ b/splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
@@ -31,12 +31,14 @@ import com.splicemachine.db.iapi.services.monitor.Monitor;
 import com.splicemachine.db.iapi.services.property.PropertyUtil;
 import com.splicemachine.db.iapi.sql.Activation;
 import com.splicemachine.db.iapi.sql.ResultColumnDescriptor;
+import com.splicemachine.db.iapi.sql.StatementType;
 import com.splicemachine.db.iapi.sql.conn.Authorizer;
 import com.splicemachine.db.iapi.sql.conn.ConnectionUtil;
 import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
 import com.splicemachine.db.iapi.sql.depend.DependencyManager;
 import com.splicemachine.db.iapi.sql.dictionary.TableDescriptor;
 import com.splicemachine.db.iapi.sql.dictionary.*;
+import com.splicemachine.db.iapi.sql.execute.ConstantAction;
 import com.splicemachine.db.iapi.sql.execute.ExecPreparedStatement;
 import com.splicemachine.db.iapi.sql.execute.ExecRow;
 import com.splicemachine.db.iapi.store.access.TransactionController;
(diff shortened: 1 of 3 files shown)


splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
```diff
@@ -1718,6 +1724,72 @@ public static void SYSCS_GET_ACTIVE_SESSIONS(ResultSet[] resultSet) throws SQLEx
...
+                            if (!sessionIDSet.contains(creatingSessionID)) {
+                                dropTempTable(lcc, td);
+                                res.newRow();
+                                tblNameCol.set(tableName);
+                            }
+                        } catch (StandardException se) {
+                            if (se.getSQLState().equals(SQLState.LANG_INVALID_INTERNAL_TEMP_TABLE_NAME)) {
+                                // for temporary tables created using instance numbers
+                                dropTempTable(lcc, td);
+                                res.newRow();
```
jyuanca, line 1767: why execute same logic in try and catch block?

martinrupp (approved):
+1 given comment by @jyuanca is addressed

db-engine/src/main/java/com/splicemachine/db/iapi/sql/conn/LanguageConnectionContext.java
```diff
@@ -282,6 +282,15 @@ void removeActivation(Activation a)
...
     boolean isVisibleToCurrentSession(TableDescriptor td) throws StandardException;
 
+    /**
+     * Return the creating session ID of a local temporary table
+     *
+     * @param td TableDescriptor of a local temporary table
+     * @return The creating session ID of this local temporary table
+     * @throws StandardException
+     */
+    java.util.UUID getLocalTempTableSessionID(TableDescriptor td) throws StandardException;
```
dgomezferro, line 292: I don't think this method really belongs here, but I don't have a better suggestion.

dgomezferro (approved):
Change looks good, I don't like the placement of `getLocalTempTableSessionID` but I don't have a better suggestion right now.

ipraznik-splice (changes requested):
please answer Jun's question

splice_machine/src/main/java/com/splicemachine/derby/procedures/SpliceAdmin.java
ascend1: Good catch, logic is revised in the latest commit into a `finally` block.

ipraznik-splice (approved):

carolp-503 (approved):
+1

ipraznik-splice:

PR #5646, opened 2021-07-02
DB-12224 Permutate system index columns to prioritize schemaId column 

arnaud-splice:
## Short description

Replace `objectName, schemaId` index by `schemaId,objectName` index for SYSTABLES, SYSTRIGGERS, SYSSTATEMENTS, SYSFILES, SYSCONSTRAINTS

## Long description

SYSTABLES, SYSTRIGGERS, SYSSTATEMENTS, SYSFILES and SYSCONSTRAINTS all had an index defined on objectName, schemaId. 
It makes little sense to lookup an object base on its name only, without checking for the schema it's in. It is however useful to query all objects which are located in a schema.
This PR recreates those indexes with schemaId as their first column.

## How to test
`show tables in schema` should still return a correct result, and should return that result in a faster way for very large databases.
I tested upgrade from 3.2.0.1982 

## All DB-12224 PRs

[spliceengine master](https://github.com/splicemachine/spliceengine/pull/5646)
[spliceengine-ee master](https://github.com/splicemachine/spliceengine-ee/pull/653)
[spliceengine branch-3.1](https://github.com/splicemachine/spliceengine/pull/5647)
[spliceengine-ee branch-3.1](https://github.com/splicemachine/spliceengine-ee/pull/654)


arnaud-splice:

cloudspliceci:
TEST SUCCEEDED +1

hatyo:
Why not adding a new index as @dgomezferro suggested in the original ticket? I am guessing the older index with `objectName` as the first column could be useful for other queries, right?

arnaud-splice:
> Why not adding a new index as @dgomezferro suggested in the original ticket? I am guessing the older index with `objectName` as the first column could be useful for other queries, right?
I don't think the old index makes any sense. Either you query an object by its UUID, or you query it by its name AND the schema it is in. Querying an object by name only is ambiguous and would return multiple objects with the same name, that could live in various schemas of various databases. 
The comment from @dgomezferro was based on a previous PR, related to SYSCONGLOMERATES which did not contain any index on conglomerate number, so we had to create a new one.

 
> nitpick: I would change "permute" word in the PR, to something like: change order of columns to make it clearer, permute makes it sound like we have now n! variations of an index with n columns.
What you're describing here would be "generate all permutations" though. Here, I operated one permutation on an index that contains two columns.

hatyo:
> Querying an object by name only is ambiguous and would return multiple objects with the same name, that could live in various schemas of various databases.

Which is useful in certain contexts such as UI tools allowing you to e.g. search for all tables whose name starts with "Emp*" across all schemas, but I agree that having the a second index column of the schema name seems pointless.

> What you're describing here would be "generate all permutations" though. Here, I operated one permutation on an index that contains two columns.

That's why it is a nitpick, it is up to you.

arnaud-splice:

arnaud-splice:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3310/

cloudspliceci:
TEST SUCCEEDED +1

martinrupp (approved):

dgomezferro (approved):

ascend1 (approved):
Several index fields in those `*RowFactory` classes are now public instead of protected. Is this only because of the upgrade script? I wonder how we can maintain access control in upgrading. This is kind of one-time requirement but permanent change to the code.

carolp-503 (approved):
+1

arnaud-splice:
> Several index fields in those `*RowFactory` classes are now public instead of protected. Is this only because of the upgrade script? I wonder how we can maintain access control in upgrading. This is kind of one-time requirement but permanent change to the code.

I think a lot more of those should be public instead of protected, because we also want to use them in datadictionaryImpl. There are a couple places where we use the number directly instead of the constant that's defined in *RowFactory classes.

PR #5652, opened 2021-07-06
DB-12304 Fix call sites of getKeyColumnPosition

ascend1:
## Short Description
This PR fixes a regression after DB-9898.

3.1 PR: https://github.com/splicemachine/spliceengine/pull/5653.

## Long Description
After DB-9898, `getKeyColumnPosition()` expects a column storage position instead of a column (logical) position. Some call sites have been adapted, but some are left untouched.

This change adapts the remaining call sites and add an IT.

System views and stored procedures should always be refreshed in an upgrade.

## How to test
For all test cases, the common setup is to have a column having different position and storage position:
```
create table t (c int);
alter table t add column d int;
alter table t drop column d;
alter table t add column d int;  // add again, now d has pos = 2 and storage pos = 3
alter table t add column e int;  // pos = 3, storage pos = 4
```
Then the test cases:
```
// add a check constraint (binary op):
alter table t add check (d >= 0);

// add a pk constraint and show it in system views
alter table t add constraint tpk primary key(d);
select * from sysibm.systables where name = 'T';
select * from sysibm.syscolumns where name = 'D';
select * from sysibm.syskeycoluse where constname = 'TPK';

// create an index and show it in system view
create index tidx on t(e);
select * from syscat.indexcoluse where indname='TIDX';
```


ascend1:

arnaud-splice (approved):

martinrupp (approved):

dgomezferro (approved):

jyuanca (approved):

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5655, opened 2021-07-08
DB-12324 NSDS v2 Changes for Streaming Resampling

jpanko1:
## Short Description
Some changes made during work on streaming resampling.

## Long Description
1. Continuous topic cleanup default value set to false to reduce traffic on Kafka.
2. Support for upsert added in the setTable method so that streaming code can do upsert.
3. Some logging updates.

## How to test
1. Run eNSDS and note that topics are not automatically cleaned up in Kafka when transactions are completed.
2. Run streaming in upsert mode and note upsert behavior with the incoming records.
3. Note sendData log as an info instead of debug.  Note insert topicCount logs are gone.

### DIFF
--- a/splice_spark2/src/main/scala/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
+++ b/splice_spark2/src/main/scala/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
@@ -590,13 +590,13 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
       val sqlText = insertSql(topicInfo)
       debug(s"SMC.inss sql $sqlText")
       
-      trace( s"SMC.inss topicCount preex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
+      //trace( s"SMC.inss topicCount preex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
 
       debug("SMC.inss executeUpdate")
       executeUpdate(sqlText)
       debug("SMC.inss done")
 
-      debug( s"SMC.inss topicCount postex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
+      //debug( s"SMC.inss topicCount postex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
     } catch {
       case e: java.sql.SQLNonTransientConnectionException => 
         if( retries < 2 ) {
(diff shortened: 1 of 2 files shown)


jpanko1:

cloudspliceci:
TEST SUCCEEDED +1

splice_spark2/src/main/scala/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
```diff
@@ -589,13 +590,13 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
       val sqlText = insertSql(topicInfo)
       debug(s"SMC.inss sql $sqlText")
       
-      trace( s"SMC.inss topicCount preex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
+      //trace( s"SMC.inss topicCount preex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
```
jyuanca, line 594: Do you want to remove commented out logging?

splice_spark2/src/main/scala/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
```diff
@@ -589,13 +590,13 @@ class SplicemachineContext(options: Map[String, String]) extends Serializable {
...
 
       debug("SMC.inss executeUpdate")
       executeUpdate(sqlText)
       debug("SMC.inss done")
 
-      debug( s"SMC.inss topicCount postex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
+      //debug( s"SMC.inss topicCount postex ${KafkaUtils.messageCount(kafkaServers, topicName)}")
```
jyuanca, line 599: ditto

jyuanca (approved):

splice_spark2/src/main/scala/com/splicemachine/spark2/splicemachine/SplicemachineContext.scala
jpanko1: These logs were useful during some of the performance test runs.  It may be best to eventually add a "diagnostic" flag which would activate these logs.  I currently don't want them to call KafkaUtils.messageCount() unnecessarily.

arnaud-splice (approved):

carolp-503 (approved):
+1

PR #5656, opened 2021-07-08
DB-12324 NSDS v2 Changes for Streaming Resampling (3.1)

jpanko1:
## Short Description
Some changes made during work on streaming resampling.

## Long Description
1. Continuous topic cleanup default value set to false to reduce traffic on Kafka.
2. Support for upsert added in the setTable method so that streaming code can do upsert.
3. Some logging updates.

## How to test
1. Run eNSDS and note that topics are not automatically cleaned up in Kafka when transactions are completed.
2. Run streaming in upsert mode and note upsert behavior with the incoming records.
3. Note sendData log as an info instead of debug.  Note insert topicCount logs are gone.


jpanko1:

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
Approved for 3.1

PR #5657, opened 2021-07-08
DB-12158 Automatic view refreshing for alter table add/drop column

ascend1:
## Short Description
This change introduces a new automatic view refreshing mode for alter table add/drop column.

3.1 PR: https://github.com/splicemachine/spliceengine/pull/5660.
3.0 PR: https://github.com/splicemachine/spliceengine/pull/5662.

## Long Description

This change introduces a new automatic view refreshing mode for alter table add/drop column.

When this mode is turned on, an `ALTER TABLE T ADD/DROP COLUMN ...` statement automatically refreshes all views defined both directly and indirectly on `T` whose top-level select list is a `SELECT *`. Other views are not refreshed.

This mode is off by default. To turn it on, run the following:
```
call syscs_util.syscs_set_global_database_property('splice.execution.alterTable.autoViewRefreshing', true)
```

To turn it off, run the following:
```
call syscs_util.syscs_set_global_database_property('splice.execution.alterTable.autoViewRefreshing', null)
```

### Detailed behavior description

- A view `V` defined using a top-level explicit select column list, i.e., `select a, b, ...` on top of `T`
	- this kind of views are not refreshed no matter view refreshing mode is on or off, even if it has subqueries that has `select *` (see explanation 1)
	- add column: the new column is not shown in `select * from v`
	- drop a column that is added after the `V` is defined: no effect
	- drop a column that exists when `V` is defined: throw (previous behavior: assertion failure if column number mismatch or wrong column names if column numbers match but it's a different set of columns)
- A view `V` defined using a top-level `select *` on top of `T`
	- when view refreshing mode is off, it has the same behavior as the explicit select column case
	- when view refreshing mode is on
		- add column: new column is shown in `select * from v`
		- drop column: dropped column is removed from `select * from v`
- A view `VV` defined using a top-level `select *` on another view `V` on top of `T`
	- when auto view refreshing mode is on, `VV` gets refreshed when a column is added to or dropped from `T`
	- if `V` is not a `select *` view, refreshing `VV` has no effect

### Explanation 1

We only need to refresh views defined with top-level "select *". View definitions could be arbitrarily complex, having "select *" in a subquery. However, as long as the top-level select items are explicitly fixed (i.e., "select a, b, ..."), newly added column will not be selected even if we refresh the view. Also, if one of the explicitly selected column is dropped, the view definition becomes invalid even if we refresh the view. As a result, in both cases, there is no need to refresh the view if its top-level select columns are explicit.

## How to test
It would be too long to copy all test cases here. For references, please check the following ITs in `DropAddColumnIT`:
- testAutomaticViewRefreshingOn_SelectStarView
- testAutomaticViewRefreshingOn_SecondLevelSelectStarView
- testAutomaticViewRefreshingOn_SelectExplicitColumnsView
- testAutomaticViewRefreshingOff

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromSubquery.java
+++ b/db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromSubquery.java
@@ -33,6 +33,7 @@ package com.splicemachine.db.impl.sql.compile;
 
 
 import com.splicemachine.db.iapi.error.StandardException;
+import com.splicemachine.db.iapi.reference.SQLState;
 import com.splicemachine.db.iapi.services.context.ContextManager;
 import com.splicemachine.db.iapi.services.sanity.SanityManager;
 import com.splicemachine.db.iapi.sql.compile.C_NodeTypes;
@@ -44,6 +45,7 @@ import com.splicemachine.db.iapi.sql.dictionary.SchemaDescriptor;
 import com.splicemachine.db.iapi.util.JBitSet;
 import com.splicemachine.db.impl.ast.CollectingVisitor;
 import com.splicemachine.db.impl.ast.ColumnCollectingVisitor;
+import com.splicemachine.db.impl.sql.misc.SQLCommentStripper;
 import splice.com.google.common.base.Predicates;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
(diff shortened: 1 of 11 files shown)


ascend1:

cloudspliceci:
TEST SUCCEEDED +1

hatyo (approved):
Great work.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/ModifyColumnConstantOperation.java
```diff
@@ -38,12 +40,14 @@
...
-import com.splicemachine.db.impl.sql.compile.ColumnDefinitionNode;
-import com.splicemachine.db.impl.sql.compile.StatementNode;
+import com.splicemachine.db.impl.sql.GenericStatement;
+import com.splicemachine.db.impl.sql.GenericStorablePreparedStatement;
+import com.splicemachine.db.impl.sql.compile.*;
 import com.splicemachine.db.impl.sql.execute.ColumnInfo;
 import com.splicemachine.pipeline.ErrorState;
 import org.apache.log4j.Logger;
 
+import javax.swing.text.View;
```
dgomezferro, line 50: Remove

dgomezferro (changes requested):
Minor issue, wrong import. Looks good otherwise

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/FromSubquery.java
```diff
@@ -310,14 +312,26 @@ public void bindExpressions(FromList fromListParam)
...
+                }
+            } else if (resultColumns.size() > subqueryRCL.size()) {
+                // columns dropped from underlying table
+                throw StandardException.newException(SQLState.LANG_DERIVED_COLUMN_LIST_MISMATCH, correlationName);
+            } else {
+                // number of columns is OK, check column names
+                for (int i = 0; i < resultColumns.size(); ++i) {
+                    ResultColumn thisRC = resultColumns.elementAt(i);
+                    ResultColumn subqRC = subqueryRCL.elementAt(i);
+                    if (thisRC.getName() != null && !thisRC.getName().equals(subqRC.getName())) {
```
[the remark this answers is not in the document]
ascend1: Having `thisRC.getName() != null` is mostly defensive programming since we call `.equals()` on it. Usually for an expression like `col1 / 3.14` we should generate an internal name during planning, something like `SQLCol1`. On the other side, passing `null` into `String.equals()` should be fine.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/actions/ModifyColumnConstantOperation.java
ascend1: Done.

ascend1:

dgomezferro (approved):

cloudspliceci:
TEST SUCCEEDED +1

carolp-503 (approved):
+1

PR #5658, opened 2021-07-08
DB-12323 Fix upgrade regression introduced by DB-9898 

arnaud-splice:
## Short description

Fix upgrade regression introduced by DB-9898 causing upgrades from 2021 to 2022 to fail.

## Long description

DB-9898 introduced a new field in `IndexDescriptorImpl`: `baseStorageColumnPositions`. From 2022 onward,  IndexDescriptorImpl contains both `baseStorageColumnPositions` and `baseColumnPosition`.
The protobuf message was also extended: on top of the already present `repeat baseColumnPositions`, DB-9898 introduced `repeat baseColumnLogicalPositions`.
This is already somewhat confusing and the mapping between `IndexDescriptorImpl` and the protobuf message should be as follow:
| Protobuf message          | IndexDescriptorImpl       |
|---------------------------|---------------------------|
| baseLogicalColumnPosition | baseColumnPosition        |
| baseColumnPosition        | baseStorageColumnPosition |

However, before 2022, the mapping was:
| Protobuf message   | IndexDescriptorImpl |
|--------------------|---------------------|
| baseColumnPosition | baseColumnPosition  |

This PR introduces an upgrade script to: 
1. Redirect protobuf's `baseColumnPosition` to the the logical column positions
2. Populate `baseStorageColumnPosition` properly.

Additionally, DB-9898 added `keyFormatIds` to both `IndexConglomerate` and `HBaseConglomerate`. Before 2022, we used to generate `keyFormatIds` during a scan operation, but we now take it directly from the conglomerate. However, conglomerates that were created and serialized before 2022 will not contain this field, and this led to `ArrayIndexOutOfBoundErrors` and `NullPointerExceptions`. The upgrade script now properly populates that field for every single conglomerate in the system (tables and indices). 

Until the upgrade script is run, indices with columns that differ in `baseColumnPositions` and `baseStorageColumnPositions` can behave incoherently. This however should not happen because: 
1. User tables will not be accessed until all upgrade scripts have run.
2. System tables never drop columns, so `baseColumnPositions == baseStorageColumnPositions`

This PR does **not** fix indexes that were already inconsistent before.

For example, the following example led to inconsistency on 2021:
```
create table foo (a int, b int, c int);
alter table foo drop column c;
alter table foo add column d int;
create index fooi on foo(a, d);
insert into foo values (1,2,3);
select a,d from foo;        -- We use an index scan
A          |D
-----------------------
1          |NULL              <-- Notice that NULL should be 3 instead
```
The only way to fix that index is to upgrade to 2023, drop and recreate the index.

## Tests

The upgrade was tested for the following versions:
`3.2.0.1995` -> `DB-12323`
`3.2.0.2021` -> `DB-12323`
`3.1.0.1995` -> `DB-12323-3.1`
`3.1.1.2021` -> `DB-12323-3.1` 

## All DB-12323 PRs

[spliceengine master](https://github.com/splicemachine/spliceengine/pull/5658)
[spliceengine branch-3.1](https://github.com/splicemachine/spliceengine/pull/5659)

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
@@ -72,6 +72,12 @@ public interface IndexDescriptor
     int[] baseColumnPositions();
 
     int[] baseColumnStoragePositions();
+
+    /**
+     * @return true if this index descriptor is invalid as a result of an upgrade from a pre 2023 version
+     */
+    boolean isInvalidIndexDescriptor();
+
     /**
      * Returns the postion of a column.
      * <p>
@@ -131,8 +137,15 @@ public interface IndexDescriptor
      * set the baseColumnPositions field of the index descriptor.  This
      * is for updating the field in operations such as "alter table drop
      * column" where baseColumnPositions is changed.
+     * @param baseColumnStoragePositions
      */
-    void     setBaseColumnPositions(int[] baseColumnPositions);
+    void setBaseColumnStoragePositions(int[] baseColumnStoragePositions);
+
+    void setBaseColumnPositions(int[] baseColumnsPositions);
+
+    int[] getBaseColumnStoragePositions();
+
+    int[] getBaseColumnPositions();
 
     /**
      * set the isAscending field of the index descriptor.  This
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
@@ -74,9 +74,10 @@ public interface IndexDescriptor
     int[] baseColumnStoragePositions();
 
     /**
-     * @return true if this index descriptor is invalid as a result of an upgrade from a pre 2023 version
+     * Returns true if this index descriptor is invalid as a result of an upgrade from a pre 2023 version
+     * See DB-12323 for more information
      */
-    boolean isInvalidIndexDescriptor();
+    boolean isInvalidIndexDescriptorAfter2022Upgrade();
 
     /**
      * Returns the postion of a column.
@@ -137,7 +138,6 @@ public interface IndexDescriptor
      * set the baseColumnPositions field of the index descriptor.  This
      * is for updating the field in operations such as "alter table drop
      * column" where baseColumnPositions is changed.
-     * @param baseColumnStoragePositions
      */
     void setBaseColumnStoragePositions(int[] baseColumnStoragePositions);
 
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
@@ -274,13 +274,11 @@ public class IndexDescriptorImpl implements IndexDescriptor, Formatable {
     /** @see IndexDescriptor#baseColumnPositions */
     public int[] baseColumnPositions()
     {
-        if (isInvalidIndexDescriptor()) {
-            throw new UnsupportedOperationException();
-        }
+        assert !isInvalidIndexDescriptorAfter2022Upgrade();
         return baseColumnPositions;
     }
 
-    public boolean isInvalidIndexDescriptor() {
+    public boolean isInvalidIndexDescriptorAfter2022Upgrade() {
         return baseColumnPositions.length != baseColumnStoragePositions.length;
     }
 
--- a/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
+++ b/db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
@@ -42,16 +42,13 @@ import com.splicemachine.db.iapi.sql.conn.LanguageConnectionContext;
 import com.splicemachine.db.iapi.types.DataTypeDescriptor;
 import com.splicemachine.db.iapi.types.ProtobufUtils;
 import com.splicemachine.db.iapi.util.ByteArray;
-import com.splicemachine.db.impl.ast.CollectingVisitor;
 import com.splicemachine.db.impl.sql.CatalogMessage;
 import com.splicemachine.db.impl.sql.compile.*;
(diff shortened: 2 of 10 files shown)


arnaud-splice:

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -276,9 +274,16 @@ public boolean isUnique()
     /** @see IndexDescriptor#baseColumnPositions */
     public int[] baseColumnPositions()
     {
+        if (isInvalidIndexDescriptor()) {
+            throw new UnsupportedOperationException();
```
hatyo, line 278: I would assert here since this should never happen, if you want to keep an exception, then I suggest throwing a more meaningful exception + a message.

db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
```diff
@@ -131,8 +137,15 @@
      * set the baseColumnPositions field of the index descriptor.  This
      * is for updating the field in operations such as "alter table drop
      * column" where baseColumnPositions is changed.
+     * @param baseColumnStoragePositions
```
hatyo, line 140: either remove or expand.

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeFixIndexDescriptors.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    private void fixIndexDescriptorColumns(TableDescriptor td, SchemaDescriptor sd, ConglomerateDescriptor cd) throws StandardException {
+        IndexDescriptor id = cd.getIndexDescriptor();
+        if (id.isInvalidIndexDescriptor()) {
+            sdd.dropConglomerateDescriptor(cd, tc);
+            /* At this point, baseStorageColumnDescription mistakenly contains the LOGICAL column ids, and
+             * baseColumnDescription contains nothing.
+             * We want to recreate the index descriptor and baseStorageColumnDescription should contain the STORAGE
```
hatyo, line 65: > baseStorageColumnDescription
typo

hatyo (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeFixIndexDescriptors.java
```diff
@@ -0,0 +1,85 @@
...
+                    fixIndexDescriptorColumns(td, sd, cd);
+                }
+                addKeyFormatIdsToConglomerate(cd);
+            }
+        }
+    }
+
+    private void fixIndexDescriptorColumns(TableDescriptor td, SchemaDescriptor sd, ConglomerateDescriptor cd) throws StandardException {
+        IndexDescriptor id = cd.getIndexDescriptor();
+        if (id.isInvalidIndexDescriptor()) {
```
ascend1, line 61: According to my previous index on expressions experience, `id` can be `null` for some conglomerate descriptors. Also, even `id` is not null, its internal `id` might be null. Maybe we should guard that?

Here is an example code snippet that testing something similar in `PredicateList.java`:
```
IndexRowGenerator irg = cd == null ? null : cd.getIndexDescriptor();
if (irg != null && irg.getIndexDescriptor() == null) {
    irg = null;
}
```

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
```diff
@@ -276,9 +274,16 @@ public boolean isUnique()
...
+        }
         return baseColumnPositions;
     }
 
+    public boolean isInvalidIndexDescriptor() {
+        return baseColumnPositions.length != baseColumnStoragePositions.length;
```
ascend1, line 284: Is this sufficient? Could they have the same length but position-wise not correct? Like column A has position 1 and storage position 2, and `baseColumnPositions[0] == 1` but `baseColumnStoragePositions[0] == 3`?

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
arnaud-splice: Done

db-engine/src/main/java/com/splicemachine/db/catalog/IndexDescriptor.java
arnaud-splice: Done

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeFixIndexDescriptors.java
arnaud-splice: Done

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/catalog/upgrade/UpgradeFixIndexDescriptors.java
arnaud-splice: I think IndexDescriptor cannot be null here, because we check for `cd.isIndex()` before.

db-engine/src/main/java/com/splicemachine/db/catalog/types/IndexDescriptorImpl.java
arnaud-splice: This is not sufficient to guarantee the validity of the index descriptor. But it's sufficient to catch upgrade issues from pre 2022 to post 2022. I will rename the function to reflect that.

ascend1 (approved):

carolp-503 (approved):
+1

ipraznik-splice:

cloudspliceci:
TEST SUCCEEDED +1

cloudspliceci:
TEST SUCCEEDED +1

PR #5661, opened 2021-07-08
DB-11826 DB-12226 Fix time-travel returning incorrect results due to using temporal-inconsistent index (3.1)

hatyo:
## Short Description
In this PR, we target a subtle inconsistency caused by time-travel allowing the optimizer to choose a temporal-inconsistent index, causing wrong results to be returned.

## Long Description

With time-travel, the user view the historical data of a table within a min. retention period. It could happen that the optimizer prefers an AccessPath involving an index conglomerate that is created _after_ the past transaction ID which is dangerous, because the base table could have changed between the past transaction ID and the index creation time (e.g. some rows are deleted before the index was created), if the optimizer chooses that index conglomerate we will get empty results because the index does not have any data at the past transaction ID since it was created after it.

To fix this issue we make sure, during optimization and when deciding whether an index access path is eligible or not, that the index's creation time precedes the past transaction ID, and only accept it in that case.

To achieve this, I implemented the following changes in this PR:

1. Add API extensions to `DataDictionary` that allows, for a given conglomerate, to return its creation transaction ID from the conglomerate's metadata.
2. Move `AS OF` expression evaluation from execution time to binding phase, since it is crucial to know upfront the past transaction ID during optimization to figure out whether an index is eligible or not by comparing it to the index's creation transaction ID.
3. Limit `AS OF` to expressions that can be evaluated at binding phase, for now, we limit to the subset of constant expressions, but in the future we can extend it to more expressions.
4. Implement constant evaluation for expressions that are used quite often with `AS OF`, i.e. the `TIMESTAMP` expression (in `UnaryDateTimestampOperatorNode`) and `TIMESTAMPADD` expression (in `TernaryOperatorNode`).
5. Throw a proper error message if the user attempts to force a temporal-inconsistent index.

With this fix, the optimizer is able now to choose valid indices that are temporal-consistent if an `AS OF` expression is present.

## How to test

Here is a quick example:

```sql


create table t1(col1 int);
insert into t1 values 42, 42;
CALL SYSCS_UTIL.SYSCS_GET_CURRENT_TRANSACTION();
txnId
----

19200

delete from t1;
alter table t1 add constraint c1  unique(col1);

select * from t1 as of 19200;
COL1
-----------
42
42

select * from t1 as of 19200 where col1 >= 42; –- wrong result
COL1
-----------

-– because we are using the newly introduced index, which is empty
explain select * from t1 as of 19200> where col1 >= 42;
Plan
-------------------------------------------------------------------
Cursor(n=3,rows=18,updateMode=READ_ONLY (1),engine=OLTP (default))
   ->  ScrollInsensitive(n=3,totalCost=8.216,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
    ->  IndexScan[SQLA35EC2CE0178AC35AAF500000B960C28(1729)] (n=0,totalCost=4.036,scannedRows=18,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,baseTable=T1(1712),preds=[(COL1[0:1] >= 42)]) 
    
-- with the fix we get the correct results, since the index was not chosen!
select * from t1 as of 19200 where col1 >= 42;
COL1
-----------
42
42

explain select * from t1 as of 19200 where col1 >= 42;

Cursor(n=3,rows=18,updateMode=READ_ONLY (1),engine=OLTP (default))
  ->  ScrollInsensitive(n=3,totalCost=8.22,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1)
    ->  TableScan[T1(2512)](n=0,totalCost=4.04,scannedRows=20,outputRows=18,outputHeapSize=18 B,partitions=1,parallelTasks=1,preds=[(COL1[0:1] >= 42)])
Table statistics are missing or skipped for the following tables:
SPLICE.T1

-- moreover, force-using the index throws an error now:

select * from t1 as of 19200 --splice-properties index=i1
where col1 >= 42;
ERROR 42Z03: Specified index 'I1' is created at tx '21042' which is after time-travel mapped past transaction id '19200'.
```

### DIFF
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
@@ -2276,10 +2276,4 @@ public interface DataDictionary{
      * @throws StandardException In case of e.g. connection error to HBase admin.
      */
     long getConglomerateCreationTxId(long conglom) throws StandardException;
-
-    long getTxnAt(long ts) throws StandardException;
-
-    boolean txnWithin(long period, long pastTx) throws StandardException;
-
-    boolean txnWithin(long period, Timestamp pastTx) throws StandardException;
 }
--- a/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
+++ b/db-engine/src/main/java/com/splicemachine/db/iapi/sql/dictionary/DataDictionary.java
@@ -46,6 +46,8 @@ import com.splicemachine.db.iapi.types.*;
 import com.splicemachine.db.impl.sql.catalog.DataDictionaryCache;
 import com.splicemachine.db.impl.sql.execute.TriggerEventDML;
 
+import java.io.IOException;
+import java.sql.Timestamp;
 import java.sql.Types;
 import java.util.*;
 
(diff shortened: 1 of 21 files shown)


carolp-503 (approved):
Approved for 3.1

PR #5664, opened 2021-07-09
DB-11841 Log4j version 2 + Row Trigger Performance Improvements

msirek:
## Short Description
This updates Log4j in Splice from version 1 to version 2 and reduces the CPU path length for row triggers.

## Long Description
Profiling of trigger execution paths was carried out to identify bottlenecks in need of correction.  Statement logging turned out to be a critical bottleneck, so Log4j is upgraded from version 1 to version 2 in order to pick up asychronous logger support.  The main difference in Log4j version 2 that everyone needs to know about is how to get a logger.  
In version 1 we use Logger.getLogger, but in version 2 we use LogManager.getLogger.
version 1:
> private static final Logger LOG = Logger.getLogger(GenericPreparedStatement.class);

version 2:
> private static final Logger LOG = LogManager.getLogger(GenericPreparedStatement.class);

Also, all version 2 log4j imports begin with **import org.apache.logging**, such as:
> import org.apache.logging.log4j.LogManager;
> import org.apache.logging.log4j.Logger;

Going forward, please **do not** include version 1 imports, such as:
> import org.apache.log4j.Logger;

The log4j dependency is only defined in one pom file now, utilities/pom.xml.  Please do not add direct dependencies on log4j going forward, but add the following dependency instead:

>         <dependency>
>             <groupId>com.splicemachine</groupId>
>             <artifactId>utilities</artifactId>
>             <version>${project.version}</version>
>         </dependency>

Changes done:

- Include the log4j and scala dependencies in one location only: utilities/pom.xml
- Define ${log4j.version} and ${scala.version} in the main pom file.
- Exclude different versions of log4j and scala brought in by dependencies.
- Rename info-log4j.properties to info-log4j2.properties.
- Update log4j properties files to use the new syntax (see "Configuration with Properties" at https://logging.apache.org/log4j/log4j-2.4/manual/configuration.html)
- Update log file path names to use sys property lookups (see https://logging.apache.org/log4j/2.x/manual/lookups.html)
- Use asychronous loggers on standalone and physical clusters by configuring "Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector" in appropriate locations, see https://logging.apache.org/log4j/2.x/manual/async.html.  Asychronous logging on cloud is working with the latest version of the code.
- Fix join strategy hinting so it does not apply to single-table scans.
- Make subqueries non-materializable (for correctness) if they reference the trigger NEW or OLD rows.
- Remove the rowConcurrentExecutorsMap dependence on TriggerDescriptor

### Performance Improvements
- Asychronous logging removes a large bottleneck concurrent triggers encouter due to simultaneous statement logging.
- Add a protocolKeyCache to BaseMonitor
- Add caching of ScanOperations used inside a trigger as long as there is no DML statement in the trigger and no non-deterministic functions.
- Avoid pushing nondeterministic predicates to the base table in a trigger, to allow the base table rows to be cached. 
- Always compile row triggers to run in OLTP mode.
- Only build the trigger WHEN clause subquery operation tree on the first execution.
- Change logStartExecuting and logEndExecuting to use StringBuilder for constructing the statement string to log.

## How to test
A benchmark for concurrent row triggers is added: https://github.com/splicemachine/spliceengine/blob/d8174af313fd7068117839cecb8feea64efcad87/splice_machine/src/test/java/com/splicemachine/benchmark/ConcurrentTriggerBenchmark.java

For feature testing, some test combinations to try are:

1. WHEN clause with different types of subqueries, referencing correlated columns, the NEW and OLD trigger row, and non-deterministic functions
2. Triggers with the SET clause and/or SIGNAL statement
3. Triggers which hint spark (all row triggers should now execute in control if the main DML executes in control)
4. Multiple actions in a row trigger
5. Row triggers which cause other row triggers or statement triggers to fire


msirek:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3307/

msirek:

cloudspliceci:
TEST SUCCEEDED +1

msirek:

cloudspliceci:
TEST SUCCEEDED +1

msirek:

cloudspliceci:
TEST SUCCEEDED +1

db-engine/src/main/java/com/splicemachine/db/iapi/services/compiler/MethodBuilder.java
```diff
@@ -432,7 +432,7 @@ the completeConditional() is called. See conditionalIfNull() for example
 	/**
 		Call a method. The instance (receiver or reference) for non-static methods
 		must be pushed by the caller. The instance (for non-static) and the arguments
-		are popped of the stack, and the return value (if any) is pushed onto the stack.
+		aassre popped of the stack, and the return value (if any) is pushed onto the stack.
```
martinrupp: typo?

mem_sql/src/main/resources/log4j2.properties
```diff
@@ -12,13 +12,16 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 rootLogger.level = info
+rootLogger.includeLocation = true
 rootLogger.appenderRef.file.ref = File1
 
-appender.file.type = File
+appender.file.type = RandomAccessFile
```
martinrupp, line 18: what does this do?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNode.java
```diff
@@ -1278,6 +1278,15 @@ public final LanguageConnectionContext getLanguageConnectionContext(){
         return lcc;
     }
 
+    /**
+     * True if this predicate is used in a trigger query.
+     */
+    public final boolean compilingTrigger() {
+        LanguageConnectionContext lcc = getLanguageConnectionContext();
+        boolean compilingTrigger = lcc.compilingTrigger();
+        return compilingTrigger;
```
martinrupp, line 1287: (nitpic) that's a quite long way to write `return getLanguageConnectionContext().compilingTrigger();` :-P

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/ProjectRestrictOperation.java
```diff
@@ -172,7 +172,7 @@ public void init(SpliceOperationContext context) throws StandardException, IOExc
         cloneMap = ((boolean[]) statement.getSavedObject(cloneMapItem));
 
         if (restrictionMethodName != null)
-            restriction = new SpliceMethod<>(restrictionMethodName, activation);
+            restriction = new SpliceMethod<>(restrictionMethodName, source.getActivation());
```
ascend1, line 175: It feels like we need some comment here. Why the original code is wrong? I mean, for the generated code of one query, could this PRN and its source have different activations?

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNode.java
```diff
@@ -1278,6 +1278,15 @@ public final LanguageConnectionContext getLanguageConnectionContext(){
...
+    }
+
```
martinrupp, line 1289: as far as i saw it, this is only used if we are just compiling a trigger, right? after that is done, compilingTrigger is set to false again.
(nitpic) that's a quite long way to write `return getLanguageConnectionContext().compilingTrigger();` :-P

martinrupp (approved):

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/DerbyScanInformation.java
```diff
@@ -461,8 +461,7 @@ public long getConglomerateId() {
         Qualifier[][] scanQualifiers = null;
         if (scanQualifiersFieldName != null) {
             try {
-                if (qualifiersField == null)
-                    qualifiersField = activation.getClass().getField(scanQualifiersFieldName);
```
ascend1, line 466: This is also semantically different. Why do we need to overwrite `qualifiersField` when it already holds something?

ascend1 (commented):
According to the fixes to ITs, this PR seems to change query plans a lot (join order, join strategy, etc.) even for queries that are not relevant to triggers. I'm a bit concerned about potential regressions since we don't yet have a good way to find them.

arnaud-splice:
Is it possible to separate the logger changes from the rest? It is fairly hard to review at the moment.

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
```diff
@@ -4148,13 +4148,22 @@ private String getLogHeader() {
...
             // cache formatted statement log
-            if (statement.equals(lastLogStmt)) {
+            BaseActivation baseActivation = null;
+            if (hasBaseActivation) {
+                baseActivation = (BaseActivation) activation;
+                if (statement.equals(baseActivation.getLastLogStmt())) {
+                    return baseActivation.getLastLogStmtFormat();
+                }
+            }
+            else if (statement.equals(lastLogStmt)) {
```
dgomezferro, line 4172: I think it would be cleaner to use a small guava Cache that handles concurrency for us. If performance wise it is equivalent I'd choose that.

dgomezferro (changes requested):
Some minor comments, I've only reviewed the log4j stuff so far.

mem_sql/src/main/resources/log4j2.properties
msirek: RandomAccessFile is buffered and faster than a normal file appender, even for synchronous loggers.  I think I read somewhere they are needed for full asynchronous logging support, but I can't find that article now.  
At the following link, search for "FileAppender vs. RandomAccessFileAppender" for a performance comparison using synchronous loggers:
https://logging.apache.org/log4j/log4j-2.4/manual/async.html

db-engine/src/main/java/com/splicemachine/db/iapi/services/compiler/MethodBuilder.java
msirek: Yes, fixed it, thanks.

db-engine/src/main/java/com/splicemachine/db/impl/sql/compile/QueryTreeNode.java
msirek: Right.  Simplified the code as suggested.

msirek:
@ascend1 
> According to the fixes to ITs, this PR seems to change query plans a lot (join order, join strategy, etc.) even for queries that are not relevant to triggers. I'm a bit concerned about potential regressions since we don't yet have a good way to find them.

The change that causes this is in FromTable.nextAccessPath : 
-        if(userSpecifiedJoinStrategy!=null){
+        if(userSpecifiedJoinStrategy!=null && !isSingleTableScan(optimizer)){

The reason for this change is that the joinStrategy hint currently has the affect of an implicit joinOrder=fixed.
For example, if we have:
select * from t1
,t2 --splice-properties joinStrategy=broadcast
,t3 --splice-properties joinStrategy=sortmerge
...
;
When t3 is the outer table, the first call to nextAccessPath will pick broadcast joinStrategy when planning the single table access of t3 alone.  When the feasible method is called, it returns false, so we will never pick any plans where t3 is the outer table.  Likewise with t2 since it is hinted. Therefore in this example it would force t1 to be the outer table of the first join.  I had previously added a joinStrategy session hint, but because that gets picked up for single-table scans as well, it's effectively useless without this change because all joins are found to be infeasible except for nestedloop join.

So, the question is, should the joinStrategy hint mean "pick this join strategy when planning the hinted table as the inner table.
Or, does it mean, "only pick query plans where the hinted table is the inner table (joinOrder=fixed behavior), unless the hinted joinStrategy is nestedloop".

If we think the hint should mean the first definition, then the code change is appropriate.  If it should mean the 2nd definition, then the change is not appropriate.  With the first definition we will consider more query plans, but it means we may cost a query plan with one joinStrategy hint with a different plan that picks up a different joinStrategy hint (but perhaps not both hints at the same time).
Do you have an opinion on this?

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/ProjectRestrictOperation.java
msirek: It is a theoretical bug.  If the activations ever differ, it will be broken.  Using the source activation here ensures the correct one is always used.
Added the following comment:
>         // setCurrentRow, called by the source operation, is always done in the activation
>         // used by the source.  In case this activation differs from the activation
>         // of this ProjectRestrict, use the source's activation as the activation
>         // for looking up the row in the restriction method.
>         if (restrictionMethodName != null)
>             restriction = new SpliceMethod<>(restrictionMethodName, source.getActivation());

splice_machine/src/main/java/com/splicemachine/derby/impl/sql/execute/operations/DerbyScanInformation.java
msirek: This is restoring the code to the original behavior.  I previously added the null check to avoid some overhead, but I found that qualifiersField is not rebuilt to use the new activation when an operation tree is reused (when we call TableScanOperation.init), and it may point to the wrong row.  We need to build the qualifiersField to refer to the proper activation because we get the row to work on from the row array in the activation.  If we point to the wrong activation, we get the wrong row.  A possible other change could be to rebuild qualifiersField in scanInformation.initialize (called by ScanOperation.init), but it's possible we sometimes need to do more setup before populateQualifiers is called, so it may be risky to move the location of this setup.

msirek:
@arnaud-splice 
> Is it possible to separate the logger changes from the rest? It is fairly hard to review at the moment.

Sorry about that.  It would be difficult to do this and may take a day's worth of effort to make sure all commits are applied properly and everything is tested.  I would rather not do this if possible, but could do it if you think it's needed.

splice_machine/src/main/java/com/splicemachine/derby/jdbc/SpliceTransactionResourceImpl.java
```diff
@@ -51,7 +52,8 @@
  * This class it NOT thread safe
  */
 public final class SpliceTransactionResourceImpl implements AutoCloseable{
-    private static final Logger LOG=Logger.getLogger(SpliceTransactionResourceImpl.class);
+    private static final Logger LOG= LogManager.getLogger(SpliceTransactionResourceImpl.class);
+    public static final String CONNECTION_STRING = "jdbc:splice:"+ SQLConfiguration.SPLICE_DB+";create=true";
```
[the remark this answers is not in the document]
msirek: Removed CONNECTION_STRING

db-engine/src/main/java/com/splicemachine/db/impl/sql/conn/GenericLanguageConnectionContext.java
msirek: Good point.  They probably do better than using "synchronized".  Added caches in BaseActivation and GenericLanguageConnectionContext which are lazily allocated.

msirek:
jenkins please test branch @dbaas3.1

cloudspliceci:
TEST SUCCEEDED +1

PR #5668, opened 2021-07-13
DB-12242 Improve EXCEPT performance in case of skew

hatyo:
In this PR, the performance of `EXCEPT [DISTINCT]` in Spark/OLAP is improved in case of skew in either side of its relations.

## Long Description
The execution of `EXCEPT` statement in Spark delegates to running DataFrame [`except` ](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html#except) set operation, which, in turn, translates to a left anti-join.

In case of massively skewed data set the performance degrades heavily as the join tasks might be unbalanced by a straggler task that is handling most of the data.

To alleviate this problem, we enforce a `distinct` on both sides of the `except` before sending it to Spark. 

## How to test

Here is an example:

```sql

drop table t1 if exists;
drop table t2 if exists;

create table t1(col1 varchar(10));
create table t2(col1 varchar(10));

insert into t1 values null;
insert into t1 select * from t1; -- 2
insert into t1 select * from t1; -- 4
insert into t1 select * from t1; -- 8
insert into t1 select * from t1; -- 16
insert into t1 select * from t1; -- 32
insert into t1 select * from t1; -- 64
insert into t1 select * from t1; -- 128
insert into t1 select * from t1; -- 256
insert into t1 select * from t1; -- 512
insert into t1 select * from t1; -- 1024
insert into t1 select * from t1; -- 2048
insert into t1 select * from t1; -- 4096
insert into t1 select * from t1; -- 8192
insert into t1 select * from t1; -- 16k
insert into t1 select * from t1; -- 32k
insert into t1 select * from t1; -- 64k
insert into t1 select * from t1; -- 128k
insert into t1 select * from t1; -- 256k
insert into t1 select * from t1; -- 512k
insert into t1 select * from t1; -- 1M
insert into t1 select * from t1; -- 2M
insert into t1 select * from t1; -- 4M
insert into t1 select * from t1; -- 8M
insert into t1 values 'abc';

insert into t2 values 'abc';
insert into t2 values 'def';
insert into t2 values 'ghi';
insert into t2 values 'jkl';
insert into t2 values 'mno';
insert into t2 values 'pqr';
insert into t2 values 'stu';
insert into t2 values 'vwx';
insert into t2 values 'yz';
insert into t2 values '';

insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;

insert into t2 select '' from t1 where t1.col1 is null;
```

as we can see, both tables `t1` and `t2` are heavily skewed. If we run the following statement:

```sql
select * from t1 --splice-properties useOlap=true
except
select * from t2;
```

* **before the fix**: never returns (cancelled manually after few hours)
* **after the fix**: finishes in about 28 seconds


msirek (approved):
The fix looks good.  Do intersect and union have a similar issue?  If you don't know, maybe we should open a Jira to investigate it.

hatyo:
> The fix looks good. Do intersect and union have a similar issue? If you don't know, maybe we should open a Jira to investigate it.

Thanks Mark, I can't tell for sure, so I opened ticket [DB-12343](https://splicemachine.atlassian.net/browse/DB-12343) to keep track of this.

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3347/

arnaud-splice (approved):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
```diff
@@ -623,7 +623,12 @@ private Dataset getDataset(DataSet dataSet) throws StandardException {
             Dataset<Row> right = getDataset(dataSet);
 
             //Do the subtract
-            Dataset<Row> result = left.except(right);
+            Dataset<Row> result;
+            if(isDistinct) {
+                result = left.distinct().except(right.distinct());
+            } else {
+                result = left.except(right);
+            }
```
dgomezferro, line 631: I think it'd be better to generate a distinct node during query compilation, that way we'd be able to push it down to the table scan. Nowadays I think we don't do anything fancy with it, it's implemented as a Scan + Distinct, but in the future we could use a Skipping HBase filter similarly to what we do with IndexPrefixIteratorMode

In short, I think this is a good change, but in the future it could be improved a little bit.

dgomezferro (approved):

hbase_sql/src/main/java/com/splicemachine/derby/stream/spark/NativeSparkDataSet.java
hatyo: Thanks Daniel! I thought I would attempt to keep it localized to Spark for two reasons:

- As discussed before, it looks to me like a bug in Spark `except` operation itself and it should be fixed there. With this PR I tried to isolate the fix as much as possible.
- If we have no index / PK, then there is not much difference between `DistinctScanOperation` + `DataSet.subtract` and `ScanOperation` + `DataSet.subtract`, both iterate on the entire table but use an intermediary HashSet when iterating over the values to get "notified" about a distinct value. I am guessing if there is a more efficient keyed AccessPath then optimizer should choose that anyway which can be iterated upon with a normal iterator or a skip iterator,, but I haven't tested that to be honest.

hatyo:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3350/

carolp-503 (approved):
+1

ipraznik-splice:

ipraznik-splice:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3358/

ipraznik-splice:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3359/

jyuanca:

cloudspliceci:
TEST FAILED
http://nexus.splicemachine.com:8080/job/spliceengine-PR/job/Splice-PR-Branch/3367/

jyuanca:

cloudspliceci:
TEST SUCCEEDED +1

PR #5670, opened 2021-07-13
DB-12242 Improve EXCEPT performance in case of skew (3.0)

hatyo:
In this PR, the performance of `EXCEPT [DISTINCT]` in Spark/OLAP is improved in case of skew in either side of its relations.

## Long Description
The execution of `EXCEPT` statement in Spark delegates to running DataFrame [`except` ](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-setops.html#except) set operation, which, in turn, translates to a left anti-join.

In case of massively skewed data set the performance degrades heavily as the join tasks might be unbalanced by a straggler task that is handling most of the data.

To alleviate this problem, we enforce a `distinct` on both sides of the `except` before sending it to Spark. 

## How to test

Here is an example:

```sql

drop table t1 if exists;
drop table t2 if exists;

create table t1(col1 varchar(10));
create table t2(col1 varchar(10));

insert into t1 values null;
insert into t1 select * from t1; -- 2
insert into t1 select * from t1; -- 4
insert into t1 select * from t1; -- 8
insert into t1 select * from t1; -- 16
insert into t1 select * from t1; -- 32
insert into t1 select * from t1; -- 64
insert into t1 select * from t1; -- 128
insert into t1 select * from t1; -- 256
insert into t1 select * from t1; -- 512
insert into t1 select * from t1; -- 1024
insert into t1 select * from t1; -- 2048
insert into t1 select * from t1; -- 4096
insert into t1 select * from t1; -- 8192
insert into t1 select * from t1; -- 16k
insert into t1 select * from t1; -- 32k
insert into t1 select * from t1; -- 64k
insert into t1 select * from t1; -- 128k
insert into t1 select * from t1; -- 256k
insert into t1 select * from t1; -- 512k
insert into t1 select * from t1; -- 1M
insert into t1 select * from t1; -- 2M
insert into t1 select * from t1; -- 4M
insert into t1 select * from t1; -- 8M
insert into t1 values 'abc';

insert into t2 values 'abc';
insert into t2 values 'def';
insert into t2 values 'ghi';
insert into t2 values 'jkl';
insert into t2 values 'mno';
insert into t2 values 'pqr';
insert into t2 values 'stu';
insert into t2 values 'vwx';
insert into t2 values 'yz';
insert into t2 values '';

insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;
insert into t2 select * from t2;

insert into t2 select '' from t1 where t1.col1 is null;
```

as we can see, both tables `t1` and `t2` are heavily skewed. If we run the following statement:

```sql
select * from t1 --splice-properties useOlap=true
except
select * from t2;
```

* **before the fix**: never returns (cancelled manually after few hours)
* **after the fix**: finishes in about 28 seconds


carolp-503 (approved):
Approved for 3.0

PR #5675, opened 2021-07-15
DB-12350 fix build error in ModifyColumnConstantOperation (3.1)

martinrupp:
DB-12224 changed the API for DataDictionary.getTableDescriptor, and DB-12158 still used the old API, so this is fixing this build error.


carolp-503 (approved):
Approved for 3.1

PR #5696, opened 2021-07-20
DB-12373 fix SqlshellIT.testCustomFormatsOption sporadic

martinrupp:
we test the current time against `\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d.\d\d\d\d\d\d` which is mostly correct,
however, if we’re unlucky and the current time is like `2021-07-19 17:55:24.500000`, it would be displayed
as `2021-07-19 17:55:24.5` , and we get a failed test. This fixes that.


martinrupp:
propability of this sporadic to happen is 1/10 obviously :-P

ipraznik-splice (approved):

ascend1 (approved):

jyuanca (approved):

carolp-503 (approved):
+1

